PR #87, opened 2016-06-06
Redesign schema handling and add additive-only mode

tgoyne:
Add the concept of "schema modes" which control how updates to the schema are handled. Currently there are five:

  1. Automatic: the Cocoa binding's default mode. Adding new tables and adding or removing indexes is always allowed. Adding, removing, or modifying a property, or changing the primary key requires incrementing the schema version. The migration function is called whenever the schema version is increased.
  2. Read-only: open the file in read-only mode, allow missing or extra tables, allow index mismatches, and error on all other differences.
  3. Reset file: allow adding new tables or adding/removing indexes. All other changes (including bumping the schema version) result in the file being deleted and recreated. Does not work with multi-process.
  4. Additive-only: Allowing adding new tables and new columns to tables. Do not allow modifying existing columns, or changing the primary key of existing tables. Indexes are updated only if the schema version is increased.
  5. Manual: the Java binding's default mode. Any schema changes require a schema version bump. Changes are not automatically applied; instead the migration function is called which is responsible for ensuring that the file's schema matches the declared one.

Supporting these various modes requires detangling the process of comparing a Schema object to a Group's schema from the code that does the updating. Redesign schema updating to use roughly the following high-level flow:

  1. Read the Schema from the Group.
  2. Diff that schema with the schema passed in by the user and build up a list of changes needed to make them match.
  3. Verify that all changes are legal for the current schema mode.
  4. Apply the changes.

Obviously some of these steps are skipped for some of the modes.

There are a few breaking changes to the API in this:

  1. The schema property on Realm::Config is now an `Optional<Schema>` rather than `unique_ptr<Schema>`.
  2. The schema and schema version is read from the Realm via the new member functions for that rather than reading them from the Realm's config object.
  3. Opening Realms in read-only mode is now a schema mode rather than a separate flag.
  4. Realm::read_group() now returns a reference, and the ObjectStore functions all take references when nullptr is not a legal argument.

This also adds `ObjectStore::rename_property()` for renaming a property from within a migration function.

The main remaining WIP thing is a lot more testing; this passes realm-cocoa's tests but that doesn't mean a lot as Cocoa has its own parallel schema functionality that may be masking bugs and there's some functionality here not used by Cocoa.

### DIFF
--- a/src/binding_context.hpp
+++ b/src/binding_context.hpp
@@ -106,19 +106,20 @@ public:
 
     // Change information for a single field of a row
     struct ColumnInfo {
-        // Did this column change?
-        bool changed = false;
-        // For LinkList columns, what kind of change occurred?
-        // Always None for other column types
+        // The index of this column prior to the changes in the tracked
+        // transaction, or -1 for newly inserted columns.
+        size_t initial_column_index = -1;
+        // What kind of change occurred?
+        // Always Set or None for everything but LinkList columns.
         enum class Kind {
             None,   // No change
-            Set,    // The entries at `indices` were assigned to
+            Set,    // The value or entries at `indices` were assigned to
             Insert, // New values were inserted at each of the indices given
             Remove, // Values were removed at each of the indices given
             SetAll  // The entire LinkList has been replaced with a new set of values
         } kind = Kind::None;
-        // The indices where things happened for Set, Insert and Remove
-        // Not used for None and SetAll
+        // The indices where things happened for Set, Insert and Remove on
+        // LinkList columns. Not used for other types or for None or SetAll.
         IndexSet indices;
     };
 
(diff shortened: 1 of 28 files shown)


agersoncgps:
I still see readOnly as a stand alone property in the [documentation](https://realm.io/docs/javascript/1.3.1/api/Realm.html#~Configuration) and no mention of how to set or change a schema mode to something other than "additive-only", which appears to be the default. I am trying to perform a migration and getting "The following changes cannot be made in additive-only schema mode" and if I provide a migration function I get "Realms opened in Additive-only schema mode do not use a migration function"

bdash:
The schema mode is internal to the Realm SDKs, and cannot be manually specified. It is set by the Realm SDKs based on whether not sync is being used: non-synced Realms use either automatic or manual schema modes (depending on the binding), while synced Realms always use additive schema mode. If you have further questions, please open an issue against the binding you're using.

PR #94, opened 2016-06-23
Choose the commit helper by compiling source

beeender:
Use macro to choose the implementation is not flexible enough and easy
to make mistakes.
Just use global function to return the implementation instance during
run time.

Especially for realm-java, it probably will use an empty commit helper
at the beginning which makes the marcos even more complex.


jpsim:
@beeender @bdash any reason to keep this open?

bdash:
I don't think so. @beeender, if the approach Thomas suggested above isn't suitable for some reason, let's have a conversation about your needs on Slack to figure out the right path forward.

PR #114, opened 2016-07-29
Eliminate some string copies for Schema::find()

tgoyne:
This gives a surprisingly large speedup for Realm instance init.

### DIFF
--- a/src/schema.cpp
+++ b/src/schema.cpp
@@ -52,25 +52,25 @@ Schema::Schema(base types) : base(std::move(types))
     std::sort(begin(), end(), compare_by_name);
 }
 
-Schema::iterator Schema::find(std::string const& name)
+Schema::iterator Schema::find(StringData name)
 {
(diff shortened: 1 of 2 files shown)


src/schema.cpp
```diff
@@ -52,25 +52,25 @@ Schema::Schema(base types) : base(std::move(types))
...
 
-Schema::const_iterator Schema::find(std::string const& name) const
+Schema::const_iterator Schema::find(StringData name) const
 {
     return const_cast<Schema *>(this)->find(name);
 }
 
 Schema::iterator Schema::find(ObjectSchema const& object) noexcept
 {
-    auto it = std::lower_bound(begin(), end(), object, compare_by_name);
```
bdash: Since `compare_by_name` is now only used in one place it might be worth getting rid of the function in favor of a lambda at the one place the comparison is needed.

bdash:
👍

PR #115, opened 2016-08-01
Removed unused `toString` function

JadenGeller:
@bdash @tgoyne

### DIFF
--- a/tests/schema.cpp
+++ b/tests/schema.cpp
@@ -64,15 +64,6 @@ struct SchemaChangePrinter {
 #undef REALM_SC_PRINT
 };
 
-namespace Catch {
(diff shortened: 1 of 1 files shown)


tgoyne:
The toString overload is used to format failure messages for the schema change tests. If it is actually unused then that is a bug that should be fixed.

JadenGeller:
It's causing a build warning for me. Can we remove it and whoever adds those tests can add it back?

JadenGeller:
To be more clear, it's causing a build warning to due a missing function prototype. It only shows a build warning for being unused if I mark it as static. Should we just add the function prototype?

bdash:
#119 fixes this in a way that makes the custom formatting in failure messages work as intended.

PR #116, opened 2016-08-02
Implement async handover

JadenGeller:
@tgoyne @bdash

### DIFF
--- a/src/handover.cpp
+++ b/src/handover.cpp
@@ -23,24 +23,16 @@ using namespace realm;
 AnyThreadConfined::AnyThreadConfined(const AnyThreadConfined& thread_confined)
 {
     switch (thread_confined.m_type) {
-        case Type::Row:
-            new (&m_row) Row(thread_confined.m_row);
+        case Type::Object:
+            new (&m_object) Object(thread_confined.m_object);
             break;
 
-        case Type::Query:
-            new (&m_query) Query(thread_confined.m_query);
+        case Type::List:
+            new (&m_list) List(thread_confined.m_list);
             break;
 
-        case Type::TableRef:
-            new (&m_table_ref) TableRef(thread_confined.m_table_ref);
-            break;
-
-        case Type::TableView:
-            new (&m_table_view) TableView(thread_confined.m_table_view);
-            break;
-
-        case Type::LinkViewRef:
-            new (&m_link_view_ref) LinkViewRef(thread_confined.m_link_view_ref);
+        case Type::Results:
+            new (&m_results) Results(thread_confined.m_results);
             break;
     }
     new (&m_type) Type(thread_confined.m_type);
@@ -49,24 +41,16 @@ AnyThreadConfined::AnyThreadConfined(const AnyThreadConfined& thread_confined)
 AnyThreadConfined::AnyThreadConfined(AnyThreadConfined&& thread_confined)
 {
     switch (thread_confined.m_type) {
-        case Type::Row:
-            new (&m_row) Row(std::move(thread_confined.m_row));
+        case Type::Object:
+            new (&m_object) Object(std::move(thread_confined.m_object));
             break;
 
-        case Type::Query:
-            new (&m_query) Query(std::move(thread_confined.m_query));
+        case Type::List:
+            new (&m_list) List(std::move(thread_confined.m_list));
             break;
 
-        case Type::TableRef:
-            new (&m_table_ref) TableRef(std::move(thread_confined.m_table_ref));
-            break;
-
-        case Type::TableView:
-            new (&m_table_view) TableView(std::move(thread_confined.m_table_view));
-            break;
-
-        case Type::LinkViewRef:
-            new (&m_link_view_ref) LinkViewRef(std::move(thread_confined.m_link_view_ref));
+        case Type::Results:
+            new (&m_results) Results(std::move(thread_confined.m_results));
             break;
     }
     new (&m_type) Type(std::move(thread_confined.m_type));
@@ -75,115 +59,103 @@ AnyThreadConfined::AnyThreadConfined(AnyThreadConfined&& thread_confined)
 AnyThreadConfined::~AnyThreadConfined()
 {
     switch (m_type) {
-        case Type::Row:
-            m_row.~Row();
+        case Type::Object:
+            m_object.~Object();
             break;
 
-        case Type::Query:
-            m_query.~Query();
+        case Type::List:
+            m_list.~List();
             break;
 
-        case Type::TableRef:
-            m_table_ref.~TableRef();
-            break;
-
-        case Type::TableView:
-            m_table_view.~TableView();
-            break;
-
-        case Type::LinkViewRef:
-            m_link_view_ref.~LinkViewRef();
+        case Type::Results:
+            m_results.~Results();
             break;
     }
 }
 
-AnyHandover AnyThreadConfined::export_for_handover(SharedGroup &shared_group) const
+SharedRealm AnyThreadConfined::get_realm() const
 {
     switch (m_type) {
-        case AnyThreadConfined::Type::Row:
-            return AnyHandover(shared_group.export_for_handover(m_row));
+        case Type::Object:
+            return m_object.realm();
 
-        case AnyThreadConfined::Type::Query:
-            return AnyHandover(shared_group.export_for_handover(m_query, ConstSourcePayload::Copy));
+        case Type::List:
+            return m_list.get_realm();
 
-        case AnyThreadConfined::Type::TableRef:
-            return AnyHandover(shared_group.export_table_for_handover(m_table_ref));
+        case Type::Results:
+            return m_results.get_realm();
+    }
+}
 
-        case AnyThreadConfined::Type::TableView:
-            return AnyHandover(shared_group.export_for_handover(m_table_view, ConstSourcePayload::Copy));
+AnyHandover AnyThreadConfined::export_for_handover() const
+{
+    SharedGroup& shared_group = Realm::Internal::get_shared_group(*get_realm());
+    switch (m_type) {
+        case AnyThreadConfined::Type::Object:
+            return AnyHandover(shared_group.export_for_handover(m_object.row()), &m_object.get_object_schema());
 
-        case AnyThreadConfined::Type::LinkViewRef:
-            return AnyHandover(shared_group.export_linkview_for_handover(m_link_view_ref));
+        case AnyThreadConfined::Type::List:
+            return AnyHandover(shared_group.export_linkview_for_handover(m_list.get_linkview()));
+
+        case AnyThreadConfined::Type::Results:
+            return AnyHandover(shared_group.export_for_handover(m_results.get_query(), ConstSourcePayload::Copy),
+                               m_results.get_sort());
     }
 }
 
 AnyHandover::AnyHandover(AnyHandover&& handover)
 {
     switch (handover.m_type) {
-        case Type::Row:
-            new (&m_row) RowHandover(std::move(handover.m_row));
+        case AnyThreadConfined::Type::Object:
+            new (&m_object.row_handover) RowHandover(std::move(handover.m_object.row_handover));
+            m_object.object_schema = handover.m_object.object_schema;
             break;
 
-        case Type::Query:
-            new (&m_query) QueryHandover(std::move(handover.m_query));
+        case AnyThreadConfined::Type::List:
+            new (&m_list.link_view_handover) LinkViewHandover(std::move(handover.m_list.link_view_handover));
             break;
 
-        case Type::Table:
-            new (&m_table) TableHandover(std::move(handover.m_table));
-            break;
-
-        case Type::TableView:
-            new (&m_table_view) TableViewHandover(std::move(handover.m_table_view));
-            break;
-
-        case Type::LinkView:
-            new (&m_link_view) LinkViewHandover(std::move(handover.m_link_view));
+        case AnyThreadConfined::Type::Results:
+            new (&m_results.query_handover) QueryHandover(std::move(handover.m_results.query_handover));
+            new (&m_results.sort_order) SortOrder(std::move(handover.m_results.sort_order));
             break;
     }
-    new (&m_type) Type(handover.m_type);
+    new (&m_type) AnyThreadConfined::Type(handover.m_type);
 }
 
 AnyHandover::~AnyHandover()
 {
     switch (m_type) {
-        case Type::Row:
-            m_row.~unique_ptr();
+        case AnyThreadConfined::Type::Object:
+            m_object.row_handover.~unique_ptr();
             break;
 
-        case Type::Query:
-            m_query.~unique_ptr();
+        case AnyThreadConfined::Type::List:
+            m_list.link_view_handover.~unique_ptr();
             break;
 
-        case Type::Table:
-            m_table.~unique_ptr();
-            break;
-
-        case Type::TableView:
-            m_table_view.~unique_ptr();
-            break;
-
-        case Type::LinkView:
-            m_link_view.~unique_ptr();
+        case AnyThreadConfined::Type::Results:
+            m_results.query_handover.~unique_ptr();
+            m_results.sort_order.~SortOrder();
             break;
     }
 }
 
-AnyThreadConfined AnyHandover::import_from_handover(SharedGroup &shared_group) &&
+AnyThreadConfined AnyHandover::import_from_handover(SharedRealm realm) &&
 {
+    SharedGroup& shared_group = Realm::Internal::get_shared_group(*realm);
     switch (m_type) {
-        case Type::Row:
-            return AnyThreadConfined(*shared_group.import_from_handover(std::move(m_row)));
-
-        case Type::Query:
-            return AnyThreadConfined(*shared_group.import_from_handover(std::move(m_query)));
-
-        case Type::Table:
-            return AnyThreadConfined(shared_group.import_table_from_handover(std::move(m_table)));
-
-        case Type::TableView:
-            return AnyThreadConfined(*shared_group.import_from_handover(std::move(m_table_view)));
-            
-        case Type::LinkView:
-            return AnyThreadConfined(shared_group.import_linkview_from_handover(std::move(m_link_view)));
+        case AnyThreadConfined::Type::Object: {
+            auto row = shared_group.import_from_handover(std::move(m_object.row_handover));
+            return AnyThreadConfined(Object(std::move(realm), *m_object.object_schema, std::move(*row)));
+        }
+        case AnyThreadConfined::Type::List: {
+            auto link_view_ref = shared_group.import_linkview_from_handover(std::move(m_list.link_view_handover));
+            return AnyThreadConfined(List(std::move(realm), std::move(link_view_ref)));
+        }
+        case AnyThreadConfined::Type::Results: {
+            auto query = shared_group.import_from_handover(std::move(m_results.query_handover));
+            return AnyThreadConfined(Results(std::move(realm), std::move(*query), m_results.sort_order));
+        }
     }
 }
--- a/src/handover.cpp
+++ b/src/handover.cpp
@@ -130,7 +130,7 @@ AnyHandover AnyThreadConfined::export_for_handover() const
     SharedGroup& shared_group = Realm::Internal::get_shared_group(*get_realm());
     switch (m_type) {
         case AnyThreadConfined::Type::Object:
-            return AnyHandover(shared_group.export_for_handover(m_object.row()), &m_object.get_object_schema());
+            return AnyHandover(shared_group.export_for_handover(m_object.row()), m_object.get_object_schema().name);
 
         case AnyThreadConfined::Type::List:
             return AnyHandover(shared_group.export_linkview_for_handover(m_list.m_link_view));
@@ -146,7 +146,7 @@ AnyHandover::AnyHandover(AnyHandover&& handover)
     switch (handover.m_type) {
         case AnyThreadConfined::Type::Object:
             new (&m_object.row_handover) RowHandover(std::move(handover.m_object.row_handover));
-            m_object.object_schema = handover.m_object.object_schema; // primitive
+            new (&m_object.object_schema_name) std::string(std::move(handover.m_object.object_schema_name));
             break;
 
         case AnyThreadConfined::Type::List:
@@ -166,7 +166,7 @@ AnyHandover& AnyHandover::operator=(AnyHandover&& handover)
     switch (handover.m_type) {
         case AnyThreadConfined::Type::Object:
             m_object.row_handover = std::move(handover.m_object.row_handover);
-            m_object.object_schema = handover.m_object.object_schema; // primitive
+            m_object.object_schema_name = std::move(handover.m_object.object_schema_name);
             break;
 
         case AnyThreadConfined::Type::List:
@@ -206,7 +206,9 @@ AnyThreadConfined AnyHandover::import_from_handover(SharedRealm realm) &&
     switch (m_type) {
         case AnyThreadConfined::Type::Object: {
             auto row = shared_group.import_from_handover(std::move(m_object.row_handover));
-            return AnyThreadConfined(Object(std::move(realm), *m_object.object_schema, std::move(*row)));
+            auto object_schema = realm->schema().find(m_object.object_schema_name);
+            REALM_ASSERT_DEBUG(object_schema != realm->schema().end());
+            return AnyThreadConfined(Object(std::move(realm), std::move(*object_schema), std::move(*row)));
         }
         case AnyThreadConfined::Type::List: {
             auto link_view_ref = shared_group.import_linkview_from_handover(std::move(m_list.link_view_handover));
--- a/src/handover.cpp
+++ b/src/handover.cpp
@@ -1,0 +1,189 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#include "handover.hpp"
+
+using namespace realm;
+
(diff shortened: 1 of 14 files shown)


JadenGeller:
Note that this PR is identical to https://github.com/realm/realm-cocoa/pull/3828 except
* it only includes the Object Store changes
* it adds Object Store handover tests
* it prefixes the `AnyThreadConfined` getters with `get_`

src/list.hpp
```diff
@@ -106,6 +106,8 @@ class List {
         size_t valid_count;
     };
 
+    LinkViewRef get_linkview() const { return m_link_view; }
```
JadenGeller: @tgoyne I don't think this would be improved by exposing an `AnyThreadConfined` here instead since it has a `get_linkview()` property on itself anyway. Thoughts?

JadenGeller:
I noticed we sometimes use the styling `link_view` and other times use `linkview`—which is preferred?

JadenGeller:
@bdash suggested I modify the user-facing API to utilize Object Store types rather than Core types, so I'm going to make this modification.

JadenGeller:
Also discussed removing copying overhead with @bdash. We decided it is probably possible using reference members, but becomes problematic on import. The API might need to be modified to take a closure where you store the result rather than just returning it, which is probably not worth it.

JadenGeller:
@tgoyne @bdash I've made some major changes to this, so I'd appreciate if you'd check it out!

* Handover now operates on Object Store types instead of Core types. This is a more natural API for the bindings to build on since extra information like `SortOrder` is handled by this abstraction, and the bindings don't have to understand the implementations of each type. Though the Cocoa bindings don't use the Object Store `Object` type, it's relatively cheap to construct one given a `Row`.
* Instead of using a `std::shared_ptr` and asserting that the packet wasn't imported multiple times, now we're returning a `HandoverPackage` directly. The copy constructors for this type are deleted, so it cannot be used multiple times. To avoid importing huge headers in `shared_realm.hpp`, we define a `VersionID` type inside `HandoverPackage` that stores the same information as `SharedGroup::VersionID`. It's annoying to convert between, but probably better than dragging this header in.
* Importing a `HandoverPackage` no longer ends a read transaction! This means that importing a handover on the main thread will no longer invalidate all your accessors. This makes handover useful for more than just GCD-style dispatch queues.
* Now you can import a `HandoverPackage` that was exported on either an older or a newer version. On import, if the package is older, a temporary Realm is used to advance it. If the Realm is older, it is advanced to the state of the packet. After import, the Realm is refreshed to provide less weird import refresh semantics.
* Lots more tests! I'm not checking a lot of edge cases. Let me know if you think anything is missing.

🎉

src/handover.hpp
```diff
@@ -0,0 +1,125 @@
...
+            LinkViewHandover link_view_handover;
+        } m_list;
+
+        struct {
+            QueryHandover query_handover;
+            SortOrder sort_order;
+        } m_results;
+    };
+
+    AnyHandover(RowHandover row_handover, const ObjectSchema* object_schema) : m_type(AnyThreadConfined::Type::Object), m_object({
```
tgoyne: This line is too long for github's code view, and the formatting is really weird.

src/handover.hpp
```diff
@@ -0,0 +1,125 @@
...
+#ifndef realm_handover_hpp
+#define realm_handover_hpp
+
+#include <realm/row.hpp>
+#include <realm/query.hpp>
+#include <realm/group_shared.hpp>
+#include <realm/link_view.hpp>
+#include <realm/table_view.hpp>
+
+#include "object_accessor.hpp"
```
tgoyne: objectstore headers before the core headers, and sort them

src/shared_realm.cpp
```diff
@@ -554,5 +555,110 @@ util::Optional<int> Realm::file_format_upgraded_from_version() const
     return util::none;
 }
 
+#define VERSION_CAST(version_id) {(version_id).version, (version_id).index}
```
tgoyne: This could just be a function.

src/shared_realm.cpp
```diff
@@ -554,5 +555,110 @@ util::Optional<int> Realm::file_format_upgraded_from_version() const
...
+
+// Precondition: `m_version` is not greater than `new_version`
+// Postcondition: `m_version` is equal to `new_version`
+void Realm::HandoverPackage::advance_to_version(VersionID new_version)
+{
+    if ((SharedGroup::VersionID)VERSION_CAST(new_version) == (SharedGroup::VersionID)VERSION_CAST(m_version_id)) return;
```
tgoyne: `return` on new line.

src/shared_realm.cpp
```diff
@@ -554,5 +555,110 @@ util::Optional<int> Realm::file_format_upgraded_from_version() const
...
+    REALM_ASSERT_DEBUG(((SharedGroup::VersionID)VERSION_CAST(new_version) >
+                        (SharedGroup::VersionID)VERSION_CAST(m_version_id)));
+
+    // Open `Realm` at handover version
+    Realm::Config config = m_coordinator->get_config();
+    config.cache = false;
+    SharedRealm realm = m_coordinator->get_realm(config);
+    REALM_ASSERT(!realm->is_in_transaction());
```
tgoyne: Should be is_in_read_transaction()

src/shared_realm.cpp
```diff
@@ -554,5 +555,110 @@ util::Optional<int> Realm::file_format_upgraded_from_version() const
...
+        m_shared_group->unpin_version(VERSION_CAST(handover.m_version_id));
+        handover.m_coordinator = nullptr;
+    });
+
+    verify_thread();
+    if (is_in_transaction()) {
+        throw InvalidTransactionException("Cannot accept handover during a write transaction.");
+    }
+
+    // Ensure we're on the same verison as the handover
```
tgoyne: verison

src/shared_realm.hpp
```diff
@@ -34,6 +34,8 @@ class Realm;
 class Replication;
 class SharedGroup;
 class StringData;
+class AnyThreadConfined;
+class AnyHandover;
```
tgoyne: sort declarations

src/shared_realm.cpp
```diff
@@ -554,5 +555,110 @@ util::Optional<int> Realm::file_format_upgraded_from_version() const
...
+    }
+
+    std::vector<AnyThreadConfined> objects;
+    objects.reserve(handover.m_objects.size());
+    for (auto &object : handover.m_objects) {
+        objects.push_back(std::move(object).import_from_handover(shared_from_this()));
+    }
+
+    // Avoid weird partial-refresh semantics when importing old packages by always refreshing
+    transaction::advance(*m_shared_group, m_binding_context.get(), m_config.schema_mode);
```
tgoyne: Could just be `refresh()`.

tests/handover.cpp
```diff
@@ -0,0 +1,370 @@
...
+#include "schema.hpp"
+#include "handover.hpp"
+
+#include <thread>
+#include <future>
+
+using namespace realm;
+
+static TableRef get_table(Realm& realm, const ObjectSchema &object_schema) {
+    StringData name (("class_" + object_schema.name).c_str());
```
tgoyne: This should just call `ObjectStore::table_for_object_type()`.

The .c_str() here is redundant.

tests/handover.cpp
```diff
@@ -0,0 +1,370 @@
...
+    return realm.read_group().get_table(name);
+}
+
+static Object create_object(SharedRealm realm, const ObjectSchema &object_schema) {
+    TableRef table = get_table(*realm, object_schema);
+    return Object(std::move(realm), object_schema, Row((*table)[table->add_empty_row()]));
```
tgoyne: The explicit conversion to `Row` should be unnecessary here.

tests/handover.cpp
```diff
@@ -0,0 +1,370 @@
...
+}
+
+static List get_list(const Object& object, size_t column_ndx) {
+    return List(object.realm(), object.row().get_table()->get_linklist(column_ndx, object.row().get_index()));
```
tgoyne: `object.row().get_linklist(column_ndx)`

tests/handover.cpp
```diff
@@ -0,0 +1,370 @@
...
+}
+
+TEST_CASE("handover") {
+    InMemoryTestFile config;
+    config.cache = false;
+    config.automatic_change_notifications = false;
+
+    SharedRealm r = Realm::get_shared_realm(config);
+
+    static const ObjectSchema string_object ({"string_object", {
```
tgoyne: I don't see any reason for these to be static.

No space before (.

tests/handover.cpp
```diff
@@ -0,0 +1,370 @@
...
+        SECTION("import") {
+            auto h = r->package_for_handover({});
+            r->begin_transaction();
+            REQUIRE_THROWS(r->accept_handover(std::move(h)));
+        }
+    }
+
+    SECTION("cleanup properly unpins version") {
+        auto durability = config.in_memory ? SharedGroup::durability_MemOnly :
+        SharedGroup::durability_Full;
```
tgoyne: This line needs to be indented.

tgoyne: But since config will only ever be in-memory you should just pass that directly to the constructor.

tests/handover.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+        auto handover_version = get_current_version();
+        auto h = r->package_for_handover({});
+        r->begin_transaction(); r->commit_transaction(); // Advance version
+
+        REQUIRE(get_current_version() != handover_version); // Ensure advanced
+        REQUIRE_NOTHROW(shared_group.begin_read(handover_version)); shared_group.end_read(); // Ensure pinned
+
+        SECTION("destroyed without being imported") {
+            h.~HandoverPackage(); // Destory handover, unpinning version
```
tgoyne: This results in the destructor being called twice on the same object, which is UB. Could use a util::Optional<Handover> instead to let you destroy it here in a more natural way.

tests/handover.cpp
```diff
@@ -0,0 +1,370 @@
...
+        }
+        SECTION("import into newer version") {
+            r->begin_transaction();
+            Object num = create_object(r, int_object);
+            num.row().set_int(0, 7);
+            r->commit_transaction();
+
+            REQUIRE(num.row().get_int(0) == 7);
+            auto h = r->package_for_handover({{{num}}});
+            std::thread([h {std::move(h)}, config]() mutable {
```
tgoyne: no space between h and {

tests/handover.cpp
```diff
@@ -0,0 +1,370 @@
...
+                SharedRealm r = Realm::get_shared_realm(config);
+                auto h_import = r->accept_handover(std::move(h));
+                Results results = h_import[0].get_results();
+
+                REQUIRE(results.size() == 3);
+                REQUIRE(results.get(0).get_string(0) == "D");
+                REQUIRE(results.get(1).get_string(0) == "B");
+                REQUIRE(results.get(2).get_string(0) == "A");
+                r->begin_transaction();
+                get_table(*r, string_object)->move_last_over(results.get(2).get_index());
```
tgoyne: results.get(2).move_last_over()

src/list.hpp
tgoyne: The goal of avoiding having get_linkview() is just to ensure that people add any required functionality to List rather than being lazy and doing things on the linkview directly.

tgoyne:
Overall design seems pretty good now.

src/shared_realm.cpp
JadenGeller: With `{` brackets `}` or no?

src/shared_realm.cpp
JadenGeller: This is slightly more efficient since it doesn't have to perform a bunch of checks again, but it might not be worth the added code complexity. Is that your reasoning?

tests/handover.cpp
JadenGeller: It's static so it doesn't need to be captured by the lambda. Is this bad practice?

src/shared_realm.cpp
tgoyne: The overhead of the extra checks is utterly inconsequential. We do similar checks on every single access of a property and even there they have a small impact.

tests/handover.cpp
tgoyne: There's no upside from capturing fewer locals here.

src/shared_realm.cpp
tgoyne: I don't have strong opinions on the use of braces for one-line ifs.

src/shared_realm.cpp
JadenGeller: It doesn't cast to either type, just gives the braced initializer list representation.

JadenGeller: I guess I could provide two overloads though, one for each direction.

JadenGeller: I guess I could do a template function too.

JadenGeller: Ah, the problem is I cannot define a function that operates on `Realm::HandoverPackage::VersionID` since it's a private member. Any solutions? @tgoyne

tests/handover.cpp
JadenGeller: It fails to compile with error "Call to delete constructor of `realm::StringData`" when `.c_str()` is omitted.

tgoyne: Oh, it's not redundant, it's just invalid. c_str() returns a pointer to the buffer owned by the std::string, which doesn't continue to exist past the end of the statement. In the std::string case we're able to detect that you're passing in a temporary and reject that.

JadenGeller: I'm unable to use `ObjectStore::table_for_object_type()` since I cannot obtain the `Group` from a `Realm` since it is a private member.

src/shared_realm.cpp
tgoyne: templated private static function on HandoverPackage?

tests/handover.cpp
tgoyne: read_group() is public and is there specifically because everything that uses the object store needs to be able to call the ObjectStore functions on a Realm's group.

src/shared_realm.cpp
JadenGeller: If I could define this conversion on `HandoverPackage`, it'd probably define implicit conversions on `VersionID`. I can't put anything about `SharedGroup::VersionID` in the header though, so I'm a bit out of luck.

JadenGeller: As far as I can tell, a template function that does the equivalent of this macro is also not useful because the compiler is unable to infer the template arguments.

src/list.hpp
JadenGeller: Thoughts on making `AnyThreadConfined` a friend of `List`?

src/handover.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+        case Type::List:
+            return m_list.get_realm();
+
+        case Type::Results:
+            return m_results.get_realm();
+    }
+}
+
+AnyHandover AnyThreadConfined::export_for_handover() const
```
JadenGeller: There might be potential to reduce copying here a bit. For example, we could `std::move` the `Query` and `Row` objects maybe (if we gave the function move semantics). We might also benefit from using `MutableSourcePayload::Move` since we already have a copy of the `Query` (I assume there's no shared state). I'm not sure how the performance of `std::move` vs. `MutableSourcePayload::Move` would differ though. Thoughts? @tgoyne

JadenGeller:
@tgoyne @bdash I realized that a bad version exception was thrown when a handover package was imported after its source `Realm` was destroyed. Instead of storing the `RealmCoordinator`, I now store a shared pointer to the source `Realm`, only using it for ownership and to access the `RealmCoordinator`. https://github.com/realm/realm-object-store/pull/116/commits/739178213df24cc8da8de28860530eece24492ea

JadenGeller:
@tgoyne @bdash Can you check out the most recent commits? I'm hoping to merge this in soon.

Major changes:
* `HandoverPackage` now stores a shared pointer to the source `Realm` so that the version stays pinned even if all other shared pointers to the `Realm` are destroyed.
* `ObjectSchema` is now handed over by name instead of by pointer since each `Realm` has its own thread-confined instance.

src/handover.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+#include <realm/group_shared.hpp>
+#include <realm/link_view.hpp>
+#include <realm/query.hpp>
+#include <realm/row.hpp>
+#include <realm/table_view.hpp>
+
+namespace realm {
+class AnyHandover;
+class AnyThreadConfined;
```
bdash: No need to forward declare `AnyThreadConfined`.

src/handover.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+AnyThreadConfined AnyHandover::import_from_handover(SharedRealm realm) &&
+{
+    SharedGroup& shared_group = Realm::Internal::get_shared_group(*realm);
+    switch (m_type) {
+        case AnyThreadConfined::Type::Object: {
+            auto row = shared_group.import_from_handover(std::move(m_object.row_handover));
+            auto object_schema = realm->schema().find(m_object.object_schema_name);
+            REALM_ASSERT_DEBUG(object_schema != realm->schema().end());
+            return AnyThreadConfined(Object(std::move(realm), std::move(*object_schema), std::move(*row)));
```
tgoyne: `object_schema` should not be moved here. It happens to be pointless here (`Object` takes a const ref), but if it wasn't pointless it'd be stealing the Realm's ObjectSchema.

src/shared_realm.cpp
```diff
@@ -554,5 +555,107 @@ util::Optional<int> Realm::file_format_upgraded_from_version() const
...
+    if (is_in_transaction()) {
+        throw InvalidTransactionException("Cannot package handover during a write transaction.");
+    }
+
+    HandoverPackage handover;
+    auto version_id = m_shared_group->pin_version();
+    handover.m_version_id = VERSION_CAST(version_id);
+    handover.m_source_realm = shared_from_this();
+    // Since `m_coordinator` is used to determine if we need to unpin when destroyed,
+    // `m_coordinator` should only be set after `pin_version` succeeds in case it throws.
```
tgoyne: m_coordinator isn't set anywhere in this function, so it seems unlikely that this comment is accurate.

JadenGeller: It needs to be updated to `m_source_realm` after https://github.com/realm/realm-object-store/pull/116/commits/580bc86d57ac5ad7290e64c5cadbffd99101be3f. Good catch.

src/shared_realm.cpp
```diff
@@ -554,5 +555,107 @@ util::Optional<int> Realm::file_format_upgraded_from_version() const
...
+
+std::vector<AnyThreadConfined> Realm::accept_handover(Realm::HandoverPackage handover)
+{
+    REALM_ASSERT(handover.is_awaiting_import()); // Enforced by move semantics
+    auto unpin_version = util::make_scope_exit([&]() noexcept {
+        m_shared_group->unpin_version(VERSION_CAST(handover.m_version_id));
+        handover.mark_not_awaiting_import();
+    });
+
+    verify_thread();
```
tgoyne: Thread check needs to be before unpin_version is set up, as unpinning the version isn't thread-safe.

src/handover.hpp
```diff
@@ -0,0 +1,120 @@
...
+    Type m_type;
+    union {
+        Object  m_object;
+        List    m_list;
+        Results m_results;
+    };
+};
+
+// Type-erased wrapper for a `Handover` of an `AnyThreadConfined` value
+class AnyHandover {
```
bdash: `AnyHandover` is only used within the implementation so it should be possible to move it into its own header in the `impl` directory, and move the type within the `realm::_impl` namespace. That will leave only `AnyThreadConfined` in a public header to be used by the bindings.

tgoyne:
It'd be good to test importing multiple handover packages with different source versions into a single realm, as well as multiple types of handed-over objects in a single package.

src/impl/handover.hpp
```diff
@@ -0,0 +1,82 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#ifndef realm_handover_hpp
```
bdash: Object store uses all-uppercase for include guards.

src/impl/handover.cpp
```diff
@@ -0,0 +1,102 @@
...
+        case AnyThreadConfined::Type::List:
+            m_list.link_view_handover = std::move(handover.m_list.link_view_handover);
+            break;
+
+        case AnyThreadConfined::Type::Results:
+            m_results.query_handover = std::move(handover.m_results.query_handover);
+            m_results.sort_order = std::move(handover.m_results.sort_order);
+            break;
+    }
+    m_type = std::move(handover.m_type);
```
bdash: No reason to move a primitive type.

src/impl/handover.cpp
```diff
@@ -0,0 +1,102 @@
...
+            new (&m_results.query_handover) QueryHandover(std::move(handover.m_results.query_handover));
+            new (&m_results.sort_order) SortOrder(std::move(handover.m_results.sort_order));
+            break;
+    }
+    new (&m_type) AnyThreadConfined::Type(handover.m_type);
+}
+
+AnyHandover& AnyHandover::operator=(AnyHandover&& handover)
+{
+    switch (handover.m_type) {
```
bdash: This implementation doesn't appear to handle destroying the existing state before it initializes the new state.

src/thread_confined.cpp
```diff
@@ -0,0 +1,146 @@
...
+        case Type::Results:
+            new (&m_results) Results(thread_confined.m_results);
+            break;
+    }
+    new (&m_type) Type(thread_confined.m_type);
+}
+
+AnyThreadConfined& AnyThreadConfined::operator=(const AnyThreadConfined& thread_confined)
+{
+    switch (thread_confined.m_type) {
```
bdash: This also doesn't appear to destroy the existing state before initializing the new state.

src/thread_confined.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+        case Type::List:
+            m_list = std::move(thread_confined.m_list);
+            break;
+
+        case Type::Results:
+            m_results = std::move(thread_confined.m_results);
+            break;
+    }
+    m_type = std::move(thread_confined.m_type);
```
bdash: No need to move this either.

src/impl/handover.cpp
JadenGeller: So for both more and copy `operator=`, I should call the destructors for all [relevant] members before I move? Can I instead call `this->~AnyHandover()`?

bdash: You can probably call the destructor then copy construct using placement new.

tests/handover.cpp
```diff
@@ -0,0 +1,455 @@
...
+
+        auto handover_version = get_current_version();
+        auto h = util::make_optional(r->package_for_handover({}));
+        r->begin_transaction(); r->commit_transaction(); // Advance version
+
+        REQUIRE(get_current_version() != handover_version); // Ensure advanced
+        REQUIRE_NOTHROW(shared_group.begin_read(handover_version)); shared_group.end_read(); // Ensure pinned
+
+        SECTION("destroyed without being imported") {
+            h = {}; // Destory handover, unpinning version
```
bdash: "Destory"

src/shared_realm.hpp
```diff
@@ -201,6 +205,43 @@ class Realm : public std::enable_shared_from_this<Realm> {
...
+        friend std::vector<AnyThreadConfined> Realm::accept_handover(Realm::HandoverPackage handover);
+
+        struct VersionID { // SharedGroup::VersionID without including header
+            uint_fast64_t version;
+            uint_fast32_t index;
+        };
+
+        VersionID m_version_id;
+        std::vector<_impl::AnyHandover> m_objects;
+        SharedRealm m_source_realm; // Strong reference keeps alive so version stays pinnned! Don't touch!!
```
bdash: "pinnned"

src/shared_realm.cpp
```diff
@@ -554,5 +558,109 @@ util::Optional<int> Realm::file_format_upgraded_from_version() const
...
+    config.cache = false;
+    SharedRealm realm = coordinator.get_realm(config);
+    REALM_ASSERT(!realm->is_in_read_transaction());
+    realm->m_group = &const_cast<Group&>(realm->m_shared_group->begin_read(m_version_id));
+
+    // Import handover, advance version, and then repackage for handover
+    auto objects = realm->accept_handover(std::move(*this));
+    transaction::advance(*realm->m_shared_group, realm->m_binding_context.get(),
+                         realm->m_config.schema_mode, new_version);
+    *this = realm->package_for_handover(objects);
```
bdash: Move `objects` here?

bdash:
I'm really happy with how this has evolved! 👍

src/shared_realm.cpp
```diff
@@ -554,5 +558,109 @@ util::Optional<int> Realm::file_format_upgraded_from_version() const
...
+    }
+
+    return handover;
+}
+
+std::vector<AnyThreadConfined> Realm::accept_handover(Realm::HandoverPackage handover)
+{
+    verify_thread();
+
+    REALM_ASSERT(handover.is_awaiting_import()); // Can only be imported once
```
jpsim: Shouldn't this be an exception since it should only happen in case of user error?

JadenGeller: Do we always use exceptions, not assertions, for misuses of the Object Store API?

jpsim: Unless you have a duplicate check in the binding for this which also throws an exception, it'd be preferable to just keep the check in one place, no?

jpsim: Assuming this can be triggered by a user of the binding.

JadenGeller: There is a duplicate check in the binding currently, but I'll just make this throw.

tgoyne:
👍

src/shared_realm.cpp
JadenGeller: @jpsim Please review https://github.com/realm/realm-object-store/pull/116/commits/31b13009ef92da8f733d9fca1dbd28f2473b70d3 and give me a thumbs up if it's good to go!

src/shared_realm.hpp
```diff
@@ -384,6 +384,11 @@ class InvalidEncryptionKeyException : public std::logic_error {
 public:
     InvalidEncryptionKeyException() : std::logic_error("Encryption key must be 64 bytes.") {}
 };
+
+class InvalidDuplicateHandoverException : public std::logic_error {
```
bdash: I don't think we need a distinct exception class for this unless we expect it to require special handling in the bindings. Using `std::logic_error` directly is fine.

jpsim:
👍

PR #118, opened 2016-08-03
Pull in remaining changes from JavaScript

bdash:
This brings in the remaining handful of changes from realm-js's copy of the object store code. It'll allow realm-js to pull in object store's master branch directly.

### DIFF
--- a/src/binding_context.hpp
+++ b/src/binding_context.hpp
@@ -80,7 +80,7 @@ public:
 
     struct ObserverState;
 
-    // Override this function if you want to recieve detailed information about
+    // Override this function if you want to receive detailed information about
     // external changes to a specific set of objects.
     // This is called before each operation which may advance the read
     // transaction to include
--- a/src/impl/android/external_commit_helper.cpp
+++ b/src/impl/android/external_commit_helper.cpp
@@ -23,6 +23,7 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <sstream>
+#include <stdlib.h>
 #include <sys/epoll.h>
 #include <sys/time.h>
 #include <sys/stat.h>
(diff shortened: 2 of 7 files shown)


tgoyne:
I don't understand what exact the thread_id thing is trying to solve. Our instances aren't *actually* tied to the thread in any way, so reusing an instance from a previous thread that happens to have the same ID shouldn't be a problem (and if it was reasonably possible, intentionally reusing instances from dead threads could be useful). Adding a TLS lookup to every call to verify_thread() (i.e. every access of a property) has a pretty significant performance impact.

More superficially, it should be named ThreadID.

bdash:
There's not a lot of information in the original PR to explain why the change was made (realm/realm-js#480). From looking at the realm-js code, I think the problem was related to realm-js's assumption that there's a single JavaScript context per thread. `RealmClass<T>::constructor` stashes the JavaScript context in `Realm::m_binding_context`. If the thread a `Realm` instance belongs to exits and its identifier is reused before the `Realm` instance is deallocated, a new thread calling `RealmClass<T>::constructor` can get back the existing `Realm` instance, and will find that it is associated with a different JavaScript context than it expects.

Changing the definition of thread ID doesn't seem like the right way to fix that issue.

bdash:
My initial thought for addressing the problem would be for realm-js to have the `Realm` instance cache be partitioned by `(thread, JavaScript Context)` rather than only by thread.

alazier:
Why not just use the JSContext? Those are always disposed/recreated between app runs and refreshes.

bdash:
I'm going to file a new issue in realm-js to track revisiting this thread ID stuff, and will drop the relevant commit from this branch. That will allow us to get the remaining changes merged while we work out a better fix than changing the concept of thread ID.

alazier:
Sounds good @bdash - there is already a ticket https://github.com/realm/realm-js/issues/494 to revisit the design for the threadlocal stuff which I was not really happy with to begin with. Could just modify that one.

bdash:
Wrote up realm/realm-js#552.

tgoyne:
👍

PR #119, opened 2016-08-03
Fix warning about toString being unused when building tests

bdash:
The overload of `toString` we were providing wasn't being picked up by Catch, and so was never being called. Switch to their suggested alternative approach of providing a specialization of `StringMaker`. This specialization is correctly picked up by Catch.

/cc @tgoyne


tgoyne:
👍

Issue #117, opened 2016-08-03
`shared_from_this()` is potentially unsafe with public constructors


JadenGeller:
*Problem*: Calling `shared_from_this` produces undefined behavior unless the object is managed by a `shared_ptr`. We don't have any type-level enforcement that a `Realm` is always `shared_ptr` managed.

*Solution*: Make constructors private and provide static member functions that mirror the constructor behavior returning a`shared_ptr` to the object.

Update: @bdash pointed out that the copy/move constructor/assignment operator should also be deleted.

*Problem*: We cannot use `make_shared` if the constructor is private, and `make_shared` is a significant performance optimization.

*Solution*:
```c++
static std::shared_ptr<Realm> create(Configuration config) {
	struct make_shared_enabler : public Realm {
		make_shared_enabler(Configuration config) : Realm(config) {}
	};
	return std::make_shared<make_shared_enabler>(config);	
}
```

PR #121, opened 2016-08-04
Ensure manipulation of `Realm` must go through a `SharedRealm`

JadenGeller:
Addresses https://github.com/realm/realm-object-store/issues/117 //cc @bdash @tgoyne

### DIFF
--- a/src/shared_realm.hpp
+++ b/src/shared_realm.hpp
@@ -196,9 +196,18 @@ public:
     // returns the file format version upgraded from if an upgrade took place
     util::Optional<int> file_format_upgraded_from_version() const;
 
+    Realm(const Realm&) = delete;
+    Realm& operator=(const Realm&) = delete;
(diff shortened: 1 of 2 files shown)


src/shared_realm.hpp
```diff
@@ -196,9 +196,18 @@ class Realm : public std::enable_shared_from_this<Realm> {
...
+    Realm(const Realm&) = delete;
+    Realm& operator=(const Realm&) = delete;
+    Realm(Realm&&) = delete;
+    Realm& operator=(Realm&&) = delete;
     ~Realm();
 
-    Realm(Config config);
+    static SharedRealm make_shared_realm(Config config) {
+        struct make_shared_enabler : public Realm {
+            make_shared_enabler(Config config) : Realm(config) {}
```
tgoyne: Need to std::move() the config both here and below.

PR #124, opened 2016-08-08
Fix incorrect results when removing multiple properties from a single object

tgoyne:
Removing a column invalidates the cached table columns for all columns after
that, so group removes after all other changes and sort them in descending
order of table column to avoid ever invalidating a yet-to-be-applied change.

See https://github.com/realm/realm-cocoa/issues/3943

@bdash

### DIFF
--- a/src/schema.cpp
+++ b/src/schema.cpp
@@ -88,6 +88,19 @@ void Schema::validate() const
     }
 }
 
+namespace {
+struct IsNotRemoveProperty {
+    bool operator()(SchemaChange sc) const { return sc.visit(*this); }
+    bool operator()(schema_change::RemoveProperty) const { return false; }
+    template<typename T> bool operator()(T) const { return true; }
+};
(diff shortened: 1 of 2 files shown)


src/schema.cpp
```diff
@@ -88,6 +88,19 @@ void Schema::validate() const
     }
 }
 
+namespace {
+struct IsNotRemoveProperty {
+    bool operator()(SchemaChange sc) const { return sc.visit(*this); }
```
bdash: Is `SchemaChange` a cheap enough type to copy that it's not worth taking by reference?

tgoyne: It's three pointers and an int, so with zero inlining it'd probably be faster as a reference but only very marginally so.

bdash: Ok. I only ask because I see it being taken by reference in other places when used as an argument.

bdash:
👍

PR #125, opened 2016-08-09
Add RealmFriend for java-binding and fix other build issues

beeender:
Ideally Java should use the async query logic from OS, but since the
behaviours are not exact the same right now, Java will keep the current
async implementation for a bit while.

Expose SharedGroup and Group pointer by RealmFriend to make java's async
query work.

### DIFF
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -446,7 +446,9 @@ bool Realm::compact()
 
 void Realm::write_copy(StringData path, BinaryData key)
 {
-    REALM_ASSERT(!key.data() || key.size() == 64);
+    if ((key.data() && key.size() != 64)) {
+        throw InvalidEncryptionKeyException();
+    }
     verify_thread();
     try {
         read_group().write(path, key.data());
--- a/src/shared_realm.hpp
+++ b/src/shared_realm.hpp
@@ -26,6 +26,8 @@
 #include <memory>
 #include <thread>
 
+#include <realm/group_shared.hpp>
+
 namespace realm {
 class BinaryData;
 class BindingContext;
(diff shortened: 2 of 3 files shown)


src/shared_realm.cpp
```diff
@@ -446,7 +446,9 @@ bool Realm::compact()
 
 void Realm::write_copy(StringData path, BinaryData key)
 {
-    REALM_ASSERT(!key.data() || key.size() == 64);
+    if ((key.data() && key.size() != 64)) {
```
bdash: There's an unnecessary set of parens here.

src/shared_realm.hpp
```diff
@@ -26,6 +26,8 @@
 #include <memory>
 #include <thread>
 
+#include <realm/group_shared.hpp>
```
bdash: This is unfortunate as we've been trying to avoid pulling core headers into object store headers where possible.

src/shared_realm.hpp
```diff
@@ -328,6 +333,24 @@ class InvalidEncryptionKeyException : public std::logic_error {
 public:
     InvalidEncryptionKeyException() : std::logic_error("Encryption key must be 64 bytes.") {}
 };
+
+// FIXME Those are exposed for Java async queries, mainly because of handover related methods.
+class _impl::RealmFriend {
```
bdash: The `_impl` namespace is internal to the object store, while this class is explicitly intended to be exposed. Those two don't go together.

bdash:
@beeender, it'd be preferable to separate build fixes from changes of substance in the future so build fixes can be landed without delay.

It'd be great if you could look at @JadenGeller's work on exposing a handover API from object store (#116) to see if it'll fit Java's needs. If it looks like it'll be a good fit it may be preferable to keep the change to expose`Realm` methods on a Java branch of the object store until you've been able to  adopt enough of the object store API to use the new handover API.

beeender:
See #128 and #127 
Others stay in java branch for now.

PR #126, opened 2016-08-09
Set up notifier immediately after creating results table view

JadenGeller:
We previously did not set up the notifier until `update_tableview` was called while `m_mode` was `Mode::TableView`. This led to us not creating the notifier until the second access (since `m_mode` is query on the first access when the table view is created). This PR corrects this buggy behavior to what was intended.

Note that this fall through leads to `sync_if_needed` being unnecessarily called, but it's supposedly cheap. We could duplicate the notifier setup code in the query case if we're worried about this.

//cc @bdash @tgoyne

### DIFF
--- a/src/results.cpp
+++ b/src/results.cpp
@@ -267,7 +267,7 @@ void Results::update_tableview(bool wants_notifications)
                 m_table_view.sort(m_sort);
             }
             m_mode = Mode::TableView;
-            break;
+            REALM_FALLTHROUGH;
         case Mode::TableView:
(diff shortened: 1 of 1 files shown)


bdash:
The code change looks fine. @tgoyne, can you comment as to whether it was intentional to defer creating the notifier in this case? I dug around the PR that added this code and couldn't find any confirmation one way or the other.

tgoyne:
I had thought that it was being created on the first access that required a TV, and I don't remember ever thinking that it should only be on the second. I guess it'd be worth making sure that the cocoa notification tests still pass with this change, but assuming they do 👍 .

JadenGeller:
Seems to pass after a rebase! 💃

PR #129, opened 2016-08-10
Update to core 1.5.0

tgoyne:
@bdash

### DIFF
--- a/src/results.cpp
+++ b/src/results.cpp
@@ -33,14 +33,13 @@ using namespace realm;
 Results::Results() = default;
 Results::~Results() = default;
 
-Results::Results(SharedRealm r, Query q, SortOrder s)
+Results::Results(SharedRealm r, Query q, SortDescriptor s)
 : m_realm(std::move(r))
 , m_query(std::move(q))
 , m_table(m_query.get_table().get())
 , m_sort(std::move(s))
 , m_mode(Mode::Query)
 {
-    REALM_ASSERT(m_sort.column_indices.size() == m_sort.ascending.size());
 }
 
 Results::Results(SharedRealm r, Table& table)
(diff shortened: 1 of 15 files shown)


tests/results.cpp
```diff
@@ -320,7 +320,7 @@ TEST_CASE("results: notifications") {
     }
 
     // Sort in descending order
-    results = results.sort({{0}, {false}});
+    results = results.sort({*table, {{0}}, {false}});
```
bdash: It's a little unfortunate that this change requires callers of `sort` to restate the table, when both `Results` and `List` already know it.

src/results.cpp
```diff
@@ -484,9 +481,8 @@ TableView Results::get_tableview()
     REALM_UNREACHABLE();
 }
 
-Results Results::sort(realm::SortOrder&& sort) const
+Results Results::sort(realm::SortDescriptor&& sort) const
 {
-    REALM_ASSERT(sort.column_indices.size() == sort.ascending.size());
```
bdash: Is it worth asserting that the sort descriptor's table is correct here?

bdash:
Unfortunate that `SortDescriptor` is a little less ergonomic for the bindings to use, but it's nice to not have to reinvent it in the object store.

👍

tests/results.cpp
tgoyne: It's definitely gross in the tests, but obj-c was using the table when constructing SortDescriptors anyway, so a convenience wrapper wouldn't help it. Might be worth adding something for the other bindings, though.

src/results.cpp
tgoyne: That should probably be done *somewhere*, but the SortDescriptor currently doesn't actually know what table it's for. Maybe I can make it check on use at least.

JadenGeller:
If I'm understanding correctly, constructor of `SortDescriptor` is gross because you have to state the `Table` even if the table is evident by context. Maybe it's worth introducing a `TransientSortDescriptor` (that stores column indices) that is taken as an argument by member functions of types associated with a table.

JadenGeller:
Something like this is what I had in mind. I'm not sure it's an API improvement though. Duplicate types is kind of gross. https://github.com/realm/realm-object-store/commit/c02591e5f5d9da71a035bd917a9f71c0d4ef21d8

JadenGeller:
At that point, it almost makes sense to skip introducing a new type and just make `sort` take the column indices and ascending vectors as arguments.

PR #130, opened 2016-08-10
Fix include path for handover.hpp

tgoyne:
All include paths need to be relative to `src` for non-Xcode/CMake builds to work.


bdash:
👍

JadenGeller:
👍

PR #131, opened 2016-08-11
Reapply the change from #97

kristiandupont:
..which was lost somewhere in the schema updates.

### DIFF
--- a/src/object_store.cpp
+++ b/src/object_store.cpp
@@ -250,7 +250,7 @@ StringData ObjectStore::object_type_for_table_name(StringData table_name) {
 }
 
 std::string ObjectStore::table_name_for_object_type(StringData object_type) {
-    return std::string(c_object_table_prefix) + static_cast<std::string>(object_type);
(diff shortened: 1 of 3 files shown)


bdash:
Any chance that you can add a test so that this can't break in the same way in the future?

src/object_store.cpp
```diff
@@ -250,7 +250,7 @@ StringData ObjectStore::object_type_for_table_name(StringData table_name) {
 }
 
 std::string ObjectStore::table_name_for_object_type(StringData object_type) {
-    return std::string(c_object_table_prefix) + object_type.data();
+    return std::string(c_object_table_prefix) + static_cast<std::string>(object_type);
```
bdash: Seems odd to use `static_cast<std::string>(…)` when we're doing`std::string(…)` on the same line of code.

kristiandupont: Good point, I'll align that.

kristiandupont:
@bdash is this ok?

bdash:
👍

Please squash down to one commit prior to merging.

PR #133, opened 2016-08-19
Extract the event loop functionality from WeakRealmNotifier

tgoyne:
This makes it possible to reuse it for other things and cuts down on the duplicated code a bit.

@bdash


src/util/android/event_loop_signal.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+#define LOGE(fmt...) do { \
+    fprintf(stderr, fmt); \
+    __android_log_print(ANDROID_LOG_ERROR, "REALM", fmt); \
+} while (0)
+
+namespace realm {
+namespace util {
+template<typename Callback>
+class EventLoopSignal : std::enable_shared_from_this() {
```
bdash: I think this should be `std::enable_shared_from_this<EventLoopSignal<Callback>>`?

bdash:
👍

PR #134, opened 2016-08-22
Add partial Android build system support

tgoyne:
Only supports targeting x86_64 with a prebuilt core binary, and while it
compiles the tests, it doesn't address finding a way to actually run them.

The linker changes for the tests are due to that the android object-store
library doesn't re-export the core symbols, so the tests need to explicitly
link core. All of the code changes are fixes for GCC 4.9 compatibility.

The toolchain file is the same one as is used by realm-java.


bdash:
👍

We should get CI set up to build for Android once this lands so we can keep things in a good state.

bmunkholm:
@emanuelez: Will you create an issue for that?

PR #135, opened 2016-08-22
Add support for building with libuv rather than the native event loop

tgoyne:
Mostly just for the sake of being able to actually test it.


src/util/node/event_loop_signal.hpp
```diff
@@ -30,8 +29,6 @@ class EventLoopSignal {
 
         // This assumes that only one thread matters: the main thread (default loop).
         uv_async_init(uv_default_loop(), m_handle, [](uv_async_t* handle) {
-            // The v8::Local handles need a "scope" to be present or will crash.
-            Nan::HandleScope scope;
```
tgoyne: The rational for this removal is that it's v8-specific while the rest of this file works with libuv on any platform, so it should be done specifically in the JS binding.

bdash: Should the directory this file is in be renamed from `node` to `uv` to better reflect that?

bdash:
👍

bmunkholm:
@tgoyne Who else needs to review this if any? Please ping them or merge.
/cc @jpsim

PR #136, opened 2016-08-23
Fix running the tests on OS X with locally-built core

bdash:
Don't link the core static library directly into the tests binary. This results in two copies of core being present (one in tests, one in librealm-object-store.dylib), causing havoc.

This was broken by #134. I don't have a working Android setup at the moment so I cannot tell what impact this fix has on the work done in #134.

/cc @tgoyne

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -21,7 +21,7 @@ set(SOURCES
 )
 
 add_executable(tests ${SOURCES} ${HEADERS})
-target_link_libraries(tests realm-object-store realm ${PLATFORM_LIBRARIES})
+target_link_libraries(tests realm-object-store ${PLATFORM_LIBRARIES})
 
 create_coverage_target(generate-coverage tests)
 


bdash:
This didn't appear to fail when using a prebuilt release of core. I suspect that's more due to luck than anything though.

tgoyne:
The realm-object-store library for Android doesn't re-export the core symbols, so there's a bunch of undefined symbols otherwise. Even with the symbols being exported linking it again shouldn't cause any problems as long as all of the settings are identical, since if no symbols are actually needed from the static library nothing will be imported from it.

bdash:
As-is you end up with duplicate `typeinfo` symbols being present in both `tests` and `librealm-object-store.dylib`, causing `dynamic_cast`s to fail at runtime.

bdash:
A better solution may be to switch the tests to linking to the object store static library. That will ensure that only a single copy of the `typeinfo` symbols is present, while also working for Android where core is built with `-fvisibility=hidden`.

bdash:
```
$ nm -m tests/tests | c++filt | grep 'typeinfo for realm::Array$'
000000010080b0b0 (__DATA,__const) external typeinfo for realm::Array
$ nm -m src/librealm-object-store.dylib | c++filt | grep 'typeinfo for realm::Array$'
0000000000895000 (__DATA,__const) external typeinfo for realm::Array
```

I'm not sure _why_ `tests` gets its own definition of that symbol rather than referencing the one provided by librealm-object-store.dylib. This happens for a handful of symbols from core:

```
$ nm -m tests/tests | c++filt | grep '__DATA,__const.*external typeinfo for realm::'
000000010080b230 (__DATA,__const) external typeinfo for realm::TrivialReplication
000000010080b0b0 (__DATA,__const) external typeinfo for realm::Array
000000010080b160 (__DATA,__const) external typeinfo for realm::Table::Parent
```

The end result is that a `dynamic_cast<Table::Parent*>(…)` in `Table::get_parent_ptr` fails unexpectedly when trying to cast a `Group*` to a `Parent*`, which should always succeed as `Table::Parent` is a base class of `Group`.

tgoyne:
`tests` has its own copy of the vtable for all of the core types defined entirely within headers that are imported and used within the tests. This is the sort of thing that -Wweak-vtables is there to help you avoid, but making core anchor all of its vtables was going to be a very large amount of work.

I'm guessing it works when not linking core because the linker doesn't see a definitive definition of the vtable and leaves them as weak symbols, but when linking against core it converts them to normal exported symbols?

bdash:
> tests has its own copy of the vtable for all of the core types defined entirely within headers that are imported and used within the tests. This is the sort of thing that -Wweak-vtables is there to help you avoid, but making core anchor all of its vtables was going to be a very large amount of work.

This would all be fine if the vtable _was_ weak. It has normal linkage in both `tests` and `librealm-object-store.dylib`.

When not linking against core the symbol is instead referenced from librealm-object-store.dylib:

```
$ nm -m tests/tests | c++filt | grep -E 'typeinfo for realm::(Array|TrivialReplication|Table::Parent) '
                 (undefined) external typeinfo for realm::Array (from librealm-object-store)
                 (undefined) external typeinfo for realm::Table::Parent (from librealm-object-store)
```

Though weirdly enough the reference to `TrivialReplication`'s type info disappears entirely.

bdash:
I poked at this for a while and figured out what's going on. With the OS X SDK in Xcode 8, `std::basic_ostream` is no longer declared as `extern template` when building with a deployment target lower than 10.10. This results in weak, hidden definitions of the relevant symbols being emitted into `librealm-dbg.a`. When `librealm-dbg.a` is linked into `librealm-object-store.dylib`, these `std::basic_ostream` symbols become internal symbols. When `ld` is linking `tests`, it searches `librealm-object-store.dylib` for the `std::basic_ostream` symbols but does not find them, so it proceeds to search `librealm-dbg.a`, where it _does_ find them. This results in it loading a handful of objects that provide the definitions (`array.dbg.o`, `replication.dbg.o`, and `table.dbg.o` being the relevant ones), bringing some typeinfo / vtable symbols along for the ride. The end result is that objects of the same type can end up with different type IDs depending on whether they're created from within `tests` or `librealm-object-store.dylib`.

I think the solution here is to just kill the object store shared library. I created it initially to catch link errors when building the object store by itself. That's no longer necessary as they'll be caught when building the `tests` executable.

PR #137, opened 2016-08-23
Accommodate API changes in newer version of core

bdash:
An upcoming version of core has some small API differences. For now we want to support building with both the currently-released version (1.5.x) and the upcoming version (2.x), so we key off the `REALM_VER_MAJOR` define to use the appropriate code path.


tgoyne:
👍

PR #140, opened 2016-08-24
Fix build error, extra ;

beeender:


bdash:
It's helpful to mention which compiler version and platform the problem you're fixing was seen with.

👍

bdash:
CI: Retest this.

beeender:
Commit message updated.

PR #143, opened 2016-08-24
workaround VS ternary expression compilation bug

AndyDentFree:
@bdash, @kristiandupont or @fealebenpae  please review.
I think the problem lies in the handling of ternary expressions - I was unable to get any variation on them to compile for this `StringData` creation.

### DIFF
--- a/src/object_store.cpp
+++ b/src/object_store.cpp
@@ -457,7 +457,20 @@ static void create_initial_tables(Group& group, std::vector<SchemaChange> const&
         void operator()(RemoveProperty op) { table(op.object).remove_column(op.property->table_column); }
         void operator()(MakePropertyNullable op) { make_property_optional(group, table(op.object), *op.property); }
         void operator()(MakePropertyRequired op) { make_property_required(group, table(op.object), *op.property); }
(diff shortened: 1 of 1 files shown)


tgoyne:
`op.property ? StringData{op.property->name} : ""` works on the [online compiler](http://webcompiler.cloudapp.net/) but I don't know if that's the same compiler version.

AndyDentFree:
As I said in the original issue description, I think the problem was partly due to matching up `std::string` vs `template<class T, class A> StringData(const std::basic_string<char, T, A>&);` because I couldn't get the ternary expression to work without the error message.

As a general comment, VS in its various incarnations over the last N years has **always** been the most _creative_ in the way it interpreted which overloads to use, out of a very wide range of compilers I have built with. It is vastly better nowadays but this little problem was almost pleasurably nostalgia-inducing.
(and a little nausea-inducing)

kristiandupont:
:+1:

bdash:
@AndyDentFree, this would have been a candidate for squashing your changes when merging or rebasing to clean up the commit history, as the intermediate state commits are just noise.

PR #144, opened 2016-08-24
Update js with latest master

fealebenpae:
/cc @bdash @alazier


alazier:
👍 assuming realm-js still compiles and tests pass with these changes.

fealebenpae:
@alazier there will have to be a couple of changes to realm-js to accommodate for object store changes, but that will have to happen in the realm-js PR that updates the object store submodule.

alazier:
sounds good lets merge 👍

Issue #142, opened 2016-08-24
Add schema change handler after handover begin read 


JadenGeller:
After I started implementing handover, this line 
https://github.com/realm/realm-object-store/commit/f0024cca6f0def4c6859a3381e96c276875bbf5e#diff-55a15b0c9125ec5f7989470e6eb8b217R156 was added to the codebase as part of beginning a transaction. I think handover ought to also call this function upon beginning a transaction.

PR #145, opened 2016-08-24
Add schema change handler after handover begin read

JadenGeller:
https://github.com/realm/realm-object-store/issues/142
@bdash @tgoyne


tgoyne:
👍

PR #148, opened 2016-08-29
Fix incorrect row index updating for List notifications

tgoyne:
Non-end row insertions and move_last_over() can move the row containing a List being observed, which was not being handled correctly.


jpsim:
This looks right to me, but I can't easily test it again due to changes in the app I was using to trigger this issue.

PR #149, opened 2016-08-30
Switch to building the object store as a static library only

bdash:
The presence of the dynamic library predates the object store having its own test suite, and was used to catch link errors due to bugs in object store or missing symbols in core. It is no longer needed for this purpose now that object store has a proper test suite.

Removing the dynamic library also works around some linker problems that reveal themselves when using a version of core built with Xcode 8. See https://github.com/realm/realm-object-store/pull/136#issuecomment-242897308 for the gory details.

/cc @tgoyne

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,8 +11,7 @@ CMakeScripts/
 realm-object-store.xcodeproj/
 
 # Build products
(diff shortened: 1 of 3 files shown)


tgoyne:
👍

PR #150, opened 2016-08-30
Update object store to support Core v2.*

austinzheng:
@bdash @tgoyne @jpsim 

All tests pass.

### DIFF
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -30,7 +30,12 @@
 
 #include "util/format.hpp"
 
+#if REALM_VER_MAJOR >= 2
+#include <realm/history.hpp>
+#else
 #include <realm/commit_log.hpp>
+#endif
+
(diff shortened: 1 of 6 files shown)


src/shared_realm.cpp
```diff
@@ -30,7 +30,12 @@
 
 #include "util/format.hpp"
 
+#if REALM_VER_MAJOR < 2
```
bdash: In these situations I typically prefer to structure the condition so that the code for the new version is first, with the older code second. It makes it a little easier to skim just the _current_ code.

bdash:
👍

src/shared_realm.cpp
austinzheng: I agree. Let me flip the conditions around.

jpsim:
👍

austinzheng:
I switched around all the conditional statements. Once the CI passes I'll merge.

PR #151, opened 2016-08-30
Handle Lists moving due to unique key reuse

tgoyne:
Requires https://github.com/realm/realm-core/pull/2093 to actually work in a useful way.

All other notification types also have issues related to objects being subsumed which need to be addressed separately.


bdash:
Is the test failure due to the dependency on realm/realm-core#2093?

tgoyne:
Yes.

bdash:
👍

PR #155, opened 2016-09-02
update shared_realm.cpp for new SharedGroupOptions

jpsim:
/cc @bdash

### DIFF
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -151,7 +151,11 @@ void Realm::open_with_config(const Config& config,
         else {
 #if REALM_VER_MAJOR >= 2
             history = realm::make_in_realm_history(config.path);
+#else
+            history = realm::make_client_history(config.path, config.encryption_key.data());
+#endif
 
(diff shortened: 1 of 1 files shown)


tgoyne:
SharedGroupOptions is also on core master, so checking for >= 2 is wrong. The lack of any useful versioning for the releases from next-major is pretty awkward too since this won't compile with the most recent release and you can't check that...

bdash:
As a temporary measure, we could do something cheeky like key off the presence of the header guard macro for the new header that `SharedGroupOptions` is declared in.

jpsim:
> As a temporary measure, we could do something cheeky like key off the presence of the header guard macro for the new header that `SharedGroupOptions` is declared in.

Good idea, I'll give that a shot.

jpsim:
I updated this to only use sg options if `REALM_GROUP_SHARED_OPTIONS_HPP` is defined. /cc @bdash @fealebenpae

bdash:
@jpsim, it looks like you neglected to fix the same issue in the tests.

jpsim:
Sorry about that. Thanks @austinzheng for fixing.

PR #157, opened 2016-09-07
Replace `HandoverPackage` collection with `ThreadSafeReference`

JadenGeller:
[Changes in the handover API design](https://github.com/realm/realm-cocoa/pull/3946#discussion_r76896594) have made it unnecessary for the Object Store to support passing a collection of handovers. Instead, it simply must be able to pass a single handover, now coined a `ThreadSafeReference`. This PR makes the smallest possible set of changes to accomplish this purpose.

EDIT: Another commit was pushed that reworks this PR to use an abstract class instead of a tagged union. This gets rid of all tagged unions, gets rid of some gross unnecessary types (like `AnyThreadConfined`), and makes it possible to use `ThreadSafeReference` in a type-safe way!

@bdash @jpsim @tgoyne


src/shared_realm.hpp
```diff
@@ -247,10 +247,10 @@ class Realm : public std::enable_shared_from_this<Realm> {
         };
 
         VersionID m_version_id;
-        std::vector<_impl::AnyHandover> m_objects;
+        std::unique_ptr<_impl::AnyHandover> m_handover; // FIXME: Remove this extra layer of indirection.
```
JadenGeller: There are a few things that I don't _love_ about this implementation, but I'm not sure how to improve. I don't like that this extra `unique_ptr` indirection is necessary, but removing it requires we drag in HUGE headers. I also don't love that we've kept `AnyHandover` and `AnyThreadConfined`, robbing the Object Store API of type safety. It would be better to use a template here, BUT without the extra indirection provided by `AnyHandover`, we'd have to drag in HUGE headers. Also, this would simply place more burden on the binding; the Objective-C API would have to use an abstract class with concrete subclasses for each handover type. I think this is a better design, but it's problematic in the ways previously discussed.

JadenGeller:
Note that this PR doesn't provide any optimizations with regards to handing over many objects. @tgoyne believes the speed gains are probably not worth the implementation complexity for the first iteration of this API, but it'd be worthwhile to benchmark this. An "easy" optimization would involve providing a single shared realm for advancing handover packages on each `RealmCoordinator` since the most expensive part of handover is not advancing the version, but opening the Realm.

src/thread_safe_reference.cpp
```diff
@@ -0,0 +1,118 @@
...
+}
+
+ThreadSafeReference<List>::ThreadSafeReference(List list)
+: ThreadSafeReferenceBase(list.get_realm())
+, m_link_view(get_source_shared_group().export_linkview_for_handover(list.m_link_view)) { }
+
+List ThreadSafeReference<List>::import_into_realm(SharedRealm realm) && {
+    return invalidate_after_import<List>(*realm, [&](SharedGroup& shared_group) {
+        LinkViewRef link_view = shared_group.import_linkview_from_handover(std::move(m_link_view));
+        return List(std::move(realm), link_view);
```
tgoyne: Could move `link_view` here as well.

src/shared_realm.cpp
```diff
@@ -585,113 +584,87 @@ util::Optional<int> Realm::file_format_upgraded_from_version() const
...
 
-Realm::HandoverPackage::HandoverPackage(HandoverPackage&&) = default;
-Realm::HandoverPackage& Realm::HandoverPackage::operator=(HandoverPackage&&) = default;
-Realm::HandoverPackage::VersionID::VersionID() : VersionID(SharedGroup::VersionID()) { }
-
-// Precondition: `m_version` is not greater than `new_version`
-// Postcondition: `m_version` is equal to `new_version`
-void Realm::HandoverPackage::advance_to_version(VersionID new_version)
+template <typename T>
+realm::ThreadSafeReference<T> realm::Realm::obtain_thread_safe_reference(T value)
```
tgoyne: Don't need the `realm::` qualifications.

src/thread_safe_reference.hpp
```diff
@@ -0,0 +1,149 @@
...
+    ThreadSafeReferenceBase& operator=(const ThreadSafeReferenceBase&) = delete;
+    ThreadSafeReferenceBase(ThreadSafeReferenceBase&&) = default;
+    ThreadSafeReferenceBase& operator=(ThreadSafeReferenceBase&&) = default;
+    ThreadSafeReferenceBase();
+    virtual ~ThreadSafeReferenceBase();
+
+    bool is_invalidated() const { return m_source_realm == nullptr; };
+
+protected:
+    struct VersionID { // SharedGroup::VersionID without including header
```
tgoyne: This file is including `group_shared.hpp` anyway so this wrapper is unneeded.

src/thread_safe_reference.hpp
```diff
@@ -0,0 +1,149 @@
...
+    SortDescriptor::HandoverPatch m_sort_order;
+
+    // Precondition: The associated Realm is for the current thread and is not in a write transaction;.
+    ThreadSafeReference(Results value);
+
+    // Precondition: Realm and handover are on same version.
+    Results import_into_realm(SharedRealm realm) &&;
+
+    template <typename F>
+    void execute_as_imported(Realm& intermediate_realm, F block);
```
bdash: These `execute_as_imported` methods appear to be unimplemented.

JadenGeller: I removed this code. Guess I forgot the headers.

jpsim:
I've rebased this and addressed @tgoyne and @bdash's feedback. Could either of you take another look at this please?

src/thread_safe_reference.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+#include "thread_safe_reference.hpp"
+
+#include "impl/realm_coordinator.hpp"
+
+#include <realm/util/scope_exit.hpp>
+
+using namespace realm;
+
+ThreadSafeReferenceBase::ThreadSafeReferenceBase(SharedRealm source_realm) : m_source_realm(source_realm)
```
bdash: Should move `source_realm`.

src/thread_safe_reference.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+ThreadSafeReferenceBase::~ThreadSafeReferenceBase()
+{
+    if (!is_invalidated())
+        invalidate();
+}
+
+template <typename V, typename T>
+V ThreadSafeReferenceBase::invalidate_after_import(Realm& destination_realm, T construct_with_shared_group) {
+    REALM_ASSERT_DEBUG(destination_realm.thread_id() == std::this_thread::get_id());
```
bdash: This may want to delegate to `Realm::verify_thread` instead.

src/shared_realm.cpp
```diff
@@ -657,112 +656,87 @@ util::Optional<int> Realm::file_format_upgraded_from_version() const
...
 }
 
-Realm::HandoverPackage::HandoverPackage(HandoverPackage&&) = default;
-Realm::HandoverPackage& Realm::HandoverPackage::operator=(HandoverPackage&&) = default;
-
-// Precondition: `m_version` is not greater than `new_version`
-// Postcondition: `m_version` is equal to `new_version`
-void Realm::HandoverPackage::advance_to_version(VersionID new_version)
+template <typename T>
+realm::ThreadSafeReference<T> Realm::obtain_thread_safe_reference(T value)
```
bdash: I think we should take `value` by const reference here to avoid creating an unnecessary copy.

src/thread_safe_reference.cpp
```diff
@@ -0,0 +1,116 @@
...
+}
+
+void ThreadSafeReferenceBase::invalidate() {
+    REALM_ASSERT_DEBUG(m_source_realm);
+    SharedRealm thread_local_realm = Realm::Internal::get_coordinator(*m_source_realm).get_realm();
+    Realm::Internal::get_shared_group(*thread_local_realm).unpin_version(m_version_id);
+    m_source_realm = nullptr;
+}
+
+ThreadSafeReference<List>::ThreadSafeReference(List list)
```
bdash: Const reference for `list` here, since we're not consuming it?

src/thread_safe_reference.cpp
```diff
@@ -0,0 +1,116 @@
...
+, m_link_view(get_source_shared_group().export_linkview_for_handover(list.m_link_view)) { }
+
+List ThreadSafeReference<List>::import_into_realm(SharedRealm realm) && {
+    return invalidate_after_import<List>(*realm, [&](SharedGroup& shared_group) {
+        LinkViewRef link_view = shared_group.import_linkview_from_handover(std::move(m_link_view));
+        return List(std::move(realm), std::move(link_view));
+    });
+}
+
+ThreadSafeReference<Object>::ThreadSafeReference(Object object)
```
bdash: Ditto about const reference.

src/thread_safe_reference.cpp
```diff
@@ -0,0 +1,116 @@
...
+Object ThreadSafeReference<Object>::import_into_realm(SharedRealm realm) && {
+    return invalidate_after_import<Object>(*realm, [&](SharedGroup& shared_group) {
+        Row row = *shared_group.import_from_handover(std::move(m_row));
+        auto object_schema = realm->schema().find(m_object_schema_name);
+        REALM_ASSERT_DEBUG(object_schema != realm->schema().end());
+        return Object(std::move(realm), *object_schema, row);
+    });
+}
+
+ThreadSafeReference<Results>::ThreadSafeReference(Results results)
```
bdash: Ditto about const reference.

src/thread_safe_reference.cpp
```diff
@@ -0,0 +1,116 @@
...
+: ThreadSafeReferenceBase(results.get_realm())
+, m_query(get_source_shared_group().export_for_handover(results.get_query(), ConstSourcePayload::Copy))
+, m_sort_order([&]() {
+    SortDescriptor::HandoverPatch sort_order;
+    SortDescriptor::generate_patch(results.get_sort(), sort_order);
+        return sort_order;
```
bdash: The indentation here is a bit off.

src/thread_safe_reference.cpp
jpsim: I think the idea here was that since this is an internal API, that we shouldn't verify the thread in release builds, and that we would catch any API misuse in debug builds/tests.

jpsim:
@tgoyne do you think you could review this?

bdash:
Despite there being no conflicts with master, this currently fails to build when merging with master. Please be sure to address those build failures before merging.

jpsim:
> this currently fails to build when merging with master

This was due to the changes from #297. I fixed this in 96bc79e, at the expense of also performing the check in release builds. Though that shouldn't be too costly.

jpsim:
@tgoyne I've incorporated the changes you made, merged back with `master` and updated for the recent distinct changes that landed (#287).

By the way, there were now delivery guarantee tests weren't actually added in #200. They were all added in cocoa in realm/realm-cocoa#4253 so I'll follow suit in my cocoa PR (realm/realm-cocoa#4114).

PR #141, opened 2016-08-24
Changes needed by java-binding for now.

beeender:
This PR is not ready to be merged to the master since we haven't decided how to deal with async-query and others.
This PR exposes more things to java-binding to use `SharedRealm` and current java async query implementation.
Create a PR to make changes more visible since object-store is a submodule in realm-java, those changes are not visible to realm-java PR.

PR #159, opened 2016-09-11
export ObjectStore::set_schema_version()

zaki50:
export `set_schema_version()` for https://github.com/realm/realm-java/issues/3417

### DIFF
--- a/src/object_store.hpp
+++ b/src/object_store.hpp
@@ -48,6 +48,7 @@ public:
     static uint64_t get_schema_version(Group const& group);
 
     // set schema version
+    // FIXME remove this after integrating OS's migration related logic into Realm java
     static void set_schema_version(Group& group, uint64_t version);
 
     // check if all of the changes in the list can be applied automatically, or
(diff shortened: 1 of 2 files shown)


zaki50:
please review this @realm/java 

There is one failing test case. I'm ignore it for now since the case is failing in other PRs (for example in #141 , https://ci.realm.io/job/ObjectStore%20Pull%20Request/314/)

src/object_store.hpp
```diff
@@ -47,6 +47,10 @@ class ObjectStore {
     // get the last set schema version
     static uint64_t get_schema_version(Group const& group);
 
+    // set schema version
+    // FIXME remove this after integrating OS's migration related logic into Realm java
+    static void set_schema_version(Group& group, uint64_t version);
```
cmelchior: Can this throw in some cases? E.g. if the version number if lower than the existing. This should be documented.

cmelchior:
Apart from a little comment 👍

src/object_store.hpp
zaki50: added more comment

kneth:
:+1:

PR #160, opened 2016-09-12
Create metadata and pk table when set the schema version

zaki50:
Object store always creates both metadata and pk table when creating a new Realm database.
This change aligns the behavior of Realm Java to ObjectStore. It is also required from realm/realm-java#3424

At first, I was thinking that I should expose `create_metadata_tables()` separately to be able to call it only at the creation of Realm, but finally I always call `create_metadata_tables()` from the `ObjectStore::set_schema_version()` since current OS always calls it on schema change.

### DIFF
--- a/src/object_store.cpp
+++ b/src/object_store.cpp
@@ -207,7 +207,9 @@ void validate_primary_column_uniqueness(Group const& group)
 }
 } // anonymous namespace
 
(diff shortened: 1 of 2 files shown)


cmelchior:
👍

kneth:
:+1:

PR #161, opened 2016-09-12
Handle the new add_int instruction

fealebenpae:
Introduced in core 2.0.0-rc4

Change courtesy of @cmelchior 

/cc @bdash @tgoyne


tgoyne:
:+1:

PR #163, opened 2016-09-14
Merge pk fixes to js branch

alazier:
@bdash


bdash:
Looks fine.

I'd probably push your branch directly to `js` rather than merging it into `js`, since merging it like a normal PR will give you a pointless extra merge commit.

alazier:
Will do that. Thanks.

PR #165, opened 2016-09-16
Send notifications on refresh and begin write

tgoyne:
This is a prerequisite for suppressing specific notifications from a write transaction, as we can't have notifications cover a range of commits with a gap, and generally makes it harder to get into inconsistent states.

The new behavior is that autorefresh works as it always did (advancing the most recent version with notifications ready), but now `refresh()` and `begin_transaction()` block if the notifications for the newest version are not ready yet. In the case of `refresh()`, the "newest version" means "the version that was newest at the time when refresh() was called or any version after that (which may still not be the actual latest version)". For `begin_transaction()` it blocks after acquiring the write lock, so it actually will be notifications for the latest version.

This makes `refresh()` within a notifications a no-op, and does not send notifications if `begin_transaction()` is called from within a notification, as I was unable to come up with a model for nested notifications which was clearly better than not doing them at all.

This does not attempt to make version advances from importing handover objects robustly produce notifications.

### DIFF
--- a/src/impl/collection_notifier.cpp
+++ b/src/impl/collection_notifier.cpp
@@ -368,9 +368,8 @@ NotifierPackage::NotifierPackage(Realm& realm, std::exception_ptr error,
 
 void NotifierPackage::package_and_wait(SharedGroup& sg)
 {
-    if (!m_lock || m_error)
+    if (!m_lock || m_error || !*this)
         return;
-    REALM_ASSERT(*this);
 
     using sgf = SharedGroupFriend;
     auto target_version = sgf::get_version_of_latest_snapshot(sg);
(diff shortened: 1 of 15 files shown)


astigsen:
> The new behavior is that autorefresh works as it always did (advancing the most recent version with notifications ready), but now refresh() and begin_transaction() block if the notifications for the newest version are not ready yet.

How is this different from the previous behavior? As I read the docs, the `begin_transaction` should already trigger any outstanding notifications:

> Before beginning the write transaction, beginWrite updates the Realm instance to the latest Realm version, as if refresh() had been called, and generates notifications if applicable

Did we not send notifications covering all the way up to the latest version before?

tgoyne:
We sent Realm-level and KVO notifications, but not collection notifications. This was an oversight that could be considered a bug.

Issue #154, opened 2016-09-01
Static analysis warning


jpsim:
```
Realm/ObjectStore/src/impl/transact_log_handler.cpp:286:9: Called C++ object pointer is null

        context->did_change(m_observers, invalidated);
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

tgoyne:
That line isn't reachable if context is nullptr, so this is just a false-positive that should be silenced.

austinzheng:
Verified that the static analyzer (in Xcode 8.2.1) no longer emits this warning.

PR #167, opened 2016-09-22
Fix possible null pointer dereference

austinzheng:


jpsim:
According to @tgoyne in #154, that's a false positive that should be silenced. Not sure how to do that though other than how you're doing it here.

austinzheng:
I'll look into silencing static analyzer warnings once I have some free time in the future.

tgoyne:
The warning happens to be eliminated by one of my notifications PRs waiting for review.

PR #168, opened 2016-09-22
Use set_string_unique() for the pk table

tgoyne:
The schema version table doesn't need it because we always read and write row zero.


beeender:
`ObjectStore::rename_property` seems not handling the renaming in pktable, is it correct?

tgoyne:
Yes, it is correct. rename_property may change which column will be the PK, but it does not change the name of the PK.

PR #171, opened 2016-09-27
Update to core 2.0.0

tgoyne:
And drop support for 1.x now that there's a non-beta release of 2.0.


tests/util/test_file.cpp
```diff
@@ -68,11 +63,7 @@ InMemoryTestFile::InMemoryTestFile()
 
 std::unique_ptr<Replication> TestFile::make_history() const
```
bdash: This helper doesn't seem particularly useful anymore.

PR #172, opened 2016-09-27
Add sync-related file management functionality to object store

austinzheng:
Tested and ready for review. Ari has been asking for this, so we should consider pulling it in. Otherwise, this can wait until after we get out of beta.

@alazier @jpsim @tgoyne @bdash 

Supersedes https://github.com/realm/realm-object-store-private/pull/47


austinzheng:
Superseded by #193

PR #173, opened 2016-09-27
Update cmake to use Core 2.0

austinzheng:


bdash:
#171 will take care of this.

PR #174, opened 2016-09-27
Add metadata support for Realm file path info

austinzheng:
WIP, tests needed

Intended for use with the 'fetch Realm' functionality. Supersedes https://github.com/realm/realm-object-store-private/pull/54

Changes:
- Added `SyncFileActionMetadata` class, representing file metadata
- Refactored some of the metadata support code
- Added additional checks for validation

### DIFF
--- a/src/sync_metadata.cpp
+++ b/src/sync_metadata.cpp
@@ -33,22 +33,36 @@
 namespace realm {
 
 static const char * const c_sync_userMetadata = "UserMetadata";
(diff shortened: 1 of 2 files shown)


austinzheng:
It's not clear what form this feature is going to take, if it's going to be included at all. I'll close and/or update this PR when we have more clarity.

austinzheng:
Closing this, since we don't want the associated feature anymore and the object store has changed drastically, meaning we'd have to refactor this code heavily even if we wanted to use it.

PR #175, opened 2016-09-28
Update to core 2.1.0 and add more tests

tgoyne:


AndyDentFree:
👍 although not hugely familiar with this code, looks straightforward

PR #176, opened 2016-09-28
Conditionally build sync

jpsim:
This PR merges `master` into `cocoa` and introduces a `REALM_ENABLE_SYNC` CMake option, which defaults to `OFF`.

Would love to know if there's a better way to do this @bdash. I'd rather not check for the presence of `REALM_SYNC_PREFIX` since when building realm cocoa the source for sync may not be available.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,11 @@ include(Sanitizers)
 include(RealmCore)
 set(REALM_CORE_VERSION "1.5.0" CACHE STRING "")
 use_realm_core(${REALM_CORE_VERSION})
-build_realm_sync(${REALM_SYNC_PREFIX})
+
+option(REALM_ENABLE_SYNC "Build realm object store with sync extensions" OFF)
+if(REALM_ENABLE_SYNC)
+  build_realm_sync(${REALM_SYNC_PREFIX})
+endif()
 
 set(PEGTL_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/external/pegtl)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,6 @@ include(RealmCore)
 set(REALM_CORE_VERSION "1.5.0" CACHE STRING "")
 use_realm_core(${REALM_CORE_VERSION})
 
-option(REALM_ENABLE_SYNC "Build realm object store with sync extensions" OFF)
 if(REALM_SYNC_PREFIX)
   set(REALM_ENABLE_SYNC ON)
   build_realm_sync(${REALM_SYNC_PREFIX})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,8 @@ set(REALM_CORE_VERSION "1.5.0" CACHE STRING "")
 use_realm_core(${REALM_CORE_VERSION})
 
 option(REALM_ENABLE_SYNC "Build realm object store with sync extensions" OFF)
-if(REALM_ENABLE_SYNC)
+if(REALM_SYNC_PREFIX)
+  set(REALM_ENABLE_SYNC ON)
   build_realm_sync(${REALM_SYNC_PREFIX})
 endif()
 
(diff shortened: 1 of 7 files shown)


bdash:
1. It's hard to review a PR when it includes a merge of master
2. Nothing seems to set the `REALM_ENABLE_SYNC` preprocessor define that this PR looks for.
3. I'd suggest having CMake set the `REALM_ENABLE_SYNC` preprocessor define when `REALM_SYNC_PREFIX` is specified. Other build systems can set it explicitly when building object store with prebuilt sync libraries.

jpsim:
> It's hard to review a PR when it includes a merge of master

Sorry, I incorrectly assumed that a single commit after a merge would be easy to review.

I'll drop those commits from this branch and rebase against `cocoa` shortly.

> Nothing seems to set the `REALM_ENABLE_SYNC` preprocessor define that this PR looks for.

The `option()` entry in `CMakeLists.txt` sets it to `OFF` if it hasn't already been explicitly set. It can be explicitly set when passing options to CMake, as demonstrated in the README changes I made.

> I'd suggest having CMake set the `REALM_ENABLE_SYNC` preprocessor define when `REALM_SYNC_PREFIX` is specified. Other build systems can set it explicitly when building object store with prebuilt sync libraries.

Works for me.

bdash:
> > Nothing seems to set the `REALM_ENABLE_SYNC` preprocessor define that this PR looks for.

> The `option()` entry in `CMakeLists.txt` sets it to `OFF` if it hasn't already been explicitly set. It can be explicitly set when passing options to CMake, as demonstrated in the README changes I made.

`option` generates a CMake setting. What turns it into a C++ preprocessor define?

jpsim:
> `option` generates a CMake setting. What turns it into a C++ preprocessor define?

I was unaware that CMake flags weren't also passed as C++ preprocessor defines. Is the best way to do that via [`add_definitions`](https://cmake.org/cmake/help/v3.0/command/add_definitions.html)?

bdash:
Yeah, `add_definitions` or `target_compile_definitions`, depending on at which level we think it makes sense to set the define.

jpsim:
Thanks for your help. How's this?

CMakeLists.txt
```diff
@@ -18,7 +18,16 @@ include(Sanitizers)
 include(RealmCore)
 set(REALM_CORE_VERSION "1.5.0" CACHE STRING "")
 use_realm_core(${REALM_CORE_VERSION})
-build_realm_sync(${REALM_SYNC_PREFIX})
+
+option(REALM_ENABLE_SYNC "Build realm object store with sync extensions" OFF)
```
bdash: Since we're now setting `REALM_ENABLE_SYNC` based on whether or not `REALM_SYNC_PREFIX` was set, I don't think there's any reason for `REALM_ENABLE_SYNC` to be independently configurable. You can likely just remove this line.

src/impl/realm_coordinator.cpp
```diff
@@ -101,6 +103,7 @@ std::shared_ptr<Realm> RealmCoordinator::get_realm(Realm::Config config)
         }
     }
 
+    #if REALM_ENABLE_SYNC
```
bdash: Please don't indent `#if`s.

tests/CMakeLists.txt
```diff
@@ -26,7 +26,9 @@ set(SOURCES
 add_executable(tests ${SOURCES} ${HEADERS})
 target_compile_definitions(tests PRIVATE ${PLATFORM_DEFINES})
 target_link_libraries(tests realm-object-store ${PLATFORM_LIBRARIES})
+if(REALM_ENABLE_SYNC)
 target_link_libraries(tests realm-sync)
```
bdash: I don't think this `target_link_libraries` is necessary. `realm-object-store` lists `realm-sync` as one of its public dependencies, which means that anyone that links to the `realm-object-store` target will automatically pull in `realm-sync`.

CMakeLists.txt
bdash: Unless this is part of your plan to work towards using prebuilt versions of sync in the object store, in which case maybe it's fine to leave this.

jpsim: That was not part of my plan. I'll remove.

tests/CMakeLists.txt
jpsim: You're right, removing this explicit link still builds and runs the tests just fine.

PR #179, opened 2016-09-30
Changes needed by java-binding for now.

beeender:


tests/util/event_loop.hpp
```diff
@@ -18,6 +18,7 @@
 
 #include <functional>
 #include <memory>
+#include <chrono>
```
kneth: Can be removed?

src/impl/realm_coordinator.hpp
```diff
@@ -145,7 +146,9 @@ class RealmCoordinator : public std::enable_shared_from_this<RealmCoordinator> {
     std::unique_ptr<_impl::ExternalCommitHelper> m_notifier;
     std::function<void(VersionID, VersionID)> m_transaction_callback;
 
+#ifdef REALM_ENABLE_SYNC
     std::shared_ptr<SyncSession> m_sync_session;
```
bdash: Why #if this?

src/shared_realm.hpp
```diff
@@ -166,8 +171,10 @@ class Realm : public std::enable_shared_from_this<Realm> {
         // speeds up tests that don't need notifications.
         bool automatic_change_notifications = true;
 
+#ifdef REALM_ENABLE_SYNC
```
bdash: Why #if this?

src/shared_realm.hpp
```diff
@@ -31,6 +31,8 @@
 #include <memory>
 #include <thread>
 
+#include <realm/group_shared.hpp>
```
bdash: If this is only needed because of the `_impl::RealmFriend` function bodies being inlined, it'd be preferable to move them, and this include, to the .cpp file instead.

src/shared_realm.hpp
```diff
@@ -40,7 +42,9 @@ class Realm;
 class Replication;
 class SharedGroup;
 class StringData;
+#ifdef REALM_ENABLE_SYNC
```
bdash: Why #if this?

.gitignore
```diff
@@ -18,3 +18,6 @@ Debug/
 Release/
 RelWithDebInfo/
 MinSizeRel/
+
+# IntelliJ
+.idea/
```
bdash: Please add a newline at the end of file.

src/shared_realm.cpp
```diff
@@ -183,7 +183,10 @@ void Realm::open_with_config(const Config& config,
             read_only_group = std::make_unique<Group>(config.path, config.encryption_key.data(), Group::mode_ReadOnly);
         }
         else {
-            bool server_synchronization_mode = bool(config.sync_config);
+            bool server_synchronization_mode = false;
+#if REALM_ENABLE_SYNC
```
bdash: Why #if this?

src/impl/realm_coordinator.hpp
cmelchior: Because the Java binding distribute the core library without any sync code. So we are stripping all references to sync code to reduce the size.

beeender:
#287 instead.

PR #181, opened 2016-09-30
Attempt to resolve object store deadlocks

### LINKED ISSUE #178: Sync-related object store deadlocks
I believe these are possible deadlocks, and related issues. Making this ticket to track my progress as I investigate further:
- When a session dies, it takes the session lock while it waits for changes to upload. If, while waiting, an error occurs, it is reported via the error handler. Certain errors may require a change in state, thus the error handler attempts to acquire the lock the dying-session lambda is already holding, and the program stops. (EDIT: this applies more generally to the error handler. The contract reads as follows:)

> CAUTION: The specified callback function may be called before the call to bind() returns, and it may be called (or continue to execute) after the session object is destroyed. The application must pass a handler that can be safely called, and can safely continue to execute from the point in time where bind() starts executing, and up until the point in time where the last invocation of `clint.run()` returns. Here, `client` refers to the associated Client object.
- A couple of the session handle APIs (`isValid`, `logOut`) seem to sporadically deadlock the program on acquiring the state change lock.
- (Not a deadlock) If a session is dying (waiting for pending changes to upload) and the user logs out, there should be a way to kill the session immediately. Right now, the session is only killed immediately if it is active (or not otherwise dying).
- (Not a deadlock) The object store `Session` should only run the `wait_for_download` (or its upload equivalent) APIs if the state makes sense to do so.
- (Not a deadlock) `testImmediateSessionTerminationWhenLoggingOut` sometimes causes a `bad_server_file_ident` error during the cleanup phase, which uncovered one of the deadlocks above.

austinzheng:
@bdash @tgoyne 

Fixes #178

This PR does two things:
- The session's state lock is never held through the duration of a call to `wait_for_upload_complete_or_client_stopped()` or `wait_for_download_complete_or_client_stopped()`. This prevents certain deadlocks from happening: for example, if the error handler is invoked while the binding is waiting (and the lock is being held), it will attempt to move the state to `error` but will never be able to acquire the lock, and the blocking call won't return and relinquish the lock because the error handler can't return.
- Any time those two APIs are called the lifetime of the `sync::Session` is now guaranteed until the call returns.

Maybe there's a better way to do this? Ideas welcome...

### DIFF
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -51,8 +51,10 @@ using namespace realm::_impl::sync_session_states;
 ///    * WAITING_FOR_ACCESS_TOKEN: if the session is informed (through the error
 ///                                handler) that the token expired
 ///    * INACTIVE: if asked to log out, or if asked to close and the stop policy
-///                is Immediate.
-///    * DYING: if asked to close and the stop policy is AfterChangesUploaded
+///                is Immediate. There must be no pending threads waiting on uploads
+///                or downloads.
+///    * DYING: if asked to close and the stop policy is AfterChangesUploaded, or
+///             there are any pending threads waiting on uploads or downloads.
 ///    * ERROR: if a fatal error occurs
 ///
 /// DYING: the session is performing clean-up work in preparation to be destroyed.
@@ -60,7 +62,8 @@ using namespace realm::_impl::sync_session_states;
 /// To:
 ///    * INACTIVE: when the clean-up work completes, if the session wasn't
 ///                revived, or if explicitly asked to log out before the
-///                clean-up work begins
+///                clean-up work begins, or otherwise once all pending threads
+///                waiting on uploads or downloads have completed waiting.
 ///    * ACTIVE: if the session is revived
 ///    * ERROR: if a fatal error occurs
 ///
@@ -98,8 +101,6 @@ struct SyncSession::State {
 
     virtual void log_out(std::unique_lock<std::mutex>&, SyncSession&) const { }
 
-    virtual void close_if_connecting(std::unique_lock<std::mutex>&, SyncSession&) const { }
-
     virtual void close(std::unique_lock<std::mutex>&, SyncSession&) const { }
 
     static const State& waiting_for_access_token;
@@ -107,6 +108,10 @@ struct SyncSession::State {
     static const State& dying;
     static const State& inactive;
     static const State& error;
+};
+
+enum class SyncSession::NetworkMode {
+    Upload, Download
 };
 
 struct sync_session_states::WaitingForAccessToken : public SyncSession::State {
@@ -138,7 +143,11 @@ struct sync_session_states::WaitingForAccessToken : public SyncSession::State {
 
     void log_out(std::unique_lock<std::mutex>& lock, SyncSession& session) const override
     {
-        session.advance_state(lock, inactive);
+        if (session.m_pending_download_threads == 0 && session.m_pending_upload_threads == 0) {
+            session.advance_state(lock, inactive);
+        } else {
+            session.advance_state(lock, dying);
+        }
     }
 
     void nonsync_transact_notify(std::unique_lock<std::mutex>&,
@@ -147,12 +156,6 @@ struct sync_session_states::WaitingForAccessToken : public SyncSession::State {
     {
         // Notify at first available opportunity.
         session.m_deferred_commit_notification = version;
-    }
-
-    void close_if_connecting(std::unique_lock<std::mutex>& lock, SyncSession& session) const override
-    {
-        // Ignore the sync configuration's stop policy as we're not yet connected.
-        session.advance_state(lock, inactive);
     }
 
     void close(std::unique_lock<std::mutex>&, SyncSession& session) const override
@@ -180,7 +183,11 @@ struct sync_session_states::Active : public SyncSession::State {
 
     void log_out(std::unique_lock<std::mutex>& lock, SyncSession& session) const override
     {
-        session.advance_state(lock, inactive);
+        if (session.m_pending_download_threads == 0 && session.m_pending_upload_threads == 0) {
+            session.advance_state(lock, inactive);
+        } else {
+            session.advance_state(lock, dying);
+        }
     }
 
     void nonsync_transact_notify(std::unique_lock<std::mutex>&, SyncSession& session,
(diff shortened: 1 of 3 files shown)


bdash:
I think change would make it possible for the object store to create two `sync::Session`s that are working with the same Realm file, which isn't permitted.

austinzheng:
Yeah, I forgot about that. I'll have to think about it some more...

austinzheng:
I won't be able to test this update until I fix my broken environment, but I wanted to see if this alternate approach was worth pursuing.

austinzheng:
I think this can be reviewed. The binding and object store tests that exist pass, and manual inspection suggests that the tests are moving the state into `inactive` appropriately now.

src/sync_session.cpp
```diff
@@ -183,42 +192,58 @@ struct sync_session_states::Active : public SyncSession::State {
...
-            if (session->m_state != &State::dying) {
-                // The session was revived. Don't kill it.
-                --session->m_pending_upload_threads;
-                return;
+        switch (session.m_config.stop_policy) {
+            case SyncSessionStopPolicy::Immediately:
+                if (session.m_pending_upload_threads == 0 && session.m_pending_upload_threads == 0) {
+                    session.advance_state(lock, inactive);
+                }
+                // Otherwise, wait for any pending uploads or downloads to complete, and then allow them to handle
```
jpsim: this should be mirrored in `enter_state`'s header doc comment

src/sync_session.cpp
```diff
@@ -183,42 +192,58 @@ struct sync_session_states::Active : public SyncSession::State {
...
+                // killing the session.
+                break;
+            case SyncSessionStopPolicy::AfterChangesUploaded: {
+                if (session.m_pending_upload_threads > 0) {
+                    // Already someone waiting for uploads to finish. Just have them handle killing the session.
```
jpsim: this should be mirrored in `enter_state`'s header doc comment

src/sync_session.cpp
```diff
@@ -183,42 +192,58 @@ struct sync_session_states::Active : public SyncSession::State {
...
+                    return;
+                }
+                // Spin up a thread to wait for any pending uploads to finish, and then die.
```
jpsim: this should be mirrored in `enter_state`'s header doc comment

src/sync_session.cpp
```diff
@@ -230,7 +255,9 @@ struct sync_session_states::Dying : public SyncSession::State {
 
     void log_out(std::unique_lock<std::mutex>& lock, SyncSession& session) const override
     {
-        session.advance_state(lock, inactive);
+        if (session.m_pending_download_threads == 0 && session.m_pending_upload_threads == 0) {
+            session.advance_state(lock, inactive);
+        }
```
jpsim: earlier, we `advance_state` to `dying` if there are pending threads, why are we not doing that here?

austinzheng: This is `dying`'s log out function, there's an assertion that forbids advancing a state to itself. (We also don't want to rerun the 'enter state' work if the binding asks the session to log out while it's already in the process of dying.)

src/sync_session.cpp
```diff
@@ -421,31 +448,54 @@ void SyncSession::unregister(std::unique_lock<std::mutex>& lock)
...
-            std::unique_lock<std::mutex> lock(m_state_mutex);
-            if (m_session) {
-                m_session->wait_for_upload_complete_or_client_stopped();
+    auto thread = std::thread([this, callback=std::move(callback)]() {
+        std::unique_lock<std::mutex> lock(m_state_mutex);
+        if (m_state != &State::active && m_state != &State::dying) {
+            lock.unlock();
+            callback();
+        } else {
+            ++m_pending_upload_threads;
```
tgoyne: This needs to be done before creating the thread to ensure that `this` still exists by the time the thread runs.

src/sync_session.cpp
```diff
@@ -421,31 +448,54 @@ void SyncSession::unregister(std::unique_lock<std::mutex>& lock)
...
+            // Once callback is finished, if necessary perform cleanup work and kill session.
+            lock.lock();
+            --m_pending_upload_threads;
+            if (m_state == &State::dying
+                && m_pending_upload_threads == 0
+                && m_pending_download_threads == 0) {
+                // Advance state to 'inactive'
+                advance_state(lock, State::inactive);
             }
+            lock.unlock();
```
tgoyne: This needs to be done only if it did not advance to `inactive` as switching to `inactive` may unlock it or even delete the lock.

src/sync_session.cpp
```diff
@@ -421,31 +448,54 @@ void SyncSession::unregister(std::unique_lock<std::mutex>& lock)
...
 
 void SyncSession::wait_for_download_completion(std::function<void()> callback)
 {
     REALM_ASSERT(shared_from_this());
-    auto thread = std::thread([this, callback=std::move(callback), self=shared_from_this()]() {
-        {
-            std::unique_lock<std::mutex> lock(m_state_mutex);
-            if (m_session) {
-                m_session->wait_for_download_complete_or_client_stopped();
+    auto thread = std::thread([this, callback=std::move(callback)]() {
```
tgoyne: There's a lot of nontrivial logic here that's entirely duplicated from the upload version.

src/sync_session.cpp
```diff
@@ -423,42 +444,66 @@ bool SyncSession::can_wait_for_network_completion() const
...
+            ++m_pending_upload_threads;
+            break;
+        case NetworkMode::Download:
+            ++m_pending_download_threads;
+            break;
+    }
+    auto thread = std::thread([this, callback=std::move(callback), self=shared_from_this(), mode]() {
+        std::unique_lock<std::mutex> lock(m_state_mutex);
+        if (!can_wait_for_network_completion()) {
+            lock.unlock();
```
bdash: You need to decrement the pending upload threads here as well.

src/sync_session.cpp
```diff
@@ -183,42 +190,58 @@ struct sync_session_states::Active : public SyncSession::State {
...
+                            // The session was revived. Don't kill it.
+                            --session->m_pending_upload_threads;
+                            return;
+                        }
+                    }
+                    session->m_session->wait_for_upload_complete_or_client_stopped();
+                    {
+                        std::unique_lock<std::mutex> lock(session->m_state_mutex);
+                        --session->m_pending_upload_threads;
+                        if (session->m_pending_upload_threads == 0 && session->m_pending_download_threads == 0) {
```
austinzheng: Yes we do.

austinzheng:
Would we still like to purse this approach? I'm happy to make any other changes necessary to get this approved and/or verify its functionality.

austinzheng:
I rebased this change onto my latest work, and re-enabled the test Mark stubbed out.

src/sync/sync_session.cpp
```diff
@@ -451,42 +475,64 @@ bool SyncSession::can_wait_for_network_completion() const
...
+        case NetworkMode::Upload:
+            ++m_pending_upload_threads;
+            break;
+        case NetworkMode::Download:
+            ++m_pending_download_threads;
+            break;
+    }
+    auto thread = std::thread([this, callback=std::move(callback), self=shared_from_this(), mode]() {
+        std::unique_lock<std::mutex> lock(m_state_mutex);
+        if (can_wait_for_network_completion()) {
```
jpsim: `lock.unlock();` is called in both branches of this conditional.

jpsim:
Should this be targeting the `master` branch now?

austinzheng:
@jpsim yeah, it should.

src/sync/sync_session.cpp
austinzheng: The lock can't be released until the if statement's predicate function has been called. However, I've slightly rearranged things to make my intent clearer.

austinzheng:
I'm going to resurrect this PR and fix the merge conflicts, because I think it provides a workaround for a couple of issues that we're seeing WRT resetting sync state for tests, and bugs in `sync::Session`/`sync::Client`'s handling of threads waiting on uploads/downloads. More details later...

austinzheng:
Since this PR deals exclusively with the blocking wait-for-X methods, it seems like https://github.com/realm/realm-sync/pull/983 supersedes it.

PR #185, opened 2016-10-04
Proposal for API to move sync functionality into object store

austinzheng:
I want to start a conversation, per @alazier's request, about eventually moving most of the sync functionality into the object store. This commit isn't meant to be merged, but rather to propose API changes through header file changes.

@jpsim @bdash @tgoyne @cmelchior 

This is what I envision. (Updated 10/5)

## `SyncUser`

`SyncUser` is a new object store type. It represents a user whose identity is known (e.g. the result of a binding successfully "logging in" using a Credential object). It does not own users, but does have weak pointers to the sessions associated with it.

## `SyncConfig` changes:

Rather than carrying a string "tag" to a user, the config now carries a shared pointer to the user object itself.

## `SyncManager` changes:

The `SyncManager` is now responsible for vending out users, as well as sessions.

## `SyncSession` changes:

Sync sessions now have a weak pointer to the user that owns them.

The sync session state machine should be changed such that the session does not automatically try to bind with the callback as soon as it's created, but waits for the user to ask it to do so. (The user might do so immediately, if it's logged in, or it might wait until it becomes active again if it were logged out previously.)

## Minimum binding requirement

A hypothetical binding which wanted to support the minimal sync feature set given this new object store design would do the following:

1. Create a `SyncConfiguration` type which wraps `realm::SyncConfig`, and allow the user to set it as a property on a `RealmConfiguration`, updating the underlying `realm::Realm::Config`.
2. Create a `User` type which wraps a `realm::SyncUser`. This user object needs to be able to make the login call to the auth server using a credential, and might have other APIs like `logOut()`. 
3. Set the global "bind session" callback.

The object store would transparently take care of everything else, from closing sessions upon completion to persisting and loading user info in the metadata database.

### DIFF
--- a/src/sync_manager.hpp
+++ b/src/sync_manager.hpp
@@ -32,6 +32,7 @@ namespace realm {
 
 struct SyncConfig;
 struct SyncSession;
+class SyncUser;
 
 namespace _impl {
 struct SyncClient;
@@ -42,8 +43,6 @@ enum class SyncSessionStopPolicy {
     LiveIndefinitely,               // Never stop the session.
     AfterChangesUploaded,           // Once all Realms/Sessions go out of scope, wait for uploads to complete and stop.
 };
-
-using SyncLoginFunction = std::function<void(const std::string&, const SyncConfig&)>;
 
 class SyncLoggerFactory {
 public:
@@ -58,7 +57,6 @@ public:
     void set_log_level(util::Logger::Level) noexcept;
     void set_logger_factory(SyncLoggerFactory&) noexcept;
     void set_error_handler(std::function<sync::Client::ErrorHandler>);
-    void set_login_function(SyncLoginFunction);
 
     /// Control whether the sync client attempts to reconnect immediately. Only set this to `true` for testing purposes.
     void set_client_should_reconnect_immediately(bool reconnect_immediately);
@@ -68,7 +66,11 @@ public:
     std::shared_ptr<SyncSession> get_session(const std::string& path, const SyncConfig& config);
     std::shared_ptr<SyncSession> get_existing_active_session(const std::string& path) const;
 
-    SyncLoginFunction& get_sync_login_function();
+    // Get a sync user for a given identity, or create one if none exists yet, and set its token.
+    // If a logged-out user exists, it will marked as logged back in.
+    shared_ptr<SyncUser> get_user(std::string identity, std::string refresh_token, bool is_admin=false);
+    // Get an existing user for a given identity, if one exists and is logged in.
+    shared_ptr<SyncUser> get_existing_logged_in_user(std::string identity);
 
 private:
     void dropped_last_reference_to_session(SyncSession*);
@@ -90,14 +92,15 @@ private:
 
     mutable std::mutex m_mutex;
 
-    SyncLoginFunction m_login_function;
-
     // FIXME: Should probably be util::Logger::Level::error
     util::Logger::Level m_log_level = util::Logger::Level::info;
     SyncLoggerFactory* m_logger_factory = nullptr;
     std::function<sync::Client::ErrorHandler> m_error_handler;
     sync::Client::Reconnect m_client_reconnect_mode = sync::Client::Reconnect::normal;
     bool m_client_validate_ssl = true;
+
+    // A map of user identities to (shared pointers to) SyncUser objects.
+    std::unordered_map<std::string, shared_ptr<SyncUser>> m_users;
 
     mutable std::shared_ptr<_impl::SyncClient> m_sync_client;
 
--- a/src/sync_session.hpp
+++ b/src/sync_session.hpp
@@ -29,6 +29,7 @@
 namespace realm {
 
 class SyncManager;
+class SyncUser;
 
 namespace _impl {
 class RealmCoordinator;
@@ -71,6 +72,11 @@ struct SyncSession : public std::enable_shared_from_this<SyncSession> {
 
     // Inform the sync session that it should log out.
     void log_out();
+
+    std::shared_ptr<SyncUser> user() const
+    {
+        return m_user.lock();
+    }
 
     // Expose some internal functionality to other parts of the ObjectStore
     // without making it public to everyone
@@ -128,6 +134,8 @@ private:
     size_t m_pending_upload_threads = 0;
 
     SyncConfig m_config;
+
+    std::weak_ptr<SyncUser> m_user;
 
     std::string m_realm_path;
     std::shared_ptr<_impl::SyncClient> m_client;
--- a/src/sync_user.hpp
+++ b/src/sync_user.hpp
@@ -1,0 +1,93 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#ifndef REALM_OS_SYNC_USER_HPP
+#define REALM_OS_SYNC_USER_HPP
+
+#include <string>
+#include <memory>
+#include <unordered_map>
+
+namespace realm {
+
+struct SyncSession;
+
+// A `SyncUser` represents a single user account. Each user manages the sessions that
+// are associated with it.
+class SyncUser {
+public:	
+	enum class State {
+		LoggedOut, Active, Error;
+	}
+
+	// Don't use this directly; use the `SyncManager` APIs. Public for use with `make_shared`.
+	SyncUser(std::string refresh_token, bool is_admin=false);
+
+	// Return a list of all users.
+	static std::vector<std::shared_ptr<SyncUser>> all_users();
+
+	// Return a list of all sessions belonging to this user.
+	std::vector<std::shared_ptr<SyncSession>> all_sessions();
+
+	// Return a session for a given file path.
+	std::shared_ptr<SyncSession> session_for_path(const std::string path&);
+
+	// Log the user out and mark it as invalid. This will also close its associated Sessions.
+	void log_out();
+
+	// Whether the user was configured as an 'admin user' (directly uses its user token
+	// to open Realms).
+	bool is_admin() const
+	{
+		return m_is_admin;
+	}
+
+	std::string identity() const
+	{
+		return m_identity;
+	}
+
+	State state() const
+	{
+		return m_state;
+	}
+
+	// The auth server URL. Bindings should set this appropriately when they retrieve
+	// instances of `SyncUser`s.
+	// FIXME: once the new token system is implemented, this can be removed completely.
+	std::string server_url;
+
+private:
+	State m_state;
+	
+	// Whether the user is an 'admin' user. Admin users use the admin tokens they were
+	// configured with to directly open sessions, and do not make network requests.
+	bool m_is_admin;
+	// The user's refresh token.
+	std::string m_refresh_token;
+	// Set by the server. The unique ID of the user account on the Realm Object Server.
+	std::string m_identity;
+
+    // Sessions are owned by the SyncManager, but the user keeps a map of weak references
+    // to them.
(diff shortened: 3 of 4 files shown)


src/sync_manager.hpp
```diff
@@ -58,39 +66,32 @@ friend struct SyncSession;
...
     /// Control whether the sync client attempts to reconnect immediately. Only set this to `true` for testing purposes.
     void set_client_should_reconnect_immediately(bool reconnect_immediately);
     /// Control whether the sync client validates SSL certificates. Should *always* be `true` in production use.
     void set_client_should_validate_ssl(bool validate_ssl);
 
-    std::shared_ptr<SyncSession> get_session(const std::string& path, const SyncConfig& config);
-    std::shared_ptr<SyncSession> get_existing_active_session(const std::string& path) const;
+    // Get a sync user for a given identity, or create one if none exists yet, and set its token.
+    // If a logged-out user exists, it will marked as logged back in.
+    shared_ptr<SyncUser> get_user(std::string identity, std::string user_token, bool is_admin=false);
```
bdash: What is a "user token" in this context? Why is `is_admin` necessary?

src/sync_manager.hpp
```diff
@@ -99,19 +100,10 @@ friend struct SyncSession;
     sync::Client::Reconnect m_client_reconnect_mode = sync::Client::Reconnect::normal;
     bool m_client_validate_ssl = true;
 
-    mutable std::shared_ptr<_impl::SyncClient> m_sync_client;
+    // A map of user identities to (shared pointers to) SyncUser objects.
+    std::unordered_map<std::string, shared_ptr<SyncUser>> m_users;
```
bdash: When are users removed from this map?

src/sync_session.hpp
```diff
@@ -72,6 +73,11 @@ struct SyncSession : public std::enable_shared_from_this<SyncSession> {
     // Inform the sync session that it should log out.
     void log_out();
 
+    std::shared_ptr<SyncUser> parent_user() const
```
bdash: Not sure that `parent` is meaningful here.

src/sync_user.hpp
```diff
@@ -0,0 +1,118 @@
...
+public:	
+	enum class State {
+		LoggedOut, Active, Error;
+	}
+
+	// Don't use this directly; use the `SyncManager` APIs. Public for use with `make_shared`.
+	SyncUser(std::string user_token, bool is_admin=false);
+
+	// Return a list of all users.
+	static std::vector<SyncUser> all_users();
```
bdash: This return type conflicts with your statement that there's only a single `SyncUser` for a given user.

src/sync_user.hpp
```diff
@@ -0,0 +1,118 @@
...
+    // due to someone holding a strong reference to it.
+    void unregister_session(const std::string& path);
+
+    // Active sessions are sessions which the client code holds a strong
+    // reference to. When the last strong reference is released, the session is
+    // moved to inactive sessions. Inactive sessions are promoted back to active
+    // sessions until the session itself calls unregister_session to remove
+    // itself from inactive sessions once it's done with whatever async cleanup
+    // it needs to do.
+    std::unordered_map<std::string, std::weak_ptr<SyncSession>> m_active_sessions;
```
bdash: How will we prevent the same Realm from being used by multiple sync sessions simultaneously if the sync sessions are now partitioned by user?

src/sync_manager.hpp
```diff
@@ -43,7 +44,14 @@ enum class SyncSessionStopPolicy {
...
 
-using SyncLoginFunction = std::function<void(const std::string&, const SyncConfig&)>;
+// A global callback used to indicate that a network request should be made to bind a session.
+// The callback is passed the current user token of the session's user, the unresolved URL of the
+// session to be bound, a boolean indicating whether this is the initial bind (true) or a refresh,
+// and a shared pointer to a SyncSession object upon which the bind or refresh API should be invoked
+// once the network response is parsed.
+// NOTE: the bind or refresh APIs must be called asynchronously.
+// FIXME: This can go away entirely once the new token system is implemented.
+using SyncBindSessionFunction = std::function<void(const std::string&, const std::string&, bool, std::shared_ptr<SyncSession>)>;
```
bdash: Passing a `shared_ptr<SyncSession>` to this function is problematic. We currently call the log-in callback from within `SyncSession`'s constructor, but no `shared_ptr` to the object exists at that point.

src/sync_manager.hpp
```diff
@@ -58,39 +66,32 @@ friend struct SyncSession;
...
     /// Control whether the sync client validates SSL certificates. Should *always* be `true` in production use.
     void set_client_should_validate_ssl(bool validate_ssl);
 
-    std::shared_ptr<SyncSession> get_session(const std::string& path, const SyncConfig& config);
-    std::shared_ptr<SyncSession> get_existing_active_session(const std::string& path) const;
+    // Get a sync user for a given identity, or create one if none exists yet, and set its token.
+    // If a logged-out user exists, it will marked as logged back in.
+    shared_ptr<SyncUser> get_user(std::string identity, std::string user_token, bool is_admin=false);
+    // Get an existing, valid sync user for a given identity.
+    shared_ptr<SyncUser> get_existing_valid_user(std::string identity);
```
bdash: What makes a user "valid"?

src/sync_user.hpp
```diff
@@ -0,0 +1,118 @@
...
+namespace realm {
+
+struct SyncSession;
+
+// A `SyncUser` represents a single user account. Each user manages the sessions that
+// are associated with it.
+class SyncUser {
+public:	
+	enum class State {
+		LoggedOut, Active, Error;
```
bdash: What does it mean for a user to be in an error state?

src/sync_manager.hpp
austinzheng: Good point. I'll have to think about an alternate way for the binding to feed the token/full URL back into the object store.

src/sync_manager.hpp
austinzheng: A `user_token` is what we currently call a refresh token. If/when we adopt the new token system the refresh token's name won't make sense anymore. (Although it might not make sense to name things anticipating future functionality...)

`is_admin` is for the use case where the user is created using the admin token (the one the server prints out when starting up) directly, which means that token is used to open all Realms. I think the browser and global notifier both use this mechanism. (also: cc @alazier)

src/sync_user.hpp
austinzheng: The session error handler is called with one of the error codes that indicates a user-level (as opposed to session level) error. I believe `Error::bad_authentication` is the only error right now that can cause this issue (i.e. auth and sync servers are misconfigured so the sync server doesn't accept the token issued by the auth server.)

Currently, we treat this as a 'fatal'/'unrecoverable' error that requires the app to try making a different user.

src/sync_manager.hpp
austinzheng: You're right, I'm not sure `valid` means anything in this context. It might make more sense if this method was named `get_existing_active_user` or `get_existing_logged_in_user`.

src/sync_manager.hpp
bdash: I think it's confusing to make references to a new system that has not yet been designed. The code should reflect the current design, unless we have a very clear idea of what we think the future will look like.

I was under the impression that with the current implementation it's not necessary to have any of the auth-related information in order to start a sync session, with it only being necessary at the point the log-in callback is called. This new design appears to require it up front. Am I missing something about how things currently work?

src/sync_user.hpp
austinzheng: Oh, I screwed up there. That should be a vector of `shared_ptr`s, not directly of `SyncUser`s.

src/sync_user.hpp
austinzheng: A given Realm is tied to a single user (and its `path` reflects this fact), so it should by definition be impossible for two users to open the same Realm.

astigsen: Why should two users not be able to open the same Realm? That seems like a pretty common use case for shared data.

bdash: > Why should two users not be able to open the same Realm? That seems like a pretty common use case for shared data.

There's no reason for two users to open the same Realm on disk. They can certainly open the same object server URL though.

> A given Realm is tied to a single user (and its path reflects this fact), so it should by definition be impossible for two users to open the same Realm.

What makes this impossible? AFAICT there's nothing stopping me from creating two separate `SyncUser`s and asking each to give me a session for the same path on disk.

astigsen: This is a good reason to never expose the path of the cached realm on disk to the user.

austinzheng: > What makes this impossible? AFAICT there's nothing stopping me from creating two separate `SyncUser`s and asking each to give me a session for the same path on disk.

`SyncUser` would be like `SyncSession` - instead of constructing them yourself you ask something else to vend them for you. In this case, the `SyncManager` would be responsible for vending them out, which means it could enforce ensuring that a given user (as marked by its identity) can only be instantiated once.

src/sync_manager.hpp
austinzheng: I agree. Mentioning systems that have yet to be designed doesn't help with anything and just confuses people. I'll fix that when I revise the PR later.

As for your second question, I think what you mean (please correct if I misunderstood) is that with the current system a session can be created for a logged-out user and only bound later (once the user is logged in), but with this system the session would need the user's refresh token at creation time (because of the callback). This is a good point.

I think the solution, with the proposed design, is no longer to have `SyncSession`s automatically start trying to get their access token as soon as they're created. Instead, when vended out from the user, the user (upon creating a session) can look at its current state and decide whether to ask the session to connect immediately, or to hold off until it's been logged in. This is pretty much what the binding does right now: a `RLMUser` will either make the network call to bind a session immediately, or it'll save the data necessary to do so and make that network call only once logged back in.

src/sync_user.hpp
bdash: I don't follow. I'm not talking about creating multiple `SyncUser`s for a given identity. I'm talking about creating two entirely different users and having them open the same path on disk.

src/sync_user.hpp
austinzheng: I think we should take a step back. If the app logs in with a username + incorrect password and the ROS is functioning properly, the auth server should reject that login request. The binding will then be responsible for asking the user to provide a different password and try again. In this case the binding won't even get as far as creating a `realm::SyncUser` for its binding-level User object.

The only way the `User` can get into the error state is if the server is broken and the auth and sync parts of the ROS are not configured properly. In this case I'm not sure what the best way to recover is, or if we should even attempt to recover.

src/sync_user.hpp
bdash: That may be where we want things to end up, but from what I can see it's not where things stand after this proposed refactoring.

austinzheng: Sorry, I'm not sure I understand. Regardless, would moving responsibility for sessions back into the `SyncManager` alleviate your concern?

austinzheng:
I updated this proposal based on the reviews above and Slack discussion with @bdash. (I'll have to amend it again once Mark's test changes are in, since they add the login callback to the `SyncConfig` and my proposal might change that callback's signature slightly.)

The proposal is overall much smaller now.

src/sync_user.hpp
```diff
@@ -0,0 +1,93 @@
...
+#include <memory>
+#include <unordered_map>
+
+namespace realm {
+
+struct SyncSession;
+
+// A `SyncUser` represents a single user account. Each user manages the sessions that
+// are associated with it.
+class SyncUser {
```
kneth: Realm Java is using `UserStore` as a safe place to store user information: https://github.com/realm/realm-java/blob/master/realm/realm-library/src/objectServer/java/io/realm/UserStore.java

App developers can implement their own `UserStore` but we provide Android specific stores.

kneth:
It might be out of scope, but how well does the session states here match the states in https://github.com/realm/realm-java/blob/master/realm/realm-library/src/objectServer/java/io/realm/SessionState.java

austinzheng:
@kneth The last time I checked:

Object store `WAITING_FOR_ACCESS_TOKEN` is equivalent to your `AUTHENTICATING`.

Object store `ACTIVE` is equivalent to your `BOUND`.

Object store `DYING` has no real equivalent. It's basically "`ACTIVE`, but unless instructed otherwise close the session once all pending uploads and download waiting calls have finished".

Object store `INACTIVE` is your `INITIAL` and `UNBOUND`. We don't have a separate entry state.

Object store `ERROR` is your `STOPPED`, in that it represents an invalid session that cannot be 'revived' or used ever again.

We don't have a separate state for binding. Calling `bind()` happens only in `WAITING_FOR_ACCESS_TOKEN`, and upon returning from that API we move to `ACTIVE`.

austinzheng:
Implemented in #193, so closing. Discussion should take place on that PR.

PR #186, opened 2016-10-05
Fix an issue where the "wait for upload/download" APIs could cause an error

austinzheng:
@tgoyne @bdash 

I pulled this small change out of https://github.com/realm/realm-object-store/pull/181, since it's not related to the deadlocks that issue tries to address.

`bind()` must be called before `wait_for_upload_complete_or_client_stopped()` and `wait_for_download_complete_or_client_stopped()` can be called; this change enforces that precondition

### DIFF
--- a/src/sync_session.cpp
+++ b/src/sync_session.cpp
@@ -418,13 +418,21 @@ void SyncSession::unregister(std::unique_lock<std::mutex>& lock)
     SyncManager::shared().unregister_session(m_realm_path);
 }
 
+bool SyncSession::can_wait_for_network_completion() const
+{
+    return m_state == &State::active || m_state == &State::dying;
+}
+
 void SyncSession::wait_for_upload_completion(std::function<void()> callback)
 {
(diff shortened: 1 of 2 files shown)


src/sync_session.cpp
```diff
@@ -418,13 +418,19 @@ void SyncSession::unregister(std::unique_lock<std::mutex>& lock)
     SyncManager::shared().unregister_session(m_realm_path);
 }
 
+bool SyncSession::can_wait_for_network_completion() const
+{
+    return (m_state == &State::active || m_state == &State::dying);
```
bdash: The parens are unnecessary here.

austinzheng:
I can't wait for the asynchronous replacements to those blocking methods on `sync::Session`. Maybe I'll try implementing them myself later.

bdash:
I'm not sure that asynchronous versions of these methods would address that problem? Presumably `sync::Session` would still care whether the session is bound before you call them.

austinzheng:
I would expect the API to accept callbacks even before binding (with the idea that they'd be queued up until after binding occurs), but I guess that's a design decision the sync team will need to make.

PR #187, opened 2016-10-05
Use unnested VersionID

tgoyne:
Eliminates a bit of cruft done to avoid dragging in group_shared.hpp.

### DIFF
--- a/src/impl/collection_notifier.hpp
+++ b/src/impl/collection_notifier.hpp
@@ -21,7 +21,7 @@
 
 #include "impl/collection_change_builder.hpp"
 
-#include <realm/group_shared.hpp>
+#include <realm/version_id.hpp>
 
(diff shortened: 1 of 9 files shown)


src/impl/collection_notifier.hpp
```diff
@@ -211,7 +213,7 @@ class CollectionNotifier {
 
     // Iteration variable for looping over callbacks
     // remove_callback() updates this when needed
-    size_t m_callback_index = npos;
+    size_t m_callback_index = -1;
```
jpsim: seems like `npos` is still correct here, no?

tgoyne: realm::npos is defined in `array.hpp` which isn't actually supposed to be a public header, and the code using it relies on npos being -1.

Issue #152, opened 2016-08-31
Single thread for all Realm's external commit helper


beeender:
Currently every Realm config needs a thread for external commit helper. This is a resource wasting for users who have more than one Realm in the app. (We have users have 20+ Realms in one app, so it is not a ideal situation for them.)

This becomes an issue for realm-java after https://github.com/realm/realm-java/pull/3370 merged.

bdash:
What resources are you concerned about? The commit helper thread spends most of its life idle, consuming only the small amount of memory used by its stack.

diegomontoya:
@bdash 

1. Each android thread consumes 1024kb stack by default. Pretty sure the 1024kb size must be continuous block, not fragmented memory. Android 4 is really slow to recover/defrag memory so you would get oom even if you have 500mb ram left over on the device.
2. Issue 1 is multiplied in new commit when Realm is constantly opened and closed which cause new monitor thread create/destroy thrashing.

bdash:
Is it address space that's the constrained resource, or physical memory? I'd expect that if threads default to a 1MB stack, that would require 1MB of address space with only a fraction of that backed by physical memory for the portion of the stack that is used.

beeender:
Address space consumption is a problem to us as well. Since we need continuous vm space to do mmap.

diegomontoya:
@bdash the 1024KB stack is vm address and not physical memory. At least based on the following wildly inconsistent test results. Check out the crazy low values for Huawei. Both the 2xhuawei and Samsung s6 are running 6.0.1 yet yield totally different values and slightly different error messages.

What really worries is how the Huaweis, with latest os updates, can have such small value. Did Huawei tweak the vm limits in their custom roms? Something to watch out for.

1) On both Huwei Mate8, 4G, 2.2G free ( (Android 6.0.1)) and Huawei P8, 4G, 2.2G free (Android 6.0.1) I can continue to spawn and run 475 simultaneous threads before:

```
W/libc: pthread_create failed: clone failed: Out of memory
W/art: Throwing OutOfMemoryError "pthread_create (1040KB stack) failed: Out of memory"
```

2)  Samsung S6 6.0.1, 3G, 1.3G free  (Android 6.0.1) the same test can create/run 2400 threads. 

``` 
W/libc: pthread_create failed: couldn't allocate 1069056-bytes mapped space: Out of memory
W/art: Throwing OutOfMemoryError "pthread_create (1040KB stack) failed: Try again
``` 

[...]

PR #191, opened 2016-10-07
fix support for the filesystem which does not allow mkfifo

zaki50:
fix for realm/realm-java#3140

review this please @realm/java @beeender @kneth

### DIFF
--- a/src/impl/android/external_commit_helper.cpp
+++ b/src/impl/android/external_commit_helper.cpp
@@ -110,7 +110,7 @@ ExternalCommitHelper::ExternalCommitHelper(RealmCoordinator& parent)
             // When the fifo already exists, mkfifo fails with error ENOSYS which is not correct.
             // In this case, we use stat to check if the path exists and it is a fifo.
             struct stat stat_buf;
-            if (stat(path.c_str(), &stat_buf) == 0) {
+            if (err == ENOSYS && stat(path.c_str(), &stat_buf) == 0) {
                 if ((stat_buf.st_mode & S_IFMT) != S_IFIFO) {
                     throw std::runtime_error(path + " exists and it is not a fifo.");
                 }
--- a/src/impl/android/external_commit_helper.cpp
+++ b/src/impl/android/external_commit_helper.cpp
@@ -18,6 +18,7 @@
 
 #include "impl/external_commit_helper.hpp"
 #include "impl/realm_coordinator.hpp"
+#include "util/format.hpp"
 
 #include <assert.h>
 #include <errno.h>
@@ -94,11 +95,9 @@ ExternalCommitHelper::ExternalCommitHelper(RealmCoordinator& parent)
             // Hash collisions are okay here because they just result in doing
             // extra work, as opposed to correctness problems
 
-            std::string named_pipe_dir = Realm::get_named_pipe_directory();
+            std::string named_pipe_dir = realm::get_named_pipe_directory();
             if (!named_pipe_dir.empty()) {
-                std::ostringstream ss;
-                ss << named_pipe_dir << "realm_" << std::hash<std::string>()(path) << ".note";
-                path = ss.str();
+                path = util::format("%1realm_%2.note", named_pipe_dir, std::hash<std::string>()(path));
                 ret = mkfifo(path.c_str(), 0600);
                 err = errno;
             }
--- a/src/impl/android/external_commit_helper.cpp
+++ b/src/impl/android/external_commit_helper.cpp
@@ -94,19 +94,10 @@ ExternalCommitHelper::ExternalCommitHelper(RealmCoordinator& parent)
             // Hash collisions are okay here because they just result in doing
             // extra work, as opposed to correctness problems
 
-            std::string tmp_dir = parent.get_temp_dir();
-            if (tmp_dir.empty()) {
-                auto tmp_dir_from_env = getenv("TMPDIR");
-                if (tmp_dir_from_env) {
-                    tmp_dir = std::string(tmp_dir_from_env);
-                }
-            }
-            if (!tmp_dir.empty()) {
+            std::string named_pipe_dir = Realm::get_named_pipe_directory();
+            if (!named_pipe_dir.empty()) {
                 std::ostringstream ss;
-                ss << tmp_dir;
-                if (tmp_dir.back() != '/')
-                    ss << '/';
-                ss << "realm_" << std::hash<std::string>()(path) << ".note";
+                ss << named_pipe_dir << "realm_" << std::hash<std::string>()(path) << ".note";
                 path = ss.str();
                 ret = mkfifo(path.c_str(), 0600);
                 err = errno;
--- a/src/impl/android/external_commit_helper.cpp
+++ b/src/impl/android/external_commit_helper.cpp
@@ -93,16 +93,24 @@ ExternalCommitHelper::ExternalCommitHelper(RealmCoordinator& parent)
             // Filesystem doesn't support named pipes, so try putting it in tmp instead
             // Hash collisions are okay here because they just result in doing
             // extra work, as opposed to correctness problems
-            std::ostringstream ss;
 
             std::string tmp_dir = parent.get_temp_dir();
-            ss << tmp_dir;
-            if (tmp_dir.back() != '/')
-              ss << '/';
-            ss << "realm_" << std::hash<std::string>()(path) << ".note";
-            path = ss.str();
-            ret = mkfifo(path.c_str(), 0600);
-            err = errno;
+            if (tmp_dir.empty()) {
+                auto tmp_dir_from_env = getenv("TMPDIR");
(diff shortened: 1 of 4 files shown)


src/impl/android/external_commit_helper.cpp
```diff
@@ -106,7 +106,18 @@ ExternalCommitHelper::ExternalCommitHelper(RealmCoordinator& parent)
         }
         // the fifo already existing isn't an error
         if (ret == -1 && err != EEXIST) {
-            throw std::system_error(err, std::system_category());
+            // Workaround for a mkfifo bug on Blackberry devices:
+            // When the fifo already exists, mkfifo fails with error ENOSYS which is not correct.
```
bdash: Should we be doing this extra test with `stat` only if the error is `ENOSYS`, rather than for all errors?

zaki50:
most code are copied from https://github.com/realm/realm-core/blob/v2.1.0/src/realm/util/interprocess_condvar.cpp#L100

bdash:
It looks like this code all exists in master and is also used by other binding that can target Android. this PR should really be against master and then be merged to Java's branch.

src/impl/android/external_commit_helper.cpp
zaki50: Right, I'll fix that.

One concern is that the code is from https://github.com/realm/realm-core/blob/v2.1.0/src/realm/util/interprocess_condvar.cpp#L100 and I can't test that.

zaki50:
@bdash OK, I'll update the PR to target the master.

zaki50:
@bdash fixed

src/impl/android/external_commit_helper.cpp
```diff
@@ -89,13 +89,13 @@ ExternalCommitHelper::ExternalCommitHelper(RealmCoordinator& parent)
...
         int err = errno;
-        if (err == ENOTSUP) {
+        if (err == ENOTSUP || err == EACCES) {
             // Filesystem doesn't support named pipes, so try putting it in tmp instead
             // Hash collisions are okay here because they just result in doing
             // extra work, as opposed to correctness problems
             std::ostringstream ss;
 
-            std::string tmp_dir(getenv("TMPDIR"));
+            std::string tmp_dir = parent.get_temp_dir();
```
cmelchior: There doesn't seem to be an error check if this exists? It looks like the `temp_dir` is entirely optional, in which case this is unsafe and should throw an exception i not set?

zaki50: added a check if the `tmp_dir` is empty.

src/impl/android/external_commit_helper.cpp
zaki50: done

zaki50:
All comments are addressed.
Ready for second round review.

src/impl/android/external_commit_helper.cpp
```diff
@@ -89,24 +89,43 @@ ExternalCommitHelper::ExternalCommitHelper(RealmCoordinator& parent)
...
-            if (tmp_dir.back() != '/')
-              ss << '/';
-            ss << "realm_" << std::hash<std::string>()(path) << ".note";
-            path = ss.str();
-            ret = mkfifo(path.c_str(), 0600);
-            err = errno;
+
+            std::string tmp_dir = parent.get_temp_dir();
+            if (tmp_dir.empty()) {
+                auto tmp_dir_from_env = getenv("TMPDIR");
```
cmelchior: What happens if this isn't set? Ideally it should throw an exception?

zaki50: If binding does not pass the temp directory and the `TMPDIR` environment variable also not set, this workaround is just disabled.

If we should throw an exception, I think it should be done at the initialization phase, not here.

Should the Object Store force to set the path of the temp dir?

beeender:
I really think the `temp_dir` should be a static var instead of a member var of per config.
The reason is i am thinking about #152, one possible solution could be a single named pipe for all Realms in the same process for notifications, and notify the specific Realm instance by checking `is_changed()`. That single named piped has to be created on the internal storage for the similar reason.

src/impl/android/external_commit_helper.cpp
```diff
@@ -89,24 +89,43 @@ ExternalCommitHelper::ExternalCommitHelper(RealmCoordinator& parent)
...
+                if (tmp_dir_from_env) {
+                    tmp_dir = std::string(tmp_dir_from_env);
+                }
+            }
+            if (!tmp_dir.empty()) {
+                std::ostringstream ss;
+                ss << tmp_dir;
+                if (tmp_dir.back() != '/')
```
kneth: Not all operating system use `/` as path separator.

cmelchior:
Having the `temp_dir` as a static var makes sense I think. Also having it required to be set when using the OS.

src/impl/android/external_commit_helper.cpp
zaki50: Right, but this helper class is for android only and all android uses / as a separator.

zaki50:
@cmelchior All other platforms require this workaround?

cmelchior:
I don't know. We could also throw the exception when initializing, I don't have any strong opinions on that, as long as we do throw somewhere. But the closer to initialization the better.

zaki50:
I updated the code to address the comment https://github.com/realm/realm-object-store/pull/191#issuecomment-252531089

Please give a final review @realm/java @bdash @tgoyne

src/shared_realm.cpp
```diff
@@ -33,9 +33,28 @@
...
+{
+    auto tmp_dir = getenv("TMPDIR");
+    if (!tmp_dir) {
+        return std::string();
+    }
+    std::string tmp_dir_str(tmp_dir);
+    if (tmp_dir_str.back() == '/') {
+        return tmp_dir_str;
+    }
+    std::ostringstream ss;
```
tgoyne: There's no reason to use a stringstream just to append a character; just do `tmp_dir_str += '/';`.

src/shared_realm.cpp
```diff
@@ -33,9 +33,28 @@
...
+    ss << tmp_dir_str << '/';
+    return ss.str();
+}
+
+std::string Realm::named_pipe_directory = get_initial_named_pipe_directory();
```
tgoyne: There's no reason for this to be a member of `Realm`.

src/shared_realm.cpp
```diff
@@ -181,6 +201,22 @@ Group& Realm::read_group()
...
 
+void Realm::set_named_pipe_directory(std::string named_pipe_directory)
+{
+    if (named_pipe_directory.empty()) {
+        throw std::invalid_argument("'named_pipe_directory` is empty.");
+    }
+    if (named_pipe_directory.back() != '/') {
+        throw std::invalid_argument("'named_pipe_directory` must end with '/'.");
+    }
+    Realm::named_pipe_directory = named_pipe_directory;
```
tgoyne: `std::move(named_pipe_directory)`

src/shared_realm.cpp
```diff
@@ -181,6 +201,22 @@ Group& Realm::read_group()
...
+}
+
+std::string& Realm::get_named_pipe_directory()
```
tgoyne: This should return a const reference.

src/shared_realm.hpp
```diff
@@ -155,6 +155,12 @@ class Realm : public std::enable_shared_from_this<Realm> {
         bool automatic_change_notifications = true;
     };
 
+    // Sets a path to a directory where Realm can write named pipes.
+    // This string should include a trailing slash '/'.
+    static void set_named_pipe_directory(std::string named_pipe_directory); // throw
```
tgoyne: `// throws` comments are used at the callsite. The lack of `noexcept` here indicates that it throws so there's no need to explicitly note that.

src/impl/android/external_commit_helper.cpp
```diff
@@ -89,24 +89,34 @@ ExternalCommitHelper::ExternalCommitHelper(RealmCoordinator& parent)
...
-              ss << '/';
-            ss << "realm_" << std::hash<std::string>()(path) << ".note";
-            path = ss.str();
-            ret = mkfifo(path.c_str(), 0600);
-            err = errno;
+
+            std::string named_pipe_dir = Realm::get_named_pipe_directory();
+            if (!named_pipe_dir.empty()) {
+                std::ostringstream ss;
+                ss << named_pipe_dir << "realm_" << std::hash<std::string>()(path) << ".note";
```
tgoyne: Using `util::format()` here rather than a stringstream will slightly reduce the size of the compiled binary and make it a bit more concise.

zaki50:
@tgoyne thanks! I'll update the PR

zaki50:
@tgoyne done

zaki50:
I'll merge this.
If there are further comments, I'll address them in another issue.

PR #192, opened 2016-10-10
Set the schema properly when no changes are needed in ResetFile mode

tgoyne:
The Realm's schema needs to be set to the passed-in schema even if no changes are needed as the schema from disk may include additional object types that are not in the requested schema. This was already done properly for all schema modes except for ResetFile.

See https://github.com/realm/realm-cocoa/issues/4190.

### DIFF
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -211,12 +211,18 @@ bool Realm::read_schema_from_group_if_needed()
     return true;
 }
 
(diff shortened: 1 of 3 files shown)


bdash:
Using `Fixes #XYZ` in the description of this fix will result in the realm-cocoa issue being closed when this is merged into object store master. It'd be better for the issue to remain open until realm-cocoa updates to a version of the object store that contains this fix.

PR #188, opened 2016-10-05
Add some very basic sync tests

bdash:
This is primarily to get the infrastructure in place to allow testing of new changes.

/cc @austinzheng @tgoyne @jpsim

PR #193, opened 2016-10-12
Refactor object store to handle sync functionality

austinzheng:
EDIT: now an actual review.

I went ahead and tried this refactoring in order to simplify the Cocoa binding. It follows the latest version of the API described in #185 pretty closely, but with changes and additions that became necessary during implementation.

In terms of validation: both Mark's (very small) object store level sync test suite, as well as the Cocoa test suite for the accompanying bindings change, pass. 

Changes:
- Added `SyncUser` as an object-store level type representing a user
- Changed `SyncConfig` to take a pointer to a `SyncUser` rather than a tag string
- Added APIs to `SyncManager` to get, create, manage, and store users
- `SyncSession` is now a class rather than a struct, for consistency
- `SyncSession` now starts in the `Inactive` state, and waits for its associated user to ask it to begin the binding process
- Added some code to `SyncSession` and its states to support directly binding the session with an 'admin token'
- Renamed `SyncLogInHandler` to `SyncBindSessionHandler` and added a pointer to `SyncSession` as an argument, allowing bindings to not need to store add'l state
- `SyncManager` now has getters for some of its configuration properties
- The object store now transparently manages persistence metadata
- The object store now manages file-related operations (`sync_file.cpp`)
- `SyncManager` can now be partially reset, only for testing purposes
- Modified cmake files to link with yaml-cpp


bmunkholm:
Since sync is a highly additive and separate functionality, I would suggest to make that clear even with the organization of files - place them in a separate sync folder? Maybe even namespace?

austinzheng:
Yeah, a separate folder and namespace seems like a really good idea.

austinzheng:
People who might be interested:
@bdash @tgoyne @jpsim 
@alazier @cmelchior @kneth

src/impl/realm_coordinator.cpp
```diff
@@ -27,9 +27,9 @@
 #include "schema.hpp"
 
 #if REALM_ENABLE_SYNC
-#include "sync_config.hpp"
-#include "sync_manager.hpp"
-#include "sync_session.hpp"
+#include "../sync/sync_config.hpp"
+#include "../sync/sync_manager.hpp"
+#include "../sync/sync_session.hpp"
```
simonask: Wouldn't this work without the `..`?

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,262 @@
...
+}
+
+static char decoded_char_for(const std::string& percent_encoding, size_t index)
+{
+    if (index+2 >= percent_encoding.length()) {
+        throw std::invalid_argument("Malformed string: not enough characters after '%' before end of string.");
+    }
+    REALM_ASSERT(percent_encoding[index] == '%');
+    return (16*value_of_hex_digit(percent_encoding[index + 1])) + value_of_hex_digit(percent_encoding[index + 2]);
+}
```
simonask: Suggest putting these functions in an anonymous namespace instead of using `static`, which is more consistent with the rest of our codebase. There's no semantic difference, but `static` is a little C-like. :-P

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+void remove_nonempty_dir(const std::string& path)
+{
+#ifdef _WIN32
+    // Not implemented yet.
+    REALM_ASSERT(false);
```
simonask: Could use `static_assert` or even `#error` to produce a compile-time error.

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,262 @@
...
+    }
+    while (struct dirent *file = readdir(dir_listing)) {
+        auto file_type = file->d_type;
+        std::string file_name = file->d_name;
+        if (file_name == "." || file_name == "..") {
+            continue;
+        }
+        if (file_type == DT_REG || file_type == DT_FIFO) {
+            // Regular file (this really shouldn't ever throw, so if it does don't bother catching the exception).
+            File::remove(file_path_by_appending_component(path, file_name));
```
simonask: Really? I'd say this is quite likely to throw, since there is a race condition with `readdir()` - i.e. someone else might have gone ahead and removed the file already between seeing it through `readdir()` and reaching this point in the code. :-)

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,262 @@
...
+        } else if (file_type == DT_DIR) {
+            // Directory, recurse
+            remove_nonempty_dir(file_path_by_appending_component(path, file_name, true));
```
simonask: Same.

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,262 @@
...
+        }
+    }
+    closedir(dir_listing);
+    // Delete the directory itself
+    util::remove_dir(path);
```
simonask: Same.

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,262 @@
...
+    }
+    char path_last = path[path.length() - 1];
+    char component_first = component[0];
+    if (path_last == '/' && component_first == '/') {
+        return path + component.substr(1) + terminal;
+    } else if (path_last == '/' || component_first == '/') {
+        return path + component + terminal;
+    } else {
+        return path + "/" + component + terminal;
+    }
```
simonask: There's a bit of heap allocation in temporaries here, consider reserving the memory up front.

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,262 @@
...
+    char path_last = path[path.length() - 1];
+    char extension_first = extension[0];
+    if (path_last == '.' && extension_first == '.') {
+        return path + extension.substr(1);
+    } else if (path_last == '.' || extension_first == '.') {
+        return path + extension;
+    } else {
+        return path + "." + extension;
+    }
+}
```
simonask: Same.

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+std::string SyncFileManager::get_utility_directory()
+{
+    auto util_path = file_path_by_appending_component(get_base_sync_directory(), c_utility_directory, true);
```
simonask: Consider naming the boolean argument in a local variable to ease readability (we do that all over the Core and Sync codebases).

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,262 @@
...
+}
+
+std::string SyncFileManager::get_base_sync_directory()
+{
+    auto sync_path = file_path_by_appending_component(m_base_path, c_sync_directory, true);
+    util::try_make_dir(sync_path);
+    return sync_path;
+}
+
+std::string SyncFileManager::user_directory(std::string user_identity)
```
simonask: Should this be `const std::string&` instead of by value?

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,262 @@
...
+{
+    if (user_identity.length() == 0) {
+        throw std::invalid_argument("user_identity cannot be an empty string");
+    }
+    auto user_path = file_path_by_appending_component(get_base_sync_directory(), std::move(user_identity), true);
+    util::try_make_dir(user_path);
+    return user_path;
+}
+
+void SyncFileManager::remove_user_directory(std::string user_identity)
```
simonask: By-ref/by-val?

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,262 @@
...
+        throw std::invalid_argument("user_identity cannot be an empty string");
+    }
+    auto user_path = file_path_by_appending_component(get_base_sync_directory(), std::move(user_identity), true);
+    try {
+        remove_nonempty_dir(user_path);
+    } catch(File::AccessError) {
+    }
+}
+
+bool SyncFileManager::remove_realm(std::string user_identity, std::string raw_realm_path)
```
simonask: By-ref/by-val?

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,262 @@
...
+    try {
+        remove_nonempty_dir(management_path);
+    } catch(File::NotFound) {
+    } catch(File::AccessError) {
+        success = false;
+    }
+    return success;
+}
+
+std::string SyncFileManager::path(std::string user_identity, std::string raw_realm_path)
```
simonask: By-ref/by-val?

src/sync/impl/sync_file.hpp
```diff
@@ -0,0 +1,79 @@
...
+/// Given a file path and a path component, return a new path created by appending the component to the path.
+std::string file_path_by_appending_component(const std::string& path,
+                                             const std::string& component,
+                                             bool is_directory=false);
+
+/// Given a file path and an extension, append the extension to the path.
+std::string file_path_by_appending_extension(const std::string& path, const std::string& extension);
+
+/// Remove a directory, including non-empty directories.
+void remove_nonempty_dir(const std::string& path);
```
simonask: Consider putting all of these functions in the `realm::util` namespace. :)

src/sync/impl/sync_file.hpp
```diff
@@ -0,0 +1,79 @@
...
+    /// Remove the metadata Realm.
+    bool remove_metadata_realm();
+
+private:
+    std::string m_base_path;
+
+    const std::string c_sync_directory = "realm-object-server";
+    const std::string c_utility_directory = "io.realm.object-server-utility";
+    const std::string c_metadata_directory = "metadata";
+    const std::string c_metadata_realm = "sync_metadata.realm";
```
simonask: Instead of `const std::string`, which will be initialized and heap-allocated in the constructor of this class, why not make them globally static and allocation-less, like so:

```c++
static constexpr const char c_sync_directory[] = "realm-object-server";
```

src/sync/sync_manager.cpp
```diff
@@ -32,6 +35,73 @@ SyncManager& SyncManager::shared()
...
+                                                                       std::move(custom_encryption_key));
+            break;
+        case MetadataMode::NoMetadata:
+            return;
+    }
+
+    REALM_ASSERT(m_metadata_manager);
+    // Load persisted users into the users map.
+    SyncUserMetadataResults users = m_metadata_manager->all_unmarked_users();
+    for (size_t i = 0; i < users.size(); i++) {
```
simonask: Consider using foreach-syntax (`for (auto& user_data: users)`).

src/sync/sync_manager.cpp
```diff
@@ -32,6 +35,73 @@ SyncManager& SyncManager::shared()
...
+        auto identity = user_data.identity();
+        if (user_token) {
+            users_to_add.emplace_back(std::make_tuple(identity,
+                                                      std::make_shared<SyncUser>(*user_token,
+                                                                                 std::move(identity),
+                                                                                 false)));
+        }
+    }
+    // Delete any users marked for death.
+    std::vector<SyncUserMetadata> dead_users;
```
simonask: Reserve `users_to_remove.size()` elements.

src/sync/sync_manager.cpp
```diff
@@ -32,6 +35,73 @@ SyncManager& SyncManager::shared()
...
+    SyncUserMetadataResults users_to_remove = m_metadata_manager->all_users_marked_for_removal();
+    for (size_t i = 0; i < users_to_remove.size(); i++) {
+        auto user = users_to_remove.get(i);
+        // FIXME: delete user data in a different way? (This deletes a logged-out user's data as soon as the app
+        // launches again, which might not be how some apps want to treat their data.)
+        m_file_manager->remove_user_directory(user.identity());
+        dead_users.emplace_back(std::move(user));
+    }
+    for (auto user : dead_users) {
```
simonask: Iteration by copy, perhaps you meant `auto&`?

src/sync/sync_manager.cpp
```diff
@@ -32,6 +35,73 @@ SyncManager& SyncManager::shared()
...
+        user.remove();
+    }
+    {
+        std::lock_guard<std::mutex> lock(m_user_mutex);
+        for (auto tuple : users_to_add) {
```
simonask: Iteration by copy.

src/sync/sync_manager.cpp
```diff
@@ -68,10 +138,71 @@ void SyncManager::set_client_should_reconnect_immediately(bool reconnect_immedia
...
+        return nullptr;
+    }
+    auto ptr = it->second;
+    return (ptr->state() == SyncUser::State::Active ? ptr : nullptr);
+}
+
+std::vector<std::shared_ptr<SyncUser>> SyncManager::all_users() const
+{
+    std::lock_guard<std::mutex> lock(m_user_mutex);
+    std::vector<std::shared_ptr<SyncUser>> users;
```
simonask: Presuming that being in the `Error` state is rare, `reserve()` `users.size()` elements.

src/sync/sync_manager.hpp
```diff
@@ -49,22 +52,70 @@ class SyncLoggerFactory {
...
     void set_client_should_reconnect_immediately(bool reconnect_immediately);
+    bool client_should_reconnect_immediately() const;
+
     /// Control whether the sync client validates SSL certificates. Should *always* be `true` in production use.
-    void set_client_should_validate_ssl(bool validate_ssl);
+    void set_client_should_validate_ssl(bool validate_ssl)
+    {
+        std::lock_guard<std::mutex> lock(m_mutex);
+        m_client_validate_ssl = validate_ssl;
+    }
```
simonask: I prefer inline functions outside the body of the class definition. That's the general style in Core and Sync (with some exceptions for trivial cases). It makes it easier to read the interface of a class without thinking too much about the implementation. :-)

src/sync/sync_manager.hpp
```diff
@@ -49,22 +52,70 @@ class SyncLoggerFactory {
...
+        return m_client_validate_ssl;
+    }
 
     std::shared_ptr<SyncSession> get_session(const std::string& path, const SyncConfig& config);
     std::shared_ptr<SyncSession> get_existing_active_session(const std::string& path) const;
 
+    util::Logger::Level log_level() const noexcept {
+        std::lock_guard<std::mutex> lock(m_mutex);
+        return m_log_level;
+    }
```
simonask: Note that locking the `std::mutex` is not `noexcept`, so this risks terminating the program in case there is a mutex error (which may be the reasonable thing to do).

simonask: See http://en.cppreference.com/w/cpp/thread/mutex/lock

src/sync/sync_manager.hpp
```diff
@@ -92,11 +143,20 @@ friend struct SyncSession;
     sync::Client::Reconnect m_client_reconnect_mode = sync::Client::Reconnect::normal;
     bool m_client_validate_ssl = true;
 
+    // Protects m_users
+    mutable std::mutex m_user_mutex;
```
simonask: Is that protection necessary? Do we have multiple threads interacting with this class?

If so, can they also safely interact with the contents of the `SyncUser` objects?

src/sync/sync_session.cpp
```diff
@@ -382,15 +404,17 @@ void SyncSession::nonsync_transact_notify(sync::Session::version_type version)
...
         std::unique_lock<std::mutex> lock(m_state_mutex);
         if (m_state->revive_if_needed(lock, *this)) {
-            log_in_handler = m_config.log_in_handler;
+            handler = m_config.bind_session_handler;
         }
     }
-    if (log_in_handler)
-        log_in_handler.value()(m_realm_path, m_config);
+    if (handler) {
+        REALM_ASSERT(shared_from_this());
```
simonask: This may indicate a poor design, or maybe I'm not familiar enough with `std::shared_ptr` idioms. When can this situation occur?

src/sync/sync_user.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+std::vector<std::shared_ptr<SyncSession>> SyncUser::all_sessions()
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
+    std::vector<std::string> keys_to_remove;
+    std::vector<std::shared_ptr<SyncSession>> sessions;
+    if (m_state == State::Error) {
+        return sessions;
+    }
+    for (auto it : m_sessions) {
```
simonask: Iteration by value.

src/sync/sync_user.cpp
```diff
@@ -0,0 +1,166 @@
...
+        if (auto ptr = it.second.lock()) {
+            if (ptr->is_valid()) {
+                sessions.emplace_back(std::move(ptr));
+                continue;
+            }
+        }
+        keys_to_remove.emplace_back(it.first);
+    }
+    for (auto key : keys_to_remove) {
```
simonask: Iteration by value.

src/sync/sync_user.cpp
```diff
@@ -0,0 +1,166 @@
...
+        case State::Error:
+            return;
+        case State::Active:
+            m_refresh_token = std::move(token);
+            break;
+        case State::LoggedOut: {
+            m_refresh_token = std::move(token);
+            m_state = State::Active;
+            // (Re)activate all pending sessions.
+            for (auto pair : m_waiting_sessions) {
```
simonask: Iteration by value.

src/sync/sync_user.cpp
```diff
@@ -0,0 +1,166 @@
...
+                if (auto ptr = pair.second.lock()) {
+                    m_sessions[pair.first] = ptr;
+                    lock.unlock();
```
simonask: A comment about why this is necessary/good.

src/sync/sync_user.cpp
```diff
@@ -0,0 +1,166 @@
...
+        return;
+    }
+    std::lock_guard<std::mutex> lock(m_mutex);
+    if (m_state == State::LoggedOut) {
+        return;
+    }
+    m_state = State::LoggedOut;
+    // Move all active sessions into the waiting sessions pool. If the user is
+    // logged back in, they will automatically be reactivated.
+    for (auto pair : m_sessions) {
```
simonask: Iteration by value.

src/sync/sync_user.cpp
```diff
@@ -0,0 +1,166 @@
...
+        if (auto ptr = pair.second.lock()) {
+            ptr->log_out();
+            m_waiting_sessions[pair.first] = ptr;
+        }
+    }
+    m_sessions.clear();
+    if (auto manager = SyncManager::shared().metadata_manager()) {
+        // Mark the user as 'dead' in the persisted metadata Realm.
+        auto metadata = SyncUserMetadata(*manager, m_identity, false);
+        metadata.mark_for_removal();
```
simonask: This code seems weird - so `SyncUserMetadata::mark_for_removal()` has side-effects beyond the life of its own instance? Perhaps that method should live on another object?

src/sync/sync_user.hpp
```diff
@@ -0,0 +1,117 @@
...
+    // Whether the user was configured as an 'admin user' (directly uses its user token
+    // to open Realms).
+    bool is_admin() const
+    {
+        return m_is_admin;
+    }
+
+    std::string identity() const
+    {
+        return m_identity;
```
simonask: Why is it safe for these to return directly, while other methods have to acquire a lock?

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+void SyncFileManager::remove_user_directory(std::string user_identity)
+{
+    if (user_identity.length() == 0) {
+        throw std::invalid_argument("user_identity cannot be an empty string");
+    }
+    auto user_path = file_path_by_appending_component(get_base_sync_directory(), std::move(user_identity), true);
+    try {
+        remove_nonempty_dir(user_path);
+    } catch(File::AccessError) {
```
simonask: Catch on new line.

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,262 @@
...
+    } catch(File::NotFound) {
+    } catch(File::AccessError) {
+        success = false;
+    }
+    // Remove the management directory (e.g. "example.realm.management").
+    auto management_path = file_path_by_appending_extension(realm_path, "management");
+    try {
+        remove_nonempty_dir(management_path);
+    } catch(File::NotFound) {
+    } catch(File::AccessError) {
```
simonask: Catches on new lines. :)

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,262 @@
...
+    return file_path_by_appending_component(std::move(dir_path), c_metadata_realm);
+}
+
+bool SyncFileManager::remove_metadata_realm()
+{
+    auto dir_path = file_path_by_appending_component(get_utility_directory(), c_metadata_directory, true);
+    try {
+        remove_nonempty_dir(dir_path);
+        return true;
+    } catch(File::AccessError) {
```
simonask: Catch on new line.

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,262 @@
...
+                throw std::invalid_argument("Input string is invalid: contains reserved characters.");
+            }
+            buffer.push_back(current);
+            idx++;
+        }
+    }
+    return buffer;
+}
+
+std::string file_path_by_appending_component(const std::string& path, const std::string& component, bool is_directory)
```
kneth: FAT32 is used on many Android devices. FAT32 has a few restrictions on the length. I'm not sure if this is the place to check for it.

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,262 @@
...
+{
+    auto sync_path = file_path_by_appending_component(m_base_path, c_sync_directory, true);
+    util::try_make_dir(sync_path);
+    return sync_path;
+}
+
+std::string SyncFileManager::user_directory(std::string user_identity)
+{
+    if (user_identity.length() == 0) {
+        throw std::invalid_argument("user_identity cannot be an empty string");
```
kneth: `user_identify` does not mean anything for the user. Please consider an alternative formulation.

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,262 @@
...
+    }
+    auto user_path = file_path_by_appending_component(get_base_sync_directory(), std::move(user_identity), true);
+    util::try_make_dir(user_path);
+    return user_path;
+}
+
+void SyncFileManager::remove_user_directory(std::string user_identity)
+{
+    if (user_identity.length() == 0) {
+        throw std::invalid_argument("user_identity cannot be an empty string");
```
kneth: Like above

src/sync/impl/sync_file.cpp
kneth: Is it required that this method can remove Realms create by older versions too?

src/sync/sync_manager.cpp
```diff
@@ -32,6 +35,73 @@ SyncManager& SyncManager::shared()
...
+            users_to_add.emplace_back(std::make_tuple(identity,
+                                                      std::make_shared<SyncUser>(*user_token,
+                                                                                 std::move(identity),
+                                                                                 false)));
+        }
+    }
+    // Delete any users marked for death.
+    std::vector<SyncUserMetadata> dead_users;
+    SyncUserMetadataResults users_to_remove = m_metadata_manager->all_users_marked_for_removal();
+    for (size_t i = 0; i < users_to_remove.size(); i++) {
```
kneth: for each style? like below.

src/sync/sync_manager.cpp
```diff
@@ -68,10 +138,71 @@ void SyncManager::set_client_should_reconnect_immediately(bool reconnect_immedia
...
+    return (ptr->state() == SyncUser::State::Active ? ptr : nullptr);
+}
+
+std::vector<std::shared_ptr<SyncUser>> SyncManager::all_users() const
+{
+    std::lock_guard<std::mutex> lock(m_user_mutex);
+    std::vector<std::shared_ptr<SyncUser>> users;
+    for (auto it : m_users) {
+        auto user = it.second;
+        if (user->state() != SyncUser::State::Error)
```
kneth: Please consider to add `{ ... }`

src/impl/realm_coordinator.cpp
tgoyne: All includes in OS files should be relative to the `src` directory

src/sync/sync_manager.hpp
tgoyne: There's probably some exceptions, but the general pattern in the OS code has been to either have everything inline (when the separate declaration of the class is not useful), or to have only the things which fit comfortably on one line inline (i.e. trivial getters/setters).

src/sync/sync_manager.hpp
tgoyne: AFAIK all mutex errors that can actually happen are very much terminate-worthy.

src/sync/impl/sync_file.cpp
tgoyne: Even better would be to use a two-case enum since it doesn't look like the value passed in is ever the result of a logical expression.

src/sync/sync_user.cpp
tgoyne: `it` also is an incorrect name as the value isn't an iterator.

tgoyne: This could use `std::remove_if()` to avoid building the temporary vector of keys and the hash table lookups to remove them:

```
auto it = std::remove_if(m_sessions.begin(), m_sessions.end(), [&](auto& session) {
    auto ptr = session.second.lock();
    if (ptr && ptr->is_valid()) {
        sessions.push_back(std::move(ptr));
        return false;
    }
    return true;
});
m_sessions.erase(it, m_sessions.end());
```

austinzheng: `it` is short for iterator. That makes total sense, I can't believe I didn't realize it before.

src/sync/impl/sync_file.cpp
austinzheng: Good to know. I'll prefer anonymous namespaces over `static` in the future.

src/sync/impl/sync_file.cpp
austinzheng: Yeah, most (if not all) these functions should take references.

src/sync/impl/sync_file.hpp
austinzheng: Oh, this is good to know. Thanks!

src/sync/sync_manager.hpp
austinzheng: Alright, that's a good heuristic. Thanks for the clarification!

src/sync/sync_session.cpp
austinzheng: After doing some research, I realized that this doesn't do what I think it does and I'll remove it. Thanks for pointing it out.

(It was intended as a sanity check, but it's useless since: 1. the way sessions are used is such that they can never exist without being already managed by a `shared_ptr`; 2. if no `shared_ptr` to the session already existed that method wouldn't return null, it would just do something undefined.)

src/sync/sync_user.cpp
austinzheng: There was an issue with this snippet of code that I noticed at your prompting. I've rewritten it to be correct.

src/sync/sync_user.hpp
austinzheng: `is_admin` and `identity` are internally set at creation and considered immutable. `state` should be protected by the lock.

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,262 @@
...
+        throw std::invalid_argument("Cannot get the value of a character that isn't a hex digit.");
+    }
+}
+
+static bool character_is_unreserved(char character)
+{
+    bool is_capital_letter = (character >= 'A' && character <= 'Z');
+    bool is_lowercase_letter = (character >= 'a' && character <= 'z');
+    bool is_number = (character >= '0' && character <= '9');
+    bool is_allowed_symbol = (character == '-' || character == '_');
```
austinzheng: `.` is allowed in URLs, but I think it has special significance for certain filesystems. Should it be allowed? (I'm afraid of someone specifying a Realm URL ending in something like `foo.lock` and causing an error.)

src/sync/sync_user.cpp
austinzheng: `metadata` is basically a barebones Realm object, and `mark_for_removal()` just sets a property on it. If you want I'm happy to sit down with you at our upcoming shindig and explain the metadata (and other) code I wrote to you in person.

src/sync/sync_manager.cpp
austinzheng: My custom collection type doesn't support it yet. I'll add it as a followup item.

src/sync/sync_manager.cpp
austinzheng: This type of collection doesn't support foreach iteration. I hope to add support in a followup item, and will change the loops when that happens.

src/sync/impl/sync_file.cpp
austinzheng: Right now this method is only used for Realms that the sync manager manages, so not at the moment. If people are interested in using it to remove files generated by older versions of Realm I'll modify the method.

src/sync/impl/sync_file.cpp
austinzheng: I changed these to `REALM_ASSERT`s, since they are preconditions that the binding is expected to fulfill.

src/sync/impl/sync_file.cpp
austinzheng: I'll file a ticket to ensure we properly respect name/path length restrictions for FAT32.

austinzheng:
I'm going to do some work to rebase this onto `master`. Things might get messy for a bit.

austinzheng:
Okay.

I rebased this branch onto master. In doing so, I pulled in the minimal necessary changes from the `cocoa` branch to get the sync-related functionality working.

Reviewers can ignore or skim the following files:

* Anything to do with cmake
* Any C++ source or header file not in the `src/sync` directory
* Any of the files in `tests/util`

Let me know if there are any questions.

src/sync/sync_manager.cpp
```diff
@@ -0,0 +1,341 @@
...
+    static SyncManager& manager = *new SyncManager;
+    return manager;
+}
+
+void SyncManager::configure_file_system(const std::string& base_file_path,
+                                        MetadataMode metadata_mode,
+                                        util::Optional<std::vector<char>> custom_encryption_key)
+{
+    std::vector<std::tuple<std::string, std::shared_ptr<SyncUser>>> users_to_add;
+    REALM_ASSERT(!m_file_manager && !m_metadata_manager);   // Can only call this once
```
alazier: This doesn't work for react-native as you can refresh an app causing this to get called multiple times. We need to allow multiple calls to this and to cleanup appropriately when this occurs.

alazier: No preference

alazier:
I haven't reviewed this thoroughly, but I have integrated these changes into JS and user persistence and realm path mapping seem to be working well. 👍

austinzheng:
Ari reports that last night's results weren't valid, but he has gotten the JS binding (actually) working with sync today, modulo a few bugs that need to be fixed. I will spend some time writing tests now.

austinzheng:
We have two working bindings using this codebase and our object store and binding tests are passing. Does anyone have any objection to me merging this in tomorrow?

simonask:
Why are all those unrelated changes in the diff? If they're already in master, you can merge master into this branch to remove them from the diff :)

austinzheng:
(I guess my original message didn't go through.)

A lot of the foundational work that led up to this PR was done on the `cocoa` branch, rather than `master`, and had to be pulled in as part of this PR. This was a mistake on my part.

Here are some of the merged PRs whose code is included as part of this one:

https://github.com/realm/realm-object-store/pull/188 (basic object server tests)
https://github.com/realm/realm-object-store/pull/186 (bugfixes)
https://github.com/realm/realm-object-store/pull/182 (don't fail silently when trying to use sync without sync enabled)
https://github.com/realm/realm-object-store-private/pull/41 (support for encrypting the metadata Realm)
https://github.com/realm/realm-object-store-private/pull/38 (session)
https://github.com/realm/realm-object-store-private/pull/20 (add support for sync metadata)
https://github.com/realm/realm-object-store-private/pull/13 (initial changes to object store to accommodate sync)

simonask:
It did go through, but I still don't think it's necessary to have all those changes displayed as part of the diff, and it's a significant burden to have to sift through that. If the PRs are merged, merge master into this one to remove that code from the diff view. :-)

src/impl/realm_coordinator.cpp
```diff
@@ -53,16 +59,46 @@ std::shared_ptr<RealmCoordinator> RealmCoordinator::get_coordinator(StringData p
...
+    SyncSession::Internal::set_sync_transact_callback(*m_sync_session,
+                                                      [weak_self](VersionID old_version, VersionID new_version) {
+        if (auto self = weak_self.lock()) {
+            if (self->m_transaction_callback)
+                self->m_transaction_callback(old_version, new_version);
+            self->notify_others();
+        }
+    });
+    if (m_config.sync_config->error_handler) {
+        SyncSession::Internal::set_error_handler(*m_sync_session, m_config.sync_config->error_handler);
```
kneth: For Realm Java, `m_config.sync_config->error_handler` will be a callback to a Java method - just like https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/cpp/objectserver_shared.hpp#L56

src/impl/realm_coordinator.cpp
```diff
@@ -82,9 +118,38 @@ std::shared_ptr<Realm> RealmCoordinator::get_realm(Realm::Config config)
         if (m_config.schema_version != config.schema_version && config.schema_version != ObjectStore::NotVersioned) {
             throw MismatchedConfigException("Realm at path '%1' already opened with different schema version.", config.path);
         }
+
+#if REALM_ENABLE_SYNC
+        if (bool(m_config.sync_config) != bool(config.sync_config)) {
```
kneth: I'm definitely not a C++ programmer, but how is `sync_config` casted to `bool`?

src/shared_realm.cpp
```diff
@@ -175,7 +179,21 @@ void Realm::open_with_config(const Config& config,
             read_only_group = std::make_unique<Group>(config.path, config.encryption_key.data(), Group::mode_ReadOnly);
         }
         else {
-            history = realm::make_in_realm_history(config.path);
+            // FIXME: The SharedGroup constructor, when called below, will
+            // throw a C++ exception if server_synchronization_mode is
+            // inconsistent with the accessed Realm file. This exception
+            // probably has to be transmuted to an NSError.
```
kneth: Just throw a C++ exception and let Realm Cocoa, Realm Java, etc. handles it. Maybe add a special exception for it.

src/sync/impl/sync_client.hpp
```diff
@@ -0,0 +1,70 @@
...
+    {
+        client.stop();
+        m_thread.join();
+    }
+
+private:
+    static sync::Client make_client(util::Logger& logger, Reconnect reconnect_mode, bool verify_ssl)
+    {
+        sync::Client::Config config;
+        config.logger = &logger;
```
kneth: Realm Java implements a logger which writes to the Android log.

src/sync/impl/sync_file.cpp
kneth: According to the (internal) protocol specification, `.` is allowed.

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,305 @@
...
+{
+    REALM_ASSERT(user_identity.length() > 0);
+    auto user_path = file_path_by_appending_component(get_base_sync_directory(),
+                                                      user_identity,
+                                                      util::FilePathType
+                                                      ::Directory);
+    try {
+        util::remove_nonempty_dir(user_path);
+    }
+    catch(File::AccessError) {
```
kneth: Why silently ignore `AccessError`?

src/shared_realm.cpp
austinzheng: I'll change the FIXME to be Cocoa specific

austinzheng: Hmm...it probably doesn't. If it's not too involved, I'll just go ahead and fix it.

src/impl/realm_coordinator.cpp
austinzheng: Yeah, that looks good. Are there any changes I need to make to the API to accommodate your use case?

CMake/RealmCore.cmake
```diff
@@ -167,3 +167,80 @@ function(build_existing_realm_core core_directory)
 
     define_built_realm_core_target(${core_directory})
 endfunction()
+
+function(build_realm_sync sync_directory)
```
austinzheng: Does Android use cmake to build the object store? You might need to adjust your build configuration.

austinzheng:
@bdash explained some things to me, and I'm going to make some changes to this PR.

First, I'm going to rebase it back onto `cocoa`. This won't affect the content of the PR, but it will make all the irrelevant files that showed up to bloat it go away. I think this should address your concerns, @simonask.

If/when we get it approved and merged, then I'll merge `cocoa` into `master`. Since `cocoa` should differ from `master` only in terms of the sync-related code, this shouldn't be too difficult. If there are other changes we'll merge them in first.

@alazier if we can't get this approved in time for you, please let me know and I'll push my backup branch to origin. You can use that branch as a basis for the JS stuff.

CMake/RealmCore.cmake
```diff
@@ -237,9 +237,9 @@ function(build_realm_sync sync_directory)
     set_property(TARGET realm-sync-server PROPERTY IMPORTED_LOCATION_RELEASE ${sync_server_library_release})
     set_property(TARGET realm-sync-server PROPERTY IMPORTED_LOCATION ${sync_server_library_release})
     if(APPLE)
-        set_property(TARGET realm-sync-server PROPERTY INTERFACE_LINK_LIBRARIES ${FOUNDATION} ${SECURITY})
+        set_property(TARGET realm-sync-server PROPERTY INTERFACE_LINK_LIBRARIES ${FOUNDATION} ${SECURITY} -lyaml-cpp)
```
bdash: If we're going to do this then realm-sync should really be changed to _require_ yaml-cpp, rather than only using it if it can find it. As it stands we're in this weird state where realm-sync will use yaml-cpp if it can find it, but there's no way for clients of realm-sync to know whether they need to link against yaml-cpp.

tgoyne: realm-sync really should just output a pkg-config file rather than requiring consumers to keep track of its dependencies.

austinzheng: I created https://github.com/realm/realm-sync/issues/960 to track this.

alazier:
When can we get this merged? Would like these changes for tomorrow's release if possible.

austinzheng:
I'd prefer not to rush the reviewers. I can cut you a special branch later today if necessary.

austinzheng:
Almost all the new changes are from old PRs previously reviewed and merged into the `cocoa` branch. This is mostly my fault; if I had more foresight I would have developed on `master` to begin with instead of on `cocoa`.

Let me know if you want specific PRs.

CMake/RealmCore.cmake
tgoyne: In the meantime this can check if yaml-cpp is installed and only try to link it if so with `find_package(PkgConfig); pkg_check_modules(YAML QUIET yaml-cpp); set_property(TARGET realm-sync-server PROPERTY INTERFACE_LINK_LIBRARIES ${FOUNDATION} ${SECURITY} ${YAML_LDFLAGS})`.

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,316 @@
...
+    }
+    while (struct dirent *file = readdir(dir_listing)) {
+        auto file_type = file->d_type;
+        std::string file_name = file->d_name;
+        if (file_name == "." || file_name == "..") {
+            continue;
+        }
+        if (file_type == DT_REG || file_type == DT_FIFO) {
+            try {
+                File::remove(file_path_by_appending_component(path, file_name));
```
tgoyne: `File::try_remove()` returns false rather than throwing if the file doesn't exist.

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,316 @@
...
+{
+#ifdef _WIN32
+    static_assert(false, "Not implemented for Win32 yet");
+#else
+    // Open the directory and list all the files.
+    DIR *dir_listing = opendir(path.c_str());
+    if (!dir_listing) {
+        return;
+    }
+    while (struct dirent *file = readdir(dir_listing)) {
```
tgoyne: `auto cleanup = util::make_scope_exit([=]() noexcept { closedir(dir_listing); });` here rather than the call to `closedir()` below will make it so that this doesn't leak the directory handle if something below throws.

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,316 @@
...
+std::string make_percent_encoded_string(const std::string& raw_string)
+{
+    std::string buffer;
+    buffer.reserve(raw_string.length());
+    for (char character : raw_string) {
+        if (character_is_unreserved(character)) {
+            buffer.push_back(character);
+        } else {
+            char b[4] = {0, 0, 0, 0};
+            snprintf(b, 4, "%%%2X", character);
```
tgoyne: Using `snprintf()` here is somewhere between redundant and incorrect. It's redundant in that the format string can't produce something that doesn't fit in the buffer, and incorrect in that it doesn't prevent buffer overflows if the buffer size is incorrect. `snprintf()` doesn't null-terminate the string if the string being printed doesn't fit into the buffer, so `buffer.append(b);` would read past the end.

Separately from that, this can also just do the formating directly into the output buffer: `buffer.resize(buffer.size() + 3); snprintf(&buffer.back() - 3, 3, "%%%2X", character);`

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,316 @@
...
+    REALM_ASSERT(raw_realm_path.length() > 0);
+    if (filename_is_reserved(user_identity) || filename_is_reserved(raw_realm_path)) {
+        throw std::invalid_argument("A user or Realm can't have an identifier reserved by the filesystem.");
+    }
+    auto escaped = util::make_percent_encoded_string(raw_realm_path);
+    auto realm_path = util::file_path_by_appending_component(user_directory(user_identity), std::move(escaped));
+    bool success = true;
+    // Remove the base Realm file (e.g. "example.realm").
+    try {
+        File::remove(realm_path);
```
tgoyne: `try_remove()` for all of these.

src/sync/impl/sync_file.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+bool SyncFileManager::remove_realm(const std::string& user_identity, const std::string& raw_realm_path)
+{
+    REALM_ASSERT(user_identity.length() > 0);
+    REALM_ASSERT(raw_realm_path.length() > 0);
+    if (filename_is_reserved(user_identity) || filename_is_reserved(raw_realm_path)) {
+        throw std::invalid_argument("A user or Realm can't have an identifier reserved by the filesystem.");
+    }
+    auto escaped = util::make_percent_encoded_string(raw_realm_path);
+    auto realm_path = util::file_path_by_appending_component(user_directory(user_identity), std::move(escaped));
```
tgoyne: No reason to move here as `file_path_by_appending_component` takes a const reference.

src/sync/impl/sync_file.hpp
```diff
@@ -0,0 +1,91 @@
...
+namespace realm {
+
+namespace util {
+
+enum class FilePathType {
+    File, Directory
+};
+
+/// Given a string, turn it into a percent-encoded string.
+std::string make_percent_encoded_string(const std::string& raw_string);
```
tgoyne: I suspect all of these could take `realm::StringData` as arguments rather than `std::string`.

src/sync/sync_session.cpp
```diff
@@ -286,13 +301,12 @@ void SyncSession::create_sync_session()
...
         using Error = realm::sync::Error;
 
-        SyncSessionError error_type;
+        SyncSessionError error_type = SyncSessionError::Debug;
         // Precondition: error_code is a valid realm::sync::Error raw value.
         Error strong_code = static_cast<Error>(error_code);
 
         switch (strong_code) {
             // Client errors; all ignored (for now)
-            case Error::invalid_error:
```
tgoyne: Merge error? These error types still exist.

src/sync/sync_user.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+SyncUser::SyncUser(std::string refresh_token,
+                   std::string identity,
+                   util::Optional<std::string> server_url,
+                   bool is_admin)
+: m_state(State::Active)
+, m_server_url(server_url.value_or(""))
+, m_is_admin(is_admin)
+, m_refresh_token(refresh_token)
+, m_identity(identity)
```
tgoyne: Can move token and identity.

src/sync/impl/sync_file.cpp
austinzheng: Thanks for the explanation, knowing that `snprintf()` doesn't terminate the string if the buffer size is wrong is pretty important.

tests/sync/file.cpp
```diff
@@ -0,0 +1,208 @@
...
+#include "sync_test_utils.hpp"
+
+#include "shared_realm.hpp"
+#include <realm/util/file.hpp>
+
+using namespace realm;
+using namespace realm::util;
+using File = realm::util::File;
+
+static const std::string base_path = "/tmp/realm_objectstore_sync_file/";
```
tgoyne: This should use `$TMPDIR` rather than being hardcoded to `/tmp` as `TestFile` does.

tests/sync/metadata.cpp
```diff
@@ -0,0 +1,249 @@
...
+			SyncMetadataManager first_manager(metadata_path, true, make_test_encryption_key(10));
+			REQUIRE_THROWS(SyncMetadataManager(metadata_path, true, make_test_encryption_key(11)));
+		}
+		SECTION("different encryption settings") {
+			SyncMetadataManager first_manager(metadata_path, true, make_test_encryption_key(10));
+			REQUIRE_THROWS(SyncMetadataManager(metadata_path, false));
+		}
+	}
+
+	SECTION("works when enabled") {
```
tgoyne: This should also test that the metadata file can be reopened with the same key.

tests/sync/session.cpp
```diff
@@ -60,10 +61,12 @@ std::shared_ptr<SyncSession> sync_session(SyncServer& server, const std::string&
 
 TEST_CASE("sync: log-in", "[sync]") {
     SyncServer server;
+    // Disable file-related functionality and metadata functionality for testing purposes.
+    SyncManager::shared().configure_file_system("/tmp/", SyncManager::MetadataMode::NoMetadata);
```
tgoyne: Also should use `$TMPDIR`.

tests/sync/sync_manager.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+static const std::string base_path = "/tmp/realm_objectstore_sync_manager/";
+
+namespace {
+
+bool validate_user_in_vector(std::vector<std::shared_ptr<SyncUser>> vector,
+						     const std::string& identity,
+						     const std::string& url,
+						     const std::string& token) {
+	for (auto& user : vector) {
```
tgoyne: This file is indented with tabs.

tests/sync/sync_manager.cpp
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	SECTION("should work for 'should validate SSL'") {
+		SyncManager::shared().set_client_should_validate_ssl(true);
+		REQUIRE(SyncManager::shared().client_should_validate_ssl());
+		SyncManager::shared().set_client_should_validate_ssl(false);
+		REQUIRE(!SyncManager::shared().client_should_validate_ssl());
+	}
+
+	SyncManager::shared().reset_for_testing();
```
tgoyne: The call to `reset_for_testing()` should probably be always done as a scope_exit call set up immediately after the `TEST_CASE` line. Calling it manually at the end of the test case means that it won't get called if any of the tests fail (presumably resulting in all of the subsequent tests failing and adding  a lot of noise to the report), and calling it within each `SECTION` as is done in some of the tests is redundant (only a single leaf SECTION is run per execution of the test case).

tests/sync/sync_test_utils.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include "sync_test_utils.hpp"
+
+namespace realm {
+
+bool create_dummy_realm(std::string path) {
+    Realm::Config config;
+    config.path = std::move(path);
+    try {
+        return Realm::make_shared_realm(config) != nullptr;
```
tgoyne: `make_shared_realm()` can't return null, so this check doesn't make much sense.

tests/sync/sync_test_utils.cpp
```diff
@@ -0,0 +1,56 @@
...
+    } catch(std::exception) {
+        return false;
+    }
+}
+
+void reset_test_directory(const std::string& base_path) {
+	remove_nonempty_dir(base_path);
```
tgoyne: More tabs.

tests/sync/sync_test_utils.hpp
```diff
@@ -0,0 +1,51 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#include "catch.hpp"
+
+#include "sync/impl/sync_file.hpp"
+#include "sync/impl/sync_metadata.hpp"
+
+#ifndef REALM_SYNC_TEST_UTILS_HPP
```
tgoyne: Include guard should start before the includes.

tests/sync/sync_test_utils.hpp
```diff
@@ -0,0 +1,51 @@
...
+bool results_contains_user(SyncUserMetadataResults& results, const std::string& identity);
+std::vector<char> make_test_encryption_key(const char start = 0);
+
+} // realm
+
+#define REQUIRE_DIR_EXISTS(macro_path) \
+{ \
+    DIR *dir_listing = opendir((macro_path).c_str()); \
+    REQUIRE(dir_listing); \
+    if (dir_listing) closedir(dir_listing); \
```
tgoyne: The `if` is redundant as `REQUIRE` exits the function if the condition is false.

src/sync/impl/sync_file.hpp
austinzheng: I'll add a FIXME and try doing this in another PR.

src/sync/sync_session.cpp
austinzheng: Yes, definitely a merge error. Thanks for catching this.

tgoyne:
Running the tests for this with thread sanitizer gives "WARNING: ThreadSanitizer: use of an invalid mutex (e.g. uninitialized or destroyed)".

You can build with tsan with `cmake -DREALM_CORE_VERSION=/path/to/core -DREALM_SYNC_PREFIX=/path/to/sync -DSANITIZE_THREAD=1 . && make && tests/tests 'sync*'`

[Full output from tsan with backtraces](https://gist.github.com/tgoyne/4377f939b616f69f067b8692d6a8af4d).

austinzheng:
Thanks for running these tests with tsan, and for showing me how to do it! There are a bunch of follow-up tests I wrote which ran into mutex errors, so this will be really helpful.

austinzheng:
I think this PR might be close to, if not ready to be merged. The tsan issue has been fixed in the latest commit and I don't think there are any outstanding review issues. Happy to wait if anyone else wants to take another look, though.

kneth:
@austinzheng Please check if the code adhere to our coding standards: https://github.com/realm/realm-core/blob/master/doc/development/coding_style_guide.cpp

Except that, please go ahead and merge.

simonask:
(Just to note: Most of the code in the Object Store does not adhere to the Core coding style, as far as I have seen. It's more important to be in line with surrounding code than the Core guideline IMO, but YMMV.)

austinzheng:
Thanks to all!

WRT coding style: @bdash's advice on a previous object store review was to follow the style of the surrounding code. I'd be happy to try converting the object store codebase to use core code style later this year, if people think that would be a good idea.

alazier:
Awesome! When can I get this on master/js?

austinzheng:
I'm about to merge `cocoa` into `master`. Just eyeballing the changes to make sure nothing bad happened.

PR #194, opened 2016-10-12
Initial jenkinsfile with docker

alebsack:
This PR is against cocoa simply because we (@alebsack and @fealebenpae) tried it earlier and gets us to the test phase.

Closes realm/realm-sync#909


fealebenpae:
I'm not sure this catches all existing errors right now. For instance, CentOS devtoolset-3's gcc [errors out](https://ci.realm.io/job/realm/job/realm-js-private/job/sync/140/execution/node/82/log/) on [this code](https://github.com/realm/realm-object-store/4e3679542d9d71649eeb38fc58f9c4f4dec9b8f1/cocoa/src/impl/realm_coordinator.cpp#L423-L426).

CentOS/RHEL is the oldest Linux we need to support and do builds of the Node.js binding on so I think the Linux CI for Object Store should run in a similar environment, or at least one where the same known errors are evinced. Otherwise we risk getting false positives.

CMake/RealmCore.cmake
```diff
@@ -40,7 +40,7 @@ endif()
 
 function(use_realm_core version_or_path_to_source)
     if("${version_or_path_to_source}" MATCHES "^[0-9]+(\\.[0-9])+")
-        if(APPLE OR REALM_PLATFORM STREQUAL "Android")
+        if(APPLE OR LINUX OR REALM_PLATFORM STREQUAL "Android")
```
jpsim: isn't `LINUX` also set for Android?

fealebenpae: I think it is, but the way we define `LINUX` is kind of hacky and overly broad. I think it should be `CMAKE_SYSTEM_NAME STREQUAL "Linux"` but I'm not sure that would actually include Android if you were cross-compiling on a Darwin host. Need to check the toolchain file.

alebsack:
@fealebenpae I've "updated" the build environment to use el6.  I also redefined the `LINUX` check as described.

fealebenpae:
Thanks @alebsack! The same error I linked to started appearing here, which is reassuring.

alebsack:
I'm going to step away from the PR for now, as the harness for compiling linux is complete.

alebsack:
It builds now.  💯 However, picking up Cobertura coverage reports is not supported by the pipeline plugin.  I'll look into how we can get around that.

alebsack:
I've added coverage report processing to mirror what we currently have in the "ObjectStore Pull Request" job.  I think we should now consider this as complete as the mac build.

alebsack:
Closes realm/realm-sync#909

.gitignore
```diff
@@ -12,9 +12,12 @@ realm-object-store.xcodeproj/
 
 # Build products
 src/librealm-object-store.a
+src/librealm-object-store-static.a
+src/librealm-object-store.dylib
```
bdash: Why?

docker-build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
bdash: Does this file really need to be at the root of the source tree?

src/CMakeLists.txt
```diff
@@ -62,7 +62,7 @@ set(HEADERS
 
 if(APPLE)
     list(APPEND SOURCES impl/apple/external_commit_helper.cpp impl/apple/keychain_helper.cpp)
-elseif(REALM_PLATFORM STREQUAL "Android")
+elseif(LINUX OR REALM_PLATFORM STREQUAL "Android")
     list(APPEND SOURCES impl/android/external_commit_helper.cpp)
```
bdash: Seems surprising for a file with "android" in the path to be used on anything but Android.

tests/CMakeLists.txt
```diff
@@ -27,9 +27,20 @@ if(REALM_ENABLE_SYNC)
...
 
+find_package(Threads)
+find_package(PkgConfig REQUIRED)
+pkg_search_module(OPENSSL REQUIRED openssl)
+
+if(OPENSSL_FOUND)
+    include_directories(${OPENSSL_INCLUDE_DIRS})
+    message(STATUS "Using OpenSSL ${OPENSSL_VERSION}")
+else()
+    # Error; with REQUIRED, pkg_search_module() will throw an error by it's own
```
bdash: Why bother checking `OPENSSL_FOUND` at all then?

src/CMakeLists.txt
fealebenpae: It's not Android-specific - we can rework it to say `epoll/external_commit_helper.cpp` or `linux/external_commit_helper.cpp` to just make it more general.

tests/CMakeLists.txt
```diff
@@ -27,9 +27,20 @@ if(REALM_ENABLE_SYNC)
     list(APPEND SOURCES sync.cpp)
 endif()
 
+find_package(Threads)
+find_package(PkgConfig REQUIRED)
+pkg_search_module(OPENSSL REQUIRED openssl)
```
tgoyne: Why is this adding a hard dependency on OpenSSL here? The tests don't use OpenSSL directly on any platform. If the existing logic for making OpenSSL a dependency of the objectstore library is incorrect then that should be updated instead.

fealebenpae: I believe this is needed to satisfy the OpenSSL link-time dependency in sync.

fealebenpae: So it does, thank you for pointing it out. Cases like these are why we asked for help in the first place.

docker-build.sh
```diff
@@ -0,0 +1,10 @@
...
+
+set -e
+rm -rf CMakeCache.txt
+find . -name CMakeFiles | xargs rm -rf
```
kneth: `-delete` instead of the pipe?

CMake/RealmCore.cmake
```diff
@@ -113,6 +113,15 @@ function(download_realm_core core_version)
     set_property(TARGET realm PROPERTY IMPORTED_LOCATION_RELEASE ${core_library_release})
     set_property(TARGET realm PROPERTY IMPORTED_LOCATION ${core_library_release})
 
+    if (LINUX)
+        find_package(Threads)
+        find_package(PkgConfig REQUIRED)
+        pkg_search_module(OPENSSL REQUIRED openssl)
+        include_directories(${OPENSSL_INCLUDE_DIRS})
```
fealebenpae: We shouldn't need the OpenSSL include directories.

CMake/RealmCore.cmake
```diff
@@ -113,6 +113,15 @@ function(download_realm_core core_version)
     set_property(TARGET realm PROPERTY IMPORTED_LOCATION_RELEASE ${core_library_release})
     set_property(TARGET realm PROPERTY IMPORTED_LOCATION ${core_library_release})
 
+    if (LINUX)
+        find_package(Threads)
+        find_package(PkgConfig REQUIRED)
+        pkg_search_module(OPENSSL REQUIRED openssl)
```
fealebenpae: You can get away with calling `find_package(OpenSSL)` directly, I believe.

cmelchior:
@alebsack It this still being worked on? We need to get CI running for Java as well and I would rather not modify the Jenkins directly.

alebsack:
@cmelchior My part of it (the Jenkinsfile/docker part) is done.  I suppose we need to fix the conflicts now, though. I'll work on that shortly.

.gitignore
```diff
@@ -12,9 +12,11 @@ realm-object-store.xcodeproj/
 
 # Build products
 src/librealm-object-store.a
+src/librealm-object-store-static.a
```
bdash: This shouldn't be necessary after 2412932253f25ad6c19ef6ba8c5273d7e101bd24 (late August).

CMake/RealmCore.cmake
bdash: The two aren't equivalent AFAIK, but `find_package(OpenSSL REQUIRED)` seems like it should be sufficient. Is there a reason `pkg_search_module` was used instead?

CMake/RealmCore.cmake
```diff
@@ -109,6 +113,14 @@ function(download_realm_core core_version)
     set_property(TARGET realm PROPERTY IMPORTED_LOCATION_RELEASE ${core_library_release})
     set_property(TARGET realm PROPERTY IMPORTED_LOCATION ${core_library_release})
 
+    if (LINUX)
+        find_package(Threads)
```
bdash: This should really be done unconditionally. The use of threads isn't specific to Linux.

CMake/RealmCore.cmake
```diff
@@ -109,6 +113,14 @@ function(download_realm_core core_version)
...
+        find_package(PkgConfig REQUIRED)
+        pkg_search_module(OPENSSL REQUIRED openssl)
+        message(STATUS "Using OpenSSL ${OPENSSL_VERSION}")
+        set_property(TARGET realm PROPERTY INTERFACE_LINK_LIBRARIES ${CMAKE_THREAD_LIBS_INIT} ${OPENSSL_LIBRARIES})
```
bdash: This also needs to be done in `define_built_realm_core_target` so that the correct libraries are linked in when building core and sync from source.

CMake/RealmCore.cmake
```diff
@@ -182,7 +194,7 @@ function(build_realm_sync sync_directory)
         BUILD_IN_SOURCE 1
         BUILD_ALWAYS 1
         CONFIGURE_COMMAND ""
-        BUILD_COMMAND make -C src/realm librealm-sync.a librealm-sync-dbg.a librealm-server.a librealm-server-dbg.a ${MAKE_FLAGS}
+        BUILD_COMMAND CPATH=${REALM_CORE_INCLUDE_DIR} make -C src/realm librealm-sync.a librealm-sync-dbg.a librealm-server.a librealm-server-dbg.a ${MAKE_FLAGS}
```
bdash: What does this do? AFAIK CPATH is something that CMake looks at. Nothing in core's build system appears to use it.

src/impl/realm_coordinator.cpp
```diff
@@ -420,10 +420,11 @@ class IncrementalChangeInfo {
...
             transaction::advance(m_sg, *m_current, version);
-            m_info.push_back({
-                m_current->table_modifications_needed,
-                m_current->table_moves_needed,
-                std::move(m_current->lists)});
+            TransactionChangeInfo info;
+            info.table_modifications_needed = m_current->table_modifications_needed;
+            info.table_moves_needed = m_current->table_moves_needed;
+            info.lists = std::move(m_current->lists);
+            m_info.push_back(std::move(info));
```
bdash: This change, and some others, should be superseded by what is in #206.

tests/sync.cpp
```diff
@@ -64,8 +64,8 @@ TEST_CASE("sync: log-in", "[sync]") {
     SECTION("Can log in") {
         std::atomic<int> error_count(0);
         auto session = sync_session(server, "user", "/test",
-                                    [&](auto...) { return s_test_token; },
-                                    [&](auto...) { ++error_count; });
+                                    [&](const std::string&, const std::string&) { return s_test_token; },
```
bdash: This is unfortunate. Which compiler version doesn't support variadic generic lambdas?

workflow/local_docker_build.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+# This script can be used to locally check and debug
+# the linux build process outside of CI.
+# This should be run from the root directory `packaging/manual_docker_build.sh`
```
bdash: This path is wrong.

CMake/RealmCore.cmake
bdash: I think a lot of the `LINUX` checks should really be either checking CMake's existing `UNIX` variable or something more specific, such as detecting whether epoll is supported.

bdash:
I've extracted the Linux build fixes from this into #209, and reworked them per my review comments above. We should be able to rebase the CI changes in this PR on top of master once #209 is merged.

alebsack:
Replaced by #211

PR #196, opened 2016-10-19
Pull over sync changes

alazier:
@fealebenpae

### DIFF
--- a/CMake/RealmCore.cmake
+++ b/CMake/RealmCore.cmake
@@ -167,3 +167,80 @@ function(build_existing_realm_core core_directory)
 
     define_built_realm_core_target(${core_directory})
 endfunction()
+
+function(build_realm_sync sync_directory)
+    get_filename_component(sync_directory ${sync_directory} ABSOLUTE)
+    if(APPLE)
+        find_library(FOUNDATION Foundation)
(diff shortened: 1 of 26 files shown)


fealebenpae:
Looking at the `al-js-sync` branch it doesn't look like it actually contains any sync changes that would be included in its tip. What's more, GitHub says this PR is basically change-neutral.

I'm curious about https://github.com/realm/realm-object-store/commit/158a45a7a50ff2273c70e75d84d15a6a9d51402b, though - it says I'm the author but you pushed it and I have no recollection of authoring such a commit. Did you `git commit --author`? Also, that one commit looks like it undoes all the sync changes in Object Store, why would it be there?

fealebenpae:
Also, probably someone more familiar with sync in Object Store should take a look at this PR.

alazier:
@fealebenpae not sure what is going on with the changes in this pr. Just made this because it is what you had been asking for. Perhaps I messed up when I was pushing across remote.

alazier:
This does seem a bit wrong though...

alazier:
Ok think i fixed it...

fealebenpae:
Yep, this looks better. 👍 FWIW

PR #197, opened 2016-10-20
Use a single thread for ExternalCommitHelper

### LINKED ISSUE #152: Single thread for all Realm's external commit helper

beeender:
Fix #152
We have users using multiple Realm files in one apk which creates some
overhead if every Realm file needs a dedicate thread for listening
notifications.
So we create a single thread for listening and check where is the change
from by calling SharedGroup's has_changed().


beeender:
It actually needs to open some more file descriptors
```
realm-tes 17805   cc   16uR  REG    8,3     1160 44042356 /home/cc/repo/realm/realm-core/test/Transactions_RollbackMoveTableColumns.1.path.realm.lock
realm-tes 17805   cc   17u  FIFO    8,3      0t0 44042362 /home/cc/repo/realm/realm-core/test/Transactions_RollbackMoveTableColumns.1.path.realm.management/access_control.new_commit.cv
realm-tes 17805   cc   18r  FIFO    8,3      0t0 44042362 /home/cc/repo/realm/realm-core/test/Transactions_RollbackMoveTableColumns.1.path.realm.management/access_control.new_commit.cv
```
(Android doesn't enable REALM_ASYNC_DAEMON so, only above `new_commit` pipe is needed for emulated condition var @finnschiermer please correct me if i am wrong.)

Before we need a named pipe for every Realm file in the `ExternalCommitHelper` which requires 2 file descriptors. Now it is  `3 descriptors` x `Realm file count` + `2 descriptors fro the fifo in tmp dir` .

So it is like adding one more file descriptor for every Realm file. I guess it is acceptable?

bdash:
What motivates using a single FIFO and `SharedGroup::has_changed` rather than multiple FIFOs managed by a single thread? Why is this new approach preferable?

beeender:
Multiple FIFOs would be a good idea. I will do  that then. Thanks for pointing out.

beeender:
@fealebenpae @bdash Please help to review again. Thanks.

src/impl/epoll/external_commit_helper.cpp
```diff
@@ -131,16 +129,51 @@ ExternalCommitHelper::ExternalCommitHelper(RealmCoordinator& parent)
         throw std::system_error(errno, std::system_category());
     }
 
+    if (!s_daemon_thread) {
+        s_daemon_thread = std::make_unique<DaemonThread>();
+    }
+    // Lock is inside add_commit_helper. The fifo and thread creation is fine since RealmCoordinator has lock when
+    // calling this.
```
bdash: Relying on details of how the caller performs locking seems fragile. My inclination would be to change the daemon thread to not be destroyed after it's created, and to rely on the thread-safe initialization of function-level statics to guard the initialization of the thread object. This would remove all reliance on external synchronization and simplify this code.

src/impl/epoll/external_commit_helper.hpp
```diff
@@ -53,22 +57,38 @@ class ExternalCommitHelper {
         FdHolder(FdHolder const&) = delete;
     };
 
-    void listen();
+    class DaemonThread {
```
bdash: This could be declared in the .cpp file since nothing in the header needs more than a forward declaration.

src/impl/epoll/external_commit_helper.hpp
```diff
@@ -16,7 +16,11 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
+#include <memory>
 #include <thread>
+#include <vector>
+
+#include <realm/group_shared.hpp>
```
bdash: This seems to be unneeded.

src/impl/epoll/external_commit_helper.hpp
beeender: After tired again i remembered why i didn't do it at the first place. `DaemonThread` also needs `FdHolder` which is declared as a private inner class -- I failed to get it compiled :( I could move the `FdHolder` outside and `DaemonThread` to the cpp file, but that also exposes `FdHolder`. What do you prefer? @bdash

src/impl/epoll/external_commit_helper.cpp
```diff
@@ -198,14 +253,22 @@ void ExternalCommitHelper::listen()
...
             return;
         }
-        assert(ev.data.u32 == (uint32_t)m_notify_fd);
 
-        m_parent.on_change();
+        {
+            std::lock_guard<std::mutex> lock(m_mutex);
+            for (auto it : m_helper_list) {
+                if (ev.data.u32 == (uint32_t)it->m_notify_fd) {
+                    it->m_parent.on_change();
```
beeender: It seems in the notifier thread, create/free external_commit_helper should never happen. Added an assert to detect that.

src/impl/epoll/external_commit_helper.hpp
bdash: Something like https://gist.github.com/bdash/9fab3dbe65449b324ae773adb6fc74c5 is what I had in mind.

src/impl/epoll/external_commit_helper.cpp
```diff
@@ -154,42 +178,77 @@ ExternalCommitHelper::ExternalCommitHelper(RealmCoordinator& parent)
...
 
-    event[1].events = EPOLLIN;
-    event[1].data.fd = m_shutdown_read_fd;
-    ret = epoll_ctl(m_epfd, EPOLL_CTL_ADD, m_shutdown_read_fd, &event[1]);
-    assert(ret == 0);
+    // In kernel versions before 2.6.9, the EPOLL_CTL_DEL operation required a non-NULL pointer in event, even
+    // though this argument is ignored. See man page of epoll_ctl.
+    epoll_ctl(m_epoll_fd, EPOLL_CTL_DEL, helper->m_notify_fd, &event);
+
+    return m_helpers.empty();
```
bdash: This can probably be removed now.

PR #200, opened 2016-10-21
Add support for suppressing specific local notifications from a write

tgoyne:
This adds `NotificationToken::suppress_next()` which can be called within a write transaction to make the write transaction not produce a notification for that specific callback. All other callbacks (including others for the same collection) will still happen. The rest of the changes are support functionality to make this possible.

All methods of advancing the read transaction version will now send notifications, including beginning a write transaction and explicitly calling `Realm::refresh()`. If a hard refresh is required and the notifications for that version are not ready yet, it will block until they are. The two exceptions to this are advances due to importing handover objects (just not implemented yet), and advancements due to beginning a write transaction from within a notification callback (fundamentally not possible to do in any sensible way). This also fixes a lot of edge cases and race conditions along the way which could have resulted in notifications not being delivered when they should have been.

### DIFF
--- a/src/impl/collection_notifier.hpp
+++ b/src/impl/collection_notifier.hpp
@@ -194,7 +194,7 @@ public:
 
     bool have_callbacks() const noexcept { return m_have_callbacks; }
 protected:
-    void add_changes(CollectionChangeBuilder change) { m_accumulated_changes.merge(std::move(change)); }
+    void add_changes(CollectionChangeBuilder change);
     void set_table(Table const& table);
     std::unique_lock<std::mutex> lock_target();
 
@@ -215,13 +215,12 @@ private:
 
     bool m_has_run = false;
     bool m_error = false;
-    CollectionChangeBuilder m_accumulated_changes;
-    CollectionChangeSet m_changes_to_deliver;
-
     std::vector<DeepChangeChecker::RelatedTable> m_related_tables;
 
     struct Callback {
         CollectionChangeCallback fn;
+        CollectionChangeBuilder accumulated_changes;
+        CollectionChangeSet changes_to_deliver;
         size_t token;
         bool initial_delivered;
     };
@@ -242,6 +241,9 @@ private:
     size_t m_callback_index = -1;
 
     CollectionChangeCallback next_callback(bool has_changes, bool pre);
+
+    template<typename Fn>
+    void for_each_callback(Fn&& fn);
 };
 
 // A smart pointer to a CollectionNotifier that unregisters the notifier when
--- a/src/impl/collection_notifier.hpp
+++ b/src/impl/collection_notifier.hpp
@@ -138,6 +138,7 @@ public:
 
     // Get the SharedGroup version which this collection can attach to (if it's
     // in handover mode), or can deliver to (if it's been handed over to the BG worker alredad)
+    // precondition: RealmCoordinator::m_notifier_mutex is locked
     VersionID version() const noexcept { return m_sg_version; }
 
     // Release references to all core types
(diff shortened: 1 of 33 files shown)


tgoyne:
@bdash All of the issues with this that I'm aware of are now fixed, so I'm done mucking about with it for now.

tests/handover.cpp
```diff
@@ -114,11 +115,13 @@ TEST_CASE("handover") {
     }
 
     SECTION("version mismatch") {
+        auto coordinator = _impl::RealmCoordinator::get_existing_coordinator(config.path);
         SECTION("import into older version") {
             r->begin_transaction();
             Object num = create_object(r, int_object);
             num.row().set_int(0, 7);
             r->commit_transaction();
+            coordinator->on_change();
```
jpsim: Why is it now necessary to call `on_change()` where it previously wasn't?

src/impl/collection_notifier.hpp
```diff
@@ -145,44 +151,54 @@ class CollectionNotifier {
...
     void prepare_handover();
 
     template <typename T>
     class Handle;
 
-protected:
     bool have_callbacks() const noexcept { return m_have_callbacks; }
-    void add_changes(CollectionChangeBuilder change) { m_accumulated_changes.merge(std::move(change)); }
+protected:
+    void add_changes(CollectionChangeBuilder change);// { m_accumulated_changes.merge(std::move(change)); }
```
jpsim: Implementation is commented out. On purpose?

src/impl/collection_notifier.hpp
```diff
@@ -268,6 +288,41 @@ class CollectionNotifier::Handle : public std::shared_ptr<T> {
...
+    // Get the version which this package can deliver into, or VersionID{} if
+    // it has not yet been packaged
+    util::Optional<VersionID> version() { return m_version; }
+
+    // Package the notifiers for delivery, blocking if they aren't ready for
+    // the given version.
+    // No-op if called multiple times
+    void package_and_wait(util::Optional<VersionID::version_type> target_version);
+
+    // Sent the before-change notifications
```
jpsim: "Send"?

src/impl/collection_notifier.hpp
```diff
@@ -268,6 +288,41 @@ class CollectionNotifier::Handle : public std::shared_ptr<T> {
...
+    void before_advance();
+    // Deliver the payload associated with the contained notifiers and/or the error
+    void deliver(SharedGroup& sg);
+    // Sent the after-change notifications
```
jpsim: "Send"?

src/impl/realm_coordinator.cpp
```diff
@@ -88,7 +89,8 @@ void RealmCoordinator::create_sync_session()
         if (auto self = weak_self.lock()) {
             if (self->m_transaction_callback)
                 self->m_transaction_callback(old_version, new_version);
-            self->notify_others();
+            if (self->m_notifier)
```
jpsim: `self->wake_up_notifier_worker()`?

tests/list.cpp
```diff
@@ -308,46 +345,53 @@ TEST_CASE("list") {
         }
 
         SECTION("tables-of-interest are tracked properly for multiple source versions") {
-            List lst2(r, other_lv);
-
-            // Add a callback for list1, advance the version, then add a
-            // callback for list2, so that the notifiers added at each source
-            // version have different tables watched for modifications
+            // Add notifiers for different tables at different versions to verify
+            // that the tables of interest are updated correct as we process new
```
bdash: s/correct/correctly/

src/impl/realm_coordinator.cpp
tgoyne: While `wake_up_notifier_worker()` does the same thing as this, it actually shouldn't and I've added a FIXME there to reflect that.

src/impl/transact_log_handler.cpp
```diff
@@ -267,12 +279,18 @@ class TransactLogObserver : public TransactLogValidationMixin, public MarkDirtyM
...
+            m_notifiers.after_advance();
             return;
         }
 
         std::sort(begin(m_observers), end(m_observers));
         func(*this);
-        context->did_change(m_observers, invalidated);
+        if (context)
+            context->did_change(m_observers, invalidated, old_version != sg.get_version_of_current_transaction());
+        m_notifiers.package_and_wait(sg.get_version_of_current_transaction().version); // is a no-op if parse_complete() was called
```
bdash: If `parse_complete()` is not called it looks like we'll call `deliver` / `after_advance` without first calling `begin_advance`. Is that ok?

tests/handover.cpp
```diff
@@ -114,6 +115,7 @@ TEST_CASE("handover") {
     }
 
     SECTION("version mismatch") {
+        auto coordinator = _impl::RealmCoordinator::get_existing_coordinator(config.path);
```
bdash: Why?

PR #202, opened 2016-10-25
allocate static resources on the heap

alazier:
@tgoyne


jpsim:
Shouldn't this PR target `master` and then be merged to other branches, like `js`?

alazier:
@bdash your suggestion seems to work

PR #203, opened 2016-10-27
Improve object server test suite

austinzheng:
@alazier @bdash @tgoyne 

This PR doesn't fix any bugs, it just adds some more sync-related tests and improves a few peripheral things. So there's no rush to get it reviewed.

Changes:
- Added more tests examining the interaction between `SyncUser` and `SyncSession`
- Renamed `SyncSession`'s `is_inactive()` API to `can_be_safely_destroyed()` to better reflect its purpose
- Added a `SyncSession` API to examine the current state, and removed the `is_valid()` API
- Tagged existing sync-related tests with the `[sync]` tag

### DIFF
--- a/src/sync/sync_user.cpp
+++ b/src/sync/sync_user.cpp
@@ -50,9 +50,9 @@ std::vector<std::shared_ptr<SyncSession>> SyncUser::all_sessions()
         return sessions;
     }
     for (auto it = m_sessions.begin(); it != m_sessions.end();) {
-        if (auto ptr = it->second.lock()) {
-            if (ptr->is_valid()) {
-                sessions.emplace_back(std::move(ptr));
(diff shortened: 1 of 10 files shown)


src/sync/sync_user.cpp
```diff
@@ -51,7 +51,7 @@ std::vector<std::shared_ptr<SyncSession>> SyncUser::all_sessions()
     }
     for (auto it = m_sessions.begin(); it != m_sessions.end();) {
         if (auto ptr = it->second.lock()) {
-            if (ptr->is_valid()) {
+            if (ptr->state() != SyncSession::PublicState::Error) {
```
jpsim: `if (!session_is_errored(ptr))`?

PR #209, opened 2016-11-10
Fix the Linux build

bdash:
Contains a reworking of the changes from #194 that are necessary to have the object store build on Linux with GCC 4.9 and GCC 5.4. The CI-related changes will remain in #194.

I've verified that this branch builds on Ubuntu 16.10 with GCC 4.9 and GCC 5.4, for Android NDK r13b with GCC 4.9, and on OS X with Xcode 8.1.

/cc @alebsack @fealebenpae


src/impl/external_commit_helper.hpp
```diff
@@ -23,8 +23,8 @@
 
 #if REALM_PLATFORM_APPLE
 #include "impl/apple/external_commit_helper.hpp"
-#elif REALM_ANDROID || REALM_PLATFORM_NODE
-#include "impl/android/external_commit_helper.hpp"
+#elif REALM_HAVE_EPOLL
```
bdash: One downside to this is that it requires bindings using the object store on Linux but not using object store's CMake files to explicitly define `REALM_HAVE_EPOLL`. They need to conditionally compile the .cpp file anyway so it doesn't seem like a big deal. Would love to hear other's thoughts on this though.

PR #210, opened 2016-11-10
fix for bad merge error

alazier:


alazier:
@bdash do you know why this is causing test failures? Not even sure which is the correct change. Trying to fix linux compilation errors. The test failures are as follows:
```
list
  add_notification_block()
  multiple LinkViws for the same LinkList can get notifications
/Users/ari/src/realm/realm-js/src/object-store/tests/list.cpp:39

/Users/ari/src/realm/realm-js/src/object-store/tests/list.cpp:212: FAILED:
due to unexpected exception with message:
  std::exception

list
  add_notification_block()
  tables-of-interest are tracked properly for multiple source versions
/Users/ari/src/realm/realm-js/src/object-store/tests/list.cpp:39

/Users/ari/src/realm/realm-js/src/object-store/tests/list.cpp:310: FAILED:
due to unexpected exception with message:
  std::exception

===============================================================================
test cases:   47 |   46 passed | 1 failed
assertions: 5392 | 5390 passed | 2 failed

```

tgoyne:
You removed the call to `transaction::advance(m_sg, *m_current, version);`.

alazier:
Thanks. That would definitely explain things....

PR #211, opened 2016-11-10
Added CI harness for linux tests

alebsack:
@bdash @cmelchior This replaces #194 with a proper merge to master.

### DIFF
--- a/workflow/local_docker_build.sh
+++ b/workflow/local_docker_build.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+# This script can be used to locally check and debug
+# the linux build process outside of CI.
+# This should be run from the root directory `packaging/manual_docker_build.sh`
+
+set -e
+rm -rf CMakeCache.txt
+find . -name CMakeFiles -delete
+
+docker build -t ci/realm-object-server:build .
+
+docker run -it --rm -v $(pwd):/source -w /source ci/realm-object-server:build ./workflow/test_coverage.sh
+
--- a/workflow/local_docker_build.sh
+++ b/workflow/local_docker_build.sh
@@ -12,6 +12,8 @@ docker build -t ci/realm-object-server:build .
 
 docker run --rm \
   -u $(id -u) \
+  -v "${HOME}:${HOME}" \
+  -e HOME="${HOME}" \
   -v /etc/passwd:/etc/passwd:ro \
   -v $(pwd):/source \
   -w /source \
--- a/workflow/local_docker_build.sh
+++ b/workflow/local_docker_build.sh
@@ -1,12 +1,9 @@
 #!/bin/sh
 # This script can be used to locally check and debug
 # the linux build process outside of CI.
-# This should be run from the root directory `packaging/local_docker_build.sh`
+# This should be run from the root directory: `./workflow/local_docker_build.sh`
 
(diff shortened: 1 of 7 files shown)


workflow/local_docker_build.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+# This script can be used to locally check and debug
+# the linux build process outside of CI.
+# This should be run from the root directory `packaging/manual_docker_build.sh`
```
bdash: This should presumably be `workflow/local_docker_build.sh`.

workflow/local_docker_build.sh
```diff
@@ -0,0 +1,13 @@
...
+
+set -e
+rm -rf CMakeCache.txt
+find . -name CMakeFiles -delete
```
bdash: If the `CMakeFiles` directories that are found are not empty, which they typically won't be, this will give an error rather than deleting the directories. Using something like `find . -name CMakeFiles -print0 | xargs -0 rm -rf` would address this.

workflow/test_coverage.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+cmake -DCMAKE_BUILD_TYPE=Coverage .
+# sh realm-sync/build.sh config -DREALM_SYNC_PREFIX=/source/realm-sync/
+make VERBOSE=1 -j2 generate-coverage-cobertura
```
bdash: Why `-j2`? Using a value closer to the number of processors gives a significantly faster build on my machine.

workflow/local_docker_build.sh
bdash: I'm not very familiar with Docker, but on Ubuntu 16.10 you can only run `docker` commands as root. Since this script appears to build from the root of the source tree, this leaves a bunch of root-owned files sprinkled around the source tree. Can we instead build from a directory within the docker image, or otherwise outside of the source tree, to avoid this contamination?

workflow/local_docker_build.sh
alebsack: Would using an out-of-source build work?  I don't really like this either.

bdash: Yeah, out of source builds should work fine.

alebsack:
@bdash I think I've addressed everything.  The local docker build script will actually use your current user.  If you are not able to run the docker executable without sudo, simply add your user to the `docker` group.

workflow/local_docker_build.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+# This script can be used to locally check and debug
+# the linux build process outside of CI.
+# This should be run from the root directory `packaging/local_docker_build.sh`
```
bdash: This path is still wrong (`s/packaging/workflow/`).

workflow/local_docker_build.sh
```diff
@@ -0,0 +1,21 @@
...
+
+set -e
+
+rm -rf CMakeCache.txt
+find . -name CMakeFiles -print0 | xargs -0 rm -rf
```
bdash: These don't seem necessary now that we're building out of the source tree. I think the equivalent would be to simply delete `coverage.build` instead.

PR #213, opened 2016-11-11
Reset metadata if metadata.realm could not be opened

stel:
This PR is related to https://github.com/realm/realm-cocoa/issues/4264, it allows to reset metadata in case of error, most likely in case of invalid encryption key. Other related issues: https://github.com/realm/realm-browser-osx/issues/228 and https://github.com/realm/realm-browser-osx-private/issues/141.

Any existing users directories should probably be deleted at least until we have an API to handle this as was suggested in https://github.com/realm/realm-cocoa/issues/4264.

/cc @austinzheng, @jpsim, @bdash

### DIFF
--- a/src/sync/sync_manager.cpp
+++ b/src/sync/sync_manager.cpp
@@ -45,7 +45,8 @@ SyncManager& SyncManager::shared()
 
 void SyncManager::configure_file_system(const std::string& base_file_path,
                                         MetadataMode metadata_mode,
-                                        util::Optional<std::vector<char>> custom_encryption_key)
+                                        util::Optional<std::vector<char>> custom_encryption_key,
+                                        bool reset_metadata_on_error)
 {
     std::vector<UserCreationData> users_to_add;
     {
@@ -68,9 +69,19 @@ void SyncManager::configure_file_system(const std::string& base_file_path,
                                                                            false);
                 break;
             case MetadataMode::Encryption:
-                m_metadata_manager = std::make_unique<SyncMetadataManager>(m_file_manager->metadata_path(),
-                                                                           true,
-                                                                           std::move(custom_encryption_key));
+                try {
+                    m_metadata_manager = std::make_unique<SyncMetadataManager>(m_file_manager->metadata_path(),
+                                                                               true,
+                                                                               std::move(custom_encryption_key));
+                } catch (RealmFileException const& ex) {
+                    if (reset_metadata_on_error && m_file_manager->remove_metadata_realm()) {
+                        m_metadata_manager = std::make_unique<SyncMetadataManager>(m_file_manager->metadata_path(),
+                                                                                   true,
+                                                                                   std::move(custom_encryption_key));
+                    } else {
+                        throw;
+                    }
+                }
                 break;
             case MetadataMode::NoMetadata:
                 return;
--- a/src/sync/sync_manager.hpp
+++ b/src/sync/sync_manager.hpp
@@ -66,7 +66,8 @@ public:
     // Configure the metadata and file management subsystems. This MUST be called upon startup.
     void configure_file_system(const std::string& base_file_path,
                                MetadataMode metadata_mode=MetadataMode::Encryption,
-                               util::Optional<std::vector<char>> custom_encryption_key=none);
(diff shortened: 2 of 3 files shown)


src/sync/sync_manager.cpp
```diff
@@ -68,9 +69,19 @@ void SyncManager::configure_file_system(const std::string& base_file_path,
...
+                    m_metadata_manager = std::make_unique<SyncMetadataManager>(m_file_manager->metadata_path(),
+                                                                               true,
+                                                                               std::move(custom_encryption_key));
+                } catch (RealmFileException const& ex) {
+                    if (reset_metadata_on_error && m_file_manager->remove_metadata_realm()) {
+                        m_metadata_manager = std::make_unique<SyncMetadataManager>(m_file_manager->metadata_path(),
+                                                                                   true,
+                                                                                   std::move(custom_encryption_key));
+                    } else {
+                        throw;
```
jpsim: this should at least rethrow the same exception: `throw ex;`.

stel:
Until we have an API for reseting metadata/users directories to me it seems like we always need to try to reset metadata in case of any errors. In cocoa `metadata.realm` is opened when initializing shared `SyncManager` so there is no way to handle this errors in public API.

tests/sync/sync_manager.cpp
```diff
@@ -172,3 +172,21 @@ TEST_CASE("sync_manager: persistent user state management") {
...
+
+TEST_CASE("sync_manager: metadata") {
+    auto cleanup = util::make_scope_exit([=]() noexcept { SyncManager::shared().reset_for_testing(); });
+    reset_test_directory(base_path);
+
+    SECTION("should be reset in case of decryption error") {
+        const size_t key_size = 64;
+        std::vector<char> encryption_key(key_size);
+
+        arc4random_buf(encryption_key.data(), key_size);
```
tgoyne: This should use `make_test_encryption_key()`. `arc4random_buf()` does not exist on most of the platforms we support.

src/sync/sync_manager.hpp
```diff
@@ -66,7 +66,8 @@ friend class SyncSession;
     // Configure the metadata and file management subsystems. This MUST be called upon startup.
     void configure_file_system(const std::string& base_file_path,
                                MetadataMode metadata_mode=MetadataMode::Encryption,
-                               util::Optional<std::vector<char>> custom_encryption_key=none);
+                               util::Optional<std::vector<char>> custom_encryption_key=none,
+                               bool reset_metadata_on_error=false);
```
stel: The only case that comes to mind is if binding specifies custom encryption key and needs to check if it's valid to try another one. My point is while binding provides encryption key it should be responsible for reseting metadata. Do you think `reset_metadata_on_error` should be `true` by default or should just be removed?

bdash: Do any bindings want that behavior? The method used for resetting the metadata (`SyncFileManager::remove_metadata_realm`) is internal to the object store so it's not clear to me how a binding could usefully reset the sync metadata when handling errors itself.

stel: Right now I don't think any bindings need this behavior at least until it needs to be informed that there was an error reading metadata. In this case it's possible to catch an exception,  perform other actions (like reseting user realms) and reset metadata by calling this method again with `reset_metadata_on_error=true`.

stel:
@austinzheng, @tgoyne, @bdash could you please look at this again? This issue blocks Browser's release.

tests/sync/sync_manager.cpp
```diff
@@ -172,3 +172,21 @@ TEST_CASE("sync_manager: persistent user state management") {
...
+    SECTION("should be reset in case of decryption error") {
+        SyncManager::shared().configure_file_system(base_path,
+                                                    SyncManager::MetadataMode::Encryption,
+                                                    make_test_encryption_key());
+
+        SyncManager::shared().reset_for_testing();
+
+        SyncManager::shared().configure_file_system(base_path,
+                                                    SyncManager::MetadataMode::Encryption,
+                                                    make_test_encryption_key(),
```
stel: Ah, yes, good catch, thanks!

PR #214, opened 2016-11-12
store table name in object schema

jpsim:
To allow having binding-level classes which map to a non-standard table name. This is especially useful for sync, where different clients may have different naming conventions. This is already in use in realm/realm-cocoa#4324.

This implementation does require `table_name` to be explicitly set by the caller if the default constructor is used. /cc @bdash

### DIFF
--- a/src/object_schema.cpp
+++ b/src/object_schema.cpp
@@ -51,6 +51,7 @@ ObjectSchema::ObjectSchema(std::string name, std::initializer_list<Property> per
 : name(std::move(name))
 , persisted_properties(persisted_properties)
 {
+    table_name = ObjectStore::table_name_for_object_type(this->name);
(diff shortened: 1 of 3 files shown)


tgoyne:
I am extremely dubious that this actually works. There are a *lot* more places where we convert between table names and object names than just the places touched. I also don't understand why it is needed. What is wrong with just changing the object name reported by the binding?

jpsim:
> What is wrong with just changing the object name reported by the binding?

Changing this causes the object name to class mapping (`s_localNameToClass`) to return the wrong class for a given object name.

tgoyne:
That sounds like a much easier problem to solve than redesigning everything which wants to get the ObjectSchema for a given table.

jpsim:
Alright, trying with a different approach now.

PR #215, opened 2016-11-14
Fix linking with CoreFoundation and Security frameworks

stel:
Got `ld: library not found for -lFoundation` error, `Foundation` and `Security` frameworks are specified as libraries when linking.

/cc @bdash

### DIFF
--- a/CMake/RealmCore.cmake
+++ b/CMake/RealmCore.cmake
@@ -23,11 +23,11 @@ find_package(PkgConfig)
 find_package(Threads)
 
 if(APPLE)
-    find_library(Foundation Foundation)
-    find_library(Foundation Security)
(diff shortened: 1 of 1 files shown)


bdash:
Hmm, I'm not sure why this would be necessary. I just did a completely clean build of c531c4467e269587b7640da63bd344c1b23b88da locally on macOS 10.12.1 without any link errors. Master has been building successfully on macOS on CI prior to this change as well (see https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-object-store/detail/PR-212/1/pipeline, for instance).

bdash:
I spoke with @stel on Slack and we figured out that when I happened to test this I wasn't building with sync enabled. CI currently never builds with sync enabled.

PR #217, opened 2016-11-14
Fix issue where default value could be moved

austinzheng:
Waiting for tests to run, since apparently `make` can't find Foundation on my machine anymore.

Changes:
- Fixed issue where default value could be erroneously moved in `SyncManager::configure_file_system()`
- `SyncMetadataManager` no longer has a default value for `encryption_key`
- Added a test and updated existing tests

### DIFF
--- a/src/sync/sync_manager.cpp
+++ b/src/sync/sync_manager.cpp
@@ -65,12 +65,15 @@ void SyncManager::configure_file_system(const std::string& base_file_path,
         switch (metadata_mode) {
             case MetadataMode::NoEncryption:
                 m_metadata_manager = std::make_unique<SyncMetadataManager>(m_file_manager->metadata_path(),
-                                                                           false);
+                                                                           false,
+                                                                           none);
                 break;
             case MetadataMode::Encryption:
                 m_metadata_manager = std::make_unique<SyncMetadataManager>(m_file_manager->metadata_path(),
                                                                            true,
-                                                                           std::move(custom_encryption_key));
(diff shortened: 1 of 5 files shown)


src/sync/sync_manager.cpp
```diff
@@ -65,12 +65,15 @@ void SyncManager::configure_file_system(const std::string& base_file_path,
...
+                                                                           false,
+                                                                           none);
                 break;
             case MetadataMode::Encryption:
                 m_metadata_manager = std::make_unique<SyncMetadataManager>(m_file_manager->metadata_path(),
                                                                            true,
-                                                                           std::move(custom_encryption_key));
+                                                                           (custom_encryption_key
+                                                                            ? std::move(custom_encryption_key)
+                                                                            : none));
```
simonask: This is pointless, as noted on Slack :)

simonask:
What? There is no problem with moving from arguments that were initialized by default value.

Default values are initialized at the call site, not (as in some languages) globally.

tests/sync/sync_manager.cpp
```diff
@@ -101,6 +101,13 @@ TEST_CASE("sync_manager: persistent user state management") {
     const std::string token_2 = "bar_token";
     const std::string token_3 = "baz_token";
 
+    SECTION("configure_file_system should allow configuration with no encryption multiple times") {
+        // Ensure a bug does not occur where default `none`s are improperly `std::move()`d
```
bdash: This test passes without any of the other changes.

PR #219, opened 2016-11-15
fix check for already-registered sessions in SyncUser::register_session

jpsim:
due to `m_sessions`/`m_waiting_sessions` being a map to `weak_ptr`, there could be a key entry with no corresponding `SyncSession`.

Thanks @bdash for finding this.


bdash:
Can we write a test that demonstrates the bug this is fixing?

jpsim:
I'm sure. Given that there are also tests in realm/realm-cocoa#4324 that would fail if this regressed, and client permissions needs to ship asap, so I'll write an object store regression test for this later this week.

austinzheng:
I can try writing an object store test for this today.

PR #222, opened 2016-11-15
Allow caller of update_schema() to handle transaction

kneth:
Realm Java has a bit of a complex flow when initializing a Realm file. Breaking it up to smaller transactions might leave the Realm file in an inconsistent state. 

@cmelchior @beeender @bdash

### DIFF
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -300,7 +300,7 @@ bool Realm::reset_file_if_needed(Schema& schema, uint64_t version, std::vector<S
     return false;
 }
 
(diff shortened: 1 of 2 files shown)


tgoyne:
This is reasonable as a temporary workaround, but note that it won't work correctly in a multi-process context unless you're duplicating all of the schema validation logic in the binding.

kneth:
@tgoyne Good point. And since I need this for simplifying Realm Java - and in particular the manual migration mode - it might be fine as it is. But when `update_schema()` does the callback to the migration function, we might not need it.

kneth:
I think I need to retarget it - to be merged into `master`. @cmelchior, is `master` ready?

PR #223, opened 2016-11-15
Add a test for sync session re-registration

austinzheng:
Tests fix in https://github.com/realm/realm-object-store/pull/219

Before the fix, this test never completes. After the fix, it works.

There's probably a more idiomatic way to pass a second value out of a function...

@bdash @jpsim @tgoyne

### DIFF
--- a/tests/sync/session.cpp
+++ b/tests/sync/session.cpp
@@ -38,7 +38,8 @@ using namespace realm::util;
 template <typename FetchAccessToken, typename ErrorHandler>
 std::shared_ptr<SyncSession> sync_session(SyncServer& server, std::shared_ptr<SyncUser> user, const std::string& path,
                                           FetchAccessToken&& fetch_access_token, ErrorHandler&& error_handler,
-                                          SyncSessionStopPolicy stop_policy=SyncSessionStopPolicy::AfterChangesUploaded)
+                                          SyncSessionStopPolicy stop_policy=SyncSessionStopPolicy::AfterChangesUploaded,
+                                          std::string* on_disk_path=nullptr)
 {
     std::string url = server.base_url() + path;
     SyncTestFile config({user, url, std::move(stop_policy),
@@ -46,6 +47,9 @@ std::shared_ptr<SyncSession> sync_session(SyncServer& server, std::shared_ptr<Sy
             auto token = fetch_access_token(path, config.realm_url);
             session->refresh_access_token(std::move(token), config.realm_url);
         }, std::forward<ErrorHandler>(error_handler)});
+    if (on_disk_path) {
+        *on_disk_path = config.path;
+    }
 
     std::shared_ptr<SyncSession> session;
     {
@@ -78,11 +82,11 @@ TEST_CASE("SyncSession: management by SyncUser", "[sync]") {
     SECTION("a SyncUser can properly retrieve its owned sessions") {
         auto user = SyncManager::shared().get_user("user1a", "not_a_real_token");
         auto session1 = sync_session(server, user, "/test1a-1",
-                                     [&](auto...) { return s_test_token; },
-                                     [&](auto...) { });
+                                     [&](auto&, auto&) { return s_test_token; },
+                                     [&](auto, auto, auto) { });
         auto session2 = sync_session(server, user, "/test1a-2",
-                                     [&](auto...) { return s_test_token; },
-                                     [&](auto...) { });
+                                     [&](auto&, auto&) { return s_test_token; },
+                                     [&](auto, auto, auto) { });
         EventLoop::main().run_until([&] { return session_is_active(*session1) && session_is_active(*session2); });
 
         // Check the sessions on the SyncUser.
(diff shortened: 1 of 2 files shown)


tests/sync/session.cpp
```diff
@@ -138,10 +142,10 @@ TEST_CASE("SyncSession: management by SyncUser", "[sync]") {
     SECTION("a SyncUser properly rebinds existing sessions upon logging back in") {
         const std::string user_id = "user1d";
         auto user = SyncManager::shared().get_user(user_id, "not_a_real_token");
-        auto session1 = sync_session(server, user, "/test1b-1",
+        auto session1 = sync_session(server, user, "/test1d-1",
                                      [&](auto...) { return s_test_token; },
                                      [&](auto...) { });
-        auto session2 = sync_session(server, user, "/test1b-2",
+        auto session2 = sync_session(server, user, "/test1d-2",
                                      [&](auto...) { return s_test_token; },
```
bdash: Using `auto...` like this isn't compatible with GCC 4.9. CI doesn't catch this as it does not yet run sync tests.

tests/sync/session.cpp
```diff
@@ -161,6 +165,38 @@ TEST_CASE("SyncSession: management by SyncUser", "[sync]") {
...
+            on_disk_path = *ptr_to_path;
+            free(ptr_to_path);
+            REQUIRE(weak_session.lock() != nullptr);
+        }
+        // Session is dead, so the SyncUser's weak pointer to it should be nulled out.
+        REQUIRE(weak_session.lock() == nullptr);
+        // The next time we request it, it'll be created anew.
+        // The call to `get_session()` should result in `SyncUser::register_session()` being called.
+        auto session = SyncManager::shared().get_session(on_disk_path, config);
+        CHECK(session != nullptr);
```
bdash: Is it necessary to explicitly compare against `nullptr`, or does `CHECK(session)` compile here?

tests/sync/session.cpp
```diff
@@ -38,14 +38,18 @@ using namespace realm::util;
 template <typename FetchAccessToken, typename ErrorHandler>
 std::shared_ptr<SyncSession> sync_session(SyncServer& server, std::shared_ptr<SyncUser> user, const std::string& path,
                                           FetchAccessToken&& fetch_access_token, ErrorHandler&& error_handler,
-                                          SyncSessionStopPolicy stop_policy=SyncSessionStopPolicy::AfterChangesUploaded)
+                                          SyncSessionStopPolicy stop_policy=SyncSessionStopPolicy::AfterChangesUploaded,
+                                          std::string** on_disk_path=nullptr)
```
bdash: This can just be a `std::string *`. There's no need for the double indirection.

An alternative to an output parameter if most callers were interested in both outputs would be to return a `std::pair` or `std::tuple`. Since this is rarely used I think the output parameter is ok.

tests/sync/session.cpp
```diff
@@ -161,6 +165,38 @@ TEST_CASE("SyncSession: management by SyncUser", "[sync]") {
         EventLoop::main().run_until([&] { return session_is_active(*session1) && session_is_active(*session2); });
         REQUIRE(user->all_sessions().size() == 2);
     }
+
+    SECTION("a SyncUser properly allows registration of sessions that were recreated") {
+        const std::string user_id = "user1e";
```
bdash: This appears to only be used the once. Doesn't seem worth having a local variable.

tests/sync/session.cpp
austinzheng: Ugh, you're right. Still in an Objective-C mindset.

tests/sync/session.cpp
austinzheng: Would it be a good idea to remove all uses of `auto...` in lambdas from the sync test code?

tests/sync/session.cpp
austinzheng: I think `CHECK(x)` should work. I'll fix all occurrences.

tests/sync/session.cpp
bdash: Yes. I did this for all occurrences I could find in 8c03a7d568b7fdefb5d2dddee141c2f49d0d9f91, but it looks like some new ones were added since. I think you may be able to get away with using `auto, auto`/`auto, auto, auto` for the argument types rather than being explicit about it like I was in that commit.

austinzheng: Perfect. I'll go through the object store commits from the past two weeks and see what else I missed. Thanks for the pointer.

tests/sync/session.cpp
```diff
@@ -161,6 +165,34 @@ TEST_CASE("SyncSession: management by SyncUser", "[sync]") {
...
+                                        [&](auto, auto, auto) { },
+                                        SyncSessionStopPolicy::Immediately,
+                                        &on_disk_path);
+            weak_session = session;
+            config = session->config();
+            REQUIRE(on_disk_path.size() > 0);
+            REQUIRE(weak_session.lock());
+        }
+        // Session is dead, so the SyncUser's weak pointer to it should be nulled out.
+        REQUIRE(!weak_session.lock());
```
bdash: It'd be clearer to express this as `REQUIRE(weak_session.expired());`.

jpsim:
Thanks for taking care of this @austinzheng!

PR #226, opened 2016-11-17
Have CI test with sync both enabled and disabled

bdash:
This requires building realm-sync from Git if sync is enabled and REALM_SYNC_PREFIX is not specified.
    
This also lays the ground work for downloading prebuilt sync binaries rather than building realm-sync from source. Sadly it's not yet possible to use this as the sync server static library isn't available for download.

This commit changes behavior in a few ways:

* To build with sync it is now necessary to explicitly pass `-DREALM_ENABLE_SYNC=1` to your `cmake` invocation.
* To build against a local version of core you now need to set `REALM_CORE_PREFIX` rather than `REALM_CORE_VERSION`.
* The version of dependencies is now specified in dependencies.list rather than directly in CMakeLists.txt.


bdash:
This PR revealed two issues:
* a crash due to code that wasn't compatible with GCC, fixed by #228.
* an issue with the tests on Linux, fixed by #230.

bdash:
@alebsack, the change to publish reports after all builds is [failing on CI](https://ci.realm.io/job/realm/job/realm-object-store/view/Pull%20Requests/job/PR-226/20/execution/node/168/log/):

> Required context class hudson.FilePath is missing
> Perhaps you forgot to surround the code with a step that provides this, such as: node

The problem appears to be the call to `unstash` within `publishReport`.

bdash:
Thanks, @alebsack!

This is now ready for review.

CMake/RealmCore.cmake
```diff
@@ -127,81 +163,117 @@ function(download_realm_core core_version)
 
     set_property(TARGET realm PROPERTY INTERFACE_LINK_LIBRARIES Threads::Threads ${CRYPTO_LIBRARIES})
 
-    set(REALM_CORE_INCLUDE_DIR ${core_directory}/include PARENT_SCOPE)
-endfunction()
+    set(REALM_CORE_INCLUDE_DIR ${core_directory}/${include_direcotry} PARENT_SCOPE)
```
austinzheng: "directory"

CMake/RealmCore.cmake
```diff
@@ -54,70 +67,93 @@ if (${CMAKE_VERSION} VERSION_GREATER "3.4.0")
...
+
+        if(APPLE)
+            set(core_platform "")
+        elseif(REALM_PLATFORM STREQUAL "Android")
+            set(core_platform "-android-x86_64")
         endif()
-        file(COPY ${core_temp_tarball} DESTINATION ${core_directory_parent})
+    else()
+        set(library_directory "src/realm/")
+        set(include_direcotry "src/")
```
austinzheng: "directory"

CMake/RealmCore.cmake
```diff
@@ -54,70 +67,93 @@ if (${CMAKE_VERSION} VERSION_GREATER "3.4.0")
...
 
-function(use_realm_core version_or_path_to_source)
-    if("${version_or_path_to_source}" MATCHES "^[0-9]+(\\.[0-9])+")
-        if(APPLE OR REALM_PLATFORM STREQUAL "Android")
-            download_realm_core(${version_or_path_to_source})
-        else()
-            clone_and_build_realm_core("v${version_or_path_to_source}")
-        endif()
+function(use_realm_core enable_sync core_prefix sync_prefix)
+  if(core_prefix)
```
tgoyne: The rest of this file is four-space indented.

tgoyne:
Logic changes all seem reasonable to me.

CMake/RealmCore.cmake
bdash: I'll fix my Emacs configuration so it stops doing this to me.

README.md
```diff
@@ -22,8 +22,7 @@ Please report bugs against the Realm product that you're using:
 
 ## Supported Platforms
 
-The object store's CMake build system currently only suports building for OS X, Linux, and Android. Building for Linux requires
-[building against a local version of core](#building-against-a-local-version-of-core).
+The object store's CMake build system currently only suports building for OS X, Linux, and Android.
```
kneth: "suports" -> "supports"

CMake/RealmCore.cmake
```diff
@@ -127,81 +163,117 @@ function(download_realm_core core_version)
...
-    add_dependencies(realm realm-core)
+function(download_realm_sync sync_version)
+    if(APPLE)
+        set(basename "realm-sync-cocoa")
+        set(compression "xz")
+        set(platform "")
+    elseif(REALM_PLATFORM STREQUAL "Android")
+        set(basename "realm-sync-android")
+        set(compression "gz")
+        set(platform "-android-x86_64")
```
kneth: It might be better to use `x86` instead of `x86_64`. Just if we wish to run the code on an emulator.

PR #230, opened 2016-11-18
Use `libuv` for event loops on certain platforms

### LINKED ISSUE #229: Linux needs a event loop implementation to support sync tests
Currently Linux uses the default event loop implementation for which every operation is simply a no-op.  Some sync tests currently fail on Linux due to them relying on having a working event loop. It'd be useful to be able to run these tests on CI as it's currently much easier to do on Linux than Android.

The easiest path may be to default to the libuv event loop implementation used by Node when on Linux. I'm not sure if there's any downsides to doing that.

austinzheng:
Many thanks to @bdash, who did most of the work. We think this provides us with a path to quickly supporting tests on many prospective new platforms, including maybe Windows in the future.

Fixes #229.

@tgoyne @jpsim


austinzheng:
I'm not sure that Linux is running the sync tests on CI (or ever was). The number of test cases & assertions is significantly lower than when I ran it locally on my Linux VM.

bdash:
CI does not run sync tests at all (yet). See #226.

PR #233, opened 2016-11-18
Add support for “file action” sync metadata

austinzheng:
Changes
- Added `SyncFileActionMetadata`, which represents an action the file subsystem should take on a sync’ed Realm the next time the app launches
- Added tests

Part of client support for client reset.

### DIFF
--- a/src/sync/impl/sync_metadata.cpp
+++ b/src/sync/impl/sync_metadata.cpp
@@ -38,31 +38,48 @@ static const char * const c_sync_identity = "identity";
 static const char * const c_sync_auth_server_url = "auth_server_url";
 static const char * const c_sync_user_token = "user_token";
 
+static const char * const c_sync_fileActionMetadata = "FileActionMetadata";
+static const char * const c_sync_original_name = "original_name";
+static const char * const c_sync_new_name = "new_name";
+static const char * const c_sync_action = "action";
+static const char * const c_sync_url = "url";
+
+
 SyncMetadataManager::SyncMetadataManager(std::string path,
                                          bool should_encrypt,
                                          util::Optional<std::vector<char>> encryption_key)
 {
     std::lock_guard<std::mutex> lock(m_metadata_lock);
 
-    auto nullable_string_property = [](const char *name) -> Property {
+    auto make_nullable_string_property = [](const char *name) -> Property {
         Property p = {name, PropertyType::String};
         p.is_nullable = true;
         return p;
     };
 
-    Property primary_key = {c_sync_identity, PropertyType::String};
-    primary_key.is_indexed = true;
-    primary_key.is_primary = true;
+    auto make_primary_key_property = [](const char *name) -> Property {
+        Property p = {name, PropertyType::String};
+        p.is_indexed = true;
+        p.is_primary = true;
+        return p;
+    };
 
     Realm::Config config;
     config.path = std::move(path);
     config.schema = Schema{
         {c_sync_userMetadata, {
-            primary_key,
+            make_primary_key_property(c_sync_identity),
             {c_sync_marked_for_removal, PropertyType::Bool},
-            nullable_string_property(c_sync_auth_server_url),
-            nullable_string_property(c_sync_user_token),
-        }}
+            make_nullable_string_property(c_sync_auth_server_url),
+            make_nullable_string_property(c_sync_user_token),
+        }},
+        {c_sync_fileActionMetadata, {
+            make_primary_key_property(c_sync_original_name),
+            {c_sync_action, PropertyType::Int},
+            make_nullable_string_property(c_sync_new_name),
+            {c_sync_url, PropertyType::String},
+            {c_sync_identity, PropertyType::String},
+        }},
     };
     config.schema_mode = SchemaMode::Additive;
     config.schema_version = 0;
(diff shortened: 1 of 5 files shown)


src/sync/impl/sync_metadata.cpp
```diff
@@ -38,32 +38,47 @@ static const char * const c_sync_identity = "identity";
...
-        Property p = { name, PropertyType::String };
+        Property p = { std::move(name), PropertyType::String };
         p.is_nullable = true;
         return p;
     };
 
-    Property primary_key = { c_sync_identity, PropertyType::String };
-    primary_key.is_indexed = true;
-    primary_key.is_primary = true;
+    auto primary_key_property = [](std::string name)->Property {
```
jpsim: seems like `make_primary_key` would be a clearer name for this?

austinzheng:
I rebased this and removed the extraneous property. Reviews would be appreciated, whenever you have the time.

src/sync/impl/sync_metadata.cpp
```diff
@@ -230,4 +264,81 @@ void SyncUserMetadata::remove()
     m_realm = nullptr;
 }
 
+
+SyncFileActionMetadata::SyncFileActionMetadata(const SyncMetadataManager& manager,
+                                               Action action,
+                                               std::string original_name,
```
bdash: I don't think we should take these strings by value since we're not consuming them. Taking const references instead can probably avoid a copy.

src/sync/impl/sync_metadata.hpp
```diff
@@ -76,6 +76,48 @@ class SyncUserMetadata {
...
+    };
+
+    enum class Action {
+        // The Realm files at the given directory will be deleted.
+        DeleteRealm,
+        // The primary Realm file will be copied to a 'recovery' directory, and the original Realm files will be deleted.
+        HandleRealmForClientReset
+    };
+
+    // The absolute path to the file in question (if this is a Realm, this is the "primary" Realm file).
```
bdash: What does it mean to be a "primary" Realm file?

src/sync/impl/sync_metadata.hpp
```diff
@@ -76,6 +76,48 @@ class SyncUserMetadata {
...
+    std::string original_name() const;
+    // The meaning of this parameter depends on the `Action` specified.
```
bdash: Can we document somewhere what it means?

src/sync/impl/sync_metadata.hpp
austinzheng: I mean the `.realm` file itself. Is that just called the "Realm file", or does it have a particular name?

jpsim: Perhaps referring to it as the "user-specified Realm" would clear this up?

bdash: I'd think that it is called the Realm file.

bdash: In what sense is it "user-specified"?

simonask:
Out of curiosity, I would love to know more about the rationale for this design. On the surface it doesn't seem to me that having a local, persisted notion of a "File Action" is necessary in order to implement client reset, but I could be missing something (crash safety?). Is it to allow the client reset to happen at a more deterministic point in time (app startup), or?

austinzheng:
@simonask That is a great question. Right now, the Cocoa binding cannot safely do anything with a Realm file once it's been opened, unless the user takes care to write their app in a very specific way. So our file-related actions are 'saved' and then carried out the next time the app is launched, before any sync'ed Realms can be opened by the app. Bindings that support a notion of `close()`ing a Realm might not need this specific functionality.

austinzheng:
Tests pass locally. Test failure on CI is because Win32 is configured incorrectly or something.

simonask:
@austinzheng Thanks for your explanation, that clears it up! Maybe it would be worth putting design considerations like this in doc-comments in the code? Just reading it without this extra context, it isn't very easy to deduce (at least for me). :-)

Also, is this functionality fully opt-in for bindings that don't require it? (They may still want to use it, because calling `close()` is completely optional.)

austinzheng:
@simonask no worries!

> Maybe it would be worth putting design considerations like this in doc-comments in the code? Just reading it without this extra context, it isn't very easy to deduce (at least for me). :-)

Definitely. I'm so deep in the weeds at this point that it's hard to remember to keep the overall context of this feature in mind.

> Also, is this functionality fully opt-in for bindings that don't require it?

I believe so. I may add a switch to allow file actions to be disabled independent of the other metadata, if that's something the other bindings think is necessary.

bdash:
Even though other bindings may have a way to close a `Realm` instance, it's difficult to do so safely from a different thread than the one that instance is bound to. It would likely require the app developer to coordinate with all threads of their application in order to ensure all instances are closed, which is quite an ask. This would likely make this deferred-cleanup approach useful even in bindings where it is possible to close `Realm` instances.

simonask:
@bdash Depends on the binding. In Node.js, it's not hard to coordinate which thread is closing the Realm file (because there's only one thread).

On the server side, there are several use cases for calling `close()` proactively simply to reclaim resources, and in situations where we do indeed track the lifetime of the open Realm files very closely (no pun intended...). In those situations it could be quite dangerous if `close()` implies a write transaction.

bdash:
I'm not sure how we've jumped to `close()` requiring a write transaction?

simonask:
Sorry, I may have assumed too much about this machinery, but my thought was that saving these file actions in the metadata would constitute a write transaction. I see that you would probably want to store it upon discovering that they are necessary.

Anyway, that's a minor detail, I'm just trying to ensure that all use cases are taken into account. We can't assume that things are multithreaded, or that all users of the Object Store are limited in their desire or ability to reason about thread-safety. Let's not get sidetracked further, it looks like we're fine. :+1:

PR #235, opened 2016-11-21
Exposing core's distinct method

### LINKED ISSUE #183: Implement `Results.distinct()`
Realm Core supports finding objects with a field/property which is unique/distinct. We should expose this feature in the `Results` class.

kneth:
Closes #183 

To do:

 - [x] basic functionality w/tests
 - [x] chaining distinct
 - [x] sorting and distinct

@bdash @cmelchior @ironage

### DIFF
--- a/src/results.cpp
+++ b/src/results.cpp
@@ -491,6 +491,13 @@ Results Results::filter(Query&& q) const
     return Results(m_realm, get_query().and_query(std::move(q)), m_sort);
 }
 
+Results Results::distinct(realm::SortDescriptor&& sort)
+{
+    auto tv = this->get_tableview();
+    tv.distinct(sort);
+    return Results(m_realm, tv);
+}
+
 Results Results::snapshot() const &
 {
     validate_read();
--- a/src/results.cpp
+++ b/src/results.cpp
@@ -155,11 +155,9 @@ size_t Results::size()
         case Mode::LinkView: return m_link_view->size();
         case Mode::Query:
             m_query.sync_view_if_needed();
-            if (m_distinct) {
-                update_tableview();
-                return m_table_view.size();
-            }
-            return m_query.count();
+            if (!m_distinct)
+                return m_query.count();
+            REALM_FALLTHROUGH;
         case Mode::TableView:
             update_tableview();
             return m_table_view.size();
@@ -523,7 +521,7 @@ Results Results::distinct(realm::SortDescriptor&& uniqueness)
 {
     auto tv = get_tableview();
     tv.distinct(uniqueness);
-    return Results(m_realm, tv, std::move(m_sort), std::move(uniqueness));
+    return Results(m_realm, tv, m_sort, std::move(uniqueness));
 }
 
 Results Results::snapshot() const &
--- a/src/results.cpp
+++ b/src/results.cpp
@@ -264,7 +264,7 @@ bool Results::update_linkview()
 {
     REALM_ASSERT(m_update_policy == UpdatePolicy::Auto);
 
-    if (m_sort) {
+    if (m_sort || m_distinct) {
         m_query = get_query();
         m_mode = Mode::Query;
         update_tableview();
--- a/src/results.cpp
+++ b/src/results.cpp
@@ -491,11 +491,11 @@ Results Results::filter(Query&& q) const
     return Results(m_realm, get_query().and_query(std::move(q)), m_sort);
 }
 
-Results Results::distinct(realm::SortDescriptor&& sort)
+Results Results::distinct(realm::SortDescriptor&& uniqueness)
 {
-    auto tv = this->get_tableview();
-    tv.distinct(sort);
-    return Results(m_realm, tv);
+    auto tv = get_tableview();
+    tv.distinct(uniqueness);
+    return Results(m_realm, tv, m_sort);
 }
 
 Results Results::snapshot() const &
--- a/src/results.cpp
+++ b/src/results.cpp
@@ -62,6 +62,15 @@ Results::Results(SharedRealm r, LinkViewRef lv, util::Optional<Query> q, SortDes
     }
 }
 
+Results::Results(SharedRealm r, Query q, SortDescriptor s, SortDescriptor d)
+: m_realm(std::move(r))
+, m_query(std::move(q))
+, m_sort(std::move(s))
+, m_distinct(std::move(d))
+, m_mode(Mode::Query)
+{
+}
+
 Results::Results(SharedRealm r, TableView tv, SortDescriptor s)
 : m_realm(std::move(r))
 , m_table_view(std::move(tv))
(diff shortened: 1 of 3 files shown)


kneth:
@ironage @rrrlasse Please review my test to see if my understanding of core's `distinct()` is correct.

ironage:
The tests look good to me!

Maybe you were going to do this, but I think it would be nice to copy the design of sort and make a `m_distinct` member that is copied around in the constructors and assignment of the Results class, then you don't lose the distinct part when passing around the Results and after sorting again and such.

beeender:
But this doesn't work for async distinct?

kneth:
@beeender Not so far. I was hoping to be able to piggyback on your `Results` work :-)

src/results.cpp
```diff
@@ -483,19 +511,19 @@ TableView Results::get_tableview()
...
-    return Results(m_realm, get_query().and_query(std::move(q)), m_sort);
+    return Results(m_realm, get_query().and_query(std::move(q)), m_sort, m_distinct);
 }
 
 Results Results::distinct(realm::SortDescriptor&& uniqueness)
 {
     auto tv = get_tableview();
     tv.distinct(uniqueness);
-    return Results(m_realm, tv, m_sort);
+    return Results(m_realm, tv, std::move(m_sort), std::move(uniqueness));
```
bdash: You don't want to move `m_sort` as that may change the sort order of `this`.

bdash:
One question this raises is whether `.sort(…).distinct(…)` should produce identical results to `.distinct(…).sort(…)`.

kneth:
@bdash I have added a section to clarify it. And one of the comments was wrong - leading you to believe that `sort().distinct() != distinct().sort()` - sorry.

bdash:
@kneth, my point is that I don't see why `sort().distinct()` and `distinct().sort()` _should_ give the same result. Performing the distinct operation before sorting gives a different result than after it, and there are potential use cases for each.

src/results.cpp
```diff
@@ -134,6 +155,10 @@ size_t Results::size()
         case Mode::LinkView: return m_link_view->size();
         case Mode::Query:
             m_query.sync_view_if_needed();
+            if (m_distinct) {
```
tgoyne: `if (!m_distinct) return m_query.count(); REALM_FALLTHROUGH;` avoids duplicating the (admittedly trivial) logic for the tableview case.

src/results.cpp
```diff
@@ -483,12 +511,19 @@ TableView Results::get_tableview()
...
 {
-    return Results(m_realm, get_query().and_query(std::move(q)), m_sort);
+    return Results(m_realm, get_query().and_query(std::move(q)), m_sort, m_distinct);
+}
+
+Results Results::distinct(realm::SortDescriptor&& uniqueness)
+{
+    auto tv = get_tableview();
+    tv.distinct(uniqueness);
+    return Results(m_realm, tv, std::move(m_sort), std::move(uniqueness));
```
bdash: This shouldn't be moving `m_sort`.

tests/results.cpp
```diff
@@ -1802,3 +1802,190 @@ TEST_CASE("results: snapshots") {
...
+    auto coordinator = _impl::RealmCoordinator::get_existing_coordinator(config.path);
+    auto table = r->read_group().get_table("class_object");
+
+    r->begin_transaction();
+    table->add_empty_row(N);
+    for (int i = 0; i < N; ++i) {
+        std::stringstream ss;
+        ss << "Foo_ " << i % 3;
+        table->set_int(0, i, i % 3);
+        table->set_string(1, i, StringData(ss.str().c_str()));
```
tgoyne: `table->set_string(1, i, util::format("Foo_%1", i % 3).c_str());` for all of these.

tests/results.cpp
```diff
@@ -1802,3 +1802,190 @@ TEST_CASE("results: snapshots") {
         CHECK_THROWS(snapshot.add_notification_callback([](CollectionChangeSet, std::exception_ptr) {}));
     }
 }
+
+TEST_CASE("distict")
```
tgoyne: No newline between TEST_CASE and {.

tests/results.cpp
```diff
@@ -1802,3 +1802,190 @@ TEST_CASE("results: snapshots") {
...
+    auto r = Realm::get_shared_realm(config);
+    r->update_schema({
+        {"object", {
+            {"num1", PropertyType::Int},
+            {"string", PropertyType::String},
+            {"num2", PropertyType::Int}
+        }},
+    });
+
+    auto coordinator = _impl::RealmCoordinator::get_existing_coordinator(config.path);
```
tgoyne: Appears to be unused.

tests/results.cpp
```diff
@@ -1802,3 +1802,190 @@ TEST_CASE("results: snapshots") {
...
+        REQUIRE(reverse.get(0).get_int(2) == 8);
+        REQUIRE(reverse.get(1).get_int(2) == 9);
+        REQUIRE(reverse.get(2).get_int(2) == 10);
+    }
+
+    SECTION("Chaining distinct") {
+        Results first = results.distinct(SortDescriptor(results.get_tableview().get_parent(), {{0}}));
+        REQUIRE(first.size() == 3);
+        
+        // distinct() will discard the previous applied distinct() calls
```
jpsim: This isn't how I'd expect chaining distinct calls to work... Typically chaining queries always returns a subset of the previous query. This is inconsistent with all other queries in Realm...

kneth: I tend to agree with you. 

Let's consider T = [(A, 1, alpha), (A, 2, beta), (B, 1, delta), (B, 2, gamma)]. 

V = T.distinct(0) = [(A, 1, alpha), (B, 1, delta)] and V.distinct(1) = [(A, 1, alpha)]
W = T.distinct(1) = [(A, 1, alpha), (A, 2, beta)] and W.distinct(0) = [(A, 1, alpha)]

On the other hand, T.distinct(0, 1) = [(A, 1, alpha), (A, 2, beta), (B, 1, delta), (B, 2, gamma)]

In short T.distinct(0).distinct(1) != T.distinct(0, 1) (I'm sure you can construct cases where equality is true).

This PR expose core's `distinct` method. From `table_view.hpp` in core:

```
// Each time you call distinct() it will first fetch the full original TableView contents and then apply
// distinct() on that. So it distinct() does not filter the result of the previous distinct().
```

The question is if we wish to change core's functionality? We might have good reasons for that but I believe that it is better to take that discussion in a core issue :-)

jpsim: It's a core issue to change it, it's an object store and binding issue to document it.

tests/results.cpp
```diff
@@ -1803,8 +1804,7 @@ TEST_CASE("results: snapshots") {
     }
 }
 
-TEST_CASE("distict")
-{
+TEST_CASE("distict") {
```
kneth: I need spell checker in my editor - thanks!

kneth:
@bdash Can you give an example where `distinct().sort() != sort().distinct()` is what is expected?

jpsim:
> Can you give an example where `distinct().sort() != sort().distinct()` is what is expected?

```
one = {a: 0, b: 0}
two = {a: 1, b: 1}
three = {a: 1, b: 0}
arr = [one, two, three]
arr.distinct(a).sort(b) => [one, two]
arr.sort(b).distinct(a) => [one, three]
```

src/results.cpp
```diff
@@ -483,12 +509,19 @@ TableView Results::get_tableview()
...
 {
-    return Results(m_realm, get_query().and_query(std::move(q)), m_sort);
+    return Results(m_realm, get_query().and_query(std::move(q)), m_sort, m_distinct);
+}
+
+Results Results::distinct(realm::SortDescriptor&& uniqueness)
+{
+    auto tv = get_tableview();
+    tv.distinct(uniqueness);
+    return Results(m_realm, tv, m_sort, std::move(uniqueness));
```
bdash: This should move `tv`.

PR #238, opened 2016-11-22
Remove duplicate files

radu-tutueanu:
@alazier @mrackwitz 
Having these files twice causes confusion.

Right now we are(should be) using the ones in sync, not sure if there is any difference that should go in there.


mrackwitz:
The files definitely differ.

kristiandupont:
As @mrackwitz says, there are significant variations there. The question is whether anything was using the files outside the sync folder?

radu-tutueanu:
Only by https://github.com/realm/realm-js-private/blob/master/src/sync_logger.hpp#L30 
and it looks like it was by mistake.  Don't know what effects that has had.

bdash:
IMO the best way to confirm that this change is ok is to:
* Verify that any JS tests related to sync still pass with this change.
* Double-check that there were no changes made to the files being removed on the `js` branch that didn't later make it into `master`.

radu-tutueanu:
Thanks. These files have not been used for a while, but I will make upstream PRs to make sure this does not break anything. :)

PR #239, opened 2016-11-23
Fix property change migration involving LinkingObjects

ironage:
Backlink columns should not be inserted directly to tables. This happens automatically when the corresponding Link or LinkList column is inserted in the linked table. This caused the stack trace in realm core [2291](https://github.com/realm/realm-core/issues/2291). The added test in this PR can produce the same stack trace.

@bdash @tgoyne please review

### DIFF
--- a/src/object_schema.cpp
+++ b/src/object_schema.cpp
@@ -107,12 +107,8 @@ const Property *ObjectSchema::property_for_name(StringData name) const {
 }
 
 bool ObjectSchema::property_is_computed(Property const& property) const {
-    for (auto& prop : computed_properties) {
-        if (prop == property) {
-            return true;
-        }
-    }
-    return false;
+    auto end = computed_properties.end();
+    return std::find(computed_properties.begin(), end, property) != end;
 }
 
 void ObjectSchema::set_primary_key_property()
--- a/src/object_schema.cpp
+++ b/src/object_schema.cpp
@@ -106,6 +106,15 @@ const Property *ObjectSchema::property_for_name(StringData name) const {
     return const_cast<ObjectSchema *>(this)->property_for_name(name);
 }
 
+bool ObjectSchema::property_is_computed(Property const& property) const {
+    for (auto& prop : computed_properties) {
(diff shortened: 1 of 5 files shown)


tgoyne:
The code in the test is not valid. LinkingObjects properties can only appear in computed_properties, and not persisted_properties.

jpsim:
The fix does look right to me, just the test that should be refactored to move the LinkingObjects properties in `computed_properties`.

tgoyne:
The fix is incorrect as well because it should not be possible for a LinkingObjects property to be passed to insert_column() or replace_column().

ironage:
@tgoyne I updated the test to use a computed property. I also moved the fix to a higher level where changes are computed. Can you recheck please?

ironage:
It could happen when users change their model code using the same variable name from a persisted type to a LinkingObjects type. 
e.g. from 
    `var owner: Person?`
to
    `let owner = LinkingObjects(fromType: Person.self, property: "dogs")`

src/object_store.cpp
```diff
@@ -93,6 +93,9 @@ void add_index(Table& table, size_t col)
 
 void insert_column(Group& group, Table& table, Property const& property, size_t col_ndx)
 {
+    if (property.type == PropertyType::LinkingObjects) {
+        throw std::logic_error("Cannot directly insert a LinkingObject column, it must be an artifact of an existing link column.");
```
tgoyne: This should just be an assert since it should not be possible to hit it.

The error message has a comma splice and it's "LinkingObjects".

src/object_schema.cpp
```diff
@@ -106,6 +106,15 @@ const Property *ObjectSchema::property_for_name(StringData name) const {
     return const_cast<ObjectSchema *>(this)->property_for_name(name);
 }
 
+bool ObjectSchema::property_is_computed(Property const& property) const {
+    for (auto& prop : computed_properties) {
```
ironage: Good idea! Updated.

Issue #207, opened 2016-11-08
Need reset method for metadata realm/global state for testing


alazier:
It isn't currently possible to easily test this functionality in the JS binding.

jpsim:
Should go in [`SyncManager::reset_for_testing()`](https://github.com/realm/realm-object-store/blob/b597945d068856a7b7a5bb8e5d68d6403ab87961/src/sync/sync_manager.hpp#L105-L106). Note that some metadata is already being reset in there (e.g. users). Could you be more explicit about what else you'd like to see reset?

alazier:
That seems to be what I should be calling but I get a segfault when calling it.

jpsim:
Perhaps you could take a look at the existing usage in the object store tests to find out if you're using it differently?

* [tests/sync/session.cpp](https://github.com/realm/realm-object-store/blob/b597945d068856a7b7a5bb8e5d68d6403ab87961/tests/sync/session.cpp#L121)
* [tests/sync/user.cpp](https://github.com/realm/realm-object-store/blob/b597945d068856a7b7a5bb8e5d68d6403ab87961/tests/sync/user.cpp#L33)
* [tests/sync/sync_manager.cpp](https://github.com/realm/realm-object-store/blob/b597945d068856a7b7a5bb8e5d68d6403ab87961/tests/sync/sync_manager.cpp#L49)

simonask:
I can reproduce a segfault through the JS binding when running many unit tests that start and stop many sessions.

The crash only happens when the stop policy in `js_sync.hpp` is `SyncSessionStopPolicy::AfterChangesUploaded`. If I change it to `SyncSessionStopPolicy::Immediately`, everything goes well (but of course terminates sessions immediately without waiting for pending changes to be uploaded).

It seems that `SyncManager::reset_for_testing()` must be extended to wait for all threads that are currently blocking on `wait_for_upload_complete_or_client_stopped()` to return. This way may have to happen in `SyncClient::~SyncClient()` between calling `sync::Client::stop()` and deleting the `sync::Client` object.

bdash:
One thing that stands out is the [sync transaction callback that `SyncSession` registers](https://github.com/realm/realm-object-store/blob/28ac73d8881189ac0b6782a6a36f4893f326397f/src/sync/sync_session.cpp#L374-L380) doesn't respect the warning found in the documentation for `sync::Session::set_sync_transact_callback`:

> CAUTION: The specified callback function may be called before the call
[...]

PR #241, opened 2016-11-23
Make the sync transaction and error handler callbacks safe to run after the SyncSession has been destroyed

bdash:
The documentation for `sync::Session` states that they may be run after the `sync::Session` goes away, which means they may also be run after the `SyncSession` goes away. Capture a `weak_ptr` to `this` in the two callback lambdas so we can detect whether the `SyncSession` exists and guarantee it remains alive for the duration of the callback.

Fixes one of the problems noted in #207.

Needs:
- [x] To do something besides abort in the error handler when called after the `SyncSession` goes away.
- [x] ~~Tests~~
- [x] Verify that this approach is safe throughout the entirety of `SyncSession`'s lifecycle, including when it transitions to being owned by a `unique_ptr`.

### DIFF
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -299,7 +299,15 @@ void SyncSession::create_sync_session()
     m_session = std::make_unique<sync::Session>(m_client->client, m_realm_path);
 
     // Set up the wrapped handler
-    auto wrapped_handler = [this](int error_code, std::string message) {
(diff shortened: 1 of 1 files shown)


bdash:
I've not had any luck in writing a test case that crashes without this fix. We don't have precise enough control over when sync invokes callbacks so it's difficult to trigger on demand.

bdash:
Since I cannot see how to write a test for this at present, this is now ready for review.

PR #242, opened 2016-11-24
Don't deadlock when SyncUser::register_session complains about registering multiple sessions for the same URL

bdash:
Reported by @mrackwitz in #178.

/cc @austinzheng

### DIFF
--- a/tests/sync/session.cpp
+++ b/tests/sync/session.cpp
@@ -196,6 +196,19 @@ TEST_CASE("SyncSession: management by SyncUser", "[sync]") {
         session = user->session_for_url(server.base_url() + path);
         CHECK(session);
     }
+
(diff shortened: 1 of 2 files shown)


tests/sync/session.cpp
```diff
@@ -196,6 +196,19 @@ TEST_CASE("SyncSession: management by SyncUser", "[sync]") {
         session = user->session_for_url(server.base_url() + path);
         CHECK(session);
     }
+
+    SECTION("a user cannot create multiple sessions for the same URL") {
```
austinzheng: Would it be worthwhile to add a FIXME to remove this test once we remove this restriction?

bdash: My thinking is that the test will fail when we remove the restriction, which should make it obvious that it needs to be removed.

PR #247, opened 2016-11-24
Replace blocking upload/download waiters with async versions

### LINKED ISSUE #178: Sync-related object store deadlocks

### LINKED ISSUE #244: Refactor `SyncSession` to use async waiters
With https://github.com/realm/realm-sync/pull/983 merged, we can probably greatly simplify the `SyncSession` state machine and get rid of some testing issues that result from `SyncSession`s being destroyed while threads are still blocked on the sync waiting APIs. @bdash

austinzheng:
Fixes #178, fixes #244
Depends upon: #241 (otherwise the tests will still sporadically fail)
@bdash 

Changes:
- Replaced use of blocking waiter APIs with their new async variants
- Updated tests
- Test resetter now clears sessions between test runs

### DIFF
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -210,24 +210,21 @@ struct sync_session_states::Active : public SyncSession::State {
 struct sync_session_states::Dying : public SyncSession::State {
     void enter_state(std::unique_lock<std::mutex>&, SyncSession& session) const override
     {
-        ++session.m_pending_upload_threads;
-        std::thread([session=&session] {
+        size_t current_death_count = ++session.m_death_count;
+        session.m_session->async_wait_for_upload_completion([session=&session, current_death_count](std::error_code error_code) {
+            if (error_code == util::error::operation_aborted) {
+                // Session was killed beneath us. Don't do anything.
+                return;
+            }
+            // FIXME: It's possible for the session to be destroyed while the callback is running,
+            // which is something we should address in the future. It is only possible when 
+            // SyncManager::reset_for_testing() is called.
+            // c.f. https://github.com/realm/realm-object-store/issues/269
             std::unique_lock<std::mutex> lock(session->m_state_mutex);
-            if (session->m_pending_upload_threads != 1) {
-                --session->m_pending_upload_threads;
-                return;
+            if (session->m_state == &State::dying && session->m_death_count == current_death_count) {
+                session->advance_state(lock, inactive);
             }
-
-            if (session->m_state != &State::dying) {
-                // The session was revived. Don't kill it.
-                --session->m_pending_upload_threads;
-                return;
-            }
-
-            session->m_session->wait_for_upload_complete_or_client_stopped();
-            --session->m_pending_upload_threads;
-            session->advance_state(lock, inactive);
-        }).detach();
+        });
     }
 
     bool revive_if_needed(std::unique_lock<std::mutex>& lock, SyncSession& session) const override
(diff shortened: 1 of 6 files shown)


tests/results.cpp
```diff
@@ -932,12 +932,12 @@ TEST_CASE("notifications: sync") {
...
-        // Make sure that the notifications still get delivered rather than
-        // waiting forever due to that we don't get a commit notification from
-        // the commits sync makes to store the upload progress
-        r->refresh();
+        SyncManager::shared().get_session(config.path, *config.sync_config)->wait_for_upload_completion([r=std::move(r)](auto) {
+            // Make sure that the notifications still get delivered rather than
+            // waiting forever due to that we don't get a commit notification from
+            // the commits sync makes to store the upload progress
+            r->refresh();
+        });
```
bdash: Since this no longer blocks I think you'd need to spin the runloop until the callback is called?

tests/results.cpp
```diff
@@ -932,12 +933,15 @@ TEST_CASE("notifications: sync") {
...
-        SyncManager::shared().get_session(config.path, *config.sync_config)->wait_for_upload_completion([r=std::move(r)](auto) {
-            // Make sure that the notifications still get delivered rather than
-            // waiting forever due to that we don't get a commit notification from
-            // the commits sync makes to store the upload progress
-            r->refresh();
+        std::atomic<bool> did_refresh(false);
+        SyncManager::shared().get_session(config.path, *config.sync_config)->wait_for_upload_completion([&](auto) {
+            did_refresh = true;
         });
+        EventLoop::main().run_until([&] { return did_refresh == true; });
```
tgoyne: This test should just continue to use the synchronous wait function. Running the event loop makes it not test what it's trying to test, and while that could be fixed by using a condition variable instead, there's no reason to reinvent that functionality when it already exists in sync::Session.

austinzheng:
@alazier has mentioned that the lack of this is a pretty critical blocker for him. Given that it seems to allow tests to run with a high level of reliability, would it be possible to merge this as a workaround while we continue thinking about a more sound solution?

tgoyne:
What about this is still a WIP?

austinzheng:
It doesn't fully get rid of all possible race conditions involved with waiting for a session to finish uploading and close, since it's possible the session could be destroyed in the middle of the callback (after acquiring the lock). @bdash and I spent part of Monday trying to figure out how to make it sound, and then decided to set it aside to work on other stuff.

src/sync/sync_session.cpp
```diff
@@ -461,44 +454,35 @@ bool SyncSession::can_wait_for_network_completion() const
...
-    std::thread([callback=std::move(callback), self=shared_from_this()]() {
-        self->wait_for_upload_completion();
-        callback();
-    }).detach();
-}
-
-void SyncSession::wait_for_download_completion(std::function<void()> callback)
+void SyncSession::wait_for_upload_completion_blocking()
 {
-    // FIXME: If the session is waiting for a token, the `wait_for_upload` should be deferred until the `bind()`
```
bdash: Why are these FIXMEs no longer relevant?

austinzheng: The tests that were depending on that behavior (in Cocoa) no longer use this API. That being said, it's probably a good idea to ensure the async API actually respects this behavior (i.e. doesn't call the callback immediately if `bind()` hasn't yet been called).

bdash:
```
    /// It is an error to call these functions before bind() has been called,
    /// and has returned.
```

austinzheng:
@bdash and I discussed this further in person, and I made some changes in response.

austinzheng:
Something's wrong with CI and it has nothing to do with the tests themselves.

PR #250, opened 2016-11-25
Validate primary column uniqueness if migration function is not specified

stel:
This PR is related to the second issue in https://github.com/realm/realm-cocoa/issues/4370: in case if the migration function is not specified the validation of the primary column should still be performed.

/cc @bdash, @jpsim

### DIFF
--- a/src/object_store.cpp
+++ b/src/object_store.cpp
@@ -647,6 +647,7 @@ void ObjectStore::apply_schema_changes(Group& group, Schema& schema, uint64_t& s
     }
     else {
         apply_post_migration_changes(group, changes, {});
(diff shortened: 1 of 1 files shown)


tgoyne:
I'm not sure this makes sense. The purpose of re-checking all primary keys is to detect bugs in the migration function, since primary key uniqueness is not enforced during a migration. I think instead we would want to require that any schema changes which could produce duplicate PKs include a migration function to set them. I think this is just changing a PK from optional to non-optional and creating an entirely new column which is the new PK.

stel:
Created a separate issue for this and closing this PR.

PR #251, opened 2016-11-25
Merge latest master into JS

alazier:
@kristiandupont @bdash


bdash:
The title suggests this is merging `master` into `js`, but the PR appears to do the reverse.

kristiandupont:
We definitely want to converge with master as well but we should probably get everyone on board with the threadid issue before we do.

radu-tutueanu:
I changed the base, as we want to merge js-latest to js

alazier:
@bdash @radu-tutueanu does this look good to merge?

radu-tutueanu:
👍 LGTM

PR #252, opened 2016-11-25
Copy property values when migrating from optional to required property

stel:
This PR is addressed to the first issue in https://github.com/realm/realm-cocoa/issues/4370.

To me it seems that `non-null` values could be copied automatically while migrating from optional to required property. It allows users to not specify the migration block for this type of migrations in case if no `null` values exist in a table.

In case if there are any `null` values before the migration the `Attempted to insert null into non-nullable column` exception is generated, that will cause the user to setup migration block manually.

Another improvement that can be done here is addressed in https://github.com/realm/realm-cocoa/issues/1793, but it's not a part of this PR.

The test that checks that an exception is thrown when copying `null` value shout be added here as well.

/cc @jpsim, @bdash, @tgoyne

### DIFF
--- a/src/object_store.cpp
+++ b/src/object_store.cpp
@@ -182,6 +182,7 @@ void make_property_required(Group& group, Table& table, Property property)
 {
     property.is_nullable = false;
     insert_column(group, table, property, property.table_column);
+    copy_property_values(property, table);
     table.remove_column(property.table_column + 1);
 }
 
--- a/tests/migrations.cpp
+++ b/tests/migrations.cpp
@@ -455,7 +455,7 @@ TEST_CASE("migration: Automatic") {
                 REQUIRE(table->get_int(0, i) == i);
         }
 
-        SECTION("values for nullable properties are discarded when converitng to required") {
+        SECTION("non-null values for nullable properties are not discarded when converitng to required") {
             Schema schema = {
                 {"object", {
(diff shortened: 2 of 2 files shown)


tgoyne:
We've explicitly rejected doing this several times already because it's very error prone. It would be very easy to test the migration locally on data that happens to have no null values and not even realize that there is an unhandled case.

The automatic copying for non-null to null is there only because when we added support for null we changed the default nullability for several property types and we did not want data loss to be the default result of upgrading to the new version of Realm.

stel:
That makes sense.

What about copying `non-null` values and reseting `null` to the default (type specific) values? It means skipping `null` values in `copy_property_values` method. This will prevent data loss. Or is data loss used as "indicator" that manual migration is required?

jpsim:
> Or is data loss used as "indicator" that manual migration is required?

To put it bluntly, yes. Failing in a big way is easier to catch in development than failing in a subtle, possibly unnoticed way.

The ideal scenario I think would be to track accessor setters to throw an exception if a nullable->required converted property wasn't re-set during a migration block. But that would involve messing with the guts of the dynamic object accessors.

A simple improvement that should be done is to throw an exception if no migration block is provided and the schema version is bumped for a nullable->required schema change.

stel:
Makes sense, closing this.

PR #253, opened 2016-11-28
Merge master into js

radu-tutueanu:
@mrackwitz @bdash @alazier 

We need to cherrypick https://github.com/realm/realm-object-store/pull/242 in order to be able to merge https://github.com/realm/realm-js/pull/679 and use it upstream.


radu-tutueanu:
This is already in `js-latest`

PR #257, opened 2016-11-28
Fix a LogicError(bad_version) error on refresh()

tgoyne:
refresh() and begin_write() don't block for notifiers if the notifiers exist only for implict async queries and there are no callbacks, which means that the Realm can get ahead of the notifier version. When this happened, it was still trying to advance to the notifier version even though nothing from that notifier would be delivered and the advance would fail due to it being an older version.

Probably is the issue reported at https://github.com/realm/realm-cocoa/issues/4345.


xiaodao:
Is this change released in RealmCocoa 2.1.0?

PR #258, opened 2016-11-28
Add more validation for Realm Config parameters

### LINKED ISSUE #249: Fail if a migration function is specified in additive mode
If a Realm is opened in additive schema mode, then migration functions aren't executed. Instead of simply ignoring them, I think we should raise an exception instead as it is a misconfiguration which we should already prevent on this level to avoid confusion.

tgoyne:
Reject a few more cases such as supplying a schema without a schema version or a migration function in a schema mode that doesn't do migrations.

Closes #249.

### DIFF
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -175,9 +175,19 @@ void Realm::open_with_config(const Config& config,
                              std::unique_ptr<Group>& read_only_group,
                              Realm* realm)
 {
-    if (config.encryption_key.data() && config.encryption_key.size() != 64) {
+    if (config.encryption_key.data() && config.encryption_key.size() != 64)
         throw InvalidEncryptionKeyException();
-    }
+    if (config.schema && config.schema_version == ObjectStore::NotVersioned)
+        throw std::logic_error("A schema version must be specified when the schema is specified");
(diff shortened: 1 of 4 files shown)


src/shared_realm.cpp
```diff
@@ -175,9 +175,19 @@ void Realm::open_with_config(const Config& config,
...
+    if (config.schema && config.schema_version == ObjectStore::NotVersioned)
+        throw std::logic_error("A schema version must be specified when the schema is specified");
+    if (config.schema_mode == SchemaMode::ReadOnly && config.sync_config)
+        throw std::logic_error("Synchronized Realms cannot be opened in read-only mode");
+    if (config.schema_mode == SchemaMode::Additive && config.migration_function)
+        throw std::logic_error("Realms opened in Additive-only schema mode do not use a migration function");
+    if (config.schema_mode == SchemaMode::ReadOnly && config.migration_function)
+        throw std::logic_error("Realms opened in read-only mode do not use a migration function");
+    // ResetFile also won't use the migration function, but specifying one is
+    // allowed to simplify switching temporarily switching modes during development
```
bdash: This comment ended up a little bit mangled.

PR #262, opened 2016-11-29
Add support for an event notifications mechanism

austinzheng:
@bdash
Implement a notifier for the `SyncManager` for use by bindings.

Changes:
- Implemented a notifier that bindings can override to get sync-wide event notifications
- A few internal API changes to facilitate this new feature
- Object store sync test infrastructure slightly overhauled to facilitate testing
- Added tests to exercise the new functionality

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -79,6 +79,7 @@ if(REALM_ENABLE_SYNC)
     list(APPEND HEADERS
         sync/sync_config.hpp
         sync/sync_manager.hpp
+        sync/sync_notifier.hpp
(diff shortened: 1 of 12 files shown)


austinzheng:
Note that some unrelated test is failing, but I think that's because `sync` changed its behavior from under us.

bdash:
What is this needed for?

austinzheng:
A way to register notifiers has been a semi-frequent request by product and some of the demo app engineers. More importantly, it provides us with a way for client reset to inform the application that a Realm has been deleted + backed up to a particular location.

PR #265, opened 2016-11-30
Pass the session as an argument to its error handler

fealebenpae:


src/sync/sync_session.cpp
```diff
@@ -374,7 +374,7 @@ void SyncSession::create_sync_session()
                 break;
         }
         if (m_error_handler) {
-            m_error_handler(error_code, message, error_type);
+            m_error_handler(self, error_code, message, error_type);
```
austinzheng: `std::move()` `self`?

fealebenpae:
@austinzheng: in .NET we need the session object as a "sender" of when firing the session error event. I used to use `SyncUser::session_for_url` to look up the session in a lambda that captures the user and the url, but that doesn't work for errors that invalidate the user because then `session_for_url` always returns `nullptr`.

austinzheng:
Excellent. I ask because we're probably going to get rid of `SyncUser::session_for_url` in some form or another in the near future, and the changes in this PR would be a logical step towards making that happen.

Issue #243, opened 2016-11-24
Reevaluate the restriction that a given user can only have the same synced URL open once


bdash:
`SyncUser::register_session` currently enforces a restriction that a given user can only have the same synced URL open once. There are plausible use cases where a user would have the same synced URL open at multiple, different local paths. I'm not aware of a reason for the restriction to exist.

@mrackwitz is running into this restriction in realm/realm-js#679, which was masked by the deadlock fixed in #242.

/cc @austinzheng

austinzheng:
The most straightforward way to fix this is to have `SyncUser` store sessions based on their local path, not on their URL. We may or may not be able to replace `SyncUser`'s map member variables with calls to `SyncManager`, since `SyncManager` also stores sessions based on local path. We can then leave it up to the bindings if they want to provide a feature to map URLs to sessions.

austinzheng:
@mrackwitz, is this ticket blocking client permissions?

mrackwitz:
Not directly, it is only breaking the integration test on realm-sync-services (& realm-js). For now we can probably just disable this test.

PR #270, opened 2016-12-02
Support multiple on-disk files per logical synced Realm

austinzheng:
@jpsim @tgoyne @bdash @alazier 

Fixes https://github.com/realm/realm-object-store/issues/243

Changes:
- `SyncConfig` now has an optional custom-path property
- `Realm::Config` now has a method to properly configure itself with a `SyncConfig`
- Users now track "default-path" Realms (indexed by URL, used by clients) separately from "custom-path" Realms
- Minor API changes to allow default-path and custom-path Realms to be created properly
- Added and upgraded tests

### DIFF
--- a/src/sync/sync_user.cpp
+++ b/src/sync/sync_user.cpp
@@ -45,45 +45,68 @@ SyncUser::SyncUser(std::string refresh_token,
 std::vector<std::shared_ptr<SyncSession>> SyncUser::all_sessions()
 {
     std::lock_guard<std::mutex> lock(m_mutex);
-    std::vector<std::shared_ptr<SyncSession>> sessions;
+    std::vector<std::shared_ptr<SyncSession>> sessions_vector;
     if (m_state == State::Error) {
-        return sessions;
+        return sessions_vector;
     }
-    for (auto it = m_sessions.begin(); it != m_sessions.end();) {
-        if (auto ptr_to_session = it->second.lock()) {
-            if (!ptr_to_session->is_in_error_state()) {
-                sessions.emplace_back(std::move(ptr_to_session));
-                it++;
-                continue;
+    auto add_sessions_to_vector = [&](SyncSessionMap& sessions){
+        for (auto it = sessions.begin(); it != sessions.end();) {
+            if (auto ptr_to_session = it->second.lock()) {
+                if (!ptr_to_session->is_in_error_state()) {
+                    sessions_vector.emplace_back(std::move(ptr_to_session));
+                    it++;
+                    continue;
+                }
             }
+            // This session is bad, destroy it.
+            it = sessions.erase(it);
         }
-        // This session is bad, destroy it.
-        it = m_sessions.erase(it);
-    }
-    return sessions;
+    };
+    add_sessions_to_vector(m_sessions);
+    add_sessions_to_vector(m_custom_sessions);
+    return sessions_vector;
 }
 
-std::shared_ptr<SyncSession> SyncUser::session_for_url(const std::string& url)
+std::shared_ptr<SyncSession> SyncUser::session_for_key(const std::string& key, SyncSessionMap& sessions)
 {
     std::lock_guard<std::mutex> lock(m_mutex);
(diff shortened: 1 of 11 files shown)


austinzheng:
Some discussion follows:

== Changes to `Realm::Config` and `SyncConfig` ==

Before, it wasn't explicit that a binding that set a `SyncConfig` value on a `Realm::Config` had to also set that `Realm::Config`'s `path` property at the same time. The new API ensures that this mistake can't happen.

`SyncConfig` has a new optional property that allows you to set a custom path on disk where the Realm file should live. This is important for enabling other features in the future (like moving the Realm file to a 'recovery' directory if a client reset becomes necessary).

== "Default-path sessions" ==

Default-path sessions are sessions for a given URL which use the location on disk generated by the sync code to store the on-disk file. There can be at most one default-path session per user per URL, and these can be looked up using the existing `SyncUser::session_for_url()` API.

Clients will almost always use default-path sessions (with the exception of the Realm Browser). The point is that clients should never have to care about where a particular synced Realm's files live on disk; that's an internal implementation detail.

Synced Realms opened at the default path are the only ones whose files are deleted by the metadata subsystem when their owning users are deleted.

== "Custom-path sessions" ==

Custom-path sessions are sessions opened using a `SyncConfig` whose `custom_file_path` property wasn't `none`. There can be as many custom-path sessions per user per URL as desired. They can be looked up using the new `SyncUser::session_for_custom_file_path()` API.

This functionality is intended for the Realm Browser (previously, it was hacked together at the Cocoa binding layer), and for servers using the sync facilities within the object store. It's assumed that such consumers will want to play a bigger role in managing the location of the Realm files on-disk.

The metadata subsystem will properly handle Realms opened at custom paths for things like deleting a specific Realm, backing up the Realm file after client reset, or any future functionality that we come up with.

== All sessions ==

The existing `SyncUser::all_sessions()` API returns both types of sessions.

Both types of sessions properly unbind and rebind when their users log out and/or their corresponding Realms are created and deleted. Apart from the differences described above, they function completely identically.

src/sync/sync_user.cpp
```diff
@@ -199,4 +244,14 @@ void SyncUser::register_session(std::shared_ptr<SyncSession> session)
     }
 }
 
+void SyncUser::register_session(std::shared_ptr<SyncSession> session)
+{
+    std::unique_lock<std::mutex> lock(m_mutex);
+    if (auto custom_path = session->config().custom_file_path) {
```
simonask: Use `auto&` to avoid a copy (maybe even `const auto&` if you prefer).

tests/sync/session.cpp
```diff
@@ -39,22 +39,24 @@ template <typename FetchAccessToken, typename ErrorHandler>
 std::shared_ptr<SyncSession> sync_session(SyncServer& server, std::shared_ptr<SyncUser> user, const std::string& path,
                                           FetchAccessToken&& fetch_access_token, ErrorHandler&& error_handler,
                                           SyncSessionStopPolicy stop_policy=SyncSessionStopPolicy::AfterChangesUploaded,
+                                          bool treat_as_custom=false,
                                           std::string* on_disk_path=nullptr)
 {
     std::string url = server.base_url() + path;
-    SyncTestFile config({user, url, std::move(stop_policy),
+    SyncConfig sync_config = {user, url, std::move(stop_policy),
```
simonask: `=` superfluous, but fine if you prefer this style :)

austinzheng: Oh, I forgot that you can do `variable{a, b, c}`. I like that better.

austinzheng:
I made both the style corrections that were requested, and I did a minor rewrite of three functions to avoid the "pass around a lock" pattern.

austinzheng:
I talked to Mark "in person", and we decided that a simpler approach (get rid of indexing by URLs completely, and use `custom_file_path` instead of adding a property to the `SyncConfig`). Stay tuned...

austinzheng:
Closing in favor of #280.

PR #274, opened 2016-12-02
Remove redundant Results constructors and set m_table in all of them

tgoyne:
One of the constructors added in #235 did not set m_table, which broke `sort()` even when not using `distinct()`.

@jpsim


tgoyne:
There's currently only Results tests for the functionality added after porting it from obj-c. Should fix this at some point, but this is the first time it's been an issue so I've never prioritized it.

PR #276, opened 2016-12-02
Fix a crash when calling SyncManager::reset_for_testing

### LINKED ISSUE #269: `SyncSession` might be killed in the middle of the wait-for-upload callback
It's possible that `SyncSession` might be killed in the middle of its "wait for upload" callback if `SyncManager::reset_for_testing()` is called. If this is an issue, we should fix it.

bdash:
`SyncClient`'s lifetime was managed by a `shared_ptr`, resulting in `SyncManager::reset_for_testing` failing to destroy it when expected. This meant that its runloop was still running, and that it could continue to deliver callbacks to sessions that `reset_for_testing` was in the process of destroying.

To address this `SyncClient` is now owned by `SyncManager`, with `SyncSession`'s holding only a reference to their client. This is safe because sync asserts that all sessions have been destroyed when the client is destroyed.

/cc @austinzheng @alazier 

Fixes #269.

### DIFF
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -211,15 +211,7 @@ struct sync_session_states::Dying : public SyncSession::State {
     void enter_state(std::unique_lock<std::mutex>&, SyncSession& session) const override
     {
         size_t current_death_count = ++session.m_death_count;
-        session.m_session->async_wait_for_upload_completion([session=&session, current_death_count](std::error_code error_code) {
-            if (error_code == util::error::operation_aborted) {
-                // Session was killed beneath us. Don't do anything.
-                return;
-            }
-            // FIXME: It's possible for the session to be destroyed while the callback is running,
-            // which is something we should address in the future. It is only possible when 
-            // SyncManager::reset_for_testing() is called.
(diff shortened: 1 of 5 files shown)


src/sync/sync_session.cpp
```diff
@@ -211,15 +211,7 @@ struct sync_session_states::Dying : public SyncSession::State {
     void enter_state(std::unique_lock<std::mutex>&, SyncSession& session) const override
     {
         size_t current_death_count = ++session.m_death_count;
-        session.m_session->async_wait_for_upload_completion([session=&session, current_death_count](std::error_code error_code) {
-            if (error_code == util::error::operation_aborted) {
```
austinzheng: Should we keep this? Are there any non-test situations in which the `Client` might be killed but the application continues running?

bdash: If we leave this early-return in place then sessions whose wait is aborted won't be removed from the inactive sessions list.

austinzheng: For example, in the case where the Client is stopped but the sessions were in the inactive sessions list? Okay, that makes sense.

alazier:
This fixes the remaining issues we are seeing on JS

PR #277, opened 2016-12-05
Merge latest into js

alazier:
@kristiandupont @bdash


bdash:
CI shows the sync tests are broken on this branch. It looks like tests/util/test_file.cpp has been incorrectly merged at some point as [the version on this branch](https://github.com/realm/realm-object-store/blob/580776d3fe300b54ed769672d8c3a1e2914a259e/tests/util/test_file.cpp) is missing the implementations of the [two `SyncTestFile` constructors that are present in the version on master](https://github.com/realm/realm-object-store/blob/master/tests/util/test_file.cpp#L84-L103).

PR #278, opened 2016-12-05
Throw if accessing a closed realm

kristiandupont:


src/shared_realm.hpp
```diff
@@ -386,6 +387,11 @@ class IncorrectThreadException : public std::logic_error {
     IncorrectThreadException() : std::logic_error("Realm accessed from incorrect thread.") {}
 };
 
+class ClosedRealmException : public std::runtime_error {
```
tgoyne: This is a logic error.

bdash:
This could do with tests.

alazier:
We should either make these changes on master, or base this pr on the js branch.

bdash:
Changes that aren't specific to a given binding should really be merged to master first. We're trying to eliminate differences, not introduce more.

PR #279, opened 2016-12-05
Expose current user in sync_manager

nirinchev:
With the API exposed in OS, bindings won't have to implement it themselves.

### DIFF
--- a/src/sync/sync_manager.cpp
+++ b/src/sync/sync_manager.cpp
@@ -303,6 +303,26 @@ std::vector<std::shared_ptr<SyncUser>> SyncManager::all_logged_in_users() const
     return users;
 }
 
+std::shared_ptr<SyncUser> SyncManager::get_current_user() const
+{
+    std::lock_guard<std::mutex> lock(m_user_mutex);
+    
+    std::shared_ptr<SyncUser> result = nullptr;
+    for (auto& it : m_users) {
+        auto user = it.second;
+        if (user->state() == SyncUser::State::Active) {
+            if (result == nullptr) {
+                result = std::move(user);
+            }
+            else {
+                throw std::logic_error("get_current_user cannot be called if more that one valid, logged-in user exists.");
+            }
+        }
+    }
+    
+    return result;
+}
+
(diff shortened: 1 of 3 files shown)


src/sync/sync_manager.cpp
```diff
@@ -303,6 +303,26 @@ std::vector<std::shared_ptr<SyncUser>> SyncManager::all_logged_in_users() const
     return users;
 }
 
+std::shared_ptr<SyncUser> SyncManager::get_current_user() const
+{
+    std::lock_guard<std::mutex> lock(m_user_mutex);
+    
+    std::shared_ptr<SyncUser> result = nullptr;
```
bdash: I'd be inclined to write this as something like:
```c++
auto is_active_user = [](auto& el) { return el.second->state() == SyncUser::State::Active; };
auto it = std::find_if(m_users.begin(), m_users.end(), is_active_user);
if (it == m_users.end())
    return nullptr;

if (std::find_if(it, m_users.end(), is_active_user) != m_users.end())
    throw std::logic_error("get_current_user cannot be called if more that one valid, logged-in user exists.");

return it->second;
```

That better highlights what the function is doing.

nirinchev:
Ping @bdash - fairly new to writing OS tests, so not sure if they follow your preferred way of doing it, but I'd be happy to address any feedback.

src/sync/sync_manager.cpp
```diff
@@ -303,6 +303,21 @@ std::vector<std::shared_ptr<SyncUser>> SyncManager::all_logged_in_users() const
...
+{
+    std::lock_guard<std::mutex> lock(m_user_mutex);
+    
+    auto is_active_user = [](auto& el) { return el.second->state() == SyncUser::State::Active; };
+    auto it = std::find_if(m_users.begin(), m_users.end(), is_active_user);
+    if (it == m_users.end()) {
+        return nullptr;
+    }
+    if (std::find_if(std::next(it), m_users.end(), is_active_user) != m_users.end()) {
+        throw std::logic_error("Current user is not valid if more that one valid, logged-in user exists.");
```
nhachicha: > more that 

more than ?

PR #282, opened 2016-12-06
Add support for an event notifications mechanism

austinzheng:
Replaces #262, which I accidentally merged and then backed out. @bdash 

Changes:
- Implemented a notifier that bindings can override to get sync-wide event notifications
- A few internal API changes to facilitate this new feature
- Object store sync test infrastructure slightly overhauled to facilitate testing
- Added tests to exercise the new functionality

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -79,6 +79,7 @@ if(REALM_ENABLE_SYNC)
     list(APPEND HEADERS
         sync/sync_config.hpp
         sync/sync_manager.hpp
+        sync/sync_notifier.hpp
         sync/sync_session.hpp
         sync/sync_user.hpp
         sync/impl/sync_client.hpp
--- a/src/sync/sync_manager.cpp
+++ b/src/sync/sync_manager.cpp
@@ -47,6 +47,14 @@ void SyncManager::configure_file_system(const std::string& base_file_path,
         std::string user_token;
         util::Optional<std::string> server_url;
     };
+    {
+        // Create the notifier, if applicable.
+        std::lock_guard<std::mutex> lock(m_mutex);
+        if (m_notifier_factory) {
+            m_notifier = m_notifier_factory->make_notifier();
+        }
+        m_notifier_factory = nullptr;
+    }
 
     std::vector<UserCreationData> users_to_add;
     {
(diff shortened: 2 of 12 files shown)


jpsim:
This seems like a promising approach, but since this is squarely aimed at allowing bindings to expose similar APIs, I'd prefer actually discussing whether or not such APIs should be available at the binding level.

For example, up until now the idea to notify users of a client reset event was to report an error unique to that scenario. There might be value in decoupling general purpose events from errors, but it's not clear what shape those APIs would take.

I'd like to see a higher-level API proposal that can justify building this before we go down this route.

cmelchior:
I'm a bit concerned about the asynchronous nature of some of those events, e.g. `session_destroyed`. From the time it is sent to the time it is received there is a window where the binding might still assume the session is around and try to interact with it, which will crash.

bdash:
A big concern I have is the callbacks that are made out of the object store while locks are held. It's asking for deadlocks.

cmelchior:
Would it be possible to rethink how locking works in the OS, because the current approach prevents a lot of things. E.g. in the Java binding we could also log all state changes in the Session which was really helpful for logging. This is also imossible with the current implementation.

bdash:
I don't think that this PR is an appropriate place for a discussion about changing how we handle locking in object store as it's not really related to this PR at all.

cmelchior:
@bdash Fair enough. I created #293 to capture discussion.

austinzheng:
To step back a bit...

The original motivation of this PR-cum-proposal was to provide a mechanism needed for client reset. More specifically, sync needs to be able to do two things: 

* to ask the app whether it should go through with a proposed client reset
* to inform the app that a Realm that has been reset has had its files deleted/backed up.

The "easiest" way to do this would have been with custom handlers that the bindings could set on `SyncManager`, similar to how the (now-gone) client-level error handler worked. I think this would have led to a proliferation of handlers as we added new features, which is something I wanted to avoid with a higher-level notification system.

In terms of locking: that's a serious issue. I suppose I was assuming that each binding would use an asynchronous mechanism to expose the notification to the app through the final API. For example, Cocoa would `dispatch_async` a block which would call the handler the app registered, or something. I'm not sure whether this solution scales to our other bindings, though. Also, we're discussing reworking how locks work in the object store (since they apparently cause other issues), so perhaps that conversation should come first...

austinzheng:
I think we should be able to discuss this sort of mechanism separate from the issue of object store locking. (Since these APIs are not consumed directly by an end user, we can be careful about how we use them as a workaround.)

Either that, or maybe we could employ the object store's runloop machinery. Instead of directly calling a callback, we could send events to the runloop and have the listeners provided by the bindings listen for events on the runloop. I think that would get around the potential deadlock issue.

austinzheng:
Per comments in https://github.com/realm/product/issues/168, this PR is being abandoned for now.

PR #284, opened 2016-12-06
fix -Wundef warnings when building for cocoa

jpsim:
cocoa's build system doesn't define `REALM_PLATFORM_NODE` or `REALM_USE_UV`, so check if these are defined before reading their value. Motivated by realm/realm-cocoa#4409

/cc @bdash


src/util/event_loop_signal.hpp
```diff
@@ -21,11 +21,11 @@
 
 #include <realm/util/features.h>
 
-#if (defined(REALM_HAVE_UV) && REALM_HAVE_UV && !REALM_PLATFORM_APPLE) || REALM_PLATFORM_NODE
+#if (defined(REALM_HAVE_UV) && REALM_HAVE_UV && !REALM_PLATFORM_APPLE) || (defined(REALM_PLATFORM_NODE) && REALM_PLATFORM_NODE)
 #define REALM_USE_UV 1
 #endif
 
-#if REALM_USE_UV
+#if defined(REALM_USE_UV) && REALM_USE_UV
```
bdash: IMO it'd be preferable to add an `#else` branch to the `#if` that defines `REALM_USE_UV` to set it to 0. Ditto for other similar defines (`REALM_USE_EPOLL` in external_commit_helper.hpp, and `REALM_USE_UV` in event_loop.cpp).

jpsim: I did consider that. I'll do that. Just to be clear, the current approach for `REALM_PLATFORM_NODE` is fine because it's defined in CMake, right?

bdash: Yes, the expectation is that `REALM_PLATFORM_NODE` is defined by the build system.

jpsim:
@bdash updated

xiaodao:
may I know when will this change be released? 

our project wants to upgrade to Realm 2.1 but now is blocked by this error.

bdash:
You'll want to follow realm/realm-cocoa#4418 to see when this is integrated into realm-cocoa.

PR #285, opened 2016-12-07
Make the generic EventLoopSignal pluggable

fealebenpae:
This allows the generic `EventLoopSignal` to be configured to use an event loop abstraction not known at compile-time. By default its operation is still a no-op, but bindings can now plug their own implementations of the `get_eventloop`, `post_on_eventloop`, and `release_eventloop` operations.

This is needed for https://github.com/realm/realm-dotnet/pull/977 where we need `WeakRealmNotifier` to post notifications using an event loop API that's only accessible from C#.


src/util/event_loop_signal.hpp
```diff
@@ -27,11 +27,17 @@
 #define REALM_USE_UV 0
 #endif
 
+#if !defined(REALM_USE_CF) && REALM_PLATFORM_APPLE
+#define REALM_USE_CF 1
+#elif !defined(REALM_USE_ALOOPER) && REALM_ANDROID
+#define REALM_USE_ALOOPER 1
+#endif
```
bdash: I take it that the intent is that .NET will define REALM_USE_CF and REALM_USE_ALOOPER to 0?

fealebenpae: Yeah, that's exactly what happens now.

src/util/generic/event_loop_signal.cpp
```diff
@@ -0,0 +1,28 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#include <utility>
```
bdash: Nothing in this .cpp file appears to need this #include. Also note that it's preferable to include the .cpp file's header prior to any other headers to ensure that the header includes everything it needs.

fealebenpae: The `std::move` in the header needs it, but I felt weird about including it there. I can do that if it makes more sense, though.

bdash: It's better to include it in the header than to force anyone who may include the header to first include `<utility>`.

fealebenpae:
Thanks, @bdash!

src/CMakeLists.txt
```diff
@@ -68,6 +68,10 @@ else()
     list(APPEND SOURCES impl/generic/external_commit_helper.cpp)
 endif()
 
+if(NOT APPLE AND NOT REALM_PLATFORM STREQUAL "Android")
+    list(APPEND SOURCES util/generic/event_loop_signal.cpp)
+endif ()
```
bdash: There's typically no whitespace before the parens in CMake files.

PR #286, opened 2016-12-08
Added support for LIKE queries to predicate parser

astigsen:
This was tested with the js binding as well.

@bdash @alazier

### DIFF
--- a/tests/parser.cpp
+++ b/tests/parser.cpp
@@ -80,6 +80,9 @@ static std::vector<std::string> valid_queries = {
     "contains contains 'contains'",
     "beginswith beginswith 'beginswith'",
     "endswith endswith 'endswith'",
+    "a like '*'",
+    "a LIKE b",
+    "a LIKE '?abc*'",
     "NOT NOT != 'NOT'",
     "AND == 'AND' AND OR == 'OR'",
     // FIXME - bug
--- a/tests/parser.cpp
+++ b/tests/parser.cpp
@@ -80,9 +80,13 @@ static std::vector<std::string> valid_queries = {
     "contains contains 'contains'",
     "beginswith beginswith 'beginswith'",
     "endswith endswith 'endswith'",
+    "like like '*'",
+    "LIKE LIKE '*'",
     "a like '*'",
     "a LIKE b",
     "a LIKE '?abc*'",
+    "a LIKE '*c*n*'",
+    "a LIKE 'a???e?",
     "NOT NOT != 'NOT'",
     "AND == 'AND' AND OR == 'OR'",
     // FIXME - bug
--- a/tests/parser.cpp
+++ b/tests/parser.cpp
@@ -86,7 +86,7 @@ static std::vector<std::string> valid_queries = {
     "a LIKE b",
     "a LIKE '?abc*'",
     "a LIKE '*c*n*'",
-    "a LIKE 'a???e?",
+    "a LIKE 'a???e?'",
     "NOT NOT != 'NOT'",
     "AND == 'AND' AND OR == 'OR'",
     // FIXME - bug
@@ -152,6 +152,8 @@ static std::vector<std::string> invalid_queries = {
     "0 contains1",
     "a contains_something",
     "endswith 0",
+    "LIKE[A] 'a*b'",
+    "LIKE LIKE",
 
     // atoms/groups
     "0=0)",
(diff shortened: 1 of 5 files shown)


tests/parser.cpp
```diff
@@ -80,6 +80,9 @@ static std::vector<std::string> valid_queries = {
     "contains contains 'contains'",
     "beginswith beginswith 'beginswith'",
     "endswith endswith 'endswith'",
+    "a like '*'",
```
bdash: We should probably test that column names of `like` still work, similar to what we do above for `contains` / `beginswith` / `endswith`.

tests/query.json
```diff
@@ -201,6 +203,7 @@
         ["QueryCount", 5, "StringObject", "stringCol BEGINSWITH[c] 'A'"],
         ["QueryCount", 4, "StringObject", "stringCol ENDSWITH[c] 'c'"],
         ["QueryCount", 2, "StringObject", "stringCol CONTAINS[c] 'B'"]
+        ["QueryCount", 2, "StringObject", "stringCol LIKE[c] '*B?'"]
```
bdash: It'd be good to test handling of invalid predicates like `stringCol LIKE 1`, and substituted patterns like `stringCol LIKE $0`

austinzheng:
This seems useful. Should we revive it?

bdash:
At some point, yes. Someone just needs to spend the time expanding the test coverage of this new functionality.

austinzheng:
If tests are the only missing thing, I'll see what I can do later this week. Seems pretty straightforward.

austinzheng:
I'm writing tests; when I get home and confirm they work I'll update this PR.

austinzheng:
I'm using CI to check my work, so I'll wait until it succeeds to merge.

tests/query.json
```diff
@@ -200,7 +209,13 @@
         ["QueryCount", 3, "StringObject", "stringCol ==[c] 'a'"],
         ["QueryCount", 5, "StringObject", "stringCol BEGINSWITH[c] 'A'"],
         ["QueryCount", 4, "StringObject", "stringCol ENDSWITH[c] 'c'"],
-        ["QueryCount", 2, "StringObject", "stringCol CONTAINS[c] 'B'"]
+        ["QueryCount", 2, "StringObject", "stringCol CONTAINS[c] 'B'"],
+        ["QueryCount", 2, "StringObject", "stringCol LIKE[c] '*B?'"],
+        ["QueryCount", 0, "StringObject", "stringCol LIKE 1"],
+        ["QueryCount", 0, "StringObject", "stringCol LIKE LIKE"],
+        ["QueryCount", 0, "StringObject", "stringCol LIKE[A] '*BC'"],
```
bdash: What does `[A]` do here? Does this query execute successfully?

austinzheng: If it doesn't, I'll remove it and add it to the list of malformed queries in `parser.cpp`. It seems that if `"stringCol LIKE 1"` parses then something like this should parse too. I need to double-check.

austinzheng: (Sorry, I didn't answer your question. It's an invalid flag and the query should return 0 results.)

bdash: `[c]` / `[C]` is the only modifier that's accepted based on https://github.com/realm/realm-object-store/blob/86138ce4e55ca04e975c1165d409f02dcb729b6b/src/parser/parser.cpp#L73

I'd expect `stringCol LIKE 1` to _parse_ successfully but be rejected when we attempt to build a query from it due to the type mismatch.

bdash: Hmm, I'm surprised that we don't distinguish between a query that matched 0 objects and an error (I'm assuming an exception is being thrown).

austinzheng: I moved the bad test cases into `parser.cpp`'s bad test case list, and the tests passed, so you were right.

I'm actually not sure how `query.json` fits into the tests. It isn't referenced anywhere else within the object store project, and breaking the JSON doesn't affect the test outcomes.

bdash: It looks like query.json is used by the JS binding. Not entirely sure why it lives in the object store repo.

austinzheng:
I'll figure out how to run the `json` test file from the JS binding tonight or Monday. I want to ensure everything is passing before merging anything else.

bdash:
I suspect @ironage took care of this after moving the query parser to core.

PR #287, opened 2016-12-08
Getting ready to merge realm-java into master

kneth:
Instead of polluting `realm-java`, I have create a new branch. Almost except one unit test is working. The remaining failing test is cause by how core uses `mkdtemp` (see also https://code.google.com/p/android/issues/detail?id=74497) but I will make a workaround later. There are still a few issue with linking but I can link manually so I will fix that later too.

Please give me some feedback on the CMake parts of this PR - and only that.

Dependencies
 * https://github.com/realm/realm-core/pull/2355
 * https://github.com/realm/realm-sync/pull/1027

@cmelchior @emanuelez @bdash @radu-tutueanu


CMake/RealmCore.cmake
```diff
@@ -238,25 +237,41 @@ endfunction()
...
-        ${USES_TERMINAL_BUILD}
-        ${ARGV}
-        )
+    if(REALM_PLATFORM STREQUAL "Android")
+      ExternalProject_Add(realm-core
+            PREFIX ${core_prefix_directory}
+            BUILD_IN_SOURCE 1
+            BUILD_COMMAND sh build.sh build-android
+            INSTALL_COMMAND ""
+            CONFIGURE_COMMAND REALM_ENABLE_ENCRYPTION=YES REALM_ENABLE_ASSERTIONS= sh build.sh config
```
bdash: The expectation when using existing local source trees for core and sync is that the trees are already configured.

CMake/RealmCore.cmake
```diff
@@ -238,25 +237,41 @@ endfunction()
...
-        INSTALL_COMMAND ""
-        CONFIGURE_COMMAND ""
-        ${USES_TERMINAL_BUILD}
-        ${ARGV}
-        )
+    if(REALM_PLATFORM STREQUAL "Android")
+      ExternalProject_Add(realm-core
+            PREFIX ${core_prefix_directory}
+            BUILD_IN_SOURCE 1
+            BUILD_COMMAND sh build.sh build-android
```
bdash: Since this command should be the only difference, it'd be better to hoist it into a variable and leave only a single call to `ExternalProject_add`.

CMake/RealmCore.cmake
```diff
@@ -238,25 +237,41 @@ endfunction()
...
+    if(REALM_PLATFORM STREQUAL "Android")
+        set(CONFIG_CMD REALM_ENABLE_ENCRYPTION=YES REALM_ENABLE_ASSERTIONS= sh build.sh config)
+    else()
+        set(CONFIG_CMD REALM_ENABLE_ENCRYPTION=YES REALM_ENABLE_ASSERTIONS=YES sh build.sh config)
+    endif()
     build_realm_core(GIT_REPOSITORY "https://github.com/realm/realm-core.git"
                      GIT_TAG ${branch}
-                     CONFIGURE_COMMAND test -f src/config.mk ||
-                                       REALM_ENABLE_ENCRYPTION=YES REALM_ENABLE_ASSERTIONS=YES sh build.sh config
+                     CONFIGURE_COMMAND test -f src/config.mk || CONFIG_CMD
```
bdash: This is currently breaking the build for this branch since you're not expanding the variable.

CMake/RealmCore.cmake
```diff
@@ -272,22 +287,51 @@ endfunction()
...
-        ${USES_TERMINAL_BUILD}
-        ${ARGV}
-        )
+    if(REALM_PLATFORM STREQUAL "Android")
+        ExternalProject_Add(realm-sync-lib
+            DEPENDS realm-core
+            PREFIX ${cmake_files}/realm-sync
+            BUILD_IN_SOURCE 1
+            BUILD_COMMAND sh build.sh build-android
+            CONFIGURE_COMMAND rm -f src/config.mk && REALM_FORCE_OPENSSL=YES REALM_ENABLE_ENCRYPTION=YES REALM_ENABLE_ASSERTIONS= sh build.sh config && echo "ENABLE_ENCRYPTION    = yes" >> src/config.mk
```
bdash: Ditto about skipping configuration when using existing local source trees.

CMake/RealmCore.cmake
```diff
@@ -272,22 +287,51 @@ endfunction()
...
-        INSTALL_COMMAND ""
-        ${USES_TERMINAL_BUILD}
-        ${ARGV}
-        )
+    if(REALM_PLATFORM STREQUAL "Android")
+        ExternalProject_Add(realm-sync-lib
+            DEPENDS realm-core
+            PREFIX ${cmake_files}/realm-sync
+            BUILD_IN_SOURCE 1
+            BUILD_COMMAND sh build.sh build-android
```
bdash: Ditto about hoisting the different command into a variable rather than repeating the entire `ExternalProject_Add`.

CMake/RealmCore.cmake
```diff
@@ -342,10 +391,15 @@ endfunction()
 
 function(clone_and_build_realm_sync branch)
     set(cmake_files ${CMAKE_CURRENT_BINARY_DIR}${CMAKE_FILES_DIRECTORY})
+    if(REALM_PLATFORM STREQUAL "Android")
+        set(CONFIG_CMD, rm -f src/config.mk && REALM_CORE_PREFIX=${cmake_files}/realm-core/src/realm-core REALM_FORCE_OPENSSL=YES REALM_ENABLE_ENCRYPTION=YES REALM_ENABLE_ASSERTIONS= sh build.sh config && echo "ENABLE_ENCRYPTION    = yes" >> src/config.mk)
```
bdash: The bits related to src/config.mk here are strange. Why does it always delete src/config.mk? Why does it explicitly add a line to src/config.mk after configuring?

CMake/RealmCore.cmake
```diff
@@ -342,10 +391,15 @@ endfunction()
...
+    else()
+        set(CONFIG_CMD, test -f src/config.mk || REALM_CORE_PREFIX=${cmake_files}/realm-core/src/realm-core sh build.sh config)
```
bdash: The comma shouldn't be here.

CMake/RealmCore.cmake
```diff
@@ -342,10 +391,15 @@ endfunction()
...
+    endif()
+
     build_realm_sync(GIT_REPOSITORY "git@github.com:realm/realm-sync.git"
                      GIT_TAG ${branch}
-                     CONFIGURE_COMMAND test -f src/config.mk ||
-                                       REALM_CORE_PREFIX=${cmake_files}/realm-core/src/realm-core sh build.sh config
+                     CONFIGURE_COMMAND CONFIG_CMD
```
bdash: You need to expand the variable.

src/impl/realm_coordinator.hpp
```diff
@@ -145,7 +146,9 @@ class RealmCoordinator : public std::enable_shared_from_this<RealmCoordinator> {
     std::unique_ptr<_impl::ExternalCommitHelper> m_notifier;
     std::function<void(VersionID, VersionID)> m_transaction_callback;
 
+#if REALM_ENABLE_SYNC
```
bdash: Doesn't appear to be related to the Java build.

src/shared_realm.hpp
```diff
@@ -31,6 +31,8 @@
 #include <memory>
 #include <thread>
 
+#include <realm/group_shared.hpp>
```
bdash: I provided feedback about this, and several other parts of the code in this PR, on one of the other Java PRs. Can you please take a look at the previous PRs so I don't have to repeat it here?

beeender:
@kneth Is the cmake changes necessary? I think we'd better only have minimal changes in this PR. BTW, I will fix the `shared_group.h` header problem and push to this PR.

CMake/RealmCore.cmake
bdash: I can't see how adding `ENABLE_ENCRYPTION` to `src/config.mk` could possibly be required. No variable of that name is ever referenced in the realm-sync codebase. Whether or not encryption is enabled is determined solely by core's configuration.

CMake/RealmCore.cmake
```diff
@@ -272,22 +281,44 @@ endfunction()
...
+        endif()
+    endif()
+
+    
     ExternalProject_Get_Property(realm-sync-lib SOURCE_DIR)
     set(sync_directory ${SOURCE_DIR})
 
-    set(sync_library_debug ${sync_directory}/src/realm/librealm-sync-dbg.a)
-    set(sync_library_release ${sync_directory}/src/realm/librealm-sync.a)
+    if(REALM_PLATFORM STREQUAL "Android")
```
bdash: The path to the built libraries could also be hoisted into the Android-vs-other if block above, so we don't have to duplicate the list of libraries for both sync and the sync server.

CMake/RealmCore.cmake
```diff
@@ -342,10 +378,15 @@ endfunction()
 
 function(clone_and_build_realm_sync branch)
     set(cmake_files ${CMAKE_CURRENT_BINARY_DIR}${CMAKE_FILES_DIRECTORY})
+    if(REALM_PLATFORM STREQUAL "Android")
+        set(config_cmd test -f src/config.mk || REALM_CORE_PREFIX=${cmake_files}/realm-core/src/realm-core REALM_FORCE_OPENSSL=YES REALM_ENABLE_ENCRYPTION=YES REALM_ENABLE_ASSERTIONS= sh build.sh config && echo "ENABLE_ENCRYPTION    = yes" >> src/config.mk)
```
bdash: As stated previously, there's no need to set `REALM_ENABLE_ENCRYPTION` here, or to add `ENABLE_ENCRYPTION` to src/config.mk. realm-sync doesn't use these.

src/shared_realm.cpp
```diff
@@ -364,10 +364,13 @@ void Realm::update_schema(Schema schema, uint64_t version, MigrationFunction mig
...
 
     m_group->set_schema_change_notification_handler(nullptr);
-    transaction::begin_without_validation(*m_shared_group);
+    if (!in_transaction) {
+        transaction::begin_without_validation(*m_shared_group);
+    }
     add_schema_change_handler();
 
     // Cancel the write transaction if we exit this function before committing it
+    // FIXME: don't cancel if in_transaction == true
```
bdash: Why add this FIXME rather than adding the `!in_transaction` check to `WriteTransactionGuard`'s destructor? Won't the behavior as-is result in some subtle bugs?

src/shared_realm.hpp
```diff
@@ -168,6 +169,9 @@ class Realm : public std::enable_shared_from_this<Realm> {
 
         /// A data structure storing data used to configure the Realm for sync support.
         std::shared_ptr<SyncConfig> sync_config;
+
+        // FIXME: realm java is manage sync in the java level, but needs Realm created sync history.
```
bdash: The wording here is awkward. Perhaps something like:

> Realm Java manages sync at the Java level, so it needs to create Realms using the sync history format.

tests/sync/sync_test_utils.cpp
```diff
@@ -49,7 +49,11 @@ std::string tmp_dir() {
     const char* dir = getenv("TMPDIR");
     if (dir && *dir)
         return dir;
-    return "/tmp";
+#if REALM_ANDROID
+        return "/data/local/tmp";
```
bdash: Indentation here is strange.

CMake/RealmCore.cmake
kneth: If `ENABLE_ENCRYPTION` isn't set, OpenSSL will _not_ be downloaded, built and included. `REALM_ENABLE_ENCRYPTION=YES` is not required (it is ignored by realm-sync's `build.sh`).

bdash: Can you clarify what uses `ENABLE_ENCRYPTION`? I see nothing in realm-sync that does. AFAICT `build.sh build-android` unconditionally [downloads](https://github.com/realm/realm-sync/blob/7b311c1bc19388608e52e984f640ed7f5c98b114/build.sh#L808) and [builds](https://github.com/realm/realm-sync/blob/7b311c1bc19388608e52e984f640ed7f5c98b114/build.sh#L872-L899) OpenSSL. Are you perhaps thinking of realm-core, which does use `ENABLE_ENCRYPTION`?

PR #288, opened 2016-12-09
Results aggregate on all null rows

beeender:
* min/max should return none on rows with all null values.
* aggregate should throw UnsupportedColumnTypeException when the
  collection is empty instead of return none.
* Add test cases.


src/results.cpp
```diff
@@ -370,20 +374,24 @@ util::Optional<Mixed> Results::aggregate(size_t column, bool return_none_for_emp
...
-                     [=](auto const& table) { return table.maximum_float(column); },
-                     [=](auto const& table) { return table.maximum_double(column); },
-                     [=](auto const& table) { return table.maximum_timestamp(column); });
+    size_t return_ndx;
+    auto results = aggregate(column, true, "max",
+                     [&](auto const& table) { return table.maximum_int(column, &return_ndx); },
+                     [&](auto const& table) { return table.maximum_float(column, &return_ndx); },
+                     [&](auto const& table) { return table.maximum_double(column, &return_ndx); },
+                     [&](auto const& table) { return table.maximum_timestamp(column, &return_ndx); });
+    return return_ndx == npos ? none : results;
```
bdash: We could push this repeated test into `aggregate` since it is of the same vein as the `return_none_for_empty` work that it already handles. It'd require adding a `return_ndx` reference parameter to each of these lambdas though, so maybe that work would outweigh the benefit from removing the small amount of repetition?

tests/results.cpp
```diff
@@ -2027,3 +2027,184 @@ TEST_CASE("distinct") {
...
+            {"int", PropertyType::Int, "", "", false, false, true},
+            {"float", PropertyType::Float,  "", "", false, false, true},
+            {"double", PropertyType::Double, "", "", false, false, true},
+            {"date", PropertyType::Date, "", "", false, false, true},
+        }},
+    });
+
+    auto table = r->read_group().get_table("class_object");
+
+    SECTION("part of rows with null values") {
```
bdash: "one row with null values" would be clearer.

tests/results.cpp
```diff
@@ -2027,3 +2027,184 @@ TEST_CASE("distinct") {
...
+        // table:
+        //  null, null, null,  null,
+        //  0,    0,    0,    (0, 0)
+        //  2,    2,    2,    (2, 0)
+        r->commit_transaction();
+
+        SECTION("max") {
+            Results results;
+
+            SECTION("results built from query") {
```
bdash: This section doesn't appear in the other aggregates?

tests/results.cpp
```diff
@@ -2027,3 +2027,184 @@ TEST_CASE("distinct") {
...
+            Results results(r, table->where());
+
+            REQUIRE(results.sum(0)->get_int() == 0);
+            REQUIRE(results.sum(1)->get_double() == 0.0);
+            REQUIRE(results.sum(2)->get_double() == 0.0);
+            REQUIRE_THROWS_AS(results.sum(3), Results::UnsupportedColumnTypeException);
+        }
+    }
+
+    SECTION("empty") {
```
bdash: It'd be worth distinguishing between `Mode::Empty`, an empty table, and a query that matched no rows. They each take slightly different paths within `Results::aggregate`.

src/results.cpp
beeender: Good point, and the only exception is the average. I will just check table empty there.

tests/results.cpp
```diff
@@ -2027,3 +2027,212 @@ TEST_CASE("distinct") {
...
+
+        SECTION("max") {
+            Results results;
+
+            SECTIONS_RESULT_BUILT_FROM_TABLE_QUERY_TABLE_VIEW()
+
+            REQUIRE(results.max(0)->get_int() == 2);
+            REQUIRE(results.max(1)->get_float() == 2.f);
+            REQUIRE(results.max(2)->get_double() == 2.0);
+            REQUIRE(results.max(3)->get_timestamp() == Timestamp(2, 0));
```
beeender: Test fails because of core's `Table::maximum_timestamp().is_null() == true` which seems to be a bug. checking now.

beeender:
CI should pass with core `2.2.2` release.

beeender:
@bdash please help to review again. Thanks!

src/results.cpp
```diff
@@ -336,25 +336,25 @@ util::Optional<Mixed> Results::aggregate(size_t column, bool return_none_for_emp
         throw OutOfBoundsIndexException{column, m_table->get_column_count()};
 
     auto do_agg = [&](auto const& getter) -> util::Optional<Mixed> {
+        util::Optional<Mixed> results;
```
bdash: It doesn't seem useful to introduce this local when we can just return the values directly.

src/results.cpp
```diff
@@ -370,38 +370,42 @@ util::Optional<Mixed> Results::aggregate(size_t column, bool return_none_for_emp
...
 
 util::Optional<Mixed> Results::average(size_t column)
 {
-    return aggregate(column, true, "average",
-                     [=](auto const& table) { return table.average_int(column); },
-                     [=](auto const& table) { return table.average_float(column); },
-                     [=](auto const& table) { return table.average_double(column); },
-                     [=](auto const&) -> util::None { throw UnsupportedColumnTypeException{column, m_table, "average"}; });
+    return aggregate(column, "average",
+            [=](auto const& table) { return table.is_empty() ? none : Optional<Mixed>(table.average_int(column)); },
```
bdash: Why is it necessary to explicitly check for an empty table? Shouldn't `Table::average_foo` / `TableView::average_foo` be handling the empty case themselves?

beeender: core's `average_foo` returns `0` for empty cases. Which is different from what object store does before this PR. (BTW, java returns `0` for the empty case, so i would be happy to follow the core's results.)

bdash: Does that mean that core also returns 0 for an average of a column where all values are null?

beeender: Yes, core returns `0` for average when all values are null.

bdash: Oh, it returns `0` but the `value_count` output parameter will be set to zero. That's what we should be looking at to determine whether to return `none`, not whether the table is empty.

src/results.cpp
```diff
@@ -370,38 +367,42 @@ util::Optional<Mixed> Results::aggregate(size_t column, bool return_none_for_emp
...
 util::Optional<Mixed> Results::max(size_t column)
 {
-    return aggregate(column, true, "max",
-                     [=](auto const& table) { return table.maximum_int(column); },
-                     [=](auto const& table) { return table.maximum_float(column); },
-                     [=](auto const& table) { return table.maximum_double(column); },
-                     [=](auto const& table) { return table.maximum_timestamp(column); });
+    size_t return_ndx;
+    auto results = aggregate(column, "max",
+            [&](auto const& table) { return table.maximum_int(column, &return_ndx); },
```
bdash: The indentation of these lambdas is a bit wonky. The indentation that was there previously made more sense to me.

src/results.cpp
beeender: we cannot ... since OS wants return 0 for all null values, but return `none` for empty. In both cases, `value_count` is 0.

bdash: Why would the average of an empty column be different than the average a column containing all nulls? Either way the number of values involved is 0, meaning the average is undefined.

beeender: IMO, in both cases the results should be 0, which will be easier to use for common cases. But I don't have a strong opinion on that, and i agree in two cases they should behave in the same way. So return `none` for both cases (empty and all null)? It is going to change current OS behaviour, is it an acceptable change for other bindings?

bdash: If bindings care they can always map `none` to 0.

PR #289, opened 2016-12-09
Use Reachability API to cancel sync client reconnect delay

stel:
This PR adds the ability to use SystemConfiguration framework on Apple platforms to force sync client to reconnect without a delay after connection is back again.

It also adds `SyncManager::reconnect()` method to reset the reconnect delay for bindings manually.

`SyncClient::cancel_reconnect_delay` is not merged yet, but I'm happy to hear your feedback about this implementation.

/cc @bdash, @austinzheng, @jpsim, @cmelchior


src/sync/impl/apple/network_reachability_observer.cpp
```diff
@@ -0,0 +1,125 @@
...
+        return NotReachable;
+    }
+
+    NetworkReachabilityStatus status = NotReachable;
+
+    if (!(flags & kSCNetworkReachabilityFlagsConnectionRequired)) {
+        status = ReachableViaWiFi;
+    }
+
+    if ((flags & kSCNetworkReachabilityFlagsConnectionOnDemand) ||
```
bdash: The documentation states that on-demand connections will only be attempted if using CFSocketStream, and that lower-level APIs will not establish the connection. Is it really appropriate for us to treat such connections as reachable?

src/sync/impl/apple/network_reachability_observer.hpp
```diff
@@ -0,0 +1,63 @@
...
+    ReachableViaWWAN
+};
+
+class NetworkReachabilityObserver {
+public:
+    // An instanse to check if the default route is available.
+    NetworkReachabilityObserver();
+
+    // An instanse to check if the specific host is available.
+    NetworkReachabilityObserver(const std::string hostname);
```
bdash: This should take the hostname by const reference.

src/sync/impl/sync_client.hpp
```diff
@@ -68,6 +90,9 @@ struct SyncClient {
 
     const std::unique_ptr<util::Logger> m_logger;
     std::thread m_thread;
+#if REALM_PLATFORM_APPLE
+    std::unique_ptr<NetworkReachabilityObserver> m_reachability_observer;
```
bdash: Why heap allocate this?

src/sync/impl/apple/network_reachability_observer.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+enum NetworkReachabilityStatus {
+    NotReachable,
+    ReachableViaWiFi,
+    ReachableViaWWAN
+};
+
+class NetworkReachabilityObserver {
+public:
+    // An instanse to check if the default route is available.
```
bdash: `s/instanse/instance/`

Do we really want to monitor the default route? Monitoring reachability of the specific hosts in question seems more useful.

src/sync/impl/apple/network_reachability_observer.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+#ifndef REALM_OS_NETWORK_REACHABILITY_OBSERVER_HPP
+#define REALM_OS_NETWORK_REACHABILITY_OBSERVER_HPP
+
+#include <functional>
+#include <string>
+
+#include <realm/util/cf_ptr.hpp>
+
+#include <SystemConfiguration/SystemConfiguration.h>
```
bdash: Since we only really need `SCNetworkReachabilityRef` in the header we could get away with forward-declaring that type and only including the header in the .cpp file.

src/sync/impl/sync_client.hpp
```diff
@@ -42,6 +46,24 @@ struct SyncClient {
         client.run();
     }) // Throws
     {
+#if REALM_PLATFORM_APPLE
```
bdash: It might be cleaner to provide an empty, no-op implementation of the interface for non-Apple platforms so that `SyncClient` doesn't need to care which platforms support detecting changes in reachability.

src/sync/impl/apple/network_reachability_observer.cpp
```diff
@@ -0,0 +1,125 @@
...
+    if (SCNetworkReachabilityGetFlags(m_reachability_ref.get(), &flags)) {
+        return reachability_status_for_flags(flags);
+    }
+
+    return NotReachable;
+}
+
+bool NetworkReachabilityObserver::set_reachability_change_handler(std::function<void (const NetworkReachabilityStatus)> handler)
+{
+    stop_observing();
```
bdash: Why stop then restart observing?

Why not take the handler as an argument to the constructor since there doesn't appear to be any reason to change the handler for a given instance.

src/sync/impl/apple/network_reachability_observer.cpp
stel: Correct, will remove that.

src/sync/impl/apple/network_reachability_observer.hpp
stel: Monitoring of the specific host is too optimistic and in the current implementation is almost equal to the default route reachability because in case of the hostname reachability API generates the first notification for the default route and another notification after DNS query. I think having the ability to use this observer for both cases makes sense so it could be used on a client level as well as on the session level. Will make a unified constructor.

src/sync/impl/sync_client.hpp
stel: It looks like that reachability observing on ObjectStore level is only possible on Apple platforms, so there wont be any other implementations. But if you think it's reasonable I'm happy to do that.

src/sync/impl/sync_client.hpp
stel: Just for keeping everything in one `#ifdef` block in constructor.

src/sync/impl/sync_client.hpp
bdash: Presumably Android has an equivalent way of observing network reachability. Shouldn't we use it in the same way?

src/sync/impl/sync_client.hpp
bdash: I'm not sure what you mean by this?

src/sync/impl/sync_client.hpp
jpsim: According to @cmelchior it's only available as a Java API.

bdash: What's stopping us from having an implementation of `NetworkReachabilityObserver` that calls into the binding to provide the required functionality? The Java binding could provide an implementation, and .NET would provide an implementation using whatever API they have available to them.

jpsim: That'd be great.

stel: Yes, that's a good point.

src/sync/impl/apple/network_reachability_observer.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+namespace realm {
+
+enum NetworkReachabilityStatus {
+    NotReachable,
+    ReachableViaWiFi,
+    ReachableViaWWAN
+};
+
+class NetworkReachabilityObserver {
```
cmelchior: Talking to @stel and @fealebenpae yesterday about this. Both Java and .NET feels this is to complicated for what we need. Just having the `SyncManager::reconnect()` or similar would be sufficient for us since all code to detect network changes would be in the binding anyway. 

Having an observer pattern on the SyncManager is also slightly weird, since it is the SyncManager interested in Network changes, not the other way around.

stel:
I've updated the observer to use serial queue for the reachability callback, so now it should be save to create it on a background thread.

According to that Java and .Net guys are not really happy to mix any reachability-platform-specific code and events with OS I think we can keep `#if REALM_PLATFORM_APPLE` in sync client for apple platforms for now. There is `SyncManager::reconnect()` so it will be possible to use it to reset the delay from the bindings. But I'm happy to continue the discussion and implement any other approaches you prefer.

austinzheng:
Can we figure out what we want to do about this PR? Continue working on it, discard, etc?

stel:
I've merged master and enabled reconnect delay reset. Moving it to Review again, @austinzheng let me know what you think about the current status, if we can use it for initial reconnect issue or if you need any help with this.

austinzheng:
@stel Were you able to work out @cmelchior's earlier concerns? Otherwise, it looks good, let me think about it a little more.

stel:
The functionality is only available on Apple platforms and `SyncManager` has `reconnect()` method as requested, so I assume it should work for Java and others.

We discussed more generic approach for all other platforms that @cmelchior didn't like, but it's not implemented anyway.

src/sync/impl/apple/network_reachability_observer.cpp
```diff
@@ -0,0 +1,127 @@
...
+{
+    auto callback = reinterpret_cast<std::function<void ()> *>(info);
+    (*callback)();
+}
+
+} // (anonymous namespace)
+
+NetworkReachabilityObserver::NetworkReachabilityObserver(util::Optional<std::string> hostname,
+                                                         std::function<void (const NetworkReachabilityStatus)> handler)
+: m_callback_queue(dispatch_queue_create("io.realm.sync.reachability", nullptr))
```
austinzheng: Can you make the second argument explicitly `DISPATCH_QUEUE_SERIAL`, just for better readability?

src/sync/impl/apple/network_reachability_observer.cpp
```diff
@@ -0,0 +1,127 @@
...
+    return NotReachable;
+}
+
+bool NetworkReachabilityObserver::start_observing()
+{
+    m_previous_status = reachability_status();
+
+    SCNetworkReachabilityContext context = {0, &m_reachability_callback, nullptr, nullptr, nullptr};
+
+    if (!SCNetworkReachabilitySetCallback(m_reachability_ref.get(), reachability_callback, &context)) {
```
austinzheng: Isn't `context` going to cause a dangling pointer if `reachability_callback()` is called after `start_observing()` returns? Apple's API docs don't say whether or not `context` is copied or not.

stel: I was thinking about that too, but they use it the same way in Reachability sample: https://developer.apple.com/library/content/samplecode/Reachability/Listings/Reachability_Reachability_m.html, so I guess it or its values are copied.

bdash:
If Java and .NET aren't interested in doing reachability at the object store level, is there any reason to do so for Apple platforms?

austinzheng:
That's a good question. I suppose the main benefit we get is that it works for Xamarin and JS on Apple platforms out of the box?

Otherwise, I'd at least like to keep the refresh immediately API on `SyncManager`, so bindings can trigger it immediately if they decide to implement reachability themselves.

cmelchior:
That was also what I was advocating for. Just have a `refresh()` method somewhere, and let the bindings decide when and how to call it

bdash:
@cmelchior's comment at https://github.com/realm/realm-object-store/pull/289#discussion_r92355663 mentions that .NET wouldn't want to use this either, presumably because they'd implement reachability notifications in a portable way at the .NET layer.

austinzheng:
So, to move forward, let's have @stel remove the Apple reachability code, leave in the refresh immediately API, and then I can file a ticket for him to build a `RLMSyncReachabilityObserver` class at the Cocoa binding level?

stel:
Sounds good to me, but what about JS? @kristiandupont, @fealebenpae does it make sense to have this reachability stuff on Apple platforms in OS or do you prefer to implement it by yourself in bindings?

fealebenpae:
@stel for JS it would be best to have it in OS - unlike .NET, JS cannot (easily) call native APIs so implementing Reachability at the JS layer would be next to impossible.

bdash:
Presumably that means we'll need an implementation for Android, Linux, and Windows in the object store at some point. If we're going to do that, is there any reason for the Android and .NET bindings not to use that implementation?

fealebenpae:
To be honest I'm not against this happening at the Object Store level on all the platforms where possible, but doing this for Android (and to some extent Windows) in C++ would be painful - the relevant Android APIs are Java-based, and on Windows it's C++/CX. This is why I advocate adding a `reconnect()` method to start with - it will be easier to call C++ from Java or C# than the other way round, and it'd mean less development time and complexity for us to ship this. This shouldn't preclude our adding the Android, Windows, and Linux equivalents of `RLMSyncReachabilityObserver` to Object Store one by one later, though.

joshuapinter:
Hey guys, any update on getting this feature? For what it's worth, I agree that adding a `reconnect()` method and letting us handle internet connectivity ourselves (at least to start) is the way to go. 

Then perhaps after you can bake this into your various platform libraries. But for us to be able to call `reconnect()` is a huge **must** at this point. 

Thanks.

stel:
Because of Java is blocked on the `cancel_reconnect_delay()` method and the reachability stuff is still in discussion I can split this PR in 2 parts.

stel:
After the discussion in Slack to me it still seems like having this in OS is good idea because JS and .NET can also use it on Apple platforms.

@bdash please let me know if this looks good to you.

@fealebenpae please note that you'll need to link bindings agains `SystemConfiguration.framework`.

src/sync/impl/apple/network_reachability_observer.cpp
bdash: From looking at the disassembly of `SCNetworkReachabilitySetCallback`, the implementation copies `sizeof(SCNetworkReachabilityContext)` bytes from the pointer passed as the context into an ivar of the `SCNetworkReachabilityRef`, so this is safe (despite looking very strange).

src/sync/impl/apple/network_reachability_observer.cpp
```diff
@@ -0,0 +1,131 @@
...
+        zeroAddress.sa_family = AF_INET;
+
+        m_reachability_ref = util::adoptCF(SCNetworkReachabilityCreateWithAddress(nullptr, &zeroAddress));
+    }
+}
+
+NetworkReachabilityObserver::~NetworkReachabilityObserver()
+{
+    stop_observing();
+    dispatch_release(m_callback_queue);
```
bdash: I think there may be a data race here. The reachability subsystem may be in the process of invoking our callback at the same time this destructor is invoked. Nothing appears to guarantee that our `m_reachabilty_callback` member will remain valid in that scenario.

src/sync/impl/sync_client.hpp
```diff
@@ -26,6 +26,10 @@
 
 #include <thread>
 
+#if REALM_PLATFORM_APPLE && !TARGET_OS_WATCH
```
bdash: I don't like repeating the limitation that all Apple platforms other than watchOS support reachability in multiple places like this. It'd be preferable to include the header on all Apple platforms and have it expose a macro that we can use to determine whether reachability is available, and then use that macro to guard use of the reachability class.

src/sync/impl/apple/network_reachability_observer.cpp
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    if ((flags & kSCNetworkReachabilityFlagsConnectionOnTraffic) &&
+        !(flags & kSCNetworkReachabilityFlagsInterventionRequired)) {
+        status = ReachableViaWiFi;
+    }
+
+#if TARGET_OS_IPHONE
+    if (flags & kSCNetworkReachabilityFlagsIsWWAN) {
+        status = ReachableViaWWAN;
```
bdash: I'm not sure this logic is right. Above it looks like we're going out of our way to avoid treating `kSCNetworkReachabilityFlagsInterventionRequired` as indicating the connection is available, but we're not doing that here.

It seems like more appropriate logic would be to check whether a connection is required, and if it requires intervention (or won't be triggered due to our use of low-level socket APIs) then just return "not reachable". Then once we've established that a connection isn't required attempt to classify the connection into wifi vs cellular.

src/sync/impl/apple/network_reachability_observer.cpp
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    return NotReachable;
+}
+
+bool NetworkReachabilityObserver::start_observing()
+{
+    m_previous_status = reachability_status();
+
+    SCNetworkReachabilityContext context = {0, &m_reachability_callback, nullptr, nullptr, nullptr};
```
bdash: Is there any reason to indirect through a `std::function` here rather than just storing `this` and having `reachability_callback` invoke `NetworkReachabilityObserver::reachability_changed` directly?

src/sync/impl/apple/network_reachability_observer.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+#if !TARGET_OS_WATCH
+
+using namespace realm;
+
+namespace {
+
+static NetworkReachabilityStatus reachability_status_for_flags(SCNetworkReachabilityFlags flags)
+{
+    if (!(flags & kSCNetworkReachabilityFlagsReachable)) {
```
bdash: Typical style in the object store is to omit the braces around the bodies of if statements that have only a single line.

src/sync/impl/apple/network_reachability_observer.cpp
stel: The only reason for that is just to keep `NetworkReachabilityObserver::reachability_changed`  `private`.

src/sync/impl/apple/network_reachability_observer.cpp
stel: I've added `dispatch_sync(m_callback_queue, ^{});` into `stop_observing()` to make sure `m_callback_queue` is empty before returning from that method.

src/sync/impl/apple/network_reachability_observer.cpp
stel: To me it's actually not completely clear from the documentation what flags are combined and what are not. https://developer.apple.com/library/content/samplecode/Reachability confuses as well. But your suggestions seems reasonable, thanks!

stel:
Sorry for delay with that and thanks for feedback, @bdash!

I've made some fixes, please let me know how it looks to you now.

src/sync/impl/network_reachability.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+#if REALM_PLATFORM_APPLE
+
+#define NETWORK_REACHABILITY_AVAILABLE !TARGET_OS_WATCH
+
+#if NETWORK_REACHABILITY_AVAILABLE
+#include "sync/impl/apple/network_reachability_observer.hpp"
+#endif
+
+#endif // REALM_PLATFORM_APPLE
```
bdash: We should `#define NETWORK_REACHABILITY_AVAILABLE 0` for non-Apple platforms to avoid generating any warnings about testing an undefined macro when we later do `#if NETWORK_REACHABILITY_AVAILABLE`. I don't think object store itself is configured to build with that warning enabled, but it's quite likely that one of the bindings is.

stel:
Thanks @bdash!

I'll keep this opened to add runtime loading of SystemConfiguration.framework (see https://github.com/realm/realm-cocoa/pull/4794).

stel:
I've added the runtime loading of SystemConfiguration framework. Seems to works so far, tested it with RealmTasks.

But now I need a second round of review here..

/cc @bdash, @austinzheng, @jpsim

stel:
I've tested it on macOS and iOS, seems to work. watchOS doesn't have reachability support. will test on my appleTV as well this weekend.

austinzheng:
If CI passes go ahead and merge it. You can test tvOS at your leisure.

austinzheng:
It looks like the Android CI job is the one failing.

stel:
Checked it with tvOS last weekend - seems to work too.

That was some random issue on CI apparently, re-runing the job fixed that.

@fealebenpae, @nirinchev because JS and .NET bindings will also be affected by these changes, could you please let me know if it's fine by you to get it merged? To support it in bindings you'll probably need to make some changes like I did in Cocoa: https://github.com/realm/realm-cocoa/pull/4794/files.

src/sync/impl/apple/system_configuration.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+#include "sync/impl/network_reachability.hpp"
+
+#if NETWORK_REACHABILITY_AVAILABLE
+
+#include <SystemConfiguration/SystemConfiguration.h>
+
+namespace realm {
+
+class SystemConfiguration {
```
bdash: Things in the `impl` directory are typically contained within the `realm::_impl` namespace rather than directly in `realm`.

src/sync/impl/apple/system_configuration.hpp
```diff
@@ -0,0 +1,70 @@
...
+    bool network_reachability_set_dispatch_queue(SCNetworkReachabilityRef target,
+                                                 dispatch_queue_t queue);
+    bool network_reachability_set_callback(SCNetworkReachabilityRef target,
+                                           SCNetworkReachabilityCallBack callback,
+                                           SCNetworkReachabilityContext *context);
+    bool network_reachability_get_flags(SCNetworkReachabilityRef target,
+                                        SCNetworkReachabilityFlags *flags);
+
+private:
+    using network_reachability_create_with_name_t = SCNetworkReachabilityRef(*)(CFAllocatorRef, const char*);
```
bdash: I wonder if we can cheat here and use `decltype(SCNetworkReachabilityCreateWithName)` as the RHS rather than having to restate the type?

I'd be inclined to drop the `network_reachability` prefix from the private members here to reduce the verbosity.

src/sync/impl/sync_client.hpp
```diff
@@ -50,9 +56,22 @@ struct SyncClient {
...
     }) // Throws
+#if NETWORK_REACHABILITY_AVAILABLE
+    , m_reachability_observer(none, [=](const NetworkReachabilityStatus status) {
+        if (status != NotReachable) {
+            cancel_reconnect_delay();
+        }
+    })
     {
+        if (!m_reachability_observer.start_observing()) {
+            m_logger->error("Failed to setup network reachability observer");
```
bdash: "setup" -> "set up", since it's a verb rather than noun or adjective.

src/sync/impl/sync_client.hpp
```diff
@@ -50,9 +56,22 @@ struct SyncClient {
...
+        }
```
bdash: The object store coding style typically omits braces around the body of an if statement with only one line. There's a number of places in this PR where you've included them.

src/sync/impl/apple/system_configuration.cpp
```diff
@@ -0,0 +1,104 @@
...
+        m_network_reachability_create_with_address = (network_reachability_create_with_address_t)dlsym(m_framework_handle,
+                                                                                                       "SCNetworkReachabilityCreateWithAddress");
+        m_network_reachability_set_dispatch_queue = (network_reachability_set_dispatch_queue_t)dlsym(m_framework_handle,
+                                                                                                     "SCNetworkReachabilitySetDispatchQueue");
+        m_network_reachability_set_callback = (network_reachability_set_callback_t)dlsym(m_framework_handle,
+                                                                                         "SCNetworkReachabilitySetCallback");
+        m_network_reachability_get_flags = (network_reachability_get_flags_t)dlsym(m_framework_handle,
+                                                                                   "SCNetworkReachabilityGetFlags");
+    } else {
+        asl_log(nullptr, nullptr, ASL_LEVEL_WARNING, "network reachability is not available");
```
bdash: This generates a deprecation warning. It'd be preferable to avoid that.

src/sync/impl/apple/network_reachability_observer.cpp
tgoyne: The `std::function` is not needed for that. This can just be:

    auto callback = [](void* self) { static_cast<NetworkReachabilityObserver*>(self)->reachability_changed(); };
    SCNetworkReachabilityContext context = {0, this, nullptr, nullptr, nullptr};
    if (!SystemConfiguration::shared().network_reachability_set_callback(m_reachability_ref.get(), callback, &context)) {

stel:
Thanks a lot for the feedback, guys!

joshuapinter:
🎉  Wow, great work guys! Excited to start using this. 

This is the first time I've followed an issue from the core library, what is the methodology for getting this released? Expected version, release time, etc.

Thanks!

bdash:
This will be integrated into the Cocoa binding in realm/realm-cocoa#4794, and will be released shortly after that is merged.

joshuapinter:
Thanks @bdash. Out of curiosity, does that trickle down to your react-native / js library as well?

jpsim:
Yes, this project is mostly just beneficial internally to Realm where we use this repo as a shared layer for all our public SDKs: ObjC/Swift, Java, JS, .NET.

PR #290, opened 2016-12-12
Add row notification tests

beeender:


beeender:
@tgoyne I added some tests for row notifications since the behaviour seems to be different from what i expected. Would you please take a look? I marked questions with FIXME in the code. thanks!

tgoyne:
The current semantics are very specific to what KVO needs. The object-level notification work I'm doing now mostly comes down to exposing the same sort of functionality in a way less tied to KVO.

PR #294, opened 2016-12-13
Handover for distinct Results

beeender:
- Handover patch for Results.m_distinct
- Add tests

async distinct is supported by current java implementation, and for https://github.com/realm/realm-java/pull/3834 , it needs to be supported in OS level as well.

I am not quite sure if it is possible to make the fine grained notification work for the distinct results. Any suggestions? @bdash @tgoyne @JadenGeller


tgoyne:
Fine-grained notifications for distinct might be trivial, as the logic for sorted results should already handle it. I haven't spent much time thinking about it though, so there might be some non-obvious issues.

beeender:
@tgoyne Added tests for notifications. Please help to review.

beeender:
OK, will add them in a later PR.

PR #297, opened 2016-12-16
Introduce the concept of an abstract execution context

bdash:
This can be used by bindings that wish cached Realms to be keyed on something other than the thread they were created on, such as a JavaScript context or Python interpreter state.

Related to realm/realm-js#552, and will allow realm-js to easily fix realm/realm-js#473.

This should allow the thread ID changes from c34b3db85d86bf542fb57cd34f393efeac87b929 to be dropped from the `js` branch, making it possible for realm-js to use master directly.

### DIFF
--- a/src/execution_context_id.hpp
+++ b/src/execution_context_id.hpp
@@ -19,6 +19,8 @@
 #ifndef REALM_OS_EXECUTION_CONTEXT_ID_HPP
 #define REALM_OS_EXECUTION_CONTEXT_ID_HPP
 
+#include "util/aligned_union.hpp"
+
 #include <cstring>
 #include <thread>
 
@@ -93,7 +95,7 @@ private:
 
     template <typename> struct TypeForStorageType;
 
-    std::aligned_union_t<2, std::thread::id, AbstractExecutionContextID> m_storage;
+    util::AlignedUnion<1, std::thread::id, AbstractExecutionContextID>::type m_storage;
     Type m_type;
 };
 
--- a/src/execution_context_id.hpp
+++ b/src/execution_context_id.hpp
@@ -1,0 +1,112 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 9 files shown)


src/execution_context_id.hpp
```diff
@@ -0,0 +1,112 @@
...
+    };
+
+public:
+
+    // Convert from the representation used by Realm::Config, where the absence of an
+    // explicit abstract execution context indicates that the current thread's identifier
+    // should be used.
+    AnyExecutionContextID(util::Optional<AbstractExecutionContextID> maybe_abstract_id)
+    {
+        if (maybe_abstract_id)
```
jpsim: Take it or leave it.

```cpp
*this = AnyExecutionContextID(maybe_abstract_id ? *maybe_abstract_id : std::this_thread::get_id());
```

src/execution_context_id.hpp
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    bool operator==(const AnyExecutionContextID& other) const
+    {
+        return m_type == other.m_type && std::memcmp(&m_storage, &other.m_storage, sizeof(m_storage)) == 0;
+    }
+
+    bool operator!=(const AnyExecutionContextID& other) const
+    {
+        return !(*this == other);
```
jpsim: How is this different than this?

```cpp
return *this != other;
```

src/execution_context_id.hpp
bdash: That won't compile as each branch of the ternary expression is required to have the same type.

src/execution_context_id.hpp
bdash: We're implementing `!=` so we cannot use it here.

jpsim: lol

src/shared_realm.cpp
```diff
@@ -429,9 +432,12 @@ static void check_read_write(Realm *realm)
 
 void Realm::verify_thread() const
 {
-    if (m_thread_id != std::this_thread::get_id()) {
+    if (!m_execution_context.template contains<std::thread::id>())
```
tgoyne: These two shouldn't need `template` since `m_execution_context` isn't a derived type.

bdash:
@tgoyne, can you please take a look at the new changes in 31f87ab that are needed for Android?

PR #301, opened 2016-12-19
Allow GlobalNotifier to access the functions declared in Realm::Internal

bdash:
This was made on the `js` branch but had not made it over to `master`.


tgoyne:
The global notifier really shouldn't be mucking about with privates now that it isn't part of the object store, but I guess fixing that's a longer-term project.

bdash:
Yep, baby steps.

PR #302, opened 2016-12-20
reset file if a file format upgrade is required in ResetFile mode

jpsim:
As requested in realm/realm-cocoa#4470. An alternative would be to require that both `deleteRealmIfMigrationNeeded` _and_ `disableFormatUpgrade` be set to enable this behavior. I have a test in Realm Objective-C since we don't seem to have object store tests related to file format upgrades, or sample Realms in this repo.

/cc @bdash @tgoyne

### DIFF
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -210,7 +210,8 @@ void Realm::open_with_config(const Config& config,
             options.durability = config.in_memory ? SharedGroupOptions::Durability::MemOnly :
                                                     SharedGroupOptions::Durability::Full;
             options.encryption_key = config.encryption_key.data();
-            options.allow_file_format_upgrade = !config.disable_format_upgrade;
+            options.allow_file_format_upgrade = !(config.disable_format_upgrade ||
+                                                  config.schema_mode == SchemaMode::ResetFile);
             options.upgrade_callback = [&](int from_version, int to_version) {
                 if (realm) {
                     realm->upgrade_initial_version = from_version;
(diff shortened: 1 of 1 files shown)


src/shared_realm.cpp
```diff
@@ -210,7 +210,8 @@ void Realm::open_with_config(const Config& config,
             options.durability = config.in_memory ? SharedGroupOptions::Durability::MemOnly :
                                                     SharedGroupOptions::Durability::Full;
             options.encryption_key = config.encryption_key.data();
-            options.allow_file_format_upgrade = !config.disable_format_upgrade;
+            options.allow_file_format_upgrade = !(config.disable_format_upgrade ||
+                                                  config.schema_mode == SchemaMode::ResetFile);
```
tgoyne: IMO `!config.disable_format_upgrade && config.schema_mode != SchemaMode::ResetFile` is slightly more obvious, but this doesn't really matter.

src/shared_realm.cpp
```diff
@@ -221,6 +222,14 @@ void Realm::open_with_config(const Config& config,
             shared_group = std::make_unique<SharedGroup>(*history, options);
         }
     }
+    catch (realm::FileFormatUpgradeRequired const& ex) {
+        if (config.schema_mode != SchemaMode::ResetFile) {
+            translate_file_exception(config.path, config.read_only());
+            return;
```
tgoyne: This return isn't reachable.

jpsim: it's not unreachable if a user is catching exceptions...

tgoyne: C++ does not support jumping back to the point of a throw from within a catch handler.

PR #303, opened 2016-12-22
Updating sync code to reflect sync API changes

austinzheng:
Changes:
- The client-level error handler has been removed
- ~The session-level error handler takes a `std::error_code` instead of an `int`, and now takes a `bool` as well~ The session-level error callback now takes a new `SyncError` struct containing data and a few helper methods, rather than multiple arguments.
- Minor fixes to types that were renamed in sync
- Updated tests

cc. @simonask


src/sync/sync_config.hpp
```diff
@@ -43,7 +45,11 @@ using SyncBindSessionHandler = void(const std::string&,          // path on disk
                                     std::shared_ptr<SyncSession> // the session which should be bound.
                                     );
 
-using SyncSessionErrorHandler = void(std::shared_ptr<SyncSession>, int error_code, std::string message, SyncSessionError);
+using SyncSessionErrorHandler = void(std::shared_ptr<SyncSession>,
+                                     std::error_code error_code,
+                                     bool is_fatal,
```
cmelchior: This method signature is starting to get slightly confusing, especially because there seem to be some dependencies between the arguments not really expressed here, e.g. isn't the `SyncSystemError` determined by the `error_code`? Also only a subset of `error_code`s are fatal? Would it make sense to wrap of all of these except `shared_ptr` in an `ObjectServerError` class instead?

austinzheng: I think that makes sense. With the new `sync` API changes, there's less of a need for the enum.

alazier:
What's the status of this? Is it ready to go?

austinzheng:
It's ready to go, but needs to be reviewed by someone.

alazier:
Ok. I might be able to look at it a bit later. I tried using this with the JS binding and the only changes I had to make was to change the way the error handler is registered. All the regular js tests pass fine but unfortunately the GlobalNotifier tests are failing. Not sure if this is due to changes in this pr or in sync6.3, but I would guess the later.

austinzheng:
If you want, I can try backing out the sync version to the earliest thing that works (6.0 maybe?) so you can see if the GN problem is indeed related to the newest sync version. Let me know what you prefer.

src/sync/sync_config.hpp
```diff
@@ -30,20 +33,52 @@ class SyncSession;
...
+    }
+
+    bool is_access_denied_error() const
+    {
+        return enum_value() == ProtocolError::permission_denied;
+    }
+
+    ProtocolError enum_value() const
+    {
+        return static_cast<ProtocolError>(error_code.value());
```
simonask: This function can return garbage if the `error_code` does not have a category that represents things from the `ProtocolError` enum.

src/sync/sync_config.hpp
```diff
@@ -30,20 +33,52 @@ class SyncSession;
...
+    {
+        ProtocolError value = enum_value();
+        return !is_user_error() && realm::sync::is_session_level_error(value);
+    }
+
+    /// The error is specific to the user owning the session attached to this error. Errors are either
+    /// client session, or user errors.
+    bool is_user_error() const
+    {
+        return enum_value() == ProtocolError::bad_authentication;
```
simonask: It is much safer to just compare the `std::error_code` with the enum value directly, which implicitly creates a temporary `std::error_code` that is safe to compare against other error codes.

src/sync/sync_session.cpp
```diff
@@ -298,16 +298,18 @@ void SyncSession::create_sync_session()
         }
 
         using ProtocolError = realm::sync::ProtocolError;
+        SyncError error = SyncError{error_code, std::move(message), is_fatal};
 
-        SyncSessionError error_type = SyncSessionError::Debug;
         // Precondition: error_code is a valid realm::sync::Error raw value.
-        ProtocolError strong_code = static_cast<ProtocolError>(error_code);
+        ProtocolError strong_code = static_cast<ProtocolError>(error_code.value());
```
simonask: This cast is not safe. `std::error_code` may contain errors that are not of the category belonging to `ProtocolError`.

src/sync/sync_config.hpp
austinzheng: Is there any way to conditionally construct a enum value from an underlying error code value, short of manually writing my own switch statement that covers all the cases? I can't use methods like `realm::sync::is_session_level_error()` without a proper enum value, and likewise for some of the other APIs you've pointed out in your review.

austinzheng:
As far as I can tell, the `ErrorHandler` we can register with `sync` only gives us back `std::error_code`, and there's no safe way to convert a error code back to a `ProtocolError`. Can I add a static function in `sync` to do so?

bdash:
You could compare the error code's category against `realm::sync::protocol_error_category` to ensure you're dealing with an error code whose values are `ProtocolError`s.

austinzheng:
@bdash That seems like a good compromise between not writing tons of duplicate code, and making sure no undefined behavior can slip in during production.

austinzheng:
Sorry about the confusion, I wasn't aware there were both `ProtocolError`s and `Client::Error`s. The updated version of this PR should handle both correctly now, both in `SyncSession`'s error handling and in the new `SyncError` struct.

src/sync/sync_config.hpp
```diff
@@ -30,20 +35,57 @@ class SyncSession;
...
+        return !is_user_error() && realm::sync::is_session_level_error(static_cast<ProtocolError>(error_code.value()));
+    }
+
+    /// The error is specific to the user owning the session attached to this error.
+    bool is_user_error() const
+    {
+        return error_code == ProtocolError::bad_authentication;
+    }
+
+    bool is_access_denied_error() const
```
bdash: Why bother with this getter when the the user of `SyncError` can perform this comparison itself?

src/sync/sync_config.hpp
```diff
@@ -30,20 +35,57 @@ class SyncSession;
...
+        if (error_code.category() != realm::sync::protocol_error_category()) {
+            return false;
+        }
+        return !is_user_error() && realm::sync::is_session_level_error(static_cast<ProtocolError>(error_code.value()));
+    }
+
+    /// The error is specific to the user owning the session attached to this error.
+    bool is_user_error() const
+    {
+        return error_code == ProtocolError::bad_authentication;
```
bdash: Is `token_expired` not also specific to the user?

src/sync/sync_config.hpp
```diff
@@ -30,20 +35,57 @@ class SyncSession;
...
+        return !realm::sync::is_session_level_error(static_cast<ProtocolError>(error_code.value()));
+    }
+
+    /// The error is specific to a session.
+    bool is_session_error() const
+    {
+        if (error_code.category() != realm::sync::protocol_error_category()) {
+            return false;
+        }
+        return !is_user_error() && realm::sync::is_session_level_error(static_cast<ProtocolError>(error_code.value()));
```
bdash: Aren't user errors a subset of session errors? It seems like we're trying to force error codes into disjoint sets when they're really more of a hierarchy. 

What is the purpose of these categorization methods?

src/sync/sync_config.hpp
```diff
@@ -30,20 +35,57 @@ class SyncSession;
...
+        return error_code.category() == realm::sync::client_error_category();
+    }
+
+    /// The error is a protocol or connection-level error.
+    bool is_protocol_error() const
+    {
+        if (error_code.category() != realm::sync::protocol_error_category()) {
+            return false;
+        }
+        return !realm::sync::is_session_level_error(static_cast<ProtocolError>(error_code.value()));
```
austinzheng: I am a bit confused as well. The definition of `ProtocolError` (in `protocol.hpp`, l.93) has two sections, one marked with the comment `// Connection level and protocol errors`, and one marked with the comment `// Session level errors`.

Maybe a more accurate way to read the comments is that 1XX errors are "connection-level protocol errors", and 2XX errors are "session-level protocol errors"?

src/sync/sync_config.hpp
austinzheng: I think you're right; there's no good reason to break out "user errors" as a separate category.

The categorization methods are intended for the bindings to use to see what sort of error occurred.

src/sync/sync_config.hpp
austinzheng: I think, like you said, it makes more sense for the binding to figure out what errors it cares about and check them explicitly.

src/sync/sync_config.hpp
simonask: Sorry, this is also still dangerous. The `error_code` may contain errors that are neither `ClientError`s or `ProtocolErrors`s, but say a system error coming from the OS or something else entirely.

The `static_cast` can be made safe by also checking the category, and you could then decide to treat that third category of errors as connection level errors (which I guess is meaningful, considering such errors will not be based on any knowledge of the protocol).

bdash: The error category is being checked immediately above.

simonask: Ah sorry my bad, I read it as `==` instead of `!=`. :)

PR #304, opened 2016-12-24
Windows build

fealebenpae:
This fixes some errors on MSVC and replaces some gnu-y and unix-y code with more agnostic variants that are compatible with MSVC as a compiler and Windows as a platform.

Also adds CI build for windows (just running tests, no coverage yet).


fealebenpae:
@bdash, could you please check to see I haven't butchered the CMake scripts or anything else too badly? Thanks!

src/property.hpp
```diff
@@ -139,10 +140,8 @@ namespace realm {
                 return "array";
             case PropertyType::LinkingObjects:
                 return "linking objects";
-#if __GNUC__
```
tgoyne: This was deliberately only including the default: case for GCC and not Clang because without it clang can warn that the switch needs to be updated if another case is added to the enum, but GCC warns about not all cases being covered even when they are. If VC++ also warns without this change then it should just be checking for GCC or VC++.

src/schema.cpp
```diff
@@ -97,7 +97,7 @@ struct IsNotRemoveProperty {
 struct GetRemovedColumn {
     size_t operator()(SchemaChange sc) const { return sc.visit(*this); }
     size_t operator()(schema_change::RemoveProperty p) const { return p.property->table_column; }
-    template<typename T> size_t operator()(T) const { __builtin_unreachable(); }
+    template<typename T> size_t operator()(T) const { REALM_UNREACHABLE(); }
```
tgoyne: This change has an unfortunate impact on the compiled binary size since `REALM_UNREACHABLE()` is not equivalent to `__builtin_unreachable()` (it's `abort()` rather than UB) and clang's optimizer is not smart enough to detect that this is impossible to hit.

src/impl/windows/external_commit_helper.cpp
```diff
@@ -0,0 +1,87 @@
...
+            m_parent.on_change();
+            continue;
+        case WAIT_OBJECT_0 + 1: // mutex released
+            return; // exit the loop
+        case WAIT_FAILED:
+            throw std::system_error(GetLastError(), std::system_category());
+        }
+    }
+    REALM_UNREACHABLE();
+}
```
tgoyne: Newline at end of file.

src/impl/windows/external_commit_helper.cpp
```diff
@@ -0,0 +1,87 @@
...
+    std::this_thread::yield();
+    ResetEvent(m_event);
+}
+
+void ExternalCommitHelper::listen()
+{
+    std::array<HANDLE, 2> handles{ m_event, m_close_mutex };
+    while (true) {
+        DWORD wait_result = WaitForMultipleObjectsEx(handles.size(), handles.data(), false, INFINITE, false);
+        switch (wait_result)
```
tgoyne: No newline before brace for anything but functions.

src/impl/windows/external_commit_helper.cpp
```diff
@@ -0,0 +1,87 @@
...
+    m_thread.wait();
+
+    CloseHandle(m_event);
+    CloseHandle(m_close_mutex);
+}
+
+void ExternalCommitHelper::notify_others()
+{
+    SetEvent(m_event);
+    std::this_thread::yield();
```
tgoyne: This doesn't really work. There's no guarantee that all (or even any) of the waiting threads will be woken up before ResetEvent is called, and yielding can be actively harmful (if they're making a commit on the UI thread, we really don't want to yield to the background worker).

src/impl/windows/external_commit_helper.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+    return event;
+}
+
+ExternalCommitHelper::ExternalCommitHelper(RealmCoordinator& parent)
+: m_parent(parent)
+, m_event(CreateNotificationEvent(parent.get_path()))
+, m_close_mutex(CreateMutexEx(nullptr, nullptr, CREATE_MUTEX_INITIAL_OWNER, SYNCHRONIZE | MUTEX_MODIFY_STATE))
+{
+    m_thread = std::async(std::launch::async, std::bind(&ExternalCommitHelper::listen, this));
```
tgoyne: There's no need to use std::bind here.

src/impl/windows/external_commit_helper.hpp
```diff
@@ -0,0 +1,51 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#include <realm/group_shared.hpp>
+
+#include <future>
+#include <windows.h>
```
tgoyne: This should only need windef.h in the header, which avoids some of the obnoxious macros in windows.h. Even with that it needs `NOMINMAX` defined to not define those obnoxious macros.

src/impl/windows/external_commit_helper.hpp
```diff
@@ -0,0 +1,51 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2015 Realm Inc.
```
tgoyne: 2017

CMake/RealmCore.cmake
```diff
@@ -120,6 +128,13 @@ function(download_realm_tarball url target libraries)
             COMMAND ${CMAKE_COMMAND} -E make_directory ${target}
             COMMAND ${CMAKE_COMMAND} -E chdir ${target} tar xf ${tarball_path}
             COMMAND ${CMAKE_COMMAND} -E touch_nocreate ${libraries})
+    elseif(CMAKE_SYSTEM_NAME MATCHES "^Windows")
+        add_custom_command(
+            COMMENT "Extracting ${tarball_name}"
+            OUTPUT ${libraries}
+            COMMAND "${CMAKE_COMMAND}" -E make_directory "${target}"
```
bdash: If I'm reading this right, it could be combined with the Android case as the only difference is whether `tar` is invoked directly vs via a CMake helper. The CMake helper should work fine for Windows as well.

src/schema.cpp
bdash: Almost seems like `REALM_UNREACHABLE` should be renamed to `REALM_ASSERT_NOT_REACHED`?

src/property.hpp
fealebenpae: Um, I'm almost positive Clang also defines `__GNUC__` - or at least Microsoft's Clang/C2 does. I can revert this change if it only makes sense for GCC, though.

bdash: Clang does.

src/impl/windows/external_commit_helper.hpp
fealebenpae: Tried to replace this with `windef.h`, but this caused the build to fail with `#error:  "No Target Architecture"` in `winnt.h`, because things were missing that are defined in `windows.h`.

src/impl/windows/external_commit_helper.cpp
fealebenpae: Yup, I suspect this will fail with enough threads and/or processes, where "enough" is a frighteningly small number, but it was the best I could come up with, and IMO is better than having nothing at all.

But I want to thank you for the insight - I'm actually quite inexperienced when it comes to inter-process synchronization (and plain Win32, too, to be honest).

fealebenpae:
@bdash @tgoyne could you please give this a look when you've got the time? thanks!

fealebenpae:
Thanks, everyone!

PR #307, opened 2016-12-29
Avoid calling pipe2 on Android

beeender:
See https://github.com/realm/realm-java/issues/3945

* pipe2 was added to bionic on API 9, but apparently there are devices
  still don't have it even API > 10. Not sure if it is because of those
  devices don't support pipe2 system call or it just a bug with their
  bionic libs, for less risk, just use pipe + fnctl instead.
* O_CLOEXEC flag doesn't seem to be necessary for android since app
  should never call exec family functions.


src/util/android/event_loop_signal.hpp
```diff
@@ -44,11 +44,18 @@ class EventLoopSignal : public std::enable_shared_from_this<EventLoopSignal<Call
         }
 
         int message_pipe[2];
-        if (pipe2(message_pipe, O_CLOEXEC | O_NONBLOCK)) {
+        // pipe2 became part of bionic from API 9. But there are some devices with API > 10 still have problems.
```
jpsim: "that still have problems"

beeender:
retest this please

bdash:
CI: Retest this please

PR #309, opened 2016-12-29
Check that current schema is versioned before throwing MismatchedConfig

radu-tutueanu:
This fixes  https://github.com/realm/realm-object-server/issues/737

cc @sorenvind @kristiandupont


tgoyne:
:-1:

The current check is correct. Having an open realm without a schema version is not supposed to be possible.

radu-tutueanu:
Thanks, did not know that. In that case, could you please look over realm/realm-object-server#737 ? It should be easy to reproduce. If not, I will provide more info tomorrow. This issue is blocking us from using master@realm-object-store.

radu-tutueanu:
This is not the right approach, @mrackwitz is working around the issue in https://github.com/realm/realm-object-server/pull/745

PR #310, opened 2016-12-30
Add support for progress notification

austinzheng:
Changes:
- Progress notifiers can be registered and unregistered on `SyncSession`s


austinzheng:
_NOTE TO SELF_: hold on merging until a proper `sync` v.BETA-7.0 release is cut.

src/sync/sync_session.cpp
```diff
@@ -404,6 +404,48 @@ void SyncSession::create_sync_session()
...
+        m_current_uploadable = uploadable;
+        m_current_downloadable = downloadable;
+        m_current_downloaded = downloaded;
+        m_current_uploaded = uploaded;
+        // Invoke all registered notifiers.
+        std::vector<uint64_t> notifiers_to_remove;
+        for (auto& pair : m_notifiers) {
+            auto& package = pair.second;
+            if (package.is_streaming) {
+                package.notifier(downloaded, downloadable, uploaded, uploadable);
```
bdash: I talked with Austin about the fact that we're calling the notification callbacks with a lock held, and the potential for deadlocks if the callback wants to register or unregister progress notifications. We can avoid this by moving the calls to the callbacks outside the scope of the lock.

src/sync/sync_session.cpp
```diff
@@ -514,6 +556,41 @@ bool SyncSession::wait_for_upload_completion_blocking()
...
+    } else {
+        notifier(m_current_downloaded, *m_current_downloadable, m_current_uploaded, *m_current_uploadable);
+        if (!is_streaming 
+            && *m_current_downloadable <= m_current_downloaded
+            && *m_current_uploadable <= m_current_uploaded) {
+            return 0;
+        }
+    }
+    // Register the notifier now.
+    m_notifiers.insert({this_token, NotifierPackage{
```
bdash: I think you can do `emplace(this_token, NotifierPackage{…})`.

austinzheng:
@jpsim are you referring to the API discussions we had on Slack? Since this PR is in a semi-decent place I'm happy to freeze it for now until we finalize our APIs. Otherwise, it's _mostly_ the same as it was originally.

jpsim:
I withdrew my review because of the changes discussed in https://github.com/realm/product/issues/40

austinzheng:
I removed the use of `Optional<T>`, since @morten-krogh said he would change sync to call the notifier immediately after registration, meaning that we should always know the downloadable and uploadable bytes counts. (https://github.com/realm/product/issues/40#issuecomment-271144959)

src/sync/sync_session.hpp
```diff
@@ -155,7 +181,23 @@ class SyncSession : public std::enable_shared_from_this<SyncSession> {
...
 
+    struct NotifierPackage {
+        std::function<SyncProgressNotifierCallback> notifier;
+        bool is_streaming;
+        NotifierType direction;
+        uint64_t captured_transferrable;
+    };
+    // A counter used as a token for progress notifications.
+    uint64_t m_progress_notifier_token = 1;
+    // How many bytes are uploadable or downloadable, or `none` if we don't know yet.
```
bdash: The `none` case doesn't seem to be possible.

src/sync/sync_session.cpp
```diff
@@ -514,6 +562,47 @@ bool SyncSession::wait_for_upload_completion_blocking()
...
+    std::function<void()> invocation;
+    uint64_t token_value = 0;
+    {
+        std::lock_guard<std::mutex> lock(m_progress_notifier_mutex);
+        token_value = m_progress_notifier_token;
+        m_progress_notifier_token++;
+        auto current_transferrable = (direction == NotifierType::download ? m_current_downloadable : m_current_uploadable);
+        auto current_transferred = (direction == NotifierType::download ? m_current_downloaded : m_current_uploaded);
+        // Always invoke the notifier immediately, to give the user up-to-date info.
+        if (!current_transferrable) {
```
bdash: This used to be testing for the absence of information. Now it's testing for the value 0. Is that intentional?

src/sync/sync_session.cpp
```diff
@@ -514,6 +562,47 @@ bool SyncSession::wait_for_upload_completion_blocking()
...
+            invocation = [notifier=notifier](){ notifier(0, none); };
```
bdash: `notifier=notifier` is redundant. It can just be `notifier`.

src/sync/sync_session.cpp
```diff
@@ -404,6 +404,54 @@ void SyncSession::create_sync_session()
...
+                if (package.is_streaming) {
+                    invocations.emplace_back([=, notifier=package.notifier](){
+                        notifier(transferred, transferrable);
+                    });
+                } else {
+                    if (!package.captured_transferrable) {
+                        package.captured_transferrable = transferrable;
+                    }
+                    invocations.emplace_back([=, notifier=package.notifier,
+                                              transferrable=package.captured_transferrable](){
```
bdash: Just capture `package` instead of individual fields?

src/sync/sync_session.cpp
```diff
@@ -404,6 +404,54 @@ void SyncSession::create_sync_session()
...
+                    });
+                    if (package.captured_transferrable <= transferred) {
+                        // A notifier is expired if the captured max is no longer greater than the current value.
+                        m_notifiers.erase(pair.first);
+                    }
+                }
+            }
+        }
+        for (auto& invocation : invocations) {
+            // Run the notifiers only after we've released the lock.
```
bdash: I'd suggest moving this comment out of the loop since it's about the loop itself rather than the body of it.

src/sync/sync_session.cpp
```diff
@@ -404,6 +404,54 @@ void SyncSession::create_sync_session()
...
+            for (auto& pair : m_notifiers) {
+                auto& package = pair.second;
+                uint64_t transferred = (package.direction == NotifierType::download ? downloaded : uploaded);
+                uint64_t transferrable = (package.direction == NotifierType::download ? downloadable : uploadable);
+                if (package.is_streaming) {
+                    invocations.emplace_back([=, notifier=package.notifier](){
+                        notifier(transferred, transferrable);
+                    });
+                } else {
+                    if (!package.captured_transferrable) {
```
bdash: Is this test meaningful now the use of optionals for these values was removed?

austinzheng:
I fixed the review issues and refactored the "make an invocation" code into its own function.

src/sync/sync_session.cpp
```diff
@@ -283,6 +283,56 @@ SyncSession::SyncSession(SyncClient& client, std::string realm_path, SyncConfig
...
+        invocation();
+    }
+}
+
+// Create a notifier invocation.
+// Precondition: all four download/upload status member variables have been set to their most up-to-date values.
+std::function<void()> SyncSession::create_notifier_invocation(const NotifierPackage& package,
+                                                              bool* is_expired)
+{
+    REALM_ASSERT(is_expired);
```
bdash: Perhaps this should be a reference so it will be non-null by construction?

src/sync/sync_session.cpp
```diff
@@ -283,6 +283,56 @@ SyncSession::SyncSession(SyncClient& client, std::string realm_path, SyncConfig
...
+    uint64_t transferred = (package.direction == NotifierType::download ? m_current_downloaded : m_current_uploaded);
+    uint64_t transferrable = (package.is_streaming
+                              ? (package.direction == NotifierType::download
+                                 ? m_current_downloadable
+                                 : m_current_uploadable)
+                              : package.captured_transferrable);
```
bdash: These nested ternary statements are inscrutable.

src/sync/sync_session.cpp
```diff
@@ -283,6 +283,56 @@ SyncSession::SyncSession(SyncClient& client, std::string realm_path, SyncConfig
...
+    auto invocation = [=](){
```
bdash: There's no need to store this in a local that I can see. You can just `return [=](){ … };`.

src/sync/sync_session.cpp
```diff
@@ -283,6 +283,56 @@ SyncSession::SyncSession(SyncClient& client, std::string realm_path, SyncConfig
...
+        package.notifier(transferred, transferrable);
+    };
+    if (!package.is_streaming && package.captured_transferrable <= transferred) {
+        // A notifier is expired if the captured max is no longer greater than the current value.
+        *is_expired = true;
+    } else {
+        *is_expired = false;
```
bdash: I find it slightly easier to parse this by rewriting the condition slightly, to so that we're expired if we've transferred more than was transferrable. You can also just assign the result of the boolean expression directly to `is_expired`:

```c++
is_expired = !package.is_streaming && transferred > package.captured_transferrable;
```

src/sync/sync_session.cpp
```diff
@@ -404,6 +454,21 @@ void SyncSession::create_sync_session()
...
+                                                      uint_fast64_t downloadable,
+                                                      uint_fast64_t uploaded,
+                                                      uint_fast64_t uploadable) {
+        auto self = weak_self.lock();
+        if (!self) {
+            // An update was delivered after the session it relates to was destroyed.
+            // There's nothing useful we can do with it.
+            return;
+        }
+        self->handle_progress_update(downloaded, downloadable, uploaded, uploadable);
```
bdash: No need for `self->`.

src/sync/sync_session.cpp
```diff
@@ -514,6 +579,43 @@ bool SyncSession::wait_for_upload_completion_blocking()
...
+        m_progress_notifier_token++;
+        auto current_transferrable = (direction == NotifierType::download
+                                      ? m_current_downloadable
+                                      : m_current_uploadable);
+        NotifierPackage package{
+            std::move(notifier),
+            is_streaming,
+            direction,
+            std::move(current_transferrable)
+        };
```
bdash: This declaration can easily fit on one line. Not sure there's any reason to split it across 6 instead?

src/sync/sync_session.cpp
```diff
@@ -514,6 +579,43 @@ bool SyncSession::wait_for_upload_completion_blocking()
...
+                                                 NotifierType direction,
+                                                 bool is_streaming)
+{
+    std::function<void()> invocation;
+    uint64_t token_value = 0;
+    {
+        std::lock_guard<std::mutex> lock(m_progress_notifier_mutex);
+        bool skip_registration = false;
+        token_value = m_progress_notifier_token;
+        m_progress_notifier_token++;
```
bdash: `token_value = m_progress_notifier_token++;`?

src/sync/sync_session.cpp
```diff
@@ -514,6 +579,43 @@ bool SyncSession::wait_for_upload_completion_blocking()
...
 
+uint64_t SyncSession::register_progress_notifier(std::function<SyncProgressNotifierCallback> notifier,
+                                                 NotifierType direction,
+                                                 bool is_streaming)
+{
+    std::function<void()> invocation;
+    uint64_t token_value = 0;
+    {
+        std::lock_guard<std::mutex> lock(m_progress_notifier_mutex);
+        bool skip_registration = false;
```
bdash: This should be moved down to immediately prior to the call to `create_notifier_invocation`.

src/sync/sync_session.cpp
austinzheng: Yes, good idea. The argument was originally "optional" but it turned out to be used everywhere.

src/sync/sync_session.cpp
austinzheng: The perils of copy-paste based refactoring.

src/sync/sync_session.cpp
```diff
@@ -283,6 +283,51 @@ SyncSession::SyncSession(SyncClient& client, std::string realm_path, SyncConfig
...
+        transferrable = (package.direction == NotifierType::download ? m_current_downloadable : m_current_uploadable);
+    } else {
+        transferrable = package.captured_transferrable;
+    }
+    // A notifier is expired if at least as many bytes have been transferred
+    // as were originally considered transferrable.
+    is_expired = !package.is_streaming && transferred >= package.captured_transferrable;
+    return [=](){
+        package.notifier(transferred, transferrable);
+    };;
```
bdash: Extra semicolon.

austinzheng:
Thanks for the thorough reviews, @bdash. Unless more flaws come to light, I'll leave this PR alone until a non-"pre" version of `sync` is released with support for this feature, then update, squash, and merge.

bdash:
I'm seeing test failures on Linux after this was merged that appear to stem from a bug in realm-sync. I've written up realm/realm-sync#1098 to track the problem.

PR #311, opened 2017-01-03
Write to fifo could fail with EAGAIN

beeender:
It is just because of the fifo is full. Clear the fifo and retry.
See https://github.com/realm/realm-java/issues/3964


beeender:
@kneth From man page of `write(2)` I think `EAGAIN` is only concern right now. And core is only checking it for fifo as well.

PR #313, opened 2017-01-04
Add support for working with Realms from in-memory buffers

astigsen:
Added support for writing Realms to in-memory buffers and reading them from the same to be opened in read-only in-memory mode.

The purpose of this is to allow quick reading and in-memory processing of Realms from external sources like S3.


src/shared_realm.hpp
```diff
@@ -210,6 +212,7 @@ class Realm : public std::enable_shared_from_this<Realm> {
     void invalidate();
     bool compact();
     void write_copy(StringData path, BinaryData encryption_key);
+    BinaryData write_copy_to_mem();
```
bdash: There's no way for a caller to know that they are required to free the memory this returns. `OwnedBinaryData` would be better.

src/shared_realm.cpp
```diff
@@ -185,12 +185,20 @@ void Realm::open_with_config(const Config& config,
         throw std::logic_error("Realms opened in Additive-only schema mode do not use a migration function");
     if (config.schema_mode == SchemaMode::ReadOnly && config.migration_function)
         throw std::logic_error("Realms opened in read-only mode do not use a migration function");
+    if (!config.realm_data.is_null() && !config.read_only())
+        throw std::logic_error("In-memory realms from static buffers can only be opened in read-only mode");
```
bdash: "static" doesn't seem appropriate here.

astigsen:
@bdash good points. I have addressed you review comments.

src/shared_realm.cpp
```diff
@@ -185,12 +185,20 @@ void Realm::open_with_config(const Config& config,
         throw std::logic_error("Realms opened in Additive-only schema mode do not use a migration function");
     if (config.schema_mode == SchemaMode::ReadOnly && config.migration_function)
         throw std::logic_error("Realms opened in read-only mode do not use a migration function");
+    if (!config.realm_data.is_null() && !config.read_only())
```
bdash: We should attempt to detect and reject other invalid config combinations here as well. For instance, setting both `path` and `realm_data`, or `realm_data` while providing an encryption key.

src/shared_realm.hpp
```diff
@@ -134,6 +135,7 @@ class Realm : public std::enable_shared_from_this<Realm> {
         std::string path;
         // User-supplied encryption key. Must be either empty or 64 bytes.
         std::vector<char> encryption_key;
+        BinaryData realm_data;
```
bdash: This should be moved to immediately after `path` since the two are related and mutually exclusive. A comment to state this, and to explain what this field represents would also be useful.

src/shared_realm.hpp
```diff
@@ -134,6 +135,7 @@ class Realm : public std::enable_shared_from_this<Realm> {
...
 
         bool in_memory = false;
```
bdash: `in_memory` feels a little ambiguous now that there are two different types of in-memory Realm.

src/shared_realm.hpp
```diff
@@ -210,6 +212,7 @@ class Realm : public std::enable_shared_from_this<Realm> {
     void invalidate();
     bool compact();
     void write_copy(StringData path, BinaryData encryption_key);
+    OwnedBinaryData write_copy_to_mem();
```
bdash: Abbreviating "memory" is unnecessary.

"Writing" feels like weird terminology to use when talking about an in-memory operation. I don't have any better alternatives in mind, though.

astigsen: What would you think of it just being `write_copy()`? The overload should make it pretty clear what is going on.

src/shared_realm.hpp
astigsen: It would probably make sense to require `in_memory` to be set.

src/shared_realm.hpp
bdash: Seems fine.

PR #314, opened 2017-01-04
Add proper support for opening a Realm without a schema and then with a schema

tgoyne:
The GN needs to open Realms without specifying a schema, while in the same process the user needs to be able to open the same Realm, on the same thread/execution context and specify a schema. The code previously assumed that this switch would never happen, and so it would either throw an exception or outright ignore the supplied schema, depending on the sequence things were supplied and whether or not caching was enabled.

Fixes https://github.com/realm/realm-object-server/issues/737 and similar issues that anyone trying to use the GN would probably eventually hit.

### DIFF
--- a/src/impl/realm_coordinator.cpp
+++ b/src/impl/realm_coordinator.cpp
@@ -101,7 +101,9 @@ void RealmCoordinator::create_sync_session()
 
 void RealmCoordinator::set_config(const Realm::Config& config)
 {
-    if ((!m_config.read_only() && !m_notifier) || (m_config.read_only() && m_weak_realm_notifiers.empty())) {
+    bool no_existing_realm = std::all_of(begin(m_weak_realm_notifiers), end(m_weak_realm_notifiers),
+                                         [](auto& realm) { return realm.expired(); });
+    if (no_existing_realm) {
         m_config = config;
     }
     else {
@@ -116,9 +118,6 @@ void RealmCoordinator::set_config(const Realm::Config& config)
         }
         if (m_config.schema_mode != config.schema_mode) {
             throw MismatchedConfigException("Realm at path '%1' already opened with a different schema mode.", config.path);
-        }
-        if (m_config.schema_version != config.schema_version && config.schema_version != ObjectStore::NotVersioned) {
-            throw MismatchedConfigException("Realm at path '%1' already opened with different schema version.", config.path);
         }
 
 #if REALM_ENABLE_SYNC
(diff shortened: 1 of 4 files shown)


bdash:
CI is failing with:
```
SharedRealm: get_shared_realm()
373 	  should validate that the config is sensible
374 	  migration function for additive-only
375 	-------------------------------------------------------------------------------
376 	/home/jenkins.1/workspace/m_realm-object-store_PR-314-4QKMMNMG2MVOGG3NZ26CZNWD6QOGCZNUE7WQQEAVPEJ6NNQMSPRQ/tests/realm.cpp:36
377 	...............................................................................
378 	
379 	/home/jenkins.1/workspace/m_realm-object-store_PR-314-4QKMMNMG2MVOGG3NZ26CZNWD6QOGCZNUE7WQQEAVPEJ6NNQMSPRQ/tests/realm.cpp:78: FAILED:
380 	  REQUIRE_THROWS( Realm::get_shared_realm(config) )
381 	because no exception was thrown where one was expected:
382 	
```

src/impl/realm_coordinator.cpp
```diff
@@ -148,35 +147,52 @@ void RealmCoordinator::set_config(const Realm::Config& config)
 
 std::shared_ptr<Realm> RealmCoordinator::get_realm(Realm::Config config)
 {
-    std::lock_guard<std::mutex> lock(m_realm_mutex);
+    std::unique_lock<std::mutex> lock(m_realm_mutex);
 
     set_config(config);
 
+    auto schema = std::move(config.schema);
+    auto migration_function = std::move(config.migration_function);
```
bdash: This is causing the test failure on Linux with GCC as it's clearing `config.migration_function`, preventing the test for `config.schema_mode == SchemaMode::ReadOnly && config.migration_function` within `Realm::open_with_config` from ever being true.

src/impl/realm_coordinator.cpp
```diff
@@ -101,7 +101,9 @@ void RealmCoordinator::create_sync_session()
 
 void RealmCoordinator::set_config(const Realm::Config& config)
 {
-    if ((!m_config.read_only() && !m_notifier) || (m_config.read_only() && m_weak_realm_notifiers.empty())) {
+    bool no_existing_realm = std::all_of(begin(m_weak_realm_notifiers), end(m_weak_realm_notifiers),
+                                         [](auto& realm) { return realm.expired(); });
```
bdash: `realm` seems like a bit of a misnomer. `weak_notifier`, perhaps?

src/impl/realm_coordinator.cpp
```diff
@@ -148,35 +147,52 @@ void RealmCoordinator::set_config(const Realm::Config& config)
...
+        }
+    }
+    if (!realm) {
+        realm = Realm::make_shared_realm(std::move(config), shared_from_this());
+        if (!config.read_only() && !m_notifier && config.automatic_change_notifications) {
+            try {
+                m_notifier = std::make_unique<ExternalCommitHelper>(*this);
+            }
+            catch (std::system_error const& ex) {
+                lock.unlock();
```
bdash: I assume this is necessary to ensure that the lock is dropped prior to `realm` being destroyed as part of stack unwinding?

sorenvind:
Does this also support the inverse case (opening with schema and then without), or is that not supported / possible?

src/impl/realm_coordinator.cpp
tgoyne: Yes.

src/impl/realm_coordinator.cpp
bdash: This was addressed by 204467252841de364d6ef8fc6ca36b773cba2a88, as the the migration function is now verified prior to this line.

bdash:
👍

PR #316, opened 2017-01-06
Add before_notify to binding_context

beeender:
This is needed by java binding to integrate the Results and keep the
current realm-java collection iterator behaviour:
When write transaction begins, all collections on that thread have to be
frozen (stop calling sync_if_needed) until next event loop.
Since there is no way to inject code to be ran at the end of current
loop, realm-java needs to refresh the Results at the very beginning of
the next event loop as much as possible.

java could use binding_context::changes_available for transaction
committed by other threads, but there is no place to refresh all Results
if the transaction has been committed locally and async callback called
in the process_available_async. CollectionChangeCallback::after cannot
be used in this case since we have to ensure all the collections are
refreshed before when it will be used (eg.: in results1's async after
callback, the results2 has to be refreshed already.)

More details about java stable iterator can be seen:
https://github.com/realm/realm-java/issues/3883#issuecomment-265659475

### DIFF
--- a/src/binding_context.hpp
+++ b/src/binding_context.hpp
@@ -74,8 +74,9 @@ public:
     // actually be called?
     virtual bool can_deliver_notifications() const noexcept { return true; }
 
-    // Called by the Realm when a notification arrives which is triggered through
-    // write transaction committed by itself or a different Realm instance.
+    // Called by the Realm when refresh called or a notification arrives which
+    // is triggered through write transaction committed by itself or a different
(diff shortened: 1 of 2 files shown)


beeender:
Or maybe it's better to replace `BindingContext::changes_available()` with `BindingContext::before_notify(bool changes_available)`?

beeender:
@tgoyne Thanks for the advise! Added the `before_notify` in the `refresh()`.

PR #318, opened 2017-01-06
Handle realm close in did_change callbacks

beeender:
Bindings could close Realm in the notify callbacks.

### DIFF
--- a/src/impl/realm_coordinator.cpp
+++ b/src/impl/realm_coordinator.cpp
@@ -324,9 +324,6 @@ void RealmCoordinator::commit_write(Realm& realm)
     REALM_ASSERT(!m_config.read_only());
     REALM_ASSERT(realm.is_in_transaction());
 
-    // In case the SharedRealm gets descturcted in the did_change callback.
-    auto realm_ref = realm.shared_from_this();
-
     {
         // Need to acquire this lock before committing or another process could
         // perform a write and notify us before we get the chance to set the
@@ -343,6 +340,14 @@ void RealmCoordinator::commit_write(Realm& realm)
         }
     }
 
+#if REALM_ENABLE_SYNC
+    // Realm could be closed in did_change. So send sync notification first before did_change.
+    if (m_sync_session) {
+        auto& sg = Realm::Internal::get_shared_group(realm);
+        auto version = LangBindHelper::get_version_of_latest_snapshot(*sg);
+        SyncSession::Internal::nonsync_transact_notify(*m_sync_session, version);
+    }
+#endif
     if (realm.m_binding_context) {
         realm.m_binding_context->did_change({}, {});
     }
@@ -350,13 +355,6 @@ void RealmCoordinator::commit_write(Realm& realm)
     if (m_notifier) {
         m_notifier->notify_others();
     }
-#if REALM_ENABLE_SYNC
-    if (m_sync_session && !realm.is_closed()) {
-        auto& sg = Realm::Internal::get_shared_group(realm);
-        auto version = LangBindHelper::get_version_of_latest_snapshot(*sg);
-        SyncSession::Internal::nonsync_transact_notify(*m_sync_session, version);
-    }
-#endif
 }
 
 void RealmCoordinator::pin_version(VersionID versionid)
--- a/src/impl/realm_coordinator.cpp
+++ b/src/impl/realm_coordinator.cpp
@@ -330,15 +330,18 @@ void RealmCoordinator::commit_write(Realm& realm)
         // skip version
         std::lock_guard<std::mutex> l(m_notifier_mutex);
 
-        transaction::commit(Realm::Internal::get_shared_group(realm));
+        transaction::commit(*Realm::Internal::get_shared_group(realm));
 
         // Don't need to check m_new_notifiers because those don't skip versions
         bool have_notifiers = std::any_of(m_notifiers.begin(), m_notifiers.end(),
                                           [&](auto&& notifier) { return notifier->is_for_realm(realm); });
         if (have_notifiers) {
-            m_notifier_skip_version = Realm::Internal::get_shared_group(realm).get_version_of_current_transaction();
+            m_notifier_skip_version = Realm::Internal::get_shared_group(realm)->get_version_of_current_transaction();
         }
     }
+
+    // In case the SharedRealm gets descturcted in the did_change callback.
+    auto realm_ref = realm.shared_from_this();
 
     if (realm.m_binding_context) {
         realm.m_binding_context->did_change({}, {});
@@ -348,9 +351,9 @@ void RealmCoordinator::commit_write(Realm& realm)
         m_notifier->notify_others();
     }
 #if REALM_ENABLE_SYNC
-    if (m_sync_session) {
+    if (m_sync_session && !realm.is_closed()) {
         auto& sg = Realm::Internal::get_shared_group(realm);
-        auto version = LangBindHelper::get_version_of_latest_snapshot(sg);
+        auto version = LangBindHelper::get_version_of_latest_snapshot(*sg);
         SyncSession::Internal::nonsync_transact_notify(*m_sync_session, version);
     }
 #endif
@@ -707,7 +710,7 @@ void RealmCoordinator::advance_to_ready(Realm& realm)
     auto& sg = Realm::Internal::get_shared_group(realm);
     if (notifiers) {
         auto version = notifiers.version();
-        if (version && *version <= sg.get_version_of_current_transaction())
+        if (version && *version <= sg->get_version_of_current_transaction())
             return;
     }
 
@@ -736,11 +739,11 @@ bool RealmCoordinator::advance_to_latest(Realm& realm)
     std::unique_lock<std::mutex> lock(m_notifier_mutex);
     _impl::NotifierPackage notifiers(m_async_error, notifiers_for_realm(realm), this);
     lock.unlock();
-    notifiers.package_and_wait(sgf::get_version_of_latest_snapshot(sg));
+    notifiers.package_and_wait(sgf::get_version_of_latest_snapshot(*sg));
 
-    auto version = sg.get_version_of_current_transaction();
+    auto version = sg->get_version_of_current_transaction();
     transaction::advance(sg, realm.m_binding_context.get(), notifiers);
-    return version != sg.get_version_of_current_transaction();
+    return version != sg->get_version_of_current_transaction();
 }
 
 void RealmCoordinator::promote_to_write(Realm& realm)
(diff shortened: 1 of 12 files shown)


tests/realm.cpp
```diff
@@ -383,3 +383,95 @@ TEST_CASE("ShareRealm: in-memory mode from buffer") {
...
+
+            virtual void did_change(std::vector<ObserverState> const&, std::vector<void*> const&, bool) override
+            {
+                REQUIRE(false);
+            }
+        };
+        r1->m_binding_context.reset(new Context(r1));
+    }
+
+    SECTION("will_change") {
```
beeender: @tgoyne Are `will_change` and `did_change` still needed after your object notification work? Shall I remove the `will_change` test from this PR? It actually requires more modifications. And I don't think Realm instance should be closed in the `will_change` callback.

src/shared_realm.cpp
```diff
@@ -575,17 +575,23 @@ void Realm::notify()
     m_is_sending_notifications = true;
     auto cleanup = util::make_scope_exit([this]() noexcept { m_is_sending_notifications = false; });
 
+    // The SharedRealm could be closed in the callbacks. Keep a ref locally to check if Realm gets closed.
+    auto this_ref = shared_from_this();
     if (m_shared_group->has_changed()) { // Throws
         if (m_binding_context) {
             m_binding_context->changes_available();
+            if (this_ref->is_closed()) return;
```
tgoyne: Newline before `return` for all of these.

tests/realm.cpp
tgoyne: Yes, will_change() and did_change() are still needed and their API and semantics are unchanged. Not handling the Realm being closed from within them is perfectly acceptable.

tests/realm.cpp
```diff
@@ -383,3 +383,95 @@ TEST_CASE("ShareRealm: in-memory mode from buffer") {
...
+    SECTION("changes_available") {
+        struct Context : public TestContext {
+            Context(SharedRealm& r) : TestContext(r) { }
+            virtual void changes_available() override
+            {
+                shared_realm->close();
+                shared_realm.reset();
+            }
+
+            virtual void will_change(std::vector<ObserverState> const&, std::vector<void*> const&) override
```
tgoyne: `virtual` is redundant when the function is marked `override`.

src/shared_realm.cpp
tgoyne: Calling these via `this_ref->` suggests that it's somehow different from `this`, when it's actually fine to just call `is_closed()` directly.

beeender:
@tgoyne java is using `did_change()` for Realm notifications, so it would be quite important to handle the close case with all `did_change()` called. The problem is this requires to change 
```cpp
    TransactLogObserver(BindingContext* context, SharedGroup& sg, Func&& func,
                        _impl::NotifierPackage& notifiers, bool validate=true)
```
to take a `SharedRealm` instead of `SharedGroup`. Otherwise the sg might become an invalid reference when the Realm gets closed. Is it an acceptable change or there are some other ideas to handle this case?

tgoyne:
The transaction log functions are used in places where the SharedGroup is not owned by a Realm. An awkward option would be to instead pass in a const ref to the `unique_ptr<SharedGroup>` and check that that's non-null all over the place.

beeender:
@tgoyne PR updated
1. Java only exposes `did_change()` and collection callback to end users. So to make this PR simpler, it only handles the `did_change()`.
2. Will it be a problem if user holding multiple collection notification tokens and close the Realm in the first collection change callback?
3. The test case `table change information / modifications to untracked tables are ignored` fails, but I cannot figure out why. It doesn't seem that any changes could cause that? Would you please kindly help to take a look? thanks.

src/impl/transact_log_handler.cpp
```diff
@@ -657,20 +657,23 @@ void advance_with_notifications(BindingContext* context, SharedGroup& sg, Func&&
     if (observers.empty() && (!notifiers || notifiers.version())) {
         notifiers.before_advance();
         func(TransactLogValidator());
-        auto new_version = sg.get_version_of_current_transaction();
+        auto new_version = sg->get_version_of_current_transaction();
         if (context && old_version != new_version)
             context->did_change({}, {});
+        // did_change() could close the Realm. Just return if it does.
+        if (!sg)
```
beeender: Is it fine to return here?

src/shared_realm.cpp
```diff
@@ -607,10 +607,14 @@ void Realm::notify()
             m_coordinator->advance_to_ready(*this);
         }
         else  {
+            // In case the SharedRealm gets descturcted in the did_change callback.
+            auto realm_ref = this->shared_from_this();
```
tgoyne: This doesn't actually work, because `cleanup`'s destructor will run after this reference is released and that needs `this` to still exist. The immediate fix would be to move this to the top of the function, but I'm not sure why the `Realm` instance would be destroyed from within the callback. There's no real need for the Java `Realm` to release its reference to the OS `Realm` after calling `close()` since it's an insignificant amount of memory freed up by destroying it early.

Also "descturcted", and the SharedGroup isn't what matters here.

tests/realm.cpp
```diff
@@ -471,3 +474,67 @@ TEST_CASE("ShareRealm: in-memory mode from buffer") {
...
+        r1->notify();
+    }
+
+    SECTION("did_change with async results") {
+        r1->m_binding_context.reset(new Context());
+        Results results(r1, table->where());
+        auto token = results.add_notification_callback([&](CollectionChangeSet, std::exception_ptr) {
+                // Should not be called.
+                REQUIRE(false);
+                });
```
tgoyne: Indentation is wrong.

tests/realm.cpp
```diff
@@ -471,3 +474,67 @@ TEST_CASE("ShareRealm: in-memory mode from buffer") {
...
+    r1->begin_transaction();
+    auto table = r1->read_group().get_table("class_object");
+    auto row_idx = table->add_empty_row(1);
+    auto table_idx = table->get_index_in_group();
+    r1->commit_transaction();
+
+    // Cannot be a member var of Context since Realm.close will free the context.
+    static SharedRealm* shared_realm;
+    shared_realm = &r1;
+    struct Context : public BindingContext {
```
tgoyne: Context can just hold a pointer to a stack variable rather than making it a static.

tests/realm.cpp
```diff
@@ -471,3 +474,67 @@ TEST_CASE("ShareRealm: in-memory mode from buffer") {
...
+        void did_change(std::vector<ObserverState> const&, std::vector<void*> const&, bool) override
+        {
+            (*shared_realm)->close();
+            (*shared_realm).reset();
+        }
+    };
+
+    SECTION("did_change") {
+        r1->m_binding_context.reset(new Context());
+        r1->invalidate();
```
tgoyne: This doesn't seem relevant to the test.

tests/transaction_log_parsing.cpp
```diff
@@ -1325,8 +1325,8 @@ TEST_CASE("Transaction log parsing: changeset calcuation") {
 
         auto observe = [&](std::initializer_list<Row> rows, auto&& fn) {
             auto history = make_in_realm_history(config.path);
-            SharedGroup sg(*history, config.options());
-            auto& group = sg.begin_read();
+            std::unique_ptr<SharedGroup> sg = std::make_unique<SharedGroup>(*history, config.options());
```
tgoyne: This is a good place to use `auto` as otherwise the type of the variable is specified twice.

src/impl/transact_log_handler.cpp
tgoyne: Yes, destroying the `NotifierPackage` without delivering it should be fine and not delivering the notifications seems like the only sensible thing to do.

src/impl/realm_coordinator.cpp
```diff
@@ -330,16 +330,19 @@ void RealmCoordinator::commit_write(Realm& realm)
...
         bool have_notifiers = std::any_of(m_notifiers.begin(), m_notifiers.end(),
                                           [&](auto&& notifier) { return notifier->is_for_realm(realm); });
         if (have_notifiers) {
-            m_notifier_skip_version = Realm::Internal::get_shared_group(realm).get_version_of_current_transaction();
+            m_notifier_skip_version = Realm::Internal::get_shared_group(realm)->get_version_of_current_transaction();
         }
     }
 
+    // In case the SharedRealm gets descturcted in the did_change callback.
+    auto realm_ref = realm.shared_from_this();
```
tgoyne: My comment about this in shared_realm.cpp applies to here as well (`realm` can theoretically be the thing keeping the coordinator alive).

tests/realm.cpp
beeender: yes, the problem is when close the Realm, the context object will be freed. the `Context.m_shared_realm_ptr` itself become invalid.

src/shared_realm.cpp
beeender: hmmm, that is a good point. Let me check if it is doable to only call close in java `Realm.close()` and delete the `SharedRealm` when it gets GCed.

beeender: I think there is no way to avoid java binding free the pointer in the callback because of the GC is not predictable. Like below example:

```java
// Member var
Realm realm;
void UserDefinedCallback() {
    realm.close(); // Here is totally doable that java keeps the SharedRealm alive.
    realm = null; // But we cannot really control what will happen here. The realm instance could be GCed at any time, which means the pointer it holds could be freed at any time from a different thread.
}
```

I use the immediate fix to move the `shared_from_this()` to the top of the method body.

tests/realm.cpp
beeender: To make the `m_group` null so it can run with the below `else` routine in `notify()`:

```c++
    if (m_auto_refresh) {
        if (m_group) {
            m_coordinator->advance_to_ready(*this);
        }
        else  {
            if (m_binding_context) {
                m_binding_context->did_change({}, {});
            }
            if (!is_closed()) {
                m_coordinator->process_available_async(*this);
            }
        }
    }

```

beeender:
@tgoyne updated, please help to review again. thanks

src/impl/realm_coordinator.cpp
```diff
@@ -348,9 +351,9 @@ void RealmCoordinator::commit_write(Realm& realm)
         m_notifier->notify_others();
     }
 #if REALM_ENABLE_SYNC
-    if (m_sync_session) {
+    if (m_sync_session && !realm.is_closed()) {
```
tgoyne: Shouldn't this be moved to before the notifications? I wouldn't expect calling `close()` on a Realm to stop it from syncing the commit I just made.

beeender: Yup, you are right . Moved. @austinzheng would you please take a look of my last commit?

PR #320, opened 2017-01-06
Allow sync file manager to act on "client reset" file actions

austinzheng:
Changes:
- When receiving a 'diverging histories' error, a client reset file action is generated
- `SyncManager` now acts on file reset actions
- `SyncError` now has a user info map
- Added more tests

Note that I had a review for this branch but I accidentally completely destroyed it when I erased the `az/client-reset` branch. @jpsim

### DIFF
--- a/src/sync/impl/sync_file.cpp
+++ b/src/sync/impl/sync_file.cpp
@@ -21,6 +21,19 @@
 #include <realm/util/file.hpp>
 #include <realm/util/scope_exit.hpp>
 
+#include <iomanip>
+#include <sstream>
+
+#if WIN32
+#include <io.h>
+#include <fcntl.h>
+
+inline static int mkstemp(char* _template) { return _open(_mktemp(_template), _O_CREAT | _O_TEMPORARY, _S_IREAD | _S_IWRITE); }
+#else
+#include <unistd.h>
+#endif
+
+
 using File = realm::util::File;
 
 namespace realm {
@@ -180,20 +193,47 @@ std::string file_path_by_appending_extension(const std::string& path, const std:
     return buffer;
 }
 
+std::string create_timestamped_template(const std::string& prefix, int wildcard_count)
+{
+    constexpr int WILDCARD_MAX = 20;
+    constexpr int WILDCARD_MIN = 6;
+    wildcard_count = std::min(WILDCARD_MAX, std::max(WILDCARD_MIN, wildcard_count));
+    std::time_t time = std::time(nullptr);
+    std::stringstream stream;
+    stream << prefix << "-" << std::put_time(std::localtime(&time), "%Y%m%d-%H%M%S") << "-" << std::string(wildcard_count, 'X');
+    return stream.str();
+}
+
+std::string reserve_unique_file_name(const std::string& path, const std::string& template_string)
+{
+    REALM_ASSERT_DEBUG(template_string.find("XXXXXX") != std::string::npos);
+    std::string path_buffer = file_path_by_appending_component(path, template_string, FilePathType::File);
+    int fd = mkstemp(&path_buffer[0]);
+    if (fd < 0) {
+        int err = errno;
+        throw std::system_error(err, std::system_category());
+    }
+    // Remove the file so we can use the name for our own file.
+    close(fd);
+    unlink(path_buffer.c_str());
+    return path_buffer;
+}
+
 } // util
 
 constexpr const char SyncFileManager::c_sync_directory[];
 constexpr const char SyncFileManager::c_utility_directory[];
+constexpr const char SyncFileManager::c_recovery_directory[];
 constexpr const char SyncFileManager::c_metadata_directory[];
 constexpr const char SyncFileManager::c_metadata_realm[];
 
-std::string SyncFileManager::get_utility_directory() const
+std::string SyncFileManager::get_special_directory(std::string directory_name) const
 {
-    auto util_path = file_path_by_appending_component(get_base_sync_directory(),
-                                                      c_utility_directory,
-                                                      util::FilePathType::Directory);
-    util::try_make_dir(util_path);
-    return util_path;
+    auto dir_path = file_path_by_appending_component(get_base_sync_directory(),
+                                                     directory_name,
+                                                     util::FilePathType::Directory);
+    util::try_make_dir(dir_path);
+    return dir_path;
 }
 
 std::string SyncFileManager::get_base_sync_directory() const
@@ -230,23 +270,17 @@ void SyncFileManager::remove_user_directory(const std::string& user_identity) co
     util::remove_nonempty_dir(user_path);
 }
 
-bool SyncFileManager::remove_realm(const std::string& user_identity, const std::string& raw_realm_path) const
+bool SyncFileManager::remove_realm(const std::string& absolute_path) const
 {
-    REALM_ASSERT(user_identity.length() > 0);
-    REALM_ASSERT(raw_realm_path.length() > 0);
-    if (filename_is_reserved(user_identity) || filename_is_reserved(raw_realm_path)) {
-        throw std::invalid_argument("A user or Realm can't have an identifier reserved by the filesystem.");
-    }
-    auto escaped = util::make_percent_encoded_string(raw_realm_path);
-    auto realm_path = util::file_path_by_appending_component(user_directory(user_identity), escaped);
+    REALM_ASSERT(absolute_path.length() > 0);
     bool success = true;
     // Remove the base Realm file (e.g. "example.realm").
-    success = File::try_remove(realm_path);
+    success = File::try_remove(absolute_path);
     // Remove the lock file (e.g. "example.realm.lock").
-    auto lock_path = util::file_path_by_appending_extension(realm_path, "lock");
+    auto lock_path = util::file_path_by_appending_extension(absolute_path, "lock");
     success = File::try_remove(lock_path);
     // Remove the management directory (e.g. "example.realm.management").
-    auto management_path = util::file_path_by_appending_extension(realm_path, "management");
+    auto management_path = util::file_path_by_appending_extension(absolute_path, "management");
     try {
         util::remove_nonempty_dir(management_path);
     }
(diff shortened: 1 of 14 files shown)


src/sync/sync_session.cpp
```diff
@@ -347,11 +361,31 @@ void SyncSession::create_sync_session()
...
+                case ProtocolError::bad_client_version:
+                    break;
                 case ProtocolError::bad_server_file_ident:
                 case ProtocolError::bad_client_file_ident:
                 case ProtocolError::bad_server_version:
-                case ProtocolError::bad_client_version:
-                case ProtocolError::diverging_histories:
+                case ProtocolError::diverging_histories: {
+                    // Add a SyncFileActionMetadata marking the Realm as needing to be deleted.
+                    auto full_name = new_path + to_string(std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now().time_since_epoch()).count());
```
austinzheng: It would be better to put a more human-readable date here, but I don't think `std::chrono` has any calendar APIs.

austinzheng:
I removed the use of `Optional<T>`, since @morten-krogh said he would change sync to call the notifier immediately after registration, meaning that we should always know the downloadable and uploadable bytes counts. (https://github.com/realm/product/issues/40#issuecomment-271144959)

src/sync/sync_session.cpp
bdash: You could use `strftime` or `std::put_time`.

src/sync/sync_session.cpp
```diff
@@ -288,9 +295,16 @@ void SyncSession::create_sync_session()
...
 
+    // Constants for client reset error case.
+    const std::string user_identity = m_config.user->identity();
+    const std::string new_path = util::make_percent_encoded_string(user_identity + "_" + m_config.realm_url + "_");
+
     // Set up the wrapped handler
     std::weak_ptr<SyncSession> weak_self = shared_from_this();
-    auto wrapped_handler = [this, weak_self](std::error_code error_code, bool is_fatal, std::string message) {
+    auto wrapped_handler = [this, weak_self,
+                            realm_url=m_config.realm_url,
```
bdash: Why not get this from `m_config.realm_url` within the callback?

src/sync/sync_session.cpp
```diff
@@ -347,11 +361,31 @@ void SyncSession::create_sync_session()
...
+                    break;
                 case ProtocolError::bad_server_file_ident:
                 case ProtocolError::bad_client_file_ident:
                 case ProtocolError::bad_server_version:
-                case ProtocolError::bad_client_version:
-                case ProtocolError::diverging_histories:
+                case ProtocolError::diverging_histories: {
+                    // Add a SyncFileActionMetadata marking the Realm as needing to be deleted.
+                    auto full_name = new_path + to_string(std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now().time_since_epoch()).count());
+                    error.user_info.insert({SyncError::c_original_file_path_key, path()});
```
bdash: You could write this as `error.user_info.emplace(SyncError::c_original_file_path_key, path());` or `error.user_info[SyncError::c_original_file_path_key] = path();`.

src/sync/sync_session.cpp
```diff
@@ -347,11 +361,31 @@ void SyncSession::create_sync_session()
...
+                    error.user_info.insert({SyncError::c_recovery_file_path_key, full_name});
+                    SyncManager::shared().perform_metadata_update([this,
+                                                                   old_path=this->path(),
```
bdash: Is `this->` necessary?

src/sync/sync_session.cpp
```diff
@@ -347,11 +361,31 @@ void SyncSession::create_sync_session()
...
+                                                                   user_identity=std::move(user_identity),
+                                                                   realm_url=std::move(realm_url),
+                                                                   full_name=std::move(full_name)](const SyncMetadataManager& manager) {
+                        SyncFileActionMetadata(manager,
```
bdash: This is all drifting so far to the right that it might be a good idea to move the body of the outermost lambda into a helper function.

src/sync/sync_manager.cpp
```diff
@@ -130,6 +154,34 @@ void SyncManager::configure_file_system(const std::string& base_file_path,
     }
 }
 
+bool SyncManager::immediately_run_file_actions(const std::string& realm_path)
+{
+    SyncFileActionMetadataResults file_actions = m_metadata_manager->all_pending_actions();
+    for (size_t i = 0; i < file_actions.size(); i++) {
+        SyncFileActionMetadata file_action = file_actions.get(i);
+        if (file_action.original_name() != realm_path) {
```
bdash: Rather than retrieving all actions and looping to find one with a matching path, why not just look the action up directly by path?

src/sync/sync_session.cpp
austinzheng: The syntax you suggested is a huge improvement over the ugly pair-based `insert()`.

src/sync/impl/sync_metadata.cpp
```diff
@@ -264,6 +264,19 @@ void SyncUserMetadata::remove()
     m_realm = nullptr;
 }
 
+std::unique_ptr<SyncFileActionMetadata> SyncFileActionMetadata::metadata_for_path(const std::string& original_name, const SyncMetadataManager& manager)
+{
+    auto realm = Realm::get_shared_realm(manager.get_configuration());
+    auto schema = manager.m_file_action_schema;
+    TableRef table = ObjectStore::table_for_object_type(realm->read_group(), c_sync_fileActionMetadata);
+    realm->begin_transaction();
+    size_t row_idx = table->find_first_string(schema.idx_original_name, original_name);\
```
bdash: Extraneous `\`.

src/sync/sync_manager.cpp
```diff
@@ -130,6 +154,34 @@ void SyncManager::configure_file_system(const std::string& base_file_path,
     }
 }
 
+bool SyncManager::immediately_run_file_actions(const std::string& realm_path)
+{
+    if (!m_metadata_manager) {
+        return false;
+    }
+    auto metadata = SyncFileActionMetadata::metadata_for_path(realm_path, *m_metadata_manager);
+    if (metadata) {
```
bdash: An early-return would probably be better here.

src/sync/sync_session.cpp
```diff
@@ -19,16 +19,31 @@
...
 
 #include <realm/sync/client.hpp>
 #include <realm/sync/protocol.hpp>
 
+#include <chrono>
+
+#if WIN32
+#include <io.h>
+#include <fcntl.h>
+inline static char * mktemp(char* _template) { static_assert("Please implement me."); }
```
bdash: This needs an implementation to avoid breaking the Windows build.

src/sync/impl/sync_metadata.cpp
```diff
@@ -264,6 +264,19 @@ void SyncUserMetadata::remove()
     m_realm = nullptr;
 }
 
+util::Optional<SyncFileActionMetadata> SyncFileActionMetadata::metadata_for_path(const std::string& original_name, const SyncMetadataManager& manager)
+{
+    auto realm = Realm::get_shared_realm(manager.get_configuration());
+    auto schema = manager.m_file_action_schema;
+    TableRef table = ObjectStore::table_for_object_type(realm->read_group(), c_sync_fileActionMetadata);
+    realm->begin_transaction();
```
kneth: Why a write transaction?

src/sync/sync_session.cpp
```diff
@@ -283,12 +299,147 @@ SyncSession::SyncSession(SyncClient& client, std::string realm_path, SyncConfig
...
 
+std::string SyncSession::get_recovery_file_path()
+{
+    std::time_t now = std::time(nullptr);
+    std::stringstream stream;
+    stream << SyncManager::shared().recovery_directory_path();
+    stream << "recovered_realm-" << std::put_time(std::localtime(&now), "%Y%m%d-%H%M%S") << "-XXXXXXXX";
+    std::string path = stream.str();
+    // Now try to create the temporary file.
+    int fd = mkstemp(&path[0]);
```
kneth: Core has `make_temp_dir()` which might be useful. `mkstemp` might be better than core's `mkdtemp`.

tests/sync/session.cpp
```diff
@@ -270,4 +267,87 @@ TEST_CASE("sync: log-in", "[sync]") {
...
 
     // TODO: write a test that logs out a Realm with multiple sessions, then logs it back in?
+    // TODO: write tests that check that a Session properly handles various types of errors reported via its callback.
+}
+
+TEST_CASE("sync: error handling", "[sync]") {
+    using ProtocolError = realm::sync::ProtocolError;
+    auto cleanup = util::make_scope_exit([=]() noexcept { SyncManager::shared().reset_for_testing(); });
+    SyncServer server;
+    SyncManager::shared().configure_file_system("/tmp/", SyncManager::MetadataMode::NoMetadata);
```
kneth: `/tmp` is not available on Android (`/data/local/tmp` is).

src/sync/impl/sync_metadata.cpp
austinzheng: Oversight on my part. Thanks for pointing it out.

src/sync/impl/sync_file.cpp
```diff
@@ -180,20 +193,54 @@ std::string file_path_by_appending_extension(const std::string& path, const std:
...
+    constexpr int WILDCARD_MAX = 20;
+    constexpr int WILDCARD_MIN = 6;
+    // Clamp
+    wildcard_count = (wildcard_count > WILDCARD_MAX ? WILDCARD_MAX : wildcard_count);
+    wildcard_count = (wildcard_count < WILDCARD_MIN ? WILDCARD_MIN : wildcard_count);
+    std::time_t actual_time = time.value_or(std::time(nullptr));
+    std::stringstream stream;
+    stream << prefix << "-" << std::put_time(std::localtime(&actual_time), "%Y%m%d-%H%M%S") << "-";
+    for (int i = 0; i < wildcard_count; i++)
+        stream << "X";
```
bdash: You could use `std::string(size_t count, char ch)` constructor that constructs a string with `count` copies of `ch`.

src/sync/impl/sync_file.hpp
```diff
@@ -45,6 +48,16 @@ std::string file_path_by_appending_component(const std::string& path,
 /// Given a file path and an extension, append the extension to the path.
 std::string file_path_by_appending_extension(const std::string& path, const std::string& extension);
 
+/// Create a timestamped `mktemp`-compatible template string.
+std::string create_timestamped_template(const std::string& prefix, int wildcard_count=8, util::Optional<std::time_t> time=none);
+
+/// Reserve a unique file name based on a base directory path and a `mktemp`-compatible template string.
+/// If fd is not null, the temporary file will NOT be unlinked, and the file descriptor will be passed out.
```
bdash: Why bother with this when we don't have any need for it?

src/sync/impl/sync_file.hpp
```diff
@@ -19,6 +19,9 @@
 #ifndef REALM_OS_SYNC_FILE_HPP
 #define REALM_OS_SYNC_FILE_HPP
 
+#include <realm/util/optional.hpp>
+
+#include <chrono>
```
bdash: This now appears to be unused.

src/sync/impl/sync_file.hpp
austinzheng: I'll remove it.

src/sync/impl/sync_file.cpp
austinzheng: This is good to know, thanks.

austinzheng:
Linux sync CI failure is unrelated to the tests or code.

src/sync/impl/sync_file.cpp
```diff
@@ -180,20 +193,49 @@ std::string file_path_by_appending_extension(const std::string& path, const std:
     return buffer;
 }
 
+std::string create_timestamped_template(const std::string& prefix, int wildcard_count)
+{
+    constexpr int WILDCARD_MAX = 20;
+    constexpr int WILDCARD_MIN = 6;
+    // Clamp
+    wildcard_count = (wildcard_count > WILDCARD_MAX ? WILDCARD_MAX : wildcard_count);
+    wildcard_count = (wildcard_count < WILDCARD_MIN ? WILDCARD_MIN : wildcard_count);
```
bdash: Since we don't have [`std::clamp`](http://en.cppreference.com/w/cpp/algorithm/clamp) yet, I'd probably write this as:
```c++
wildcard_count = std::min(WILDCARD_MAX, std::max(WILDCARD_MIN, wildcard_count));
```

austinzheng:
Linux CI is broken. Tests pass locally. Going to merge.

PR #324, opened 2017-01-11
Add a single-object "collection" notifier

tgoyne:
Most of this is just refactoring to merge the two separate sets of transaction log parsing logic, so that the parser for the collection notifiers now (optionally) tracks which columns were modified, and the KVO stuff now uses that parser rather than a separate one.

With that done the object notifier itself is pretty trivial and straightforward.

### DIFF
--- a/src/impl/realm_coordinator.hpp
+++ b/src/impl/realm_coordinator.hpp
@@ -21,6 +21,8 @@
 
 #include "shared_realm.hpp"
 
+#include <realm/version_id.hpp>
+
 #include <condition_variable>
 #include <mutex>
 
--- a/src/impl/transact_log_handler.cpp
+++ b/src/impl/transact_log_handler.cpp
@@ -79,7 +79,7 @@ KVOAdapter::KVOAdapter(std::vector<BindingContext::ObserverState>& observers, Bi
         }
     }
 
-    auto max = max_element(begin(tables_needed), end(tables_needed));
+    auto max = std::max_element(begin(tables_needed), end(tables_needed));
     if (*max >= table_modifications_needed.size())
         table_modifications_needed.resize(*max + 1, false);
     if (*max >= table_moves_needed.size())
--- a/src/impl/transact_log_handler.cpp
+++ b/src/impl/transact_log_handler.cpp
@@ -27,10 +27,168 @@
 #include <realm/lang_bind_helper.hpp>
 
 #include <algorithm>
+#include <numeric>
 
 using namespace realm;
 
 namespace {
+
+class KVOAdapter : public _impl::TransactionChangeInfo {
+public:
+    KVOAdapter(std::vector<BindingContext::ObserverState>& observers, BindingContext* context);
+
+    void before(SharedGroup& sg);
+    void after(SharedGroup& sg);
+
+private:
+    BindingContext* m_context;
+    std::vector<BindingContext::ObserverState>& m_observers;
+    std::vector<void *> m_invalidated;
+
+    struct ListInfo {
+        BindingContext::ObserverState* observer;
+        size_t col;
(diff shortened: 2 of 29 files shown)


src/impl/transact_log_handler.cpp
```diff
@@ -27,10 +27,168 @@
...
+    for (auto& observer : observers) {
+        auto table = group.get_table(observer.table_ndx);
+        for (size_t i = 0, count = table->get_column_count(); i < count; ++i) {
+            if (table->get_column_type(i) != type_LinkList)
+                continue;
+            m_lists.push_back({&observer, i, {}});
+        }
+    }
+
+    auto max = max_element(begin(tables_needed), end(tables_needed));
```
bdash: Is ADL guaranteed to find `std::max_element` here? The iterator types seem like they could plausibly be raw pointer types, which aren't necessarily in the `std` namespace.

src/object.hpp
```diff
@@ -0,0 +1,91 @@
...
+    : m_realm(std::move(r)), m_object_schema(&s), m_row(o) { }
+
+    Object(SharedRealm r, ObjectSchema const& s, Row const& o)
+    : m_realm(std::move(r)), m_object_schema(&s), m_row(o) { }
+
+    ~Object();
+
+    // property getter/setter
+    template<typename ValueType, typename ContextType>
+    void set_property_value(ContextType ctx, std::string prop_name,
```
bdash: It's a little weird that these methods are declared in this class but the implementation lives over in object_accessor.hpp.

src/impl/transact_log_handler.cpp
```diff
@@ -27,10 +27,168 @@
...
+
+    for (auto& list : m_lists) {
+        if (list.builder.empty())
+            continue;
+        // column should have been marked as modified and thus already exist in `changes`
+        REALM_ASSERT(list.col < list.observer->changes.size());
+        auto& builder = list.builder;
+        auto& changes = list.observer->changes[list.col];
+
+        // KVO can't express moves (becuase NSArray doesn't have them), so
```
bdash: `s/becuase/because/`

src/impl/transact_log_handler.cpp
```diff
@@ -27,10 +27,168 @@
...
+        // transform them into a series of sets on each effected index when possible
```
bdash: affected?

src/impl/realm_coordinator.hpp
```diff
@@ -21,6 +21,8 @@
 
 #include "shared_realm.hpp"
 
+#include <realm/version_id.hpp>
```
jpsim: appears unused

src/shared_realm.hpp
```diff
@@ -33,17 +32,19 @@
...
 
 namespace realm {
-class BinaryData;
 class BindingContext;
 class Group;
 class Realm;
 class Replication;
 class SharedGroup;
 class StringData;
+class Table;
```
jpsim: appears unused

bdash: Used below:
```
+typedef BasicRow<Table> Row;
```

src/impl/realm_coordinator.hpp
bdash: Needed because it's no longer pulled in by shared_realm.hpp. `VersionID` is used in a few places in this file.

jpsim: I guess Xcode resolves these differently, since commenting this out still compiled.

src/shared_realm.hpp
jpsim: I guess Xcode resolves these differently, since commenting this out still compiled.

PR #325, opened 2017-01-11
Enable the building of Android with sync on CI

emanuelez:


bdash:
What remains before this can be merged?

emanuelez:
No work has been put into this yet. I think it's ok to put this on hold until we standardize the artifacts of the dependencies

PR #326, opened 2017-01-11
Adding gcc 4.9 support

kneth:
`std::put_time` is not supported by GCC 4.9. This is a hack to get around it.

See also http://stackoverflow.com/questions/14136833/stdput-time-implementation-status-in-gcc

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.2.0)
 if(REALM_PLATFORM STREQUAL "Android")
     # This must be before project()
     set(CMAKE_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/CMake/android.toolchain.cmake")
-    set(ANDROID_ABI "x86_64" CACHE STRING "")
+    set(ANDROID_ABI "x86" CACHE STRING "")
     set(ANDROID_NATIVE_API_LEVEL "android-16" CACHE STRING "")
 endif()
 
--- a/src/sync/impl/sync_file.cpp
+++ b/src/sync/impl/sync_file.cpp
@@ -23,6 +23,7 @@
 
 #include <iomanip>
 #include <sstream>
+#include <system_error>
 
 #if WIN32
 #include <io.h>
@@ -200,7 +201,15 @@ std::string create_timestamped_template(const std::string& prefix, int wildcard_
     wildcard_count = std::min(WILDCARD_MAX, std::max(WILDCARD_MIN, wildcard_count));
     std::time_t time = std::time(nullptr);
     std::stringstream stream;
-    stream << prefix << "-" << std::put_time(std::localtime(&time), "%Y%m%d-%H%M%S") << "-" << std::string(wildcard_count, 'X');
+    stream << prefix << "-";
+#if __GNUC__ < 5
+    char s[16];
+    strftime(s, 16, "%Y%m%d-%H%M%S", std::localtime(&time));
+    stream << s;
+#else
+    stream << std::put_time(std::localtime(&time), "%Y%m%d-%H%M%S");
+#endif
+    stream << "-" << std::string(wildcard_count, 'X');
     return stream.str();
 }
 
--- a/tests/sync/session.cpp
+++ b/tests/sync/session.cpp
@@ -335,19 +335,37 @@ TEST_CASE("sync: error handling", "[sync]") {
         CHECK(idx != std::string::npos);
         if (just_before->tm_year == just_after->tm_year) {
             std::stringstream stream;
+#if __GNUC__ < 5
+            char s[5];
+            strftime(s, 5, "%Y", just_after);
+            stream << s;
+#else
             stream << std::put_time(just_after, "%Y");
+#endif
             idx = recovery_path.find(stream.str());
             CHECK(idx != std::string::npos);
         }
         if (just_before->tm_mon == just_after->tm_mon) {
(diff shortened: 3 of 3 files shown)


CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.2.0)
 if(REALM_PLATFORM STREQUAL "Android")
     # This must be before project()
     set(CMAKE_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/CMake/android.toolchain.cmake")
-    set(ANDROID_ABI "x86_64" CACHE STRING "")
+    set(ANDROID_ABI "x86" CACHE STRING "")
```
emanuelez: Why was this change needed?

src/sync/impl/sync_file.cpp
```diff
@@ -200,7 +201,15 @@ std::string create_timestamped_template(const std::string& prefix, int wildcard_
     wildcard_count = std::min(WILDCARD_MAX, std::max(WILDCARD_MIN, wildcard_count));
     std::time_t time = std::time(nullptr);
     std::stringstream stream;
-    stream << prefix << "-" << std::put_time(std::localtime(&time), "%Y%m%d-%H%M%S") << "-" << std::string(wildcard_count, 'X');
+    stream << prefix << "-";
+#if __GNUC__ < 5
```
emanuelez: I think the limitation comes from the NDK and not from GCC 4.9 itself. Maybe we could add a check for android as well? Same goes for the other changes.

CMakeLists.txt
beeender: Just curious, why change this?

src/sync/impl/sync_file.cpp
```diff
@@ -23,6 +23,7 @@
 
 #include <iomanip>
 #include <sstream>
+#include <system_error>
```
beeender: not needed?

src/sync/impl/sync_file.cpp
beeender: `#if defined(__GNUC__) && __GNUC__ < 5`?  It seems this file will be used by other platform as well. Same for others.

CMakeLists.txt
kneth: Not really related - just easier for my personal emulator setup :-)

src/sync/impl/sync_file.cpp
kneth: That's safer. If undefined, the condition will be true.

kneth: No, not really. It's a GCC thing:

```kneth@blackbird:~$ uname -a
Linux blackbird 3.16.0-4-amd64 #1 SMP Debian 3.16.36-1+deb8u2 (2016-10-19) x86_64 GNU/Linux
kneth@blackbird:~$ g++ --version
g++ (Debian 4.9.2-10) 4.9.2
Copyright (C) 2014 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

kneth@blackbird:~$ cat pt.cpp
#include <sstream>
#include <iomanip>
#include <iostream>
#include <ctime>

int main(int argc, char *argv[]) {
    std::stringstream ss;
    std::time_t time = std::time(nullptr);
    ss << std::put_time(std::localtime(&time), "%Y%m%d-%H%M%S");
    std::cout << ss.str() << std::endl;
}
kneth@blackbird:~$ g++ pt.cpp
pt.cpp: In function 'int main(int, char**)':
pt.cpp:8:34: error: 'nullptr' was not declared in this scope
     std::time_t time = std::time(nullptr);
                                  ^
pt.cpp:9:11: error: 'put_time' is not a member of 'std'
     ss << std::put_time(std::localtime(&time), "%Y%m%d-%H%M%S");```

Compiling on Debian GNU/Linux (stable) gives the same error.

File: src/sync/impl/sync_file.cpp (hunk 3 above)
Comment (kneth) [inline src/sync/impl/sync_file.cpp]: Yes, `system_error` is declared there.

File: src/sync/impl/sync_file.cpp (hunk 2 above)
Comment (beeender) [inline src/sync/impl/sync_file.cpp]: uh, my bad. i thought it's `__GNUC__ > 5`. nevermind then.

File: tests/sync/session.cpp [hunk 4]
```diff
@@ -335,19 +335,37 @@ TEST_CASE("sync: error handling", "[sync]") {
         CHECK(idx != std::string::npos);
         if (just_before->tm_year == just_after->tm_year) {
             std::stringstream stream;
+#if __GNUC__ < 5
+            char s[5];
+            strftime(s, 5, "%Y", just_after);
```
bdash: Rather than duplicating this `#if` in  multiple places, why not define a `put_time` helper function that does the right thing for all platforms we use?

CMakeLists.txt
kneth: Match my emulator :-)

bdash: Rather than editing this file, do `-DANDROID_ABI=x86` when configuring the project via `cmake`.

tests/sync/session.cpp
kneth: I have created #328 so we don't forget your suggestion.

bdash: Oh, come on! I'm not asking you to spend hours revising things. I'm asking you to pull out a 4 line helper function.

PR #331, opened 2017-01-12
Fix resource leaking in the android event loop

beeender:
To fix https://github.com/realm/realm-java/issues/4002

The problems:
- Android thread may have a looper, but it may not be looping (eg:
  IntentService). Also the looper could quit by calling Looper.quit().
  This means if we rely on the looper callback to do clean up, many
  things will be leaked because of the looper callback might not be
  called at all.
- Passing pointers through pipe sounds dangerous. There would be many
  cases the pipe has some pointers left without deletion.

Things need to be considered:
- The EventLoopSignal can be destroyed in a different thread because of
  java GC.
- Looper callback should always have a valid pointer to use.
- We have to do clean up in the destructor.

Solution:
- Maintain a EventLoopSignal weak_ptr list and protect it with lock. So
  the callback can either get a share_ptr ref to use or detect the
  object has been destroyed.
- Use data param of the looper callback to pass the this's weak_ptr
  instead of passing it by pipe.

### DIFF
--- a/src/util/android/event_loop_signal.hpp
+++ b/src/util/android/event_loop_signal.hpp
@@ -16,8 +16,10 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-#include <atomic>
+#include <algorithm>
 #include <memory>
+#include <mutex>
+#include <vector>
 
 #include <errno.h>
 #include <fcntl.h>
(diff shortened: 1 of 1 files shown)


src/util/android/event_loop_signal.hpp
```diff
@@ -134,6 +167,37 @@ class EventLoopSignal : public std::enable_shared_from_this<EventLoopSignal<Call
         // return 1 to continue receiving events
         return 1;
     }
+
+    // Write a byte to a pipe to notify anyone waiting for data on the pipe
+    static void notify_fd(int fd)
```
beeender: Shall we put this method somewhere? It is using by multiple places.

bdash: Are the multiple places some of the different event loop signal implementations across platforms? It might be a little tricky to find a shared home as not every implementation wants it.

beeender: @bdash 
This is used by apple and epoll. It would be need by all POSIX systems i think. a `posix` dir in the `util`? It does worth it i think... at least `EAGAIN` checking has been fixed on apple version long time ago but it just got discovered recently on Android.

Anyway, i will just merge it like this for now. And create another PR for that.

beeender:
Tested with the original app, works fine even on blackberry.

src/util/android/event_loop_signal.hpp
beeender: never mind, they are not exact the same methods. my bad.

PR #332, opened 2017-01-12
Add more sync progress notifier tests

austinzheng:


austinzheng:
CI is crashing on the issue in https://github.com/realm/realm-sync/issues/1098.

austinzheng:
If this passes CI I'm just going to merge it, since NFC.

austinzheng:
@tgoyne @bdash 

I found and fixed a bug with my initial implementation that was causing these new tests to fail on Linux. Would appreciate a sanity check whenever you have time.

src/sync/sync_session.cpp
```diff
@@ -427,13 +427,19 @@ void SyncSession::handle_progress_update(uint64_t downloaded, uint64_t downloada
...
             bool should_delete = false;
-            invocations.emplace_back(create_notifier_invocation(pair.second, should_delete));
+            auto package = pair.second;
+            invocations.emplace_back(create_notifier_invocation(package, should_delete));
             if (should_delete) {
-                m_notifiers.erase(pair.first);
+                to_erase.emplace_back(pair.first);
             }
         }
+        // We need to erase after iteration since our Linux builds break if we try it the C++14 way.
```
bdash: I think it'd be better to state which compiler it breaks with, and why. Something like "std::unordered_map::erase in GCC 4.9's libstdc++ doesn't follow C++14's requirement that it preserve the order of the remaining elements, so we cannot safely erase the elements while iterating the map". (I'm speculating at the compiler version in question)

src/sync/sync_session.cpp
```diff
@@ -427,13 +427,19 @@ void SyncSession::handle_progress_update(uint64_t downloaded, uint64_t downloada
         m_current_downloadable = downloadable;
         m_current_downloaded = downloaded;
         m_current_uploaded = uploaded;
+        std::vector<uint64_t> to_erase;
         for (auto& pair : m_notifiers) {
             bool should_delete = false;
-            invocations.emplace_back(create_notifier_invocation(pair.second, should_delete));
+            auto package = pair.second;
```
bdash: This creates a copy of the package. Is that intentional? Is the local variable necessary for some reason?

src/sync/sync_session.cpp
```diff
@@ -455,8 +461,8 @@ std::function<void()> SyncSession::create_notifier_invocation(const NotifierPack
     // A notifier is expired if at least as many bytes have been transferred
     // as were originally considered transferrable.
     is_expired = !package.is_streaming && transferred >= package.captured_transferrable;
-    return [=](){
-        package.notifier(transferred, transferrable);
+    return [=, notifier=package.notifier](){
```
bdash: Why is this necessary?

src/sync/sync_session.cpp
austinzheng: This needs to be removed, oversight on my part.

src/sync/sync_session.cpp
austinzheng: What happens if `package` is destroyed after this function returns but before the closure being returned is invoked? If the closure body was `package.notifier()`, would we capture `package` by copy, or the reference to `package` by copy?

src/sync/sync_session.cpp
austinzheng: I'll note GCC 5.4.0 at least, since that was the version on my VM where I reproduced the issue.

tgoyne: The previous code was incorrect. You have to continue iteration from the iterator returned by `erase()` (which means you can't use for-range).

austinzheng: That makes a lot of sense. I'll go back to using an old-style for loop.

src/sync/sync_session.cpp
bdash: `[=]` says to capture any referenced variables by copy, so the lambda stores a copy of the object `package` references. It doesn't matter what happens to `package` after the lambda is created.

austinzheng: Ah, okay. I'll change it back then.

PR #334, opened 2017-01-13
Clear pipe for in android event loop callback

beeender:
ALooper is insteresting, the ALOOPER_EVENT_INPUT will always be triggered
whenever there are bytes left in the pipe. So read to clear the pipe in the
looper_callback.

The new changes are inside 8b948ae


beeender:
CI failure is not related with this changes. See https://github.com/realm/realm-sync/issues/1098 . Merge this first.

PR #336, opened 2017-01-13
Run the Android unit tests in an emulator

emanuelez:


Jenkinsfile
```diff
@@ -70,6 +70,39 @@ def doDockerBuild(String flavor, Boolean withCoverage, Boolean enableSync) {
...
+      def image = buildDockerEnv('ci/realm-object-store:android')
+      def emulator = docker.image('tracer0tong/android-emulator').run()
+
+      try {
+	ansiColor('xterm') {
+	  sshagent(['realm-ci-ssh']) {
+	    image.inside("-v /etc/passwd:/etc/passwd:ro -v ${env.HOME}:${env.HOME} -v ${env.SSH_AUTH_SOCK}:${env.SSH_AUTH_SOCK} -e HOME=${env.HOME} --link ${emulator.id}:emulator") {
+	      sh '''
+                rm -rf build
+                mkdir -p build
```
kneth: You don't need `-p` - `build` doesn't exist (removed in previous line).

Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM ubuntu:xenial
 
 RUN apt-get update && \
     apt-get install -y wget build-essential lcov curl cmake gcovr libssl-dev \
-      git python-cheetah libuv1-dev
+      git python-cheetah libuv1-dev ninja-build adb
```
kneth: Do we need `libuv1-dev` and `libssl-dev` when building for Android?

Jenkinsfile
```diff
@@ -74,30 +74,24 @@ def doAndroidDockerBuild() {
...
+	    sh '''
+              rm -rf build
+              mkdir build
+              cd build
+              cmake -DREALM_PLATFORM=Android -DANDROID_NDK=/opt/android-ndk -GNinja ..
+              ninja
+              adb connect emulator
+              timeout 10m adb wait-for-device
+              adb push tests/tests /data/local/tmp
+              adb shell '/data/local/tmp/tests || echo __ADB_FAIL__' | tee adb.log
```
kneth: Is the emulator deleted afterwards? Or should we clean up (remove `tests`) here?

Dockerfile
emanuelez: No, but the same Dockerfile is currently being used by all the Linux builds

Jenkinsfile
emanuelez: Yes, `withRun` takes care to stop and remove the container after use

PR #338, opened 2017-01-13
Fix a bug that crashes sync when refreshing a session’s token

austinzheng:
Fixes https://github.com/realm/realm-object-store/issues/337

Explanation: you're not allowed to call `bind()` on a `sync::Session` more than once. In one particular situation (token expires while session is active) we had a code path that violated this requirement.


tgoyne:
Shouldn't this just track whether or not we've ever called `bind()`? If we should never call `bind()` multiple times on one session I don't understand why it ever transitions back to `bind()` when given an access token.

austinzheng:
It's possible for the `sync::Session` owned by a `SyncSession` to be destroyed, and later replaced with a new one, which is why the logic is a bit more complicated. That being said, I should better document that possibility.

tgoyne:
Oh, that makes sense. I'd be inclined to push `session.m_refresh_session_upon_getting_token = false;` into `create_sync_session()` both because it eliminates a trivial amount of duplication and to make it more immediately obvious that they're inherently linked.

bdash:
I'd be inclined to do that, and to change the flag to `m_session_has_been_bound` (inverting the meaning of it).

src/sync/sync_session.hpp
```diff
@@ -177,6 +197,10 @@ class SyncSession : public std::enable_shared_from_this<SyncSession> {
 
     bool can_wait_for_network_completion() const;
 
+    // Whether or not the session object in `m_session` has been `bind()`ed before.
+    // This determines how the `SyncSession` behaves when refreshing tokens.
+    bool m_session_has_been_bound;
```
bdash: This member should be declared down near `m_session`, rather than amongst unrelated functions.

bdash:
Can we find some way to test this?

austinzheng:
I have some ideas for testing. The next commit I make to this PR will include some tests.

austinzheng:
Something is wrong with the tests on Linux. Troubleshooting now...

PR #343, opened 2017-01-17
copy_realm_to_recovery should take path as param

beeender:
The full path has been saved in the file_actions,
copy_realm_file_to_recovery_directory doesn't have to append the name to
the recovery path -- it will cause a File::NotFound thrown always.

### DIFF
--- a/src/sync/impl/sync_file.cpp
+++ b/src/sync/impl/sync_file.cpp
@@ -295,11 +295,10 @@ bool SyncFileManager::remove_realm(const std::string& absolute_path) const
     return success;
 }
 
-bool SyncFileManager::copy_realm_file_to_recovery_directory(const std::string& absolute_path, const std::string& new_name) const
+bool SyncFileManager::copy_realm_file_to_recovery_directory(const std::string& absolute_path, const std::string& new_path) const
 {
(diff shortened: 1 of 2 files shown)


beeender:
@austinzheng I am not familiar with those code, not sure how to build a test case to cover it.

nhachicha:
without this fix the client reset feature will not copy the Realm file. it will fail silently with `File::NotFound`, this is not caught by the existing unit test. 

cc @jpsim you may want to  do another cocoa release, once this is merged

austinzheng:
I thought I wrote a test for this, but it looks like I didn't. I can either write a follow-up test, or you can write a test by looking at `file.cpp` and `metadata.cpp`.

A good test for this would:
* Create a dummy realm at a location using `create_dummy_realm()`
* Manually create a file action metadata entry to move the file to a destination
* Execute the file action metadata by running `SyncManager::shared().configure_file_system`
* Assert the Realm file now exists at the path in the test file action metadata entry.

nhachicha:
add unit test here https://github.com/realm/realm-object-store/pull/346

austinzheng:
I prefer to fix this a different way.

PR #346, opened 2017-01-17
add unit test to cover the use of recovery_directory_path, for client…

nhachicha:
Adding unit test for https://github.com/realm/realm-object-store/pull/343

### DIFF
--- a/tests/sync/sync_manager.cpp
+++ b/tests/sync/sync_manager.cpp
@@ -451,48 +451,6 @@ TEST_CASE("sync_manager: file actions", "[sync]") {
     }
 }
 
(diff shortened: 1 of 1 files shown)


austinzheng:
Some of the existing tests are failing. Can you either update them if they're wrong, or change the fix if they're correct? Let me know if you have any questions about what they test; they should be self-explanatory.

bdash:
Why is this not just added as part of #343?

austinzheng:
I'm going to fix the tests, squash this PR onto Mulong's PR, and manually commit it once I've verified it.

austinzheng:
On second thought, I want to fix this a different way.

PR #349, opened 2017-01-17
Fix a retain cycle in the object notifier

### LINKED ISSUE #340: "Too many open files" exceptions when running tests on master
```
realm-object-store$ git show-ref HEAD
9ac9626f65c27306755e9cf61bd803d63a77263c refs/remotes/origin/HEAD
realm-object-store$ ulimit -n
256
realm-object-store$ ./tests/tests 

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
tests is a Catch v1.3.2 host application.
Run with -? for options

notifications: async delivery
  notifications for remote changes
realm-object-store/tests/results.cpp:55

realm-object-store/tests/results.cpp:218: FAILED:
  {Unknown expression after the reported line}
due to unexpected exception with message:
  Too many open files
```

`lsof` shows that most of the open files are Realm files (and their various locks). By adding some extra logging I determined that most of the Realm files opened by the [new tests in object.cpp](https://github.com/realm/realm-object-store/blob/0dc71fbfaf5e7361f0a06cd0db94767e6a4c5aa6/tests/object.cpp#L37-L159) are never being closed. From looking at the reference count on `r` as it goes out of scope it is apparent that there is a second reference to the Realm. A breakpoint added to `realm::_impl::ObjectNotifier::~ObjectNotifier()` is never hit, suggesting that this issue is similar in nature to a problem we had with collection notifications when they were first added, which was fixed in #77.

@tgoyne, can you please take a look?

tgoyne:
`unregister()` was never actually being called, so object notifiers would never get destroyed.

Add an assertion to the beginning of some of the test cases that there are no realms already open to catch this sort of thing before we hit the max file open limit.

Fixes #340.

### DIFF
--- a/src/impl/realm_coordinator.cpp
+++ b/src/impl/realm_coordinator.cpp
@@ -302,6 +302,14 @@ void RealmCoordinator::clear_all_caches()
     }
 }
 
+void RealmCoordinator::assert_no_open_realms() noexcept
+{
+#ifdef REALM_DEBUG
+    std::lock_guard<std::mutex> lock(s_coordinator_mutex);
+    REALM_ASSERT(s_coordinators_per_path.empty());
+#endif
+}
+
 void RealmCoordinator::wake_up_notifier_worker()
 {
     if (m_notifier) {
--- a/src/impl/realm_coordinator.hpp
+++ b/src/impl/realm_coordinator.hpp
@@ -77,6 +77,9 @@ public:
     // Clears all caches on existing coordinators
     static void clear_all_caches();
 
+    // Verify that there are no Realms open for any paths
+    static void assert_no_open_realms() noexcept;
+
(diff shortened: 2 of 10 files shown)


src/impl/realm_coordinator.cpp
```diff
@@ -302,6 +302,14 @@ void RealmCoordinator::clear_all_caches()
     }
 }
 
+void RealmCoordinator::assert_no_open_realms() noexcept
+{
+#ifdef REALM_DEBUG
```
bdash: Why `REALM_DEBUG`? `REALM_ASSERTIONS_ENABLED` would be my inclination.

PR #351, opened 2017-01-18
Fix issues with metadata and file actions

austinzheng:
Changes:
- Fixed a bug where the recovery path saved to the file action metadata was not the one expected when the metadata was read back out
- Refactored code that handles file actions into its own method
- Code that handles file actions now properly handles missing original files/a file in the intended path
- Modified some test utility methods for better resilience and easier debugging
- Added a test and fixed other tests

This commit contains code by @nhachicha and @beeender, and replaces https://github.com/realm/realm-object-store/pull/343 and https://github.com/realm/realm-object-store/pull/346

### DIFF
--- a/src/sync/impl/sync_file.cpp
+++ b/src/sync/impl/sync_file.cpp
@@ -277,7 +277,7 @@ bool SyncFileManager::remove_realm(const std::string& absolute_path) const
 {
     REALM_ASSERT(absolute_path.length() > 0);
     bool success = true;
-    // Remove the base Realm file (e.g. "example.realm").
+    // Remove the Realm file (e.g. "example.realm").
     success = File::try_remove(absolute_path);
     // Remove the lock file (e.g. "example.realm.lock").
     auto lock_path = util::file_path_by_appending_extension(absolute_path, "lock");
@@ -295,17 +295,17 @@ bool SyncFileManager::remove_realm(const std::string& absolute_path) const
     return success;
 }
 
-bool SyncFileManager::copy_realm_file_to_recovery_directory(const std::string& absolute_path, const std::string& new_name) const
+bool SyncFileManager::copy_realm_file(const std::string& old_path, const std::string& new_path) const
 {
-    REALM_ASSERT(absolute_path.length() > 0);
+    REALM_ASSERT(old_path.length() > 0);
     try {
-        auto new_path = util::file_path_by_appending_component(recovery_directory_path(), new_name);
         if (File::exists(new_path)) {
             return false;
         }
-        File::copy(absolute_path, std::move(new_path));
+        File::copy(old_path, new_path);
     } 
     catch (File::NotFound const&) {
+        return false;
     }
     catch (File::AccessError const&) {
         return false;
--- a/src/sync/sync_manager.cpp
+++ b/src/sync/sync_manager.cpp
@@ -92,21 +92,9 @@ void SyncManager::configure_file_system(const std::string& base_file_path,
         std::vector<SyncFileActionMetadata> completed_actions;
         SyncFileActionMetadataResults file_actions = m_metadata_manager->all_pending_actions();
         for (size_t i = 0; i < file_actions.size(); i++) {
-            SyncFileActionMetadata file_action = file_actions.get(i);
-            switch (file_action.action()) {
-                case SyncFileActionMetadata::Action::DeleteRealm:
-                    // Delete all the files for the given Realm.
-                    m_file_manager->remove_realm(file_action.original_name());
-                    completed_actions.emplace_back(std::move(file_action));
-                    break;
-                case SyncFileActionMetadata::Action::HandleRealmForClientReset:
-                    // Copy the primary Realm file to the recovery dir, and then delete the Realm.
-                    auto new_name = file_action.new_name();
-                    if (new_name && m_file_manager->copy_realm_file_to_recovery_directory(file_action.original_name(), *new_name)) {
-                        m_file_manager->remove_realm(file_action.original_name());
-                        completed_actions.emplace_back(std::move(file_action));
-                    }
-                    break;
+            auto file_action = file_actions.get(i);
+            if (run_file_action(file_action)) {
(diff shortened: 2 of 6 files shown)


tests/sync/sync_test_utils.hpp
```diff
@@ -31,7 +31,7 @@ bool create_dummy_realm(std::string path);
 void reset_test_directory(const std::string& base_path);
 bool results_contains_user(SyncUserMetadataResults& results, const std::string& identity);
 bool results_contains_original_name(SyncFileActionMetadataResults& results, const std::string& original_name);
-std::string tmp_dir();
+std::string tmp_dir(bool use_env_tmp=false);
```
bdash: This appears to be unused.

src/sync/sync_manager.cpp
```diff
@@ -163,24 +151,34 @@ bool SyncManager::immediately_run_file_actions(const std::string& realm_path)
...
             // Copy the primary Realm file to the recovery dir, and then delete the Realm.
             auto new_name = md.new_name();
-            if (new_name && m_file_manager->copy_realm_file_to_recovery_directory(md.original_name(), *new_name)) {
-                m_file_manager->remove_realm(md.original_name());
-                md.remove();
+            auto original_name = md.original_name();
+            if (!util::File::exists(original_name)) {
+                // The Realm file doesn't exist anymore.
+                return true;
+            } else if (new_name && !util::File::exists(*new_name) && m_file_manager->copy_realm_file(original_name, *new_name)) {
```
bdash: There's no need for this to be an `else if`.

tests/sync/sync_manager.cpp
```diff
@@ -338,9 +339,41 @@ TEST_CASE("sync_manager: file actions", "[sync]") {
...
+            std::string recovery_path = util::reserve_unique_file_name(SyncManager::shared().recovery_directory_path(),
+                                                                       util::create_timestamped_template("recovered_realm"));
+            // Ensure recovery file doesn't exist yet
+            REQUIRE(!File::exists(recovery_path));
+
+            // Open the metadata separately, so we can investigate it ourselves.
+            SyncMetadataManager manager(file_manager.metadata_path(), false);
+            auto a = SyncFileActionMetadata(manager, SyncFileActionMetadata::Action::HandleRealmForClientReset, realm_base_path, "", "", recovery_path);
+
+            SyncManager::shared().reset_for_testing();
```
bdash: Shouldn't reseting the sync manager also be resetting any sync metadata?…

tests/sync/sync_test_utils.hpp
austinzheng: I'll remove it; we can add it back in if we need it later.

tests/sync/sync_manager.cpp
austinzheng: `reset_for_testing()` only resets the sync manager's in-memory state. What the test is trying to do is to simulate the app starting up again and reloading the persisted metadata (following line). Should I rename `reset_for_testing()` to make its semantics clearer?

austinzheng:
The Linux sync build is failing, again. I haven't been able to reproduce it on my VM yet (running same version of GCC as the CI machine). I'll think about it tonight.

bdash:
I can reproduce the exception that CI is seeing some of the time. [Here's a backtrace at the point the exception is thrown](https://gist.github.com/bdash/5cb25f801a48f826ee8d873a9bf841db).

src/sync/sync_manager.cpp
```diff
@@ -163,22 +151,35 @@ bool SyncManager::immediately_run_file_actions(const std::string& realm_path)
...
-                m_file_manager->remove_realm(md.original_name());
-                md.remove();
+            auto original_name = md.original_name();
+            if (!util::File::exists(original_name)) {
+                // The Realm file doesn't exist anymore.
+                return true;
+            } 
+            if (new_name && !util::File::exists(*new_name) && m_file_manager->copy_realm_file(original_name, *new_name)) {
+                // We successfully copied the Realm file to the recovery directory.
+                m_file_manager->remove_realm(original_name);
```
beeender: If the realm file is removed here, `metadata->remove();` will operate on a deleted file. It won't be a problem, but is it expected?

bdash: `SyncFileActionMetadata::remove` just removes the metadata record from the metadata Realm.

austinzheng:
Thanks for the logs. I'm too tired to examine this in much more detail tonight, but I suspect that calling the reset function in the middle of the test might have something to do with it.

src/sync/impl/sync_file.cpp
```diff
@@ -295,15 +295,14 @@ bool SyncFileManager::remove_realm(const std::string& absolute_path) const
...
+    REALM_ASSERT(old_path.length() > 0);
     try {
-        auto new_path = util::file_path_by_appending_component(recovery_directory_path(), new_name);
         if (File::exists(new_path)) {
             return false;
         }
-        File::copy(absolute_path, std::move(new_path));
+        File::copy(old_path, new_path);
     } 
     catch (File::NotFound const&) {
```
austinzheng: Logging is a good idea. I'll think about whether NotFound should be swallowed silently, or merit an error message. (I think there's an argument for having the method report all errors, and making it the responsibility of the caller to check for error conditions they want to ignore.)

tests/sync/sync_test_utils.cpp
```diff
@@ -56,13 +56,10 @@ bool results_contains_original_name(SyncFileActionMetadataResults& results, cons
 }
 
 std::string tmp_dir() {
-    const char* dir = getenv("TMPDIR");
```
tgoyne: Why is this switching away from using $TMPDIR? Being able to override the temporary directory in the standard manner is useful.

austinzheng: I originally thought not using $TMPDIR would make things easier to debug, but upon further thought I don't think that's true. I'll change it back.

austinzheng:
Restarting test since the Android CI system failed.

PR #352, opened 2017-01-18
Fix skipping notifications for transactions with no relevant changes

tgoyne:
The skip flag was only cleared if a notification would actually have been delivered, which resulted in the *next* transaction's notification being skipped if a transaction did not actually do anything which would produce a notification.

Reported by https://github.com/realm/realm-cocoa/issues/4558.

### DIFF
--- a/src/impl/collection_notifier.cpp
+++ b/src/impl/collection_notifier.cpp
@@ -286,6 +286,12 @@ void CollectionNotifier::prepare_handover()
     m_sg_version = m_sg->get_version_of_current_transaction();
     do_prepare_handover(*m_sg);
     m_has_run = true;
+
+#ifdef REALM_DEBUG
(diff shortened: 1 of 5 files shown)


src/impl/collection_notifier.cpp
```diff
@@ -286,6 +286,12 @@ void CollectionNotifier::prepare_handover()
     m_sg_version = m_sg->get_version_of_current_transaction();
     do_prepare_handover(*m_sg);
     m_has_run = true;
+
+#ifdef REALM_DEBUG
```
bdash: Same question as in #349 about `REALM_DEBUG` vs `REALM_ASSERTIONS_ENABLED`.

tgoyne: Even if assertions are enabled in a release build we don't really want the extra lock acquisitions.

PR #354, opened 2017-01-19
Fix an issue with session progress notifications

austinzheng:
Changes:
- `SyncSession` now tracks whether the initial sync progress update has happened yet, and manages notifications registered by the binding accordingly
- Session progress notification tests now wait for pending uploads and downloads, to avoid `handle_notifications()` being "spuriously" called by the actual sync client during tests
- Removed use of `check_status()` so we can get useful line numbers should the tests fail


austinzheng:
CI passes, except for the `osx_sync` job which died because the CI machines are being reset. Merging.

PR #355, opened 2017-01-19
Fix a race condition leading to bad version errors in the notifier worker

tgoyne:
We need to decide which version to advance to in the notifier worker while the notifier lock is held, as otherwise we may advance to a version that some notifiers needed to skip over and then trying to perform that skip will fail.

The problematic sequence of events was:

  1. Notifier worker releases notifier lock
  2. Thread with notifications acquires notifier lock
  3. Thread with notifications commits write transaction
  4. Thread with notifications sets skip version
  5. Notifier worker advances to latest version
  6. Notifier worker acquires notifier lock
  7. Notifier worker tries to advance to skip version and fails

No new test for this as I don't know how to produce it other than hammering commits and hoping to hit the bad timing.

### DIFF
--- a/src/impl/realm_coordinator.cpp
+++ b/src/impl/realm_coordinator.cpp
@@ -599,6 +599,20 @@ void RealmCoordinator::run_async_notifiers()
         for (auto& notifier : new_notifiers) {
             notifier->detach();
         }
+
+        // We want to advance the non-new notifiers to the same version as the
+        // new notifiers to avoid having to merge changes from any new
+        // transaction that happen immediately after this into the new notifier
+        // changes
+        version = m_advancer_sg->get_version_of_current_transaction();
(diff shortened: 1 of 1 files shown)


src/impl/realm_coordinator.cpp
```diff
@@ -599,6 +599,20 @@ void RealmCoordinator::run_async_notifiers()
...
+
+        // We want to advance the non-new notifiers to the same version as the
+        // new notifiers to avoid having to merge changes from any new
+        // transaction that happen immediately after this into the new notifier
+        // changes
+        version = m_advancer_sg->get_version_of_current_transaction();
+        m_advancer_sg->end_read();
+    }
+    else {
+        // If we have no new notifiers we want to just advnace to the latest
```
austinzheng: 'advance'

Issue #356, opened 2017-01-25
Sporadic sync test failure


austinzheng:
Test fails as follows, after pulling in @morten-krogh's fixes to the notifier deadlock issue:

```
SyncSession: management by SyncUser
  sessions that were destroyed can be properly recreated when requested again
-------------------------------------------------------------------------------
tests/sync/session.cpp:79
...............................................................................

tests/sync/session.cpp:200: FAILED:
  REQUIRE( weak_session.expired() )
with expansion:
  false
```

and/or

```
SyncSession: management by SyncUser
  sessions that were destroyed can be properly recreated when requested again
-------------------------------------------------------------------------------
tests/sync/session.cpp:79
...............................................................................

tests/sync/session.cpp:206: FAILED:
  CHECK( session )
with expansion:
  false
```

Test failure is very rare; around once every 250 runs.

We hypothesize something related to the notifications code may be keeping the session alive longer than would be expected if the last reference to it was dropped when leaving the block scope. Using the runloop to wait for it to be destroyed fixes that bug.

We don't yet know why the second issue is happening. My hypothesis is that `sync` is reporting a particular error that puts the user in the `error` state around the time the session is being recreated, causing the user to return a null pointer instead of a pointer to a session.

cc @bdash

austinzheng:
Another test failure:

```
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~/
root/realm-object-store/src/sync/sync_manager.cpp:451: [realm-core-2.3.0] Assertion failed: it != m_inactive_sessions.end()
IMPORTANT: if you see this error, please send this log to help@realm.io.
tests is a Catch v1.3.2 host application.
Run with -? for options

-------------------------------------------------------------------------------
SyncSession: management by SyncUser
  sessions that were destroyed can be properly recreated when requested again
-------------------------------------------------------------------------------
/root/realm-object-store/tests/sync/session.cpp:79
...............................................................................

Thread name: tests/root/realm-object-store/tests/sync/session.cpp:208: FAILED:
  CHECK( session )

with expansion:
  false
  false

-------------------------------------------------------------------------------
SyncSession: management by SyncUser
-------------------------------------------------------------------------------
/root/realm-object-store/tests/sync/session.cpp:79
...............................................................................

/root/realm-object-store/tests/sync/session.cpp:181: FAILED:
[...]

PR #360, opened 2017-01-27
Apply Mark's fix to fix a sync test

austinzheng:
The sync fix in https://github.com/realm/realm-sync/pull/1112, which fixes a sync deadlock related to progress notification blocks, allowed @bdash to uncover a new issue in our sync tests.

The issue is this: one of our tests relies on a session being destroyed right after leaving a block scope, which would ordinarily happen because the last pointer to that session goes out of scope. However, it's possible that, if certain things happen (the progress notifier callback is invoked at a certain point, causing another strong reference to the session to be generated), the session might not actually be destroyed immediately.

There was also another issue, but I've not been able to reproduce it after applying this fix. We can either hold this PR open as a way to track progress, or we can merge it in and fix the other test bug (if it still exists) in a later PR.

Related: https://github.com/realm/realm-object-store/issues/356

### DIFF
--- a/tests/sync/session.cpp
+++ b/tests/sync/session.cpp
@@ -196,8 +196,10 @@ TEST_CASE("SyncSession: management by SyncUser", "[sync]") {
             REQUIRE(on_disk_path.size() > 0);
             REQUIRE(weak_session.lock());
         }
-        // Session is dead, so the SyncUser's weak pointer to it should be nulled out.
-        REQUIRE(weak_session.expired());
+        // Wait for the session to die. It may not happen immediately if a progress or error handler
+        // is called on a background thread and keeps the session alive past the scope of the above block.
+        EventLoop::main().run_until([&] { return weak_session.expired(); });
+
         // The next time we request it, it'll be created anew.
         // The call to `get_session()` should result in `SyncUser::register_session()` being called.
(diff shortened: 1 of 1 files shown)


austinzheng:
I did run into this new bug after running the offending test 20,000 times.

```
../realm/noinst/client_file_access_cache.hpp:205: [realm-core-2.3.0] Assertion failed: m_prev_open_file
IMPORTANT: if you see this error, please send this log to help@realm.io.
Thread name: tests

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
tests is a Catch v1.3.2 host application.
Run with -? for options

SyncSession: management by SyncUser
  a SyncUser properly unbinds its sessions upon logging out
/root/realm-object-store/tests/sync/session.cpp:79

/root/realm-object-store/tests/sync/session.cpp:114: FAILED:
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal

===============================================================================
test cases: 1 | 1 failed
assertions: 6 | 5 passed | 1 failed
```

bdash:
I think that particular error can happen even prior to this change. It's just a little harder to hit since the `REQUIRE(weak_session.expired())` problem was so easy to hit.

PR #361, opened 2017-02-01
Update to catch 1.7.0

tgoyne:
There's been a bunch of assorted bug fixes since the version we were on (1.3.1) and it's a bit faster now.


austinzheng:
Looks like this uncovered a bug in one of our tests. I'll run it using `lldb` tonight or tomorrow and figure out what's going on.

bdash:
I'm not able to reproduce a test failure locally. It wouldn't surprise me if it just hit an existing intermittent problem.

tgoyne:
It happened twice in a row on CI. If it's a pre-existing race condition the perf improvements in catch may have made it easier to hit.

austinzheng:
I'm making the job run one more time. If it fails again I'll try running LLDB from the CI machine remotely.

austinzheng:
I pushed a commit to try fixing the issue. If it's not appropriate we can revert it.

tgoyne:
With that change `perform()` doesn't make any sense; the only thing it can possibly do is introduce race conditions in tests (if no other threads are scheduling things on the current thread's loop it's identical to just doing the things in the block directly, and if they are then the order things happen depends on timing).

Currently `perform()` is only used in one test which passes if the use of `perform()` is removed entirely and it's not clear to me what the intention of the test even is.

austinzheng:
You're right. Mark wrote this test when our "wait for download" API was still blocking; in that light it makes sense that `perform()` would enqueue the block onto the runloop but not actually run it until `run_until()` was called. I am going to remove my change and just remove that test.

PR #363, opened 2017-02-03
Add support for passing in a changeset transformer when opening a SharedRealm

alazier:
@bdash


dependencies.list
```diff
@@ -1,2 +1,2 @@
 REALM_CORE_VERSION=2.3.0
-REALM_SYNC_VERSION=1.0.0-BETA-7.2
+REALM_SYNC_VERSION=1.1.0-alpha17
```
bdash: We probably can't merge this until it's in a real release.

src/sync/sync_config.hpp
```diff
@@ -100,6 +104,7 @@ struct SyncConfig {
     SyncSessionStopPolicy stop_policy;
     std::function<SyncBindSessionHandler> bind_session_handler;
     std::function<SyncSessionErrorHandler> error_handler;
+    std::shared_ptr<ChangesetTransformer> transformer;
```
bdash: What is the expected lifetime of the `ChangesetTransformer` with respect to other objects? Having the configuration share ownership while the `sync::Session` does nothing to keep it alive seems a bit weird.

dependencies.list
alazier: The real release is out and this file has been updated.

src/sync/sync_config.hpp
alazier: This object is owned externally to SyncConfig - an instance of this object is passed in and reused whenever the adapter opens a synced Realm which needs to export changeset information. As is this object is passed into the global notifier on creation - I guess we could instead pass a method which will create this object on demand, but since this object has no state is seems unnecessary.

bdash: What ensures that the object outlives the sync code that makes use of it? Do we ensure that that `SyncConfig` outlives the `Realm`'s `SharedGroup`? Is the `SyncConfig` the right thing to be keeping this alive?

If it's up to some external entity to keep the transformer alive, why does `SyncConfig` share ownership? How does that external entity know when the transformer is no longer in use?

dependencies.list
```diff
@@ -1,2 +1,2 @@
 REALM_CORE_VERSION=2.3.1
-REALM_SYNC_VERSION=1.0.0-BETA-7.2
+REALM_SYNC_VERSION=1.1.0
```
bdash: Looks like the newline at EOF went missing here?

alazier:
@bdash do we need to do anything else before we can merge?

bdash:
@alazier, responding to my questions above would be a start.

alazier:
Didn't see your latest comment...

src/sync/sync_config.hpp
alazier: There is nothing that ensures this at the moment. In the case that you create a `Realm` and its `SharedGroup` with a `ChangesetCookeer` then reset the `shared_ptr` or SyncConfig` after creation then you would run into issues. This would pretty much be going out of your way to break things though so not sure this is a case we need to deal with - if we do need to prevent this we can either store the `ChangesetCooker` separately from `Config` in the `Realm` object, or we should try to make it const (not sure if this is possible). 

The only other possible issue I can think of would be during destruction as both the `ChangesetCooker` and `SharedGroup` are destroyed in the `Realm` destructor - if the cooker is destroyed first and things are happening on other threads this might cause issues but not sure if this is something we need to worry about or deal with.

bdash: No-one external to the object store has the ability to modify the `SyncConfig` that `Realm` and `SyncSession` instances hold, so that's not a concern. I'm asking if those `SyncConfig` instances that the object store holds internally outlive the objects that use the `ChangesetCooker`. If they don't, the object may be deallocated while it's still being used.

src/shared_realm.hpp
```diff
@@ -326,6 +326,8 @@ class Realm : public std::enable_shared_from_this<Realm> {
     // FIXME private
     Group& read_group();
 
+    Replication *history() { return m_history.get(); }
```
bdash: What needs this?

src/sync/sync_config.hpp
alazier: The `ChangesetCooker` is used by `Sync::Client` when applying remote changesets. Remote changesets are applied after new writelogs are received from the server. So if a `Sync::Client` still exists after we destroy the `SharedGroup` then there could be an issue. I'm not sure if this is the case though - I remember earlier object-store `SyncClients` were kept alive but I'm not sure if this is still the case.

src/shared_realm.hpp
alazier: The api for accessing `ChangeSets` is exposed through `SyncHistory`.

src/sync/sync_config.hpp
bdash: Is it really tied to the client rather than a session? The sync client created by the object store lives indefinitely. One sync client services multiple sessions.

alazier: Our `ChangesetCooker` is and stored on `SyncHistory`. `SyncHistory` is created at the same time as the `shared_group` and is owned by the `SharedRealm`. I believe there is no lifetime issues with this reference to the cooker.

There is also a `ChangesetCooker` used by `Sync::Client` which it gets via a `SyncHistory` object via its `m_file_access` member. But I don't see where this would get initialized with our cooker instance so I don't think this is an issue (although i am not yet certain).

alazier: Hopefully @kspangsege can verify that the `ChangesetCooker` on `Sync::Client` is a different cooker than the one passed into `SyncHistory`.

kspangsege: > Hopefully @kspangsege can verify that the `ChangesetCooker` on `Sync::Client` is a different cooker than the one passed into `SyncHistory`.

Every `sync::Sync::History` contains a pointer to a cooker, or null when cooking is not enabled. Your cooker is called called from the internals of `sync::SyncHistory`.

Each session (`sync::Session`) has its own `sync::SyncHistory` object, and those history objects are created automatically as the sessions are created. A pointer to the cooker (or null) is passed via `sync::SyncHistory::Config` to the history constructor (`sync::make_sync_history()`).

If a session specific cooker pointer is specified via `sync::Session::Config`, then that pointer is passed to the history constructor. Otherwise if a client-wide cooker pointer is specified via `sync::Client::Config`, then that pointer is passed. Otherwise null is passed (cooking disabled).

In any case, neither the `sync::Client`, the `sync::Session`, or the `sync::SyncHistory` assumes ownership of any cooker. It is always the applications (object stores) responsibility to keep the cooker alive for sufficiently long.

Does this answer your question?

kspangsege: By the way, when you create your own `sync::SyncHistory` objects, which you do when you create `SharedGroup` objects, then it is up to you whether you want to pass a cooker to it, but it makes little sense to do that, because it will never be used. This is because the cooker is only invoked during integration of changesets received from the server, and those are always processed internally in the history objects created internally by the `sync::Session` objects.

alazier: Really? That is exactly what we are doing and it seems to work as intended.

kspangsege: Can you point me to the relevant code?

kspangsege: In particular, where are the `sync::Session::Config` objects created, and where are the `sync::SyncHistory::Config` objects created?

kspangsege: I can see that you are setting `sync::SyncHistory::Config::changeset_cooker` here https://github.com/realm/realm-object-store/pull/363/files#diff-55a15b0c9125ec5f7989470e6eb8b217R182. If I am right, and I'm pretty sure I am, you should be able to remove that line and have everything working the same as before.

kspangsege: The line at https://github.com/realm/realm-object-store/pull/363/files#diff-f3165f1bbea4d4c8abf9c03a4b4e86f9R509, on the other hand, is crucial.

alazier: How is that possible? That is the only place I actually set the `changeset_cooker`.

alazier: Nevermind, I also set it on `sync::Session::Config`...

kspangsege: Exactly.

alazier: Removing that line does indeed work. But this means that if a `sync::Session` outlives the `SharedRealm` then it could be holding onto an invalid cooker. If we removed this line and set the cooker using the history object would that still work?

alazier: That doesn't work unfortunately. So we have an object lifetime issue if the `sync::Session` outlives the `SharedRealm`. If we stored the `ChangesetCooker` as a `shared_ptr` that would resolve this. Is that possible?

alazier: Of if we could make a copy rather than passing a pointer...

kspangsege: Ok, so you are destroying the cooker when the `SharedRealm` is destroyed? This is a problem if it leads to a violation of the rule stated here https://github.com/realm/realm-sync/blob/b10fe25d4477603e3deddf06b8363d250ccd332c/src/realm/sync/client.hpp#L231. I do realize that this can be a pain.

kspangsege: Hmm. Why do you want to tie the lifetime of the cooker to the lifetime of the SharedRealm in the first place? The two have little to do with each other, as far as I can see.

alazier: It is the way it is because users specify a cooker when populating a config used to create a SharedRealm. This is why it is stored in the config. If we were able to hold onto a shared_ptr to the cooker in the `sync::Client` then I think there would no longer be an issue.

kspangsege: Alright, I will make the change you suggest. I'll say though, that I dislike the change on principal grounds, because it obscures the point of destruction, and forces you to dynamically allocate the cooker. I understand that this is no problem for the current use case. I also understand that there is little you can do on your side to solve the problem given the current limitations of the `sync::Session` API. I hope in the future we can revisit this and find a better solution.

kspangsege: Done. See https://github.com/realm/realm-sync/pull/1206.

alazier:
@bdash this has been updated to use the latest sync version which stores the cooker as a shared_ptr. @radu-tutueanu

src/shared_realm.cpp
```diff
@@ -38,6 +38,7 @@
 #include <realm/util/scope_exit.hpp>
 
 #if REALM_ENABLE_SYNC
+#include "sync/sync_config.hpp"
```
bdash: This no longer appears to be necessary.

src/shared_realm.cpp
```diff
@@ -209,7 +210,7 @@ void Realm::Internal::begin_read(Realm& realm, VersionID version_id)
 
 SharedRealm Realm::get_shared_realm(Config config)
 {
-    auto coordinator = RealmCoordinator::get_coordinator(config.path);
+    auto coordinator = RealmCoordinator::get_coordinator(config);
```
bdash: This no longer appears to be necessary.

src/sync/sync_config.hpp
```diff
@@ -29,11 +29,15 @@
...
 #include <unordered_map>
 
+#include <realm/sync/history.hpp>
+
 namespace realm {
 
 class SyncUser;
 class SyncSession;
 
+using ChangesetTransformer = sync::SyncHistory::ChangesetCooker;
```
bdash: Ugh, I hate how the use of nested classes forces us to drag in headers.

bdash:
I'm not quite sure what's happened here, but the commit history appears to show a bunch of duplicated commits. Can we sort that out before merging?

src/sync/sync_config.hpp
alazier: Me too...

PR #364, opened 2017-02-06
Do not move a variable we will use later (fixes path in exception).

ironage:
The `config.path` would otherwise be empty if we throw the exception below.


tgoyne:
Rather than copying the config this can just call `get_path()` when building the exception.

ironage:
We use `config` as if it is alive after the move, calling a method and reading its other members. It's working fine now (except for the path) but if it changes in the future (configuration is something I imagine is likely to change) then it would be hard to spot the error here again. I think a full copy is a small price to pay for the future maintenance problems this could cause.

PR #365, opened 2017-02-06
Store a strong reference to the Table in Results

tgoyne:
This fixes a use-after-free in `is_valid()` when there's nothing else keeping the Table alive after the Realm is invalidated.

### DIFF
--- a/src/results.cpp
+++ b/src/results.cpp
@@ -36,7 +36,7 @@ Results::~Results() = default;
 Results::Results(SharedRealm r, Query q, SortDescriptor s, SortDescriptor d)
 : m_realm(std::move(r))
 , m_query(std::move(q))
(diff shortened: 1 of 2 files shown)


src/results.cpp
```diff
@@ -45,18 +45,18 @@ Results::Results(SharedRealm r, Query q, SortDescriptor s, SortDescriptor d)
 
 Results::Results(SharedRealm r, Table& table)
 : m_realm(std::move(r))
-, m_table(&table)
 , m_mode(Mode::Table)
 {
+    m_table.reset(&table);
```
bdash: It's a bit surprising that it's not possible to construct a `TableRef` from a `Table*`, so we instead have to resort to this default-construct / reset dance.

PR #367, opened 2017-02-07
If a session is dying and receives a fatal error, kill it immediately

### LINKED ISSUE #292: Sync sessions with "wait for upload" policy never die if they cannot connect to the server
In realm/realm-js-private#235 it was observed that if a sync session is created with `SyncSessionStopPolicy::AfterChangesUploaded` and it is unable to connect to the server, the session will linger in the dying state indefinitely. The object store should ensure that such sessions die in a timely manner.

austinzheng:
Fixes #292

Any ideas about how to test this change would be greatly appreciated.

### DIFF
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -238,9 +238,10 @@ struct sync_session_states::Dying : public SyncSession::State {
     {
         if (error.is_fatal) {
             session.advance_state(lock, inactive);
-            return true;
         }
-        return false;
+        // If the error isn't fatal, don't change state, but don't
+        // allow it to be reported either.
+        return true;
     }
 
     bool revive_if_needed(std::unique_lock<std::mutex>& lock, SyncSession& session) const override
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -306,6 +306,13 @@ void SyncSession::handle_error(SyncError error)
     bool should_invalidate_session = error.is_fatal;
     auto error_code = error.error_code;
 
+    if (m_state == &State::dying && error.is_fatal) {
+        // A fatal error while the session is dying should immediately kill the session.
+        std::unique_lock<std::mutex> lock(m_state_mutex);
+        advance_state(lock, State::inactive);
+        return;
+    }
+
     if (error_code.category() == realm::sync::protocol_error_category()) {
         using ProtocolError = realm::sync::ProtocolError;
         switch (static_cast<ProtocolError>(error_code.value())) {
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -109,6 +109,9 @@ struct SyncSession::State {
 
     virtual void close(std::unique_lock<std::mutex>&, SyncSession&) const { }
 
+    /// Returns true iff the error has been fully handled and the error handler should immediately return.
+    virtual bool handle_error(std::unique_lock<std::mutex>&, SyncSession&, const SyncError&) const { return false; }
+
     static const State& waiting_for_access_token;
     static const State& active;
     static const State& dying;
@@ -231,6 +234,15 @@ struct sync_session_states::Dying : public SyncSession::State {
         });
     }
 
+    bool handle_error(std::unique_lock<std::mutex>& lock, SyncSession& session, const SyncError& error) const override
+    {
+        if (error.is_fatal) {
+            session.advance_state(lock, inactive);
+            return true;
+        }
+        return false;
+    }
+
     bool revive_if_needed(std::unique_lock<std::mutex>& lock, SyncSession& session) const override
     {
         // Revive.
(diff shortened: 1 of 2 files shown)


kspangsege:
I suppose you want to test two things: That a nonfatal error does not prematurely end the dying phase, and that a fatal error does end it. So you need to be able to make a nonfatal and a fatal error occur at the right time, namely after the session that initiated the dying phase.

I'm not sure what freedoms you have in you test setups, but I would be thinking in terms of:
  1. Create server, but don't start executing it's event loop just yet.
  2. Create client session with invalid token or similar to provoke error from server.
  3. End client session, and enter into dying phase.
  4. Start executing servers event loop, to make it discover the invalid token and send back an error.

austinzheng:
How does a client react to a server that's not running? Wouldn't the client just fail to properly handshake with the server in the first place?

kspangsege:
> How does a client behave when trying to connect to a server that's not running [(ed) its event loop]? Wouldn't the client just fail to properly handshake with the server in the first place?

No. After `sync::Server::start()` has been called, the server has created the listening socket. At that time clients can start trying to establish new connections. Such connections, however, will not be fully established until the server event loop is executed (`sync::Server::run()`).

src/sync/sync_session.cpp
```diff
@@ -306,6 +306,13 @@ void SyncSession::handle_error(SyncError error)
     bool should_invalidate_session = error.is_fatal;
     auto error_code = error.error_code;
 
+    if (m_state == &State::dying && error.is_fatal) {
```
bdash: a. Testing `m_state` without first taking the mutex is racy.
b. Explicitly testing the state doesn't really fit with how the existing state machine code works. It'd be better to add a `handle_error` action to the state machine, which would be a noop for every other state, and could return a boolean indicating whether to consider the error to be handled.

austinzheng: Good catch. Thanks for looking at this.

austinzheng:
Thanks for the tips, @kspangsege. After trying to write some tests, it looks like I need to figure out a way to actually insert objects into a test file or at least convince the test session that it has data it needs to upload. Otherwise, `Session::async_wait_for_upload_completion()` will immediately invoke its callback, and I can't reliably keep the session in the `Dying` state long enough to do any useful test work.

austinzheng:
I added a bunch of tests that exercise both this specific case as well as some other behavior related to `dying`. Unfortunately, if I run the tests thousands of times I eventually end up leaking file descriptors and everything breaks.

austinzheng:
Something is wrong with the linux-sync CI box (the tests pass but the post-test phase fails), but everything is working otherwise. A review would be greatly appreciated.

tests/sync/session.cpp
```diff
@@ -42,17 +47,27 @@ template <typename FetchAccessToken, typename ErrorHandler>
...
+                                          Realm::Config* out_config=nullptr)
 {
     std::string url = server.base_url() + path;
     SyncTestFile config({user, url, std::move(stop_policy),
         [&, fetch_access_token=std::forward<FetchAccessToken>(fetch_access_token)](const auto& path, const auto& config, auto session) {
             auto token = fetch_access_token(path, config.realm_url);
             session->refresh_access_token(std::move(token), config.realm_url);
         }, std::forward<ErrorHandler>(error_handler)});
+    config.schema = Schema{
+        {"sync_session_object", {
```
bdash: Do we really want to use this schema for every test? It seems preferable to specify it only for tests that care.

tests/sync/session.cpp
```diff
@@ -64,6 +79,20 @@ std::shared_ptr<SyncSession> sync_session(SyncServer& server, std::shared_ptr<Sy
 
 namespace {
 
+void add_objects(Realm::Config& config, int count=2)
```
bdash: Since we only use this in one place, why not just do it in-line? Do we think other tests will need these same test objects?

tests/sync/session.cpp
austinzheng: Agreed. I'll make it a parameter to be passed in.

tests/sync/session.cpp
austinzheng: I have some vague notions of tests that might be able to use them, but I think it makes sense to inline this code and only break it out if we do end up needing test objects for other tests.

austinzheng:
I made a few changes. Now all the session death tests create test objects (for the same reason I mentioned in the comment to Kristian above; there needs to be outstanding data for the upload waiter to not immediately fire). Since the test objects use a schema particular to that particular set of tests, I changed the helper function into a helper closure defined within the test suite.

tests/sync/session.cpp
```diff
@@ -382,6 +395,119 @@ TEST_CASE("sync: error handling", "[sync]") {
...
+    SyncServer server(false);
+    auto schema = Schema{
+        {"sync_session_object", {
+            {"value 1", PropertyType::Int},
+            {"value 2", PropertyType::Int},
+        }},
+    };
+
+    constexpr int count = 2;
+    std::function<void(Realm::Config&)> add_objects = [=](Realm::Config& config) {
```
bdash: No need to use `std::function` here. This can just be `auto`.

Why does this use `[=]`? It doesn't look like it needs to capture anything.

austinzheng: No it doesn't. I had a slightly different implementation before.

PR #369, opened 2017-02-08
Check all outgoing links for deep change checking

tgoyne:
Previously it stopped after the first non-null link. It also turned out that the check for exceeding the depth was slightly wrong in a way that was rendered irrelevant by the first bug.

Addresses https://github.com/realm/realm-cocoa/issues/4639.

### DIFF
--- a/src/impl/collection_notifier.cpp
+++ b/src/impl/collection_notifier.cpp
@@ -98,12 +98,12 @@ bool DeepChangeChecker::check_outgoing_links(size_t table_ndx,
         return false;
     };
 
-    for (auto const& link : it->links) {
+    auto linked_object_changed = [&](OutgoingLink const& link) {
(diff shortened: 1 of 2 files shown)


austinzheng:
I suspect this change might fix https://github.com/realm/realm-object-server/issues/921 as well.

tgoyne:
I don't see how it could.

austinzheng:
No, upon closer examination it doesn't seem so. I will continue working with David to figure out what is going on.

tgoyne:
I have no idea why that one test is failing on CI. It passes for me if I run the same tests in a local docker container.

siemensikkema:
@tgoyne Is the CI problem blocking the release of this fix?

tgoyne:
#383 fixes the test which failed here, which turned out to be an unrelated pre-existing issue.

PR #370, opened 2017-02-09
Fix resetting of deferred close flag, and add more sync tests

austinzheng:
Changes:
- ~Removed support for 'deferred close'; sessions will immediately close even if waiting for the access token~
- Fixed an issue where the 'deferred close' flag was not properly reset in some cases when transitioning out of 'waiting for token'
- Removed an unused `SyncSession` API
- Added tests to ensure that `SyncSession.close()` works no matter what state the session is in

cc @nhachicha

### DIFF
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -101,6 +101,8 @@ struct SyncSession::State {
 
     virtual void nonsync_transact_notify(std::unique_lock<std::mutex>&, SyncSession&, sync::Session::version_type) const { }
 
+    // Perform any work needed to reactivate a session that is not already active.
+    // Returns true iff the session should ask the binding to get a token for `bind()`.
     virtual bool revive_if_needed(std::unique_lock<std::mutex>&, SyncSession&) const { return false; }
 
     // The user that owns this session has been logged out, and the session should take appropriate action.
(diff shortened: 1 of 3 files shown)


austinzheng:
Hmm...what if you've configured the session to wait for uploads, there are pending uploads, but then the session is closed while waiting for token? What should happen?

1. Session immediately closes (new behavior)
2. Session waits for token, gets token, then waits for pending uploads before closing (existing behavior)

I think an argument can be made for either one of these. 1 is straightforward and it doesn't prolong the session's lifetime in surprising ways. 2 is better for data integrity (since it tries harder to ensure all pending data is uploaded if the opportunity exists), but can leave a session stuck in the waiting state forever if the session never makes the `bind()` callback.

bdash:
Why would it be appropriate to ignore the session's stop policy?

austinzheng:
Maybe instead of closing the Realm immediately, we should document that creating a session (by opening a Realm) implies that the session will stay alive long enough for the connection to be established and for all pending data to be uploaded, regardless of the lifetime of the Realm object itself.

In that case, I'll have to figure out why @nhachicha is running into the issue he's seeing, and which this PR was intended to address: a session whose shared pointers are all invalidated while the network request is in flight has the 'deferred close' flag set, and if it's re-retrieved before the network response returns, the session still begins dying as soon as the response is received.

I'll keep this PR open nevertheless, since the tests are something I want to keep regardless of how I resolve the problem. I'll also file a ticket to track the issue itself (we just talked about it in chat last week.)

bdash:
The issue @nhachicha was hitting is that `deferred_close` is never reset to `false` if a session is revived after `deferred_close` was set to `true`.

austinzheng:
@nhachicha Please try running the Java binding against the tip of `master` and see if you still run into problems with the session lifespan. Thanks!

nhachicha:
@austinzheng thanks, looks like the fix works 🎉

PR #371, opened 2017-02-10
Check attached for snapshot last and first

beeender:
Results's last() and first() should return none if the row is deleted.

### DIFF
--- a/src/results.cpp
+++ b/src/results.cpp
@@ -215,7 +215,9 @@ util::Optional<RowExpr> Results::first()
         case Mode::Query:
         case Mode::TableView:
             update_tableview();
-            return m_table_view.size() == 0 ? util::none : util::make_optional(m_table_view.front());
+            return (m_table_view.size() == 0) ||
+                (m_update_policy == UpdatePolicy::Never && !m_table_view.is_row_attached(0)) ?
+                util::none : util::make_optional(m_table_view.front());
     }
     REALM_UNREACHABLE();
 }
@@ -235,7 +237,10 @@ util::Optional<RowExpr> Results::last()
         case Mode::Query:
         case Mode::TableView:
             update_tableview();
-            return m_table_view.size() == 0 ? util::none : util::make_optional(m_table_view.back());
+            auto s = m_table_view.size();
+            return (s == 0) ||
+                (m_update_policy == UpdatePolicy::Never && !m_table_view.is_row_attached(s - 1)) ?
+                util::none : util::make_optional(m_table_view.back());
     }
     REALM_UNREACHABLE();
 }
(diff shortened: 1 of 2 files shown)


src/results.cpp
```diff
@@ -215,7 +215,9 @@ util::Optional<RowExpr> Results::first()
         case Mode::Query:
         case Mode::TableView:
             update_tableview();
-            return m_table_view.size() == 0 ? util::none : util::make_optional(m_table_view.front());
+            return (m_table_view.size() == 0) ||
+                (m_update_policy == UpdatePolicy::Never && !m_table_view.is_row_attached(0)) ?
+                util::none : util::make_optional(m_table_view.front());
```
beeender: maybe it makes more sense to return a `util::make_optional(RowExpr())` instead? so it returns `none` for empty snapshot, but empty Row for deleted rows?

bdash: Returning a detached row accessor would be consistent with the behavior of `Results::get`.

tests/results.cpp
```diff
@@ -2038,6 +2038,22 @@ TEST_CASE("results: snapshots") {
...
+            table->add_empty_row();
+        });
+        Results results(r, *table);
+        auto snapshot = results.snapshot();
+        write([=] {;
+            table->clear();
+        });
+
+        REQUIRE_FALSE(snapshot.get(0).is_attached());
+        REQUIRE_FALSE(snapshot.first().value().is_attached());
```
bdash: Typically we'd write `->is_attached()` rather than `.value().is_attached()`.

src/results.cpp
```diff
@@ -215,7 +215,13 @@ util::Optional<RowExpr> Results::first()
         case Mode::Query:
         case Mode::TableView:
             update_tableview();
-            return m_table_view.size() == 0 ? util::none : util::make_optional(m_table_view.front());
+            if (m_table_view.size() == 0) {
+                return util::none;
+            } else if (m_update_policy == UpdatePolicy::Never && !m_table_view.is_row_attached(0)) {
+                return util::make_optional(RowExpr());
```
bdash: Calling `util::make_optional` isn't necessary here. It was only necessary in the ternary expression as it allows each branch needs to have the same type.

src/results.cpp
```diff
@@ -215,7 +215,13 @@ util::Optional<RowExpr> Results::first()
         case Mode::Query:
         case Mode::TableView:
             update_tableview();
-            return m_table_view.size() == 0 ? util::none : util::make_optional(m_table_view.front());
+            if (m_table_view.size() == 0) {
+                return util::none;
+            } else if (m_update_policy == UpdatePolicy::Never && !m_table_view.is_row_attached(0)) {
```
bdash: Minor nit, but in object store we typically don't use braces on single-line if statements. It's also not really necessary to have an `else` when the `if`'s body ended with a `return`.

PR #375, opened 2017-02-13
Expose integer increment functionality to bindings

austinzheng:
Need to figure out how to test this...

### DIFF
--- a/src/object_accessor.hpp
+++ b/src/object_accessor.hpp
@@ -249,13 +249,14 @@ void Object::increment_integer_impl(ContextType ctx, const Property &property, l
     size_t column = property.table_column;
     switch (property.type) {
         case PropertyType::Int:
-            if (property.is_primary)
+            if (property.is_primary) {
                 throw PropertyCannotBeIncrementedException(m_object_schema->name, property.name,
                                                            util::format("Cannot increment primary key '%1.%2'",
                                                                         m_object_schema->name,
                                                                         property.name));
-            else
+            } else {
                 m_row.add_int(column, Accessor::to_long(ctx, value));
+            }
             break;
         case PropertyType::Bool:
         case PropertyType::Float:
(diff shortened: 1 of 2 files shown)


austinzheng:
This will need to wait for Core 2.3.2 at least.

src/object_accessor.hpp
```diff
@@ -280,6 +298,46 @@ ValueType Object::get_property_value_impl(ContextType ctx, const Property &prope
...
+{
+    using Accessor = NativeAccessor<long long, ContextType>;
+
+    verify_attached();
+
+    if (!m_realm->is_in_transaction()) {
+        throw MutationOutsideTransactionException("Can only increment integer values within a transaction.");
+    }
+
+    size_t column = property.table_column;
```
jpsim: why not simplify this?

```cpp
if (property.type != PropertyType::Int) {
    throw PropertyCannotBeIncrementedException(m_object_schema->name, property.name,
                                               util::format("Cannot increment non-integer property '%1.%2'",
                                                            m_object_schema->name,
                                                            property.name));
}
else if (property.is_primary) {
    throw PropertyCannotBeIncrementedException(m_object_schema->name, property.name,
                                               util::format("Cannot increment primary key '%1.%2'",
                                                            m_object_schema->name,
                                                            property.name));
}
else {
    m_row.add_int(property.table_column, Accessor::to_long(ctx, value));
}
```

jpsim:
> This will need to wait for Core 2.3.2 at least.

Really? According to the release notes, `add_int()` was added to `Row` in 2.3.1: https://github.com/realm/realm-core/blob/master/CHANGELOG.md#231-release-notes

Core 2.3.1 is available via sync 1.1.0: https://github.com/realm/realm-sync/releases/tag/v1.1.0

austinzheng:
The functionality itself was added, but the API was a private API that the object store couldn't get to. I merged in a PR to Core after 2.3.1 was cut that added a `RowFuncs` setter, just like all the other getters and setters.

austinzheng:
This is so out of date by now it would be easier to start from scratch.

Issue #379, opened 2017-02-16
sync::Session error_handler block not triggered, causing infinite 203 error


nhachicha:
scenario:

- `SyncManager::shared().get_session` creates an ObjectStore wrapper around the `sync::Session`, let's say `std::shared_ptr<SyncSession> os_session; // (state WaitingForAccessToken)` 

- The `os_session`  is revived which cause `create_sync_session` to be called, this will create an error_handling [block](https://github.com/realm/realm-object-store/blob/1e20ebfa017a0bc03ad76e3b4e0aca44110d7c55/src/sync/sync_session.cpp#L515)

- The error handler lambda uses a `weak_ptr` obtained from the Shared Pointer `os_session`. 💣

- ObjectStore trigger the binding callback to get an `access_token`

- The binding close the SharedRealm, that releases the `RealmCoordinator`, this will call `dropped_last_reference_to_session` on `os_session` since the reference count drops to 0.

- The `dropped_last_reference_to_session` _deleter_ will store `os_session` as a `unique_ptr` in the `m_inactive_sessions` (now we have an inactive session in `WaitingForAccessToken`)
 
- Now we open a new Realm i.e `RealmCoordinator` this will call `get_session`, since we have an inactive session already, the method will create a new `shared_ptr` around the previous `os_session` [pointer](https://github.com/realm/realm-object-store/blob/1e20ebfa017a0bc03ad76e3b4e0aca44110d7c55/src/sync/sync_manager.cpp#L423) 💣
  
- Binding return with a valid `access_token`, we call `refresh_access_token` now we have an `Active` session

- Now we wait until the token expires (we don't actively refresh the token, for the sake of this use case).

- Now we try to upload a new commit using the active session. 

- Server will respond with `202` 

- The `sync::Session` error handler will trigger the defined `error_handler` [callback](https://github.com/realm/realm-object-store/blob/1e20ebfa017a0bc03ad76e3b4e0aca44110d7c55/src/sync/sync_session.cpp#L328)  however this callback will return because `weak_ptr` doesn't hold a valid reference anymore, because the first `shared_ptr` was destroyed, and we created a new one.

This will cause the `sync:Session` to never ask the binding to refresh the token, & it will keep trying to connect to the server with the old token, which will result on receiving `203` continuously

### option 1:

When `dropped_last_reference_to_session` is called instead of calling `close` 
[...]

PR #380, opened 2017-02-16
Fix an issue with dying sessions that are revived

austinzheng:
Fixes https://github.com/realm/realm-object-store/issues/379.

EDIT: On hold pending rethink of how sessions are managed.

When a session is created, a bunch of lambdas are created that capture a weak pointer from the `shared_ptr` to the `SyncSession`. Those weak pointers become invalid as soon as the conditions that make the session enter `dying` are met. If the session is revived and returned to service, those weak pointers remain invalid forever. This changes the lambdas to try and update their weak pointers if the weak pointers are invalid.

### DIFF
--- a/src/sync/sync_manager.cpp
+++ b/src/sync/sync_manager.cpp
@@ -386,15 +386,15 @@ std::shared_ptr<SyncSession> SyncManager::get_existing_active_session_locked(con
     return nullptr;
 }
 
-bool SyncManager::perform_work_on_inactive_session(const std::string& path,
-                                                   std::function<void(SyncSession&, std::unique_lock<std::mutex>&)> work) {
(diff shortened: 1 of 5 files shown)


austinzheng:
The tests are failing because the linux-sync box is screwed up.

austinzheng:
I think this is ready for preliminary review.

austinzheng:
There's a potential deadlock. Back to the drawing board...

austinzheng:
I fixed the deadlock issue. As long as the sync manager's session lock is never acquired by someone holding a session's state lock, we should be fine. (The `inactive` state transition function releases the state lock before attempting to acquire the session lock.)

austinzheng:
Superseded by https://github.com/realm/realm-object-store/pull/388/files

PR #381, opened 2017-02-20
Use exception_ptr in translation exceptions

fealebenpae:
Our current rethrowing pattern seems not to work with Windows SEH - the rethrown exception points to garbage memory. Explicit use of std::exception_ptr works, though.

### DIFF
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -105,10 +105,10 @@ Realm::Realm(Config config, std::shared_ptr<_impl::RealmCoordinator> coordinator
     m_coordinator = std::move(coordinator);
 }
 
-REALM_NOINLINE static void translate_file_exception(StringData path, bool read_only=false)
+REALM_NOINLINE static void translate_file_exception(std::exception_ptr exception, StringData path, bool read_only=false)
 {
     try {
-        throw;
+        std::rethrow_exception(exception);
     }
     catch (util::File::PermissionDenied const& ex) {
         throw RealmFileException(RealmFileException::Kind::PermissionDenied, ex.get_path(),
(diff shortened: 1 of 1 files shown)


tgoyne:
SEH is unrelated to c++ exceptions.

fealebenpae:
Yes, you're right - I've read up on the differences between `try/catch` and `__try/__except` and I understand this better now.

I guess it must be a MSVC bug then - the exact same code works great in the .NET SDK compiled with MSVC 2017, but produces weird results at runtime in the JS SDK where it's compiled with MSVC 2015.

tgoyne:
I highly doubt that it is a MSVC bug, especially since I have used `throw;` without problems with MSVC many times in the past. I suspect there are configuration problems such as mismatched compiler settings.

bdash:
And if this _is_ in fact a compiler bug, is this really the only use of `throw;` that's problematic? There are a number of other uses in core and object store.

fealebenpae:
That's entirely possible. Would you be willing to look at the `cl.exe` invocation to see if anything there seems wrong, @tgoyne?

@bdash, .NET compiles with MSVC 2017, and so does Core. The Windows tests for Object Store are built with MSVC 2017, too. It's only Node.js that needs 2015, but as @tgoyne suggested it's probably an error in the gyp changes I made in realm-js rather than a bug. The JS SDK being the only one that uses MSVC 2015 threw me off.

fealebenpae:
I managed to fix this by defining `_HAS_EXCEPTIONS=1`. I'm not at all sure what the actual problem was, or how this fixes it. I'm going to chalk it up to gyp being wonderful.

Thank you for pointing me in the right direction, @tgoyne.

tgoyne:
The build settings are probably missing `/EHsc` since I think that's what defines `_HAS_EXCEPTIONS=1`.

fealebenpae:
We have `/EHsc` - I had to enable it manually in gyp. Without it MSVC complained every time it encountered a try/catch construct.

PR #385, opened 2017-02-22
Allow completion handlers to be registered before binding

austinzheng:
Changes:
- Completion handlers registered before binding are now queued instead of being silently dropped

To do:
- [x] Tests

### DIFF
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -169,8 +169,6 @@ struct sync_session_states::WaitingForAccessToken : public SyncSession::State {
             session.m_deferred_commit_notification = util::none;
         }
 
-        // Move the session into `active`. Note that we may then immediately
-        // change state again if the session needs to be closed.
         session.advance_state(lock, active);
         if (session.m_deferred_close) {
             session.m_state->close(lock, session);
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -72,6 +72,8 @@ constexpr const char SyncError::c_recovery_file_path_key[];
 ///
 /// INACTIVE: the user owning this session has logged out, the `sync::Session`
 /// owned by this session is destroyed, and the session is quiescent.
+/// Note that a session briefly enters this state before being destroyed, but
+/// it can also enter this state and stay there if the user has been logged out.
 /// From: initial, WAITING_FOR_ACCESS_TOKEN, ACTIVE, DYING
 /// To:
 ///    * WAITING_FOR_ACCESS_TOKEN: if the session is revived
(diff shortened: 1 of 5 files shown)


austinzheng:
Getting this (or something with equivalent functionality) in is a prerequisite for the download-a-Realm-asynchronously API we're planning.

austinzheng:
On hold until I can pull the test changes out.

src/sync/sync_session.cpp
```diff
@@ -608,28 +643,35 @@ bool SyncSession::can_wait_for_network_completion() const
...
+{
+    return m_state == &State::waiting_for_access_token || m_state == &State::inactive;
+}
+
 bool SyncSession::wait_for_upload_completion(std::function<void(std::error_code)> callback)
 {
     std::unique_lock<std::mutex> lock(m_state_mutex);
-    // FIXME: instead of dropping the callback if we haven't yet `bind()`ed,
-    // save it and register it when the session `bind()`s.
     if (can_wait_for_network_completion()) {
```
bdash: Rather than switching on the session state, can we instead just delegate to a function on the state?

src/sync/sync_session.hpp
```diff
@@ -263,6 +266,17 @@ class SyncSession : public std::enable_shared_from_this<SyncSession> {
     std::string m_realm_path;
     _impl::SyncClient& m_client;
 
+    enum class CompletionWaitType {
+        Upload, Download, Sync,
+    };
+
+    // For storing wait-for-completion requests if the session isn't yet ready to handle them.
+    struct CompletionWaitPackage {
+        CompletionWaitType type;
```
bdash: Since the only thing we use this enum for is to determine which function to call, we could instead directly store a member function pointer and avoid having to switch on the enum.

austinzheng:
I addressed @bdash's feedback. I want to see if I can get rid of `wait_for_upload_completion_blocking()` completely, since it is only used in one place and by a single test.

src/sync/sync_session.hpp
```diff
@@ -49,6 +49,7 @@ namespace sync {
 class Session;
 }
 
+using SessionWaiterPointer = void(sync::Session::*)(std::function<void(std::error_code)>);
```
bdash: Since this is only used internally it'd be better to put this inside the class, alongside the `private` members.

src/sync/sync_session.hpp
```diff
@@ -211,6 +214,7 @@ class SyncSession : public std::enable_shared_from_this<SyncSession> {
     // }
 
     bool can_wait_for_network_completion() const;
+    bool can_queue_for_network_completion() const;
```
bdash: These are both gone now.

tests/results.cpp
```diff
@@ -1042,7 +1042,14 @@ TEST_CASE("notifications: sync") {
...
         // makes some writes to the Realm and bumps the version
         server.start();
-        SyncManager::shared().get_session(config.path, *config.sync_config)->wait_for_upload_completion_blocking();
+        std::condition_variable cv;
+        std::mutex wait_mutex;
+        SyncManager::shared().get_session(config.path, *config.sync_config)->wait_for_upload_completion([&](auto) {
+            cv.notify_one();
+        });
+        std::unique_lock<std::mutex> lock(wait_mutex);
+        cv.wait(lock);
```
bdash: `std::condition_variable::wait` can return spuriously, so there's the potential for this code to continue executing prior to the upload completing. I think you you can use the variant that takes a predicate, combined with the upload completion handler setting a flag, to avoid problems due to spurious unblocking.

austinzheng:
I've rebased onto master and updated this PR again, @bdash. There was also an issue with a changeset cooker variable assignment that I had to fix for things to compile.

tests/sync/session/wait_for_completion.cpp
```diff
@@ -0,0 +1,256 @@
...
+        // Register the download-completion notification
+        REQUIRE(session->wait_for_download_completion([&](auto) {
+            handler_called = true;
+        }));
+        EventLoop::main().run_until([&] { return login_handler_called == true; });
+        REQUIRE(session);
+        REQUIRE(handler_called == false);
+        // Spin the loop many times.
+        std::atomic<int> spin_count(0);
+        EventLoop::main().run_until([&] { spin_count++; return spin_count > 100; });
```
bdash: This single section of the test appears to run for over a second, predominantly due to this spinning. This pattern is repeated a few times in these new tests, adding around 5 seconds to the total runtime of all object store tests. That doubles how long it takes the object store tests to run. Is it necessary to spin for so long?

Issue #382, opened 2017-02-21
Add support for 'stale' progress notifications flag


austinzheng:
See title

PR #386, opened 2017-02-22
Fix issue where non-streaming notifiers were called with stale data

austinzheng:
Fixes https://github.com/realm/realm-object-store/issues/382

To do:
- [x] Tests. Would like to wait until #385 is merged, since that PR cleans up and refactors the sync tests

### DIFF
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -552,11 +552,15 @@ void SyncSession::NotifierPackage::update(const Progress& current_progress, bool
                                                                  : current_progress.uploadable;
 }
 
+// PRECONDITION: `update()` must first be called on the same package.
 std::function<void()> SyncSession::NotifierPackage::create_invocation(const Progress& current_progress,
                                                                       bool& is_expired) const
 {
     // It's possible for a non-streaming notifier to not yet have fresh transferrable bytes data.
     // In that case, we don't call it at all.
+    // NOTE: `update()` is always called before `create_invocation()`, and will
+    // set `captured_transferrable` on the notifier package if fresh data has
+    // been received and the package is for a non-streaming notifier.
     if (!is_streaming && !captured_transferrable)
         return [](){ };
 
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -519,16 +519,17 @@ void SyncSession::handle_error(SyncError error)
 }
 
 void SyncSession::handle_progress_update(uint64_t downloaded, uint64_t downloadable,
-                                         uint64_t uploaded, uint64_t uploadable)
+                                         uint64_t uploaded, uint64_t uploadable, bool is_fresh)
 {
     std::vector<std::function<void()>> invocations;
     {
         std::lock_guard<std::mutex> lock(m_progress_notifier_mutex);
         m_current_progress = Progress{uploadable, downloadable, uploaded, downloaded};
+        m_latest_progress_data_is_fresh = is_fresh;
 
         for (auto it = m_notifiers.begin(); it != m_notifiers.end();) {
             auto& package = it->second;
-            package.update(*m_current_progress);
+            package.update(*m_current_progress, is_fresh);
 
             bool should_delete = false;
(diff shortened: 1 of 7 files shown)


src/sync/sync_session.cpp
```diff
@@ -473,18 +474,22 @@ void SyncSession::handle_progress_update(uint64_t downloaded, uint64_t downloada
...
                                                                  : current_progress.uploadable;
 }
 
-std::function<void()> SyncSession::NotifierPackage::create_invocation(const Progress& current_progress, bool& is_expired) const
+std::function<void()> SyncSession::NotifierPackage::create_invocation(const Progress& current_progress,
+                                                                      bool& is_expired) const
 {
-    REALM_ASSERT(is_streaming || captured_transferrable);
+    // It's possible for a non-streaming notifier to not yet have fresh transferrable bytes data.
+    // In that case, we don't call it at all.
```
bdash: This seems a little surprising to me. What's the reasoning behind not ever calling the callback?

austinzheng: The idea is that we don't call the callback until sync reports that it has numbers that came from the server recently ('fresh', not 'stale').

The problem was that people were opening synced Realms and immediately registering notifier blocks. Sync would happily report 0 out of 0 bytes before it actually got any info from the server, so the notifier blocks would say "your download has finished". Then the server would report fresh data and sync would report (e.g.) 0 out of 20 bytes, but by then it was too late.

bdash: That much I understand. But this seems to result in the block _never_ being called, rather than waiting for concrete information about the progress to be available?

austinzheng: Oh, I see.

Every time `create_invocation()` is called, it's preceded by a call to `update()`. `update()` is responsible for checking if the newest data is fresh or not, and then updating the `captured_transferrable` value. So the first time fresh data is sent down `captured_transferrable` is set to a non-`none` value, and the block should be called.

That being said, the fact that I had to explain the above means that the code needs to be refactored for clarity, so let me think about how best to move things around and make my intent explicit...

bdash: It's probably a result of my reading only the diff rather than looking at the surrounding context (in this case, the caller of the method). I'd incorrectly thought this was only called once when registering the notification, rather than each time new progress information is available. My mistake.

austinzheng:
I need to figure out how to test this. Unfortunately, since sync downloads some metadata when a session is established (and therefore sets the `Progress` data and marks it as fresh, it's not really possible to test the changes in this PR directly. I have at least two options: add a test-only method to `SyncSession` to reset the progress state member variables so we can test them, or add a testing flag to `SyncSession` so that we can avoid registering the progress callback on the `sync::Session` when creating it in a testing context. Both options are pretty awful.

tests/sync/session/progress_notifications.cpp
```diff
@@ -24,7 +24,128 @@
...
+    SECTION("non-streaming notifiers don't run until fresh data is available") {
+        session->register_progress_notifier([&](auto xferred, auto xferable) {
+            transferred = xferred;
+            transferrable = xferable;
+            ++call_count;
+        }, SyncSession::NotifierType::download, false);
+
+        // Send a stale notification
+        SyncSession::OnlyForTesting::handle_progress_update(*session, 10, 200, 1, 100, false);
+        std::atomic<int> spin_count(0);
```
bdash: There's no need for this to be atomic, since it can only ever be updated from the current thread. The state we update from within sync callbacks needs to be atomic since those callbacks run on the sync client's thread.

tests/sync/session/progress_notifications.cpp
```diff
@@ -24,7 +24,128 @@
...
+        EventLoop::main().run_until([&] {
+            ++spin_count;
+            return spin_count.load() > 2; 
+        });
```
bdash: I think you can make the count internal to the lambda by doing something like:
```c++
EventLoop::main().run_until([count=0] mutable { return ++count > 2; });
```

austinzheng:
`macos-sync` job is sporadically getting stuck while cloning `core`. I'm going to go ahead and merge.

PR #387, opened 2017-02-23
Reorganize sync session tests

austinzheng:
Changes:
- Moved sync session tests into their own directory
- Extracted some helper functions into their own set of utils files
- Extracted the progress notification tests into their own file

No tests were added or removed.

The goal of this change is to make it easier to add coverage for `SyncSession`, and to begin better organizing the tests by what they cover.

Prereq for https://github.com/realm/realm-object-store/pull/385


tests/sync/session/session_util.cpp
```diff
@@ -0,0 +1,31 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#include "session_util.hpp"
+
+namespace realm {
+    bool sessions_are_active(const SyncSession& session)
+    {
+        return session.state() == SyncSession::PublicState::Active;
```
bdash: Not sure there's any reason to put these in a .cpp file when the header pulls in all of the other headers this needs anyway.

tests/sync/session/session_util.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+#include "sync/sync_config.hpp"
+#include "sync/sync_manager.hpp"
+#include "sync/sync_session.hpp"
+#include "sync/sync_user.hpp"
+
+using namespace realm;
+using namespace realm::util;
+
+namespace realm {
```
bdash: Why put these helpers in the `realm` namespace when `sync_session` / `full_control_sync_session` aren't?

tests/sync/session/session_util.hpp
```diff
@@ -0,0 +1,99 @@
...
+    return sessions_are_inactive(session) && sessions_are_inactive(s...);
+}
+
+}
+
+// Identical to `sync_session(...)`, but takes a bind-session callback that is
+// passed directly into the configuration. This allows, for example, a session
+// that remains stalled in the 'waiting for token' state.
+template <typename BindCallback, typename ErrorHandler>
+std::shared_ptr<SyncSession> full_control_sync_session(SyncServer& server, std::shared_ptr<SyncUser> user, const std::string& path,
```
bdash: `full_control_sync_session` is a bit of a weird name.

tests/sync/session/session_util.hpp
```diff
@@ -0,0 +1,99 @@
...
+// session back out to the test, or configure the session more precisely.
+template <typename FetchAccessToken, typename ErrorHandler>
+std::shared_ptr<SyncSession> sync_session(SyncServer& server, std::shared_ptr<SyncUser> user, const std::string& path,
+                                          FetchAccessToken&& fetch_access_token, ErrorHandler&& error_handler,
+                                          SyncSessionStopPolicy stop_policy=SyncSessionStopPolicy::AfterChangesUploaded,
+                                          std::string* on_disk_path=nullptr,
+                                          util::Optional<Schema> schema=none,
+                                          Realm::Config* out_config=nullptr)
+{
+    return full_control_sync_session(server, std::forward<std::shared_ptr<SyncUser>>(user), path,
```
bdash: You want `std::move` here, not `std::forward`.

tests/sync/session/session_util.hpp
```diff
@@ -0,0 +1,99 @@
...
+        [&, fetch_access_token=std::forward<FetchAccessToken>(fetch_access_token)](const auto& path, const auto& config, auto session) {
+            auto token = fetch_access_token(path, config.realm_url);
+            session->refresh_access_token(std::move(token), config.realm_url);
+        }, 
+        std::forward<ErrorHandler>(error_handler),
+        stop_policy, on_disk_path, std::forward<util::Optional<Schema>>(schema), out_config);
```
bdash: Ditto about `std::move` rather than `std::forward`.

tests/sync/session/session_util.hpp
austinzheng: The arguments aren't rvalue references, right.

bdash: More specifically, `std::forward` is only meaningful when used with [forwarding references](http://en.cppreference.com/w/cpp/language/reference#Forwarding_references), which are a subset of rvalue references.

tests/sync/session/session_util.hpp
austinzheng: It's an awful name, let me think about better alternatives.

austinzheng:
Looks like the macos sync job is stuck cloning `core`. Given that the tests have never failed on my machine I'm inclined to merge this and deal with any CI fallout later.

PR #388, opened 2017-02-23
Rework the ownership of SyncSession so that it's always managed by a shared_ptr

### LINKED ISSUE #379: sync::Session error_handler block not triggered, causing infinite 203 error

bdash:
Ownership of a `SyncSession` was previously managed by a `shared_ptr` for as long as there were external references to it. When all external references were dropped and the `SyncSession` became inactive, it was transferred to a `unique_ptr` held by the `SyncManager`. This transfer proved problematic as it meant that the `weak_ptr`s that `SyncSession` used internally to refer to itself from callbacks expired, making it impossible to correctly handle the callbacks.

The new approach is for `SyncSession` to always be managed by a `shared_ptr`. To detect when external references to the `SyncSession` have gone away, we tie the lifetime of the external references to a separate object (by way of `shared_ptr`'s aliasing constructor) and use the destruction of this object to signal that the `SyncSession` has become inactive. Since the `SyncSession` is always managed by a `shared_ptr` any `weak_ptr`s continue to work as expected for the complete duration of the session's lifecycle.

This has the secondary benefit of removing the need to track active and inactive sessions separately in `SyncManager`.

Fixes #379.

### DIFF
--- a/src/sync/sync_manager.hpp
+++ b/src/sync/sync_manager.hpp
@@ -154,20 +154,17 @@ private:
 
     mutable std::unique_ptr<_impl::SyncClient> m_sync_client;
 
-    // Protects m_sessions
-    mutable std::mutex m_session_mutex;
-
     // Protects m_file_manager and m_metadata_manager
     mutable std::mutex m_file_system_mutex;
     std::unique_ptr<SyncFileManager> m_file_manager;
     std::unique_ptr<SyncMetadataManager> m_metadata_manager;
 
-    // Active sessions are sessions which the client code holds a strong
-    // reference to. When the last strong reference is released, the session is
-    // moved to inactive sessions. Inactive sessions are promoted back to active
-    // sessions until the session itself calls unregister_session to remove
-    // itself from inactive sessions once it's done with whatever async cleanup
-    // it needs to do.
+    // Protects m_sessions
+    mutable std::mutex m_session_mutex;
+
+    // Map of sessions by path name.
+    // Sessions remove themselves from this map by calling `unregister_session` once they're
+    // inactive and have performed any necessary cleanup work.
     std::unordered_map<std::string, std::shared_ptr<SyncSession>> m_sessions;
 };
 
--- a/src/sync/sync_manager.hpp
+++ b/src/sync/sync_manager.hpp
@@ -121,7 +121,6 @@ public:
 
 private:
     using ReconnectMode = sync::Client::ReconnectMode;
-    void dropped_last_reference_to_session(SyncSession*);
 
     // Stop tracking the session for the given path if it is inactive.
     // No-op if the session is either still active or in the active sessions list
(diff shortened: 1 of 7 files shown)


bdash:
Not sure why you say Linux CI is broken? https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-object-store/detail/PR-388/5/pipeline/34 shows it having passed.

austinzheng:
I restarted the job after making that comment, and this time it passed.

src/sync/sync_manager.hpp
```diff
@@ -170,8 +168,7 @@ friend class SyncSession;
     // sessions until the session itself calls unregister_session to remove
```
bdash: This comment needs to be revised.

PR #393, opened 2017-03-02
fix for core 2.3.2 change

alazier:
Since object-accessor is not yet tested in object-store this slipped through the cracks.

Needed because of this change: https://github.com/realm/realm-core/commit/4322f25731a9eb44ae133b0a7a57bb547f83d377

@bdash

### DIFF
--- a/src/object_accessor.hpp
+++ b/src/object_accessor.hpp
@@ -193,9 +193,11 @@ void Object::set_property_value_impl(ContextType ctx, const Property &property, 
                 m_row.set_string(column, string_value);
             break;
         }
-        case PropertyType::Data:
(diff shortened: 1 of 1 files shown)


ironage:
This is a case where this constructor was actually being used safely so there was no bug here. But the extra safety that not supporting the r-value constructor gives us is worth a small change like this in my opinion. Thanks for fixing it @alazier  👍

bdash:
@tgoyne has a branch in progress adding tests for a lot of the object accessor code.

alazier:
Can we merge this?

bdash:
Sure. In future please be sure to add the `S:Review` label to make it obvious that the PR is waiting for review.

PR #394, opened 2017-03-02
Add basic object accessor tests and fix a few bugs

tgoyne:
The object accessor code was never updated to reflect sync's requirements (no setting the PK after creating objects and default values have to be explicitly flagged for them to be merged correctly).

Quite a bit more is still needed here (this doesn't fully cover the object accessor code and even the parts that are covered aren't particularly thorough), but it's better than nothing.

### DIFF
--- a/src/object_accessor.hpp
+++ b/src/object_accessor.hpp
@@ -28,9 +28,11 @@
 #include "schema.hpp"
 #include "util/format.hpp"
 
-#include <string>
 #include <realm/link_view.hpp>
+#include <realm/util/assert.hpp>
 #include <realm/table_view.hpp>
+
+#include <string>
 
 namespace realm {
 //
@@ -109,6 +111,9 @@ void Object::set_property_value_impl(ContextType ctx, const Property &property, 
     verify_attached();
     m_realm->verify_in_write();
 
+    if (property.is_primary)
+        throw std::logic_error("Cannot modify primary key after creation");
+
     size_t column = property.table_column;
     if (property.is_nullable && Accessor::is_null(ctx, value)) {
         if (property.type == PropertyType::Object) {
@@ -125,10 +130,7 @@ void Object::set_property_value_impl(ContextType ctx, const Property &property, 
             m_row.set_bool(column, Accessor::to_bool(ctx, value));
             break;
         case PropertyType::Int:
-            if (property.is_primary)
-                m_row.set_int_unique(column, Accessor::to_long(ctx, value));
-            else
-                m_row.set_int(column, Accessor::to_long(ctx, value));
+            m_row.set_int(column, Accessor::to_long(ctx, value));
             break;
         case PropertyType::Float:
             m_row.set_float(column, Accessor::to_float(ctx, value));
@@ -137,11 +139,8 @@ void Object::set_property_value_impl(ContextType ctx, const Property &property, 
             m_row.set_double(column, Accessor::to_double(ctx, value));
             break;
         case PropertyType::String: {
-            auto string_value = Accessor::to_string(ctx, value);
-            if (property.is_primary)
-                m_row.set_string_unique(column, string_value);
-            else
-                m_row.set_string(column, string_value);
+            auto str = Accessor::to_string(ctx, value);
+            m_row.set_string(column, str);
             break;
         }
         case PropertyType::Data: {
@@ -249,8 +248,14 @@ Object Object::create(ContextType ctx, SharedRealm realm, const ObjectSchema &ob
         if (row_index == realm::not_found) {
             row_index = table->add_empty_row();
             created = true;
-            Object object(realm, object_schema, table->get(row_index));
-            object.set_property_value_impl(ctx, *primary_prop, primary_value, try_update);
+            if (primary_prop->type == PropertyType::Int)
+                table->set_int_unique(primary_prop->table_column, row_index, Accessor::to_long(ctx, primary_value));
+            else if (primary_prop->type == PropertyType::String) {
+                auto value = Accessor::to_string(ctx, primary_value);
+                table->set_string_unique(primary_prop->table_column, row_index, value);
+            }
+            else
+                REALM_UNREACHABLE();
         }
         else if (!try_update) {
             throw std::logic_error(util::format("Attempting to create an object of type '%1' with an existing primary key value.", object_schema.name));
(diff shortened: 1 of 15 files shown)


src/object_accessor.hpp
```diff
@@ -128,91 +94,75 @@ struct MutationOutsideTransactionException : public std::logic_error {
...
+            table.set_timestamp(column, row, Accessor::to_timestamp(ctx, value), is_default);
             break;
         case PropertyType::Object: {
-            if (Accessor::is_null(ctx, value)) {
-                m_row.nullify_link(column);
-            }
-            else {
-                m_row.set_link(column, Accessor::to_object_index(ctx, m_realm, value, property.object_type, try_update));
-            }
+            table.set_link(column, row, Accessor::to_object_index(ctx, m_realm, value, property.object_type, try_update), is_default);
```
bdash: This appears to expect `to_object_index` to return `npos` if the value is null. The test implementation of `to_object_index` doesn't appear to do this, nor does JS's existing implementation.

tests/object.cpp
```diff
@@ -18,23 +18,99 @@
...
+    static util::Any from_string(Defaults*, StringData v) { return std::string(v); }
+    static util::Any from_timestamp(Defaults*, Timestamp v) { return v; }
+    static util::Any from_list(Defaults*, List v) { return v; }
+    static util::Any from_results(Defaults*, Results v) { return v; }
+    static util::Any from_object(Defaults*, Object v) { return v; }
+
+    static bool is_null(Defaults*, util::Any& v) { return !v.has_value(); }
+    static util::Any null_value(Defaults*) { return {}; }
+
+    static size_t to_existing_object_index(Defaults*, SharedRealm, util::Any &) { return 0; }
```
bdash: It'd be clearer to have this be something like `REALM_ASSERT(false && "Not implemented");`, so it's obvious to anyone adding new tests that this needs to be fleshed out before it can be tested.

tests/object.cpp
```diff
@@ -18,23 +18,99 @@
...
+    static size_t to_object_index(Defaults* d, SharedRealm realm, util::Any& value, std::string const& object_type, bool update)
+    {
+        if (auto object = any_cast<Object>(&value)) {
```
bdash: If's with a single line body are brace-less elsewhere in this PR.

tests/util/any.hpp
```diff
@@ -0,0 +1,155 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+#include <typeinfo>
+#include <type_traits>
+#include <stdexcept>
+
+namespace realm {
+namespace util {
+
+// A naive implementation of C++17's std::Any
```
bdash: `std::any`

src/object_accessor.hpp
tgoyne: `value` can't be null here (there's a check for that before the switch statement).

tests/util/any.hpp
```diff
@@ -0,0 +1,155 @@
...
+// This does not perform the small-object optimization or make any particular
+// attempt at being performant
+class Any final {
```
bdash: This is a very clear implementation 😍

tests/object.cpp
```diff
@@ -175,4 +277,219 @@ TEST_CASE("object") {
...
+        REQUIRE(row.get_int(2) == 5);
+        REQUIRE(row.get_float(3) == 6.6f);
+        REQUIRE(row.get_double(4) == 3.3);
+        REQUIRE(row.get_string(5) == "hello");
+        REQUIRE(row.get_binary(6) == BinaryData("olleh", 5));
+        REQUIRE(row.get_timestamp(7) == Timestamp(10, 20));
+    }
+
+    SECTION("create throws for missing values if there is no default") {
+        REQUIRE_THROWS(create(AnyDict{
```
bdash: Is it worth testing that the exception that is thrown is the exception we expect?

src/object_accessor.hpp
```diff
@@ -128,91 +94,75 @@ struct MutationOutsideTransactionException : public std::logic_error {
...
-    verify_attached();
-
-    if (!m_realm->is_in_transaction()) {
-        throw MutationOutsideTransactionException("Can only set property values within a transaction.");
-    }
-
+    auto& table = *m_row.get_table();
     size_t column = property.table_column;
+    size_t row = m_row.get_index();
     if (property.is_nullable && Accessor::is_null(ctx, value)) {
```
bdash: Should we throw if the property isn't nullable but we're passed a null value?

tests/util/test_file.cpp
```diff
@@ -95,9 +95,10 @@ SyncTestFile::SyncTestFile(const SyncConfig& sync_config)
     schema_mode = SchemaMode::Additive;
 }
 
-SyncTestFile::SyncTestFile(SyncServer& server)
+SyncTestFile::SyncTestFile(SyncServer& server, std::string name)
```
bdash: The changes in test_file.{cpp,hpp} don't appear to be related to this PR?

src/object_accessor.hpp
tgoyne: The code currently expects `NativeAccessor` to handle checking for type mismatches, including throwing an appropriate error when `Accessor::to_foo()` is called on a null value.

tests/util/test_file.cpp
tgoyne: This is used in the test that verifies that default values are merged correctly.

tests/util/any.hpp
```diff
@@ -0,0 +1,155 @@
...
+}
+} // namespace util
+} // namespace realm
+
+namespace std {
+inline void swap(realm::util::Any& lhs, realm::util::Any& rhs) noexcept
+{
+    lhs.swap(rhs);
+}
+} // nsmespace std
```
bdash: namespace

tests/util/test_file.cpp
bdash: So it is. I think GitHub must've decided to collapse the new file of tests when I did my find in page to check for where this was used. Sorry about that.

src/object_accessor.hpp
bdash: Ok, after looking over JS's implementation I now see that they do type checking as you describe.

tgoyne:
I hit the linker failure that's occurring on CI once, but then it worked after that. I suspect the fact that the coverage-instrumented test binary is now over 2 GB is related.

tgoyne:
Cutting the size of the tests binary from 2GB to 1.1GB has seemingly fixed the linker failures (or it just happened to work this tie).

PR #396, opened 2017-03-03
Properly distinguish between admin users and users wrapping admin tokens

austinzheng:
Required for https://github.com/realm/realm-cocoa/pull/4699

Changes:
- Replaced 'is admin' flag with an enum member to distinguish between users that wrap an admin token, and ones that are marked by the server as admins
- Updated user metadata to store the 'is admin' flag
- Updated APIs exposed to bindings to allow a user's admin status to be specified or read out
- Updated tests and user creation code

To do:
- [x] Once object store tests aren't broken, ensure all tests pass
- [x] Write a test to ensure migrating from metadata v0 to v1 works fine
- [x] Write a few more tests to exercise `is_admin` updating logic

### DIFF
--- a/src/sync/impl/sync_metadata.cpp
+++ b/src/sync/impl/sync_metadata.cpp
@@ -37,6 +37,7 @@ static const char * const c_sync_marked_for_removal = "marked_for_removal";
 static const char * const c_sync_identity = "identity";
 static const char * const c_sync_auth_server_url = "auth_server_url";
 static const char * const c_sync_user_token = "user_token";
+static const char * const c_sync_user_is_admin = "user_is_admin";
 
 static const char * const c_sync_fileActionMetadata = "FileActionMetadata";
 static const char * const c_sync_original_name = "original_name";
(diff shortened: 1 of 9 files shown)


austinzheng:
This is done (except for a few more tests to write), but I can't run the tests for object store successfully on my machine right now (with or without this commit).

alazier:
Why don't we just make token admin users behave the same as non-token users?

austinzheng:
It doesn't make sense to persist a token user since such a user doesn't _actually_ represent a user, but is rather a wrapper around an admin token intended to be used with our APIs that do expect a user.

austinzheng:
Okay, this is ready for review.

src/sync/sync_user.hpp
```diff
@@ -64,18 +74,21 @@ friend class SyncSession;
...
     }
 
     std::string identity() const
     {
         return m_identity;
     }
 
+    // A flag which can be set by a binding to indicate whether the Realm Object Server
+    // indicates that this user has server administration privileges.
+    bool marked_as_object_server_admin;
```
bdash: Is this used anywhere?

austinzheng: No, that looks like cruft left over from the original version of the design. I'll remove it and make sure nothing else like it is still hanging around.

src/sync/sync_user.hpp
```diff
@@ -31,6 +31,16 @@ namespace realm {
...
 
+enum class SyncUserAdminMode {
+    // User is not an admin.
+    None,
+    // User wraps an admin token. Such users use the tokens they were configured
+    // with to directly open sessions, and do not make network requests.
+    WrapsAdminToken,
+    // User was marked by the authentication server as a user with admin privileges.
+    MarkedAsAdmin,
+};
```
bdash: The terminology in this enum feels awkward. Maybe using an enum like `enum ObjectServerAccessLevel { Normal, Admin }` and having `SyncUser` store an `Optional<AccessLevel>` would be clearer? The `none` case would represent the user not being authenticated via the object server (i.e., having used an admin access token).

Or maybe it should just be `enum AccessLevel { Normal, Admin }`, and admin tokens / admin users should both be `Admin`, and the authenticated-with-object-server-vs-admin-token distinction tracked separately?

austinzheng: I decided to go with something similar to your second approach.

The sync user has APIs to check whether it's an admin, and to update its admin status (as a result of the auth server sending us a response).

We also make a distinction now between users that are or aren't saved in the metadata database. Right now, the only non-persisted users are admin-token-wrapping users, so the code treats non-persisted users as admin users. Once we move the responsibility of token-wrapping users into the binding (as we discussed earlier), we can make persistence/non-persistence completely orthogonal to admin status, or we can remove support for non-persisted users altogether.

I'm happy to make further changes if this approach isn't satisfactory.

src/sync/sync_manager.hpp
```diff
@@ -104,7 +106,7 @@ friend class SyncSession;
     std::shared_ptr<SyncUser> get_user(const std::string& identity,
                                        std::string refresh_token,
                                        util::Optional<std::string> auth_server_url=none,
-                                       bool is_admin=false);
+                                       bool should_persist=true);
```
austinzheng: This should be renamed `is_ephemeral` and default to `false`, since if we reverse the polarity of this bool silently we'll just end up introducing bugs into bindings.

src/sync/impl/sync_metadata.cpp
```diff
@@ -37,52 +37,95 @@ static const char * const c_sync_marked_for_removal = "marked_for_removal";
...
+}
+
+}
+
+#pragma mark - Sync metadata manager
+
+SyncMetadataManager::SyncMetadataManager(std::string path,
+                                         bool should_encrypt,
+                                         util::Optional<std::vector<char>> encryption_key,
+                                         int schema_version)
```
bdash: It looks like the support for multiple schema versions is only needed for sake of testing the upgrade path. IMO the logic to create Realms with an older schema version should live in the unit tests, or we should test against prepackaged Realms created using the old code.

src/sync/sync_manager.hpp
```diff
@@ -104,7 +106,7 @@ friend class SyncSession;
     std::shared_ptr<SyncUser> get_user(const std::string& identity,
                                        std::string refresh_token,
                                        util::Optional<std::string> auth_server_url=none,
-                                       bool is_admin=false);
+                                       bool do_not_persist=false);
```
bdash: Is it intentional that the sense of this was inverted vs what `SyncUser`'s constructor takes?

src/sync/sync_user.cpp
```diff
@@ -177,7 +191,8 @@ void SyncUser::register_session(std::shared_ptr<SyncSession> session)
         case State::Active:
             // Immediately ask the session to come online.
             m_sessions[path] = session;
-            if (m_is_admin) {
+            // FIXME: `SyncUser`s shouldn't even wrap admin tokens; the bindings should do that.
+            if (!m_should_persist) {
```
bdash: It feels a bit odd to have a flag that controls whether or not we should persist the user be what determine whether or not the token is treated as an admin token or not.

src/sync/sync_user.cpp
```diff
@@ -26,20 +26,24 @@ namespace realm {
...
 : m_state(State::Active)
 , m_server_url(server_url.value_or(""))
-, m_is_admin(is_admin)
+, m_should_persist(should_persist)
 , m_refresh_token(std::move(refresh_token))
 , m_identity(std::move(identity))
 {
-    if (!is_admin) {
-        SyncManager::shared().perform_metadata_update([this, server_url=std::move(server_url)](const auto& manager) {
+    bool is_admin = false;
```
bdash: Why bother with this rather than just setting `m_is_admin` directly?

src/sync/impl/sync_metadata.cpp
```diff
@@ -37,52 +37,95 @@ static const char * const c_sync_marked_for_removal = "marked_for_removal";
...
+        case 0:
+            config.schema = make_v0_schema();
+            break;
+        case 1:
+            config.schema = make_v1_schema();
+            break;
+        default:
+            REALM_TERMINATE("Unsupported metadata schema version specified.");
+    }
     config.schema_mode = SchemaMode::Additive;
```
bdash: Seems like we shouldn't be using additive here, since we'll likely need to perform migrations when we change the schema?

src/sync/sync_user.hpp
```diff
@@ -64,13 +64,18 @@ friend class SyncSession;
...
     bool is_admin() const
     {
         return m_is_admin;
     }
 
+    // Specify whether the user has administrator privileges.
+    // Note that this is an internal flag meant for bindings to communicate information
+    // originating from the server. It is *NOT* possible to unilaterally change a user's
+    // administrator status from the client through this or any other API.
+    void update_admin_status(bool);
```
bdash: Is there a reason this isn't just `set_is_admin`?

austinzheng: I had originally wanted to make it clear that changing the bool wouldn't actually change the user's admin status on the server, but I think that concern is overkill.

austinzheng:
I've addressed all feedback so far. The failing CI tests are due to #417, which isn't directly related to my change.

src/sync/impl/sync_metadata.cpp
```diff
@@ -37,41 +37,40 @@ static const char * const c_sync_marked_for_removal = "marked_for_removal";
...
+    Property p = {name, PropertyType::String};
+    p.is_indexed = true;
+    p.is_primary = true;
+    return p;
+}
 
-    Realm::Config config;
-    config.path = std::move(path);
-    config.schema = Schema{
+inline Schema make_schema()
```
bdash: `inline` isn't necessary here.

PR #398, opened 2017-03-05
Integrating SyncManager with Java

nhachicha:
This PR is the continued work (partially reviewed) in https://github.com/realm/realm-object-store/compare/814beb5a1e96f0bb72cf78e206b2e710ac79e217...3300b386474af5e96569778a65d27c97e297a8d1#diff-b7147bdb5e4732db8177678d8cb49c13 that will help integrate the ObjectStore's implementation of SyncManager/SyncSession with Java https://github.com/realm/realm-java/pull/4214

Major change is probably the addition of `SyncManager::get_existing_session` method, that will return an existing session without creating a new one if it doesn't exist. 

This is needed as we only communicate with the ObjectStore using the `realm_path` we don't have to allocate a dynamic shared_ptr to the session in order to work with it over the JNI lifecycle.

Refer to the full Java implementation to see how we use it.

The rest of the changes are mostly helper/wrapper for Java and I believe were previously discussed with @cmelchior on his original [PR](https://github.com/realm/realm-java/pull/3900/)


src/sync/sync_manager.cpp
```diff
@@ -387,6 +393,12 @@ std::shared_ptr<SyncSession> SyncManager::get_existing_session_locked(const std:
     return it->second;
 }
 
+std::shared_ptr<SyncSession> SyncManager::get_existing_session(const std::string& path) const
+{
+    std::lock_guard<std::mutex> lock(m_session_mutex);
+    return get_existing_session_locked(path);
```
bdash: This needs to return an external reference (`SyncSession::external_reference`). I suspect it also needs to attempt to revive the session by calling `SyncUser::register_session` so it doesn't return a session that is in the process of dying.

src/sync/sync_config.hpp
```diff
@@ -105,6 +105,21 @@ struct SyncConfig {
     std::function<SyncBindSessionHandler> bind_session_handler;
     std::function<SyncSessionErrorHandler> error_handler;
     std::shared_ptr<ChangesetTransformer> transformer;
+#if __GNUC__ < 5
+// GCC 4.9 does not support C++14 braced-init with NSDMIs
```
bdash: I'm a little confused by this as there are no NSDMIs in this class. Why is it really necessary?

src/sync/impl/sync_client.hpp
```diff
@@ -33,11 +45,16 @@ struct SyncClient {
...
                ReconnectMode reconnect_mode = ReconnectMode::normal,
-               bool verify_ssl = true)
+               bool verify_ssl = true,
+               ClientThreadListener* client_thread_listener = nullptr)
     : client(make_client(*logger, reconnect_mode, verify_ssl)) // Throws
     , m_logger(std::move(logger))
-    , m_thread([this] {
+    , m_thread([this, client_thread_listener=std::move(client_thread_listener)] {
+                if (client_thread_listener)
+                client_thread_listener->on_client_thread_ready();
```
bdash: The indentation in this lambda is all horked.

src/sync/impl/sync_client.hpp
```diff
@@ -33,11 +45,16 @@ struct SyncClient {
 
     SyncClient(std::unique_ptr<util::Logger> logger,
                ReconnectMode reconnect_mode = ReconnectMode::normal,
-               bool verify_ssl = true)
+               bool verify_ssl = true,
+               ClientThreadListener* client_thread_listener = nullptr)
     : client(make_client(*logger, reconnect_mode, verify_ssl)) // Throws
     , m_logger(std::move(logger))
-    , m_thread([this] {
+    , m_thread([this, client_thread_listener=std::move(client_thread_listener)] {
```
bdash: Moving a pointer isn't very useful.

src/sync/sync_manager.cpp
nhachicha: Would something like the following work?

```C++
std::lock_guard<std::mutex> lock(m_session_mutex);
    if (auto session = get_existing_session_locked(path)) {
        session->m_config.user->register_session(session);
        return session->external_reference();
    }
    return nullptr;
```

bdash: It might. I'm not clear what semantics Java needs though. Can you give me some background on how you intend to use this?

src/sync/sync_manager.hpp
```diff
@@ -141,6 +147,7 @@ friend class SyncSession;
     // FIXME: Should probably be util::Logger::Level::error
     util::Logger::Level m_log_level = util::Logger::Level::info;
     SyncLoggerFactory* m_logger_factory = nullptr;
+    realm::ClientThreadListener* m_client_thread_listener;
```
bdash: This needs to be initialized to `nullptr`. Otherwise it'll start off pointing to garbage and we'll crash when `SyncClient` finds it non-null and tries to dereference it.

src/sync/impl/sync_client.hpp
```diff
@@ -24,7 +24,19 @@
...
+public:
+    // This method is called just before the client is started
+    virtual void on_client_thread_ready() = 0;
+
+    // This method is called just before the client thread is being killed
+    // The client should be stopped at this point.
+    virtual void on_client_thread_closing() = 0;
+};
+
+    namespace _impl {
```
bdash: Indentation

src/sync/impl/sync_client.hpp
```diff
@@ -33,12 +45,17 @@ struct SyncClient {
...
+               ClientThreadListener *client_thread_listener = nullptr)
+            : client(make_client(*logger, reconnect_mode, verify_ssl)) // Throws
+            , m_logger(std::move(logger)),
+              m_thread([this, client_thread_listener] {
+                  if (client_thread_listener)
+                      client_thread_listener->on_client_thread_ready();
+                  client.run();
+                  if (client_thread_listener)
+                      client_thread_listener->on_client_thread_closing();
+              }) // Throws
```
bdash: This indentation is still very weird. It should look something like:
```c++
SyncClient(std::unique_ptr<util::Logger> logger,
           ReconnectMode reconnect_mode = ReconnectMode::normal,
           bool verify_ssl = true,
           ClientThreadListener *client_thread_listener = nullptr)
: client(make_client(*logger, reconnect_mode, verify_ssl)) // Throws
, m_logger(std::move(logger)),
  m_thread([this, client_thread_listener] {
    if (client_thread_listener)
        client_thread_listener->on_client_thread_ready();
    client.run();
    if (client_thread_listener)
        client_thread_listener->on_client_thread_closing();
}) // Throws
{
}
```

This makes me notice that the calls to the client thread listener are not exception safe. If `client.run()` throws an exception, which it can, then the thread will be closed but `on_client_thread_closing` will never be called.

src/sync/sync_manager.cpp
```diff
@@ -447,5 +463,6 @@ std::unique_ptr<SyncClient> SyncManager::create_sync_client() const
     }
     return std::make_unique<SyncClient>(std::move(logger),
                                         m_client_reconnect_mode,
-                                        m_client_validate_ssl);
+                                        m_client_validate_ssl,
+                                        std::move(m_client_thread_listener));
```
bdash: No reason to `std::move` here.

src/sync/sync_manager.hpp
```diff
@@ -20,6 +20,7 @@
 #define REALM_OS_SYNC_MANAGER_HPP
 
 #include "shared_realm.hpp"
+#include "sync/impl/sync_client.hpp"
```
bdash: We avoid including `impl` headers from within public headers. The declaration of `ClientThreadListener` should be moved to a public header.

src/sync/sync_manager.hpp
bdash: No need to explicitly qualify this with `realm::`.

src/sync/sync_manager.hpp
```diff
@@ -78,6 +79,9 @@ friend class SyncSession;
 
     void set_log_level(util::Logger::Level) noexcept;
     void set_logger_factory(SyncLoggerFactory&) noexcept;
+    // Option callback invoked when the thread responsible for running the Sync Client is started
+    // and the client has been created (but not started).
+    void set_client_thread_listener(realm::ClientThreadListener& listener);
```
bdash: No need to explicitly qualify this with `realm::`.

src/sync/sync_manager.hpp
```diff
@@ -78,6 +79,9 @@ friend class SyncSession;
 
     void set_log_level(util::Logger::Level) noexcept;
     void set_logger_factory(SyncLoggerFactory&) noexcept;
+    // Option callback invoked when the thread responsible for running the Sync Client is started
```
bdash: s/Option/Optional/

src/sync/impl/sync_client.hpp
```diff
@@ -24,7 +24,19 @@
 #include <thread>
 
 namespace realm {
-namespace _impl {
+
+// Interface for bindings interested in the lifecycle of the Sync Client thread.
+class ClientThreadListener {
```
bdash: This should probably be prefixed with `Sync`.

src/sync/impl/sync_client.hpp
```diff
@@ -24,7 +24,19 @@
...
+public:
+    // This method is called just before the client is started
+    virtual void on_client_thread_ready() = 0;
```
bdash: `did_create_client_thread`?

src/sync/impl/sync_client.hpp
```diff
@@ -24,7 +24,19 @@
...
+
+    // This method is called just before the client thread is being killed
+    // The client should be stopped at this point.
+    virtual void on_client_thread_closing() = 0;
```
bdash: `will_destroy_client_thread`?

src/sync/impl/sync_client.hpp
bdash: From looking at the Java PR that builds on top of this, it doesn't look like there's anything specific to the sync client in this callback. It might be preferable to give this a name that's more general. If I understand correctly this pair of calls needs to be made on any thread created by the object store that may call into Java?

src/binding_callback_thread_observer.hpp
```diff
@@ -0,0 +1,39 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#ifndef REALM_OBJECT_STORE_BINDING_CALLBACK_THREAD_OBSERVER_HPP
```
bdash: We typically use `OS` in place of `OBJECT_STORE` in these defines.

src/binding_callback_thread_observer.hpp
```diff
@@ -0,0 +1,39 @@
...
+#define REALM_OBJECT_STORE_BINDING_CALLBACK_THREAD_OBSERVER_HPP
+
+namespace realm {
+    // Interface for bindings interested in registering callbacks before/after the ObjectStore thread runs.
```
bdash: We don't typically indent the entire body of a namespace like this.

src/binding_callback_thread_observer.hpp
```diff
@@ -0,0 +1,39 @@
...
+        // This method is called just before the thread is started
+        virtual void did_create_thread() = 0;
+
+        // This method is called just before the thread is being destroyed
+        virtual void will_destroy_thread() = 0;
+    };
+
+    extern BindingCallbackThreadObserver* g_sync_client_thread_listener;
+
+    void set_sync_client_thread_listener(BindingCallbackThreadObserver& listener);
```
bdash: I'm not sure there's any reason to have a setter when the variable itself is public. The only reason that comes to mind would be if we move the variable into the `_impl` namespace and leave the setter in the public namespace. Not sure there's any reason to bother with that though. If for some reason we do decide to keep the setter, it should be renamed to match the new class name.

src/sync/impl/sync_client.hpp
```diff
@@ -33,11 +35,23 @@ struct SyncClient {
 
     SyncClient(std::unique_ptr<util::Logger> logger,
                ReconnectMode reconnect_mode = ReconnectMode::normal,
-               bool verify_ssl = true)
+               bool verify_ssl = true,
+               BindingCallbackThreadObserver *client_thread_listener = nullptr)
```
bdash: Why take this as an argument rather than just having the lambda below use the global variable?

src/sync/impl/sync_client.hpp
```diff
@@ -33,11 +35,23 @@ struct SyncClient {
...
-    , m_thread([this] {
-        client.run();
+    , m_thread([this, client_thread_listener] {
+        if (client_thread_listener)
+            client_thread_listener->did_create_thread();
+        try {
+            client.run();
+        }
+        catch (...) {
+            if (client_thread_listener)
```
bdash: It'd be cleaner and more idiomatic to use a scope exit handler for this purpose rather than explicit try / catch.

```c++
if (client_thread_listener)
    client_thread_listener->did_create_thread();

auto will_destroy_thread = util::make_scope_exit([&]() noexcept {
    if (client_thread_listener)
        client_thread_listener->will_destroy_thread();
});

client.run(); // Throws
```

The destructor of `will_destroy_thread` will be invoked when the function returns or if the stack is unwound due to an exception, ensuring that the function is called in both cases.

src/sync/sync_manager.hpp
```diff
@@ -119,6 +119,8 @@ friend class SyncSession;
     // calling this method.
     void reset_for_testing();
 
+    std::shared_ptr<SyncSession> get_existing_session(const std::string& path) const;
```
bdash: This should be moved to alongside `get_session`.

src/binding_callback_thread_observer.hpp
```diff
@@ -0,0 +1,39 @@
...
+}
+
+#endif //REALM_OBJECT_STORE_BINDING_CALLBACK_THREAD_OBSERVER_HPP
```
bdash: Whitespace after the `//`?

src/binding_callback_thread_observer.cpp
```diff
@@ -0,0 +1,29 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#include "binding_callback_thread_observer.hpp"
+
+namespace realm {
+
+    BindingCallbackThreadObserver* g_sync_client_thread_listener;
```
bdash: We don't typically indent the entire body of a namespace like this.

src/sync/impl/sync_client.hpp
```diff
@@ -20,10 +20,13 @@
 #define REALM_OS_SYNC_CLIENT_HPP
 
 #include <realm/sync/client.hpp>
+#include "binding_callback_thread_observer.hpp"
 
 #include <thread>
+#include <realm/util/scope_exit.hpp>
```
bdash: Typically includes from this project are grouped together and come first, followed by those from core and sync, then anything from the system.

PR #399, opened 2017-03-07
Fix a move-related issue

austinzheng:
`schema` is captured from the enclosing scope by reference, and the only condition in which it's safe to move a value out of it is if the lambda returns `true` (and the enclosing function immediately returns). Otherwise, we attempt to use `schema` afterwards, causing issues since it no longer contains the values we expect.

- [x] This change needs a test


austinzheng:
I added a test. The test passes with the fix and fails without.

PR #401, opened 2017-03-08
read schema from realm when opening a realm from a custom version

alazier:
Right now we can end up with a realm with a schema referencing non-existent objects which can lead to assertion failures.

@bdash


bdash:
@tgoyne, this feels related to the issue you're looking at of missing schema changes when no read transactions are open. Do you think your solution will address this problem in a more general way?

tgoyne:
It's certainly related, but the main thing I'm working on is better tracking of when exactly we need to re-read the schema, as doing it unconditionally has an unreasonable performance impact in some cases. If that's not a problem for what this function is being used for then this change is fine.

alazier:
I'll add some tests.

PR #403, opened 2017-03-09
Avoid beginning unnecessary write transactions in Additive schema mode

tgoyne:
`schema.compare()` can return a non-empty array and yet still not require a write transaction in Additive mode if all of the differences are that the Realm file has extra columns or index changes that wouldn't be applied anyway due to the schema version not changing.

### DIFF
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -299,16 +299,17 @@ void Realm::update_schema(Schema schema, uint64_t version, MigrationFunction mig
             case SchemaMode::ResetFile:
                 return reset_file_if_needed(schema, version, required_changes);
 
-            case SchemaMode::Additive:
-                if (required_changes.empty()) {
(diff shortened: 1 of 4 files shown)


src/shared_realm.cpp
```diff
@@ -300,14 +300,13 @@ void Realm::update_schema(Schema schema, uint64_t version, MigrationFunction mig
                 return reset_file_if_needed(schema, version, required_changes);
 
             case SchemaMode::Additive:
-                if (required_changes.empty()) {
-                    if (version == m_schema_version) {
-                        set_schema(std::move(schema), version);
-                        return true;
-                    }
+                if (ObjectStore::verify_valid_additive_changes(required_changes, version > m_schema_version))
```
bdash: It might be clearer to hoist `version > m_schema_version` into a variable so it can be given a name. It's not obvious without looking at the parameter name what this would mean.

PR #406, opened 2017-03-17
Fix a bunch of issues with Realms with non-constant schema

tgoyne:
This attempts to generalize all of the schema handling logic to fully work with the idea that a file's schema may be different at different transaction versions, and different Realm instances may be working with different subsets of the schema at the same time.

RealmCoordinator no longer tries to track "the" schema for a file, and instead has a cache of the most recent schema seen, along with the range of transaction versions that it's known to be valid for. This is used to minimize how often we have to introspect the file, as that's an expensive process.

Opening a Realm with different subsets of tables or columns (in additive or read-only mode) on different threads now fully works. This is mostly for the sake of giving operations sensible semantics: it's unlikely that any of the bindings need this, but it's easier to test than trying to figure out what the exact combination of things actually can be done via the bindings and only testing those (and probably missing a few).

SharedRealm now re-reads (usually from the cache) and validates the schema when beginning a new read transaction rather than simply assuming it's still valid, which fixes a crash when it's changed by another process or sync while a Realm instance existed with no read transaction.

The value of the schema returned by `Realm::schema()` has now been made consistent: it used to be either the current schema on disk, an out-of-date version of the schema on disk (but with up-to-date table_columns) or the user-supplied schema (with up-to-date table_columns) depending on what happened to last update it. It is now either exactly the user-supplied schema (with table columns populated), or exactly the schema on disk at the realm's transaction version if no schema has been applied. This fixes some crashes due to property indexes changing when they weren't supposed to, and is a halfway step towards a proper dynamic schema mode.

Since this results in the schema being validated more often, it includes some optimizations to avoid regressions in the obj-c performance tests.

### DIFF
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -63,6 +63,8 @@ Realm::Realm(Config config, std::shared_ptr<_impl::RealmCoordinator> coordinator
         // otherwise get the schema from the group
         m_schema_version = ObjectStore::get_schema_version(read_group());
         m_schema = ObjectStore::schema_from_group(read_group());
+        if (coordinator && m_schema_version != ObjectStore::NotVersioned)
+            coordinator->cache_schema(m_schema, m_schema_version);
 
         if (m_shared_group) {
             m_schema_transaction_version = m_shared_group->get_version_of_current_transaction().version;
@@ -219,7 +221,7 @@ void Realm::set_schema(Schema schema, uint64_t version)
 {
     schema.copy_table_columns_from(m_schema);
     m_schema = schema;
-    m_coordinator->update_schema(schema, version);
+    m_coordinator->set_schema_version(version);
 }
 
 bool Realm::read_schema_from_group_if_needed()
@@ -373,7 +375,8 @@ void Realm::update_schema(Schema schema, uint64_t version, MigrationFunction mig
     if (!in_transaction) {
         commit_transaction();
     }
-    m_coordinator->update_schema(m_schema, version);
+
+    m_coordinator->set_schema_version(version);
 }
 
 void Realm::add_schema_change_handler()
@@ -383,8 +386,9 @@ void Realm::add_schema_change_handler()
             auto new_schema = ObjectStore::schema_from_group(read_group());
             auto required_changes = m_schema.compare(new_schema);
             ObjectStore::verify_valid_additive_changes(required_changes);
+            m_schema_version = ObjectStore::get_schema_version(read_group());
+            m_coordinator->cache_schema(new_schema, m_schema_version);
             m_schema.copy_table_columns_from(new_schema);
-            m_coordinator->update_schema(m_schema, m_schema_version);
         });
     }
 }
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -330,6 +330,34 @@ Schema Realm::get_full_schema()
     return actual_schema;
 }
 
+void Realm::set_schema_subset(Schema schema)
+{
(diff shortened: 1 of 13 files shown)


ironage:
Could this fix https://github.com/realm/realm-core/issues/2520 ? Looking at the stack trace seems to indicate that the object-store has an incorrect index for a link column.

tgoyne:
It's possible but it seems unlikely.

src/shared_realm.cpp
```diff
@@ -267,86 +256,116 @@ bool Realm::reset_file_if_needed(Schema& schema, uint64_t version, std::vector<S
...
 
         // Beginning the write transaction may have advanced the version and left
         // us with nothing to do if someone else initialized the schema on disk
-        if (schema_changed && no_changes_required()) {
-            cancel_transaction();
-            return;
+        if (m_new_schema) {
+            actual_schema = std::move(*m_new_schema);
+            m_new_schema = util::none;
+            m_coordinator->cache_schema(actual_schema, m_schema_version,
```
bdash: It looks like this is doing roughly the same work as `cache_new_schema()`. Is it intentional that we're not updating `m_schema_transaction_version` like `cache_new_schema()` does? Can we find a way to call `cache_new_schema()` here rather than repeating its logic?

src/shared_realm.cpp
```diff
@@ -666,16 +717,19 @@ uint64_t Realm::get_schema_version(const Realm::Config &config)
 
 void Realm::close()
 {
+    // Order matters here: if we're the last Realm instance for this path, we
+    // need to tear down the coordinator before destroying our shared group
+    // to maintain the coordinator's invariant that there be a continuous session
```
bdash: I'm not clear which invariant this is maintaining. What do you mean by "continuous session"?

src/shared_realm.hpp
```diff
@@ -295,8 +295,13 @@ class Realm : public std::enable_shared_from_this<Realm> {
 
     uint64_t m_schema_version;
     Schema m_schema;
+    util::Optional<Schema> m_new_schema;
     uint64_t m_schema_transaction_version = -1;
 
+    // FIXME: this should be a Dynamic schema mode instead, but only once
```
bdash: What's remaining before a real dynamic schema mode could be added?

src/shared_realm.cpp
tgoyne: With out-of-file transaction logs the transaction version would reset if the sharedgroup is closed entirely and reopened, which isn't handled by this code. With in-realm transaction logs the version is persisted, so I guess this no longer applies.

src/shared_realm.hpp
tgoyne: The biggest thing is just taking the time to think about how it should work and write tests. Probably need some sort of schema change notification for the binding.

src/shared_realm.cpp
tgoyne: "session" here is referring to SharedGroup's session concept.

src/shared_realm.cpp
bdash: GitHub decided this comment is outdated for some reason, but it still appears to be applicable to the most recent version of the PR.

src/shared_realm.hpp
kneth: How does it affect bindings which expose dynamic Realms? Java does, and I can't tell if anyone is using it for more than manual migrations.

tgoyne: Dynamic realms currently don't work, so at the minimum this shouldn't make things any worse.

PR #407, opened 2017-03-17
Introduce the Windows Universal platform in CMake

fealebenpae:


jpsim:
Is there any way to set up a CI environment to test this?

fealebenpae:
I'm not sure we can test the UWP binary stand-alone without a container app, much like on iOS. We could verify that it builds, though.

PR #408, opened 2017-03-21
Add reconnect() method to SyncManager

stel:
This PR adds `reconnect()` method to `SyncManager` that forces sync client to reconnect without a delay after connection is back again.

refs https://github.com/realm/realm-object-store/pull/289
/cc @cmelchior, @bdash, @austinzheng, @fealebenpae

### DIFF
--- a/src/sync/impl/sync_client.hpp
+++ b/src/sync/impl/sync_client.hpp
@@ -52,6 +52,10 @@ struct SyncClient {
     }) // Throws
     {
     }
+    
+    void cancel_reconnect_delay() {
+        client.cancel_reconnect_delay();
+    }
 
     void stop()
     {
--- a/src/sync/sync_manager.cpp
+++ b/src/sync/sync_manager.cpp
@@ -267,6 +267,14 @@ bool SyncManager::client_should_validate_ssl() const noexcept
     return m_client_validate_ssl;
 }
 
+void SyncManager::reconnect()
+{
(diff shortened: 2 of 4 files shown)


stel:
I'm actually not sure what's the best way to test it here because there are no any properties exposed from `sync::Client` that could be checked. @kvap, @kspangsege please let me know if you have any suggestions.

For now I'll add a check that just calls `reconnect()` to make sure it doesn't crash at least.

kvap:
Set reconnect mode to `never`, then cause an error which closes the connection. Then call your `reconnect()` and `wait until download completed`.

kvap:
BTW, the quick reconnection has been implemented as part of heartbeats feature. Just calling `cancel_reconnect_delay()` will cause an urgent PING to be sent. If the PONG does not return in some seconds, the reconnection will be initiated.

stel:
@kvap the problem is that there are no network connections during the tests and there is no reason to test if `sync::Client::cancel_reconnect_delay()` works properly here, we just need to make sure that it's called from `SyncManager::reconnect()`.

kvap:
@stel provide a mock client object which tracks whether `cancel_reconnect_delay` is called.

stel:
It's not an option as well because sync client is created by `SyncManager` itself.

kvap:
@stel Another crazy idea: put a custom `Logger` in `Client::Config` and make it expect `"Reconnect delay canceled"` or `"Will emit a ping"`.

PR #410, opened 2017-03-21
Permission APIs

alazier:


austinzheng:
Agreed. I'm experimenting with some stuff, I might remove it and just use `sole` on Apple platforms too if it doesn't pan out.

src/sync/sync_permission.cpp
```diff
@@ -0,0 +1,295 @@
...
+#include <sole.hpp>
+#endif
+
+#include "property.hpp"
+
+using namespace realm;
+
+namespace {
+
+// TODO: this should already exist somewhere.
```
bdash: The way this is used it could be replaced by `std::string`.

src/sync/sync_permission.cpp
```diff
@@ -0,0 +1,295 @@
...
+}
+
+Permission::Condition::~Condition()
+{
+    switch (type) {
+        case Type::UserId:
+            user_id.std::basic_string<char>::~basic_string<char>();
+            break;
+        case Type::KeyValue:
+            std::get<0>(key_value).std::basic_string<char>::~basic_string<char>();
```
bdash: This should invoke `key_value`'s destructor rather than manually invoking the destructor on each of its members.

src/sync/sync_permission.hpp
```diff
@@ -0,0 +1,129 @@
...
+    struct Condition {
+        enum class Type { UserId, KeyValue };
+        Type type;
+
+        union {
+            std::string user_id;
+            std::pair<std::string, std::string> key_value;
+        };
+
+        Condition(std::string id) : type(Type::UserId), user_id(id) {}
```
bdash: Move `id` into `user_id`?

src/sync/sync_permission.hpp
```diff
@@ -0,0 +1,129 @@
...
+    NotificationToken async(std::function<void (std::exception_ptr)> target) { return m_results->async(target); }
+
+    // Create a new PermissionResults by further filtering or sorting this PermissionResults
+    PermissionResults filter(Query&& q) const;
+
+    // Create with a results - should be private
+    PermissionResults(std::unique_ptr<Results> results) : m_results(std::move(results)) {}
+
+private:
+    std::unique_ptr<Results> m_results;
```
bdash: Why is this a `unique_ptr`?

src/sync/sync_permission.cpp
```diff
@@ -0,0 +1,295 @@
...
+}
+
+size_t PermissionResults::size() {
+    REALM_ASSERT(m_results->size());
+    return m_results->size() - 1;
+}
+
+#pragma mark - Permission
+
+Permission::Permission(const Permission& p)
```
bdash: It looks like this could just be defaulted?

src/sync/sync_permission.cpp
```diff
@@ -0,0 +1,295 @@
...
+, condition(p.condition)
+{ }
+
+Permission::Permission(std::string path, AccessLevel access, Condition condition)
+: path(std::move(path))
+, access(std::move(access))
+, condition(std::move(condition))
+{ }
+
+Permission& Permission::operator=(const Permission& p)
```
bdash: Ditto about being defaulted.

src/sync/sync_permission.cpp
```diff
@@ -0,0 +1,295 @@
...
+{
+    if (&p != this) {
+        path = p.path;
+        access = p.access;
+        condition = p.condition;
+    }
+    return *this;
+}
+
+Permission::Condition& Permission::Condition::operator=(const Permission::Condition& c)
```
bdash: This needs to destroy any existing state before assigning to our member variables, otherwise we have a leak.

src/sync/sync_permission.cpp
austinzheng: Yeah, I think once there are implementations for `Permission::Condition`'s copy constructor/assignment operator, the equivalents for `Permission` should be auto generated.

src/impl/object_accessor_impl.hpp
```diff
@@ -0,0 +1,91 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+#ifndef REALM_OS_OBJECT_ACCESSOR_IMPL_HPP
+#define REALM_OS_OBJECT_ACCESSOR_IMPL_HPP
+
+#include "object_accessor.hpp"
+#include "util/any.hpp"
+
+namespace realm {
+    using AnyDict = std::map<std::string, util::Any>;
```
bdash: We typically avoid indenting the entirety of the body of a namespace.

src/sync/sync_permission.cpp
```diff
@@ -0,0 +1,276 @@
...
+    REALM_ASSERT(m_results->size());
+    return m_results->size() - 1;
+}
+
+#pragma mark - Permission
+
+Permission::Condition& Permission::Condition::operator=(const Permission::Condition& c)
+{
+    if (&c != this) {
+        type.Condition::Type::~Type();
```
bdash: There's no need to invoke the destructor of an enum type (it's just an integer, after all).

austinzheng:
Oh, that makes sense. I was going to ask if there were different rules for RAII inside copy constructors, but what you said makes more sense.

src/sync/sync_permission.hpp
alazier: We need a pointer allocated with `new` in the JS binding.

bdash: Why?… And why does that need to impact on what the object store does? Can JS not move-construct another instance on the heap?

alazier: The JavascriptCore uses a `void *` to store data in custom objects which is how we store the `PermissionResults` internally. The code assumes all of these objects were allocated using `new` but I think we have the ability to add a custom destructor. How would you suggest move-constructing a Results?

I was also having issues at one point with the move constructor for some object store types where the `m_realm` member or notifications weren't getting copied which is why I was using unique_ptrs in some spots. Don't remember if that was the case here though.

bdash: > How would you suggest move-constructing a Results?
```c++
new Results(std::move(some_results));
```
[The JS binding already does this when dealing with object store types](https://github.com/realm/realm-js/blob/92c10c469c5d92c4044d953e9167afa477dc687c/src/js_results.hpp#L98-L101) for the reason you mention.

> I was also having issues at one point with the move constructor for some object store types where the m_realm member or notifications weren't getting copied which is why I was using unique_ptrs in some spots.

Notifications will seem to disappear if you _copy_ a `Results` (or a `List` or `Object`) then discard the original as they're tied to a specific instance. If you move the `Results` they'll come along for the ride, though.

src/sync/sync_permission.cpp
```diff
@@ -0,0 +1,218 @@
...
+    };
+}
+
+// MARK: - Permissions
+
+// A box that stores a value and an associated notification token.
+// The point of this type is to keep the notification token alive
+// until the value can be properly processed or handled.
+template<typename T>
+struct LifetimeProlongingBox {
```
alazier: I hate the name Box. Prefer wrapper. Or open to other alternatives.

austinzheng: I don't have any preference for what the name should be. If you pick a replacement I'll change it.

alazier: So this is really used to keep notifications alive until they fire. So it should be called `NotificationWrapper` or something similar. And `value` should be `object`

src/sync/sync_permission.cpp
```diff
@@ -0,0 +1,218 @@
...
+    };
+}
+
+// MARK: - Permissions
+
+// A wrapper that stores a value and an associated notification token.
+// The point of this type is to keep the notification token alive
+// until the value can be properly processed or handled.
+template<typename T>
+struct NotificationWrapper {
```
bdash: Thoughts on doing something like https://gist.github.com/bdash/ede60c67ac3ca65d43985b70dd8fdf7c to make this more ergonomic?

src/sync/sync_permission.cpp
```diff
@@ -0,0 +1,218 @@
...
+        { "updatedAt",  Timestamp(0, 0) },
+        { "userId",     permission.condition.user_id },
+        { "realmUrl",   realm_url },
+        { "mayRead",    permission.access != Permission::AccessLevel::None },
+        { "mayWrite",   permission.access == Permission::AccessLevel::Write || permission.access == Permission::AccessLevel::Admin },
+        { "mayManage",  permission.access == Permission::AccessLevel::Admin },
+    }, false);
+    realm->commit_transaction();
+
+    auto block = [object_wrapper, callback=std::move(callback)](CollectionChangeSet, std::exception_ptr ex) mutable {
```
bdash: This could do with a comment explaining what it's doing. It takes a close reading to figure out when the notification is unregistered.

src/sync/sync_permission.cpp
```diff
@@ -0,0 +1,218 @@
...
+                                  PermissionResultsCallback callback,
+                                  const ConfigMaker& make_config)
+{
+    auto realm = Permissions::permission_realm(user, make_config);
+    auto session = SyncManager::shared().get_session(realm->config().path, *realm->config().sync_config);
+    auto results_wrapper = std::make_shared<NotificationWrapper<Results>>();
+    auto table = ObjectStore::table_for_object_type(realm->read_group(), "Permission");
+    results_wrapper->object = Results(std::move(realm), *table);
+
+    auto async = [results_wrapper, callback=std::move(callback)](auto ex) mutable {
```
bdash: This could do with a comment explaining what it's doing. It takes a close reading to figure out when the notification is unregistered.

src/sync/sync_permission.cpp
```diff
@@ -0,0 +1,218 @@
...
+        Results& res = results_wrapper->object;
+        if (ex) {
+            callback(nullptr, ex);
+            results_wrapper.reset();
+        } else if (res.size() > 0) {
```
bdash: Writing this as an early return would make it the comment about the presence of `__management` easier to follow, IMO.

src/sync/sync_permission.cpp
```diff
@@ -0,0 +1,218 @@
...
+            // We monitor the raw results. The presence of a `__management` Realm indicates
+            // that the permissions have been downloaded (hence, we wait until size > 0).
+            TableRef table = ObjectStore::table_for_object_type(res.get_realm()->read_group(), "Permission");
+            size_t col_idx = table->get_descriptor()->get_column_index("path");
+            auto query = !(table->column<StringData>(col_idx).ends_with("/__permission")
+                           || table->column<StringData>(col_idx).ends_with("/__management"));
+            callback(std::make_unique<PermissionResults>(res.filter(std::move(query))), nullptr);
```
bdash: Why is this heap-allocating the `PermissionResults`?

src/sync/sync_permission.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+SharedRealm Permissions::management_realm(std::shared_ptr<SyncUser> user, const ConfigMaker& make_config)
+{
+    // FIXME: maybe we should cache the management Realm on the user, so we don't need to open it every time.
+    const auto realm_url = std::string("realm") + user->server_url().substr(4) + "/~/__management";
+    Realm::Config config = make_config(user, std::move(realm_url));
+    config.sync_config->stop_policy = SyncSessionStopPolicy::Immediately;
+    config.schema = Schema{
+        { "PermissionChange", {
+            { "id",             PropertyType::String,   "", "", true, true, false   },
```
bdash: These boolean literals are impossible to understand without referring to the type's constructor (and the type isn't even named here, making that harder too).

src/sync/sync_permission.hpp
```diff
@@ -0,0 +1,149 @@
...
+public:
+    // Consumers of these APIs need to pass in a method which creates a Config with the proper
+    // SyncConfig and associated callbacks, as well as the path and other parameters.
+    using ConfigMaker = std::function<Realm::Config(std::shared_ptr<SyncUser>, std::string url)>;
+
+    // Callback used to asynchronously vend a `PermissionResults` object.
+    using PermissionResultsCallback = std::function<void(std::unique_ptr<PermissionResults>, std::exception_ptr)>;
+
+    // Asynchronously retrieve the permissions for the provided user.
+    static void get_permissions(std::shared_ptr<SyncUser>, PermissionResultsCallback, const ConfigMaker&);
```
bdash: Why do these functions take a `shared_ptr<SyncUser>`? AFAICT they don't care about the ownership of the `SyncUser` and so could be taking a const reference to a `SyncUser` instead.

src/util/any.hpp
```diff
@@ -16,6 +16,9 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
+#ifndef REALM_ANY_HPP
```
bdash: This should probably be `REALM_OS_UTIL_ANY_HPP`

bdash: We probably want to add tests for `Any` now that we're relying on its behavior in production code, rather than just tests.

src/sync/sync_permission.hpp
```diff
@@ -0,0 +1,149 @@
...
+            // The permission is applied to a single user based on their user ID
+            UserId,
+            // The permission is based on any user that meets a criterion specified by key/value.
+            KeyValue,
+        };
+        Type type;
+
+        // FIXME: turn this back into a union type
+        std::string user_id = "";
+        std::pair<std::string, std::string> key_value = std::make_pair("", "");
```
bdash: Why are these initialized like this? The default constructors should be sufficient, no?

src/sync/sync_permission.hpp
```diff
@@ -0,0 +1,149 @@
...
+};
+
+class Permissions {
+public:
+    // Consumers of these APIs need to pass in a method which creates a Config with the proper
+    // SyncConfig and associated callbacks, as well as the path and other parameters.
+    using ConfigMaker = std::function<Realm::Config(std::shared_ptr<SyncUser>, std::string url)>;
+
+    // Callback used to asynchronously vend a `PermissionResults` object.
+    using PermissionResultsCallback = std::function<void(std::unique_ptr<PermissionResults>, std::exception_ptr)>;
```
bdash: Why does this take a unique pointer to a `PermissionResults`? If it's optional because there may have been an error then it should use `util::Optional` instead.

src/sync/sync_permission.hpp
```diff
@@ -0,0 +1,149 @@
...
+namespace realm {
+
+class Permissions;
+class SyncUser;
+class Object;
+
+// Permission object used to represent a user permission.
+// Permission objects can be passed into or returned by various permissions
+// APIs. They are immutable objects.
+struct Permission {
```
bdash: These type names are very general as is. We should likely prefix them with `Sync`.

austinzheng:
I'll take care of these review comments today & tomorrow (inc. adding tests for `Any`).

src/sync/sync_permission.hpp
austinzheng: Some code I was doing to try something out that I left in by accident. It's gone now!

src/sync/sync_permission.hpp
austinzheng: Unfortunately the sync user is passed to the `ConfigMaker`, and at least the Cocoa config maker requires a shared pointer (if we got rid of it here we'd end up using `shared_from_this` anyways). In any case the ConfigMaker should be removed in a future iteration, at which point we can get rid of the shared pointer and just use a reference.

src/sync/sync_permission.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+protected:
+    Results m_results;
+};
+
+class Permissions {
+public:
+    // Consumers of these APIs need to pass in a method which creates a Config with the proper
+    // SyncConfig and associated callbacks, as well as the path and other parameters.
+    using ConfigMaker = std::function<Realm::Config(std::shared_ptr<SyncUser>, std::string url)>;
```
tgoyne: Looking at the code using this I don't understand why it isn't just passing in a Config object.

src/sync/sync_permission.cpp
```diff
@@ -0,0 +1,218 @@
...
+{
+    const auto realm_url = user->server_url() + permission.path;
+    auto realm = Permissions::management_realm(std::move(user), make_config);
+    auto object_wrapper = std::make_shared<NotificationWrapper<Object>>();
+    CppContext context;
+
+    // Write the permission object.
+    realm->begin_transaction();
+    object_wrapper->object = Object::create<util::Any>(&context, realm, *realm->schema().find("PermissionChange"), AnyDict{
+        { "id",         util::uuid_string() },
```
tgoyne: None of the existing objectstore code puts spaces inside {}.

src/sync/sync_permission.cpp
```diff
@@ -0,0 +1,218 @@
...
+                                    const ConfigMaker& make_config)
+{
+    permission.access = Permission::AccessLevel::None;
+    set_permission(std::move(user), std::move(permission), std::move(callback), make_config);
+}
+
+SharedRealm Permissions::management_realm(std::shared_ptr<SyncUser> user, const ConfigMaker& make_config)
+{
+    // FIXME: maybe we should cache the management Realm on the user, so we don't need to open it every time.
+    const auto realm_url = std::string("realm") + user->server_url().substr(4) + "/~/__management";
```
tgoyne: `util::format("realm/%1/~/__management", user->server_url().substr(4))`

src/sync/sync_permission.cpp
```diff
@@ -0,0 +1,218 @@
...
+    config.schema_version = 0;
+    auto shared_realm = Realm::get_shared_realm(std::move(config));
+    user->register_management_session(config.path);
+    return shared_realm;
+}
+
+SharedRealm Permissions::permission_realm(std::shared_ptr<SyncUser> user, const ConfigMaker& make_config)
+{
+    // FIXME: maybe we should cache the permission Realm on the user, so we don't need to open it every time.
+    const auto realm_url = std::string("realm") + user->server_url().substr(4) + "/~/__permission";
```
tgoyne: `util::format("realm/%1/~/__permission", user->server_url().substr(4))`

src/sync/sync_permission.cpp
```diff
@@ -0,0 +1,218 @@
...
+        CppContext context;
+        auto status_code = obj.get_property_value<util::Any>(&context, "statusCode");
+        if (status_code.has_value()) {
+            auto code = any_cast<long long>(status_code);
+            std::exception_ptr exc_ptr = nullptr;
+            if (code) {
+                auto status = obj.get_property_value<util::Any>(&context, "statusMessage");
+                std::string error_str = (status.has_value()
+                                         ? any_cast<std::string>(status)
+                                         : std::string("Error code: ") + util::to_string(code));
```
tgoyne: `util::format("Error code: %1", code)`

src/sync/sync_permission.cpp
austinzheng: The subclassing solution is elegant and clever, I like it.

src/sync/sync_permission.hpp
austinzheng: Unfortunately it turns out that the bindings are responsible for setting a specific login callback on the `SyncConfig`, since the object store delegates the task of making a network request to fetch the refresh token to the client. That architecture is pretty horrible and we are planning to move responsibility for that network request into the object store in the future, at which point we should be able to get rid of the `ConfigMaker` too.

austinzheng:
I am aware the build is broken on Linux. I'm bringing up another VM to debug it and will fix it by tomorrow morning. Development on macOS platforms should not be affected.

src/sync/sync_permission.hpp
austinzheng: I'm going to punt on this until we can figure out how to address namespacing object store more generally.

austinzheng:
I'm happy to continue addressing comments in follow-up PRs, if anyone has any.

PR #414, opened 2017-03-23
Pass the Realm encryption key to the sync Session

fealebenpae:
Sync recently acquired the ability to sync encrypted realm files. We need to start passing the encryption key specified in `Realm::Config` to `sync::Session`. This is my attempt to do so. Please help me improve it and come up with tests.

/cc @austinzheng @bdash @kristiandupont


src/impl/realm_coordinator.cpp
```diff
@@ -81,7 +81,14 @@ void RealmCoordinator::create_sync_session()
     if (m_sync_session)
         return;
 
-    m_sync_session = SyncManager::shared().get_session(m_config.path, *m_config.sync_config);
+    util::Optional<std::array<char, 64>> encryption_key;
+    if (m_config.encryption_key.data()) {
+        std::array<char, 64> key;
+        std::copy_n(m_config.encryption_key.begin(), key.size(), key.begin());
+        encryption_key = std::move(key);
```
tgoyne: You can just copy the key directly into the optional without a temporary:

        encryption_key = std::array<char, 64>{};
        std::copy(m_config.encryption_key.begin(), m_config.encryption_key.end(),
                  encryption_key->begin());

src/impl/realm_coordinator.cpp
```diff
@@ -81,7 +81,14 @@ void RealmCoordinator::create_sync_session()
...
+    }
+
+    m_sync_session = SyncManager::shared().get_session(m_config.path, *m_config.sync_config, std::move(encryption_key));
```
tgoyne: std::move() on a std::array doesn't do anything useful (the data for std::array is stored inline and not in a separate stack allocation).

fealebenpae: Don't we need to move the `util::Optional` precisely because we want not to incur the cost of copying the array?

tgoyne: `Optional<std::array<char, 64>>` is just a blob of 65 bytes, so "moving" it is still memcpy(). There's no way to have the newly constructed std::array just point to the data from the old std::array because std::array doesn't point at heap-allocated data.

src/sync/sync_session.hpp
```diff
@@ -275,6 +275,7 @@ class SyncSession : public std::enable_shared_from_this<SyncSession> {
     size_t m_death_count = 0;
 
     SyncConfig m_config;
+    util::Optional<std::array<char, 64>> m_realm_encryption_key;
```
bdash: This sorta feels like it should be a member of `SyncConfig`, since that holds all of the information `SyncSession` needs. The downside there is that the binding would need to configure the encryption key in both `Realm::Config` and `SyncConfig` when opening a synced Realm.

fealebenpae: Yeah, I was exploring both options, but it feels weird whichever way I try it. Okay, I'll move it to `SyncConfig`.

src/impl/realm_coordinator.cpp
```diff
@@ -81,13 +81,7 @@ void RealmCoordinator::create_sync_session()
...
         return;
 
-    util::Optional<std::array<char, 64>> encryption_key;
-    if (m_config.encryption_key.data()) {
-        encryption_key = std::array<char, 64>();
-        std::copy_n(m_config.encryption_key.begin(), encryption_key->size(), encryption_key->begin());
-    }
-
-    m_sync_session = SyncManager::shared().get_session(m_config.path, *m_config.sync_config, std::move(encryption_key));
+    m_sync_session = SyncManager::shared().get_session(m_config.path, *m_config.sync_config);
```
bdash: We should probably validate that `m_config.encryption_key` and `m_config.sync_config.realm_encryption_key` are equal. If not here, then in `RealmCoordinator::set_config`.

PR #416, opened 2017-03-24
update to sync 1.4.2

jpsim:


src/sync/sync_session.cpp
```diff
@@ -423,6 +423,7 @@ void SyncSession::handle_error(SyncError error)
             case ProtocolError::reuse_of_session_ident:
             case ProtocolError::bound_in_other_session:
             case ProtocolError::bad_message_order:
+            case ProtocolError::pong_timeout:
```
bdash: If I'm reading this correctly, this results in `pong_timeout` being reported to the binding's error handler. Is that really what we want?

jpsim: Actually, I think you're right, that's probably not what we want. I've posted https://github.com/realm/realm-sync/pull/1187#issuecomment-290212966 to get some clarification. Thanks!

PR #419, opened 2017-03-27
Fix an assertion failure when an observed List is deleted after being modified

tgoyne:
Fixes all the `list.col < list.observer->changes.size()` assertion failures.


tgoyne:
There were some reports that mentioned that it happened when the array was deleted and that was enough of a hint about what scenarios to investigate.

PR #423, opened 2017-03-30
Building sync test for Android

kneth:
In order to perform sync tests on Android, we need to build YAML locally.

TO DO:

* [x] Download and build YAML
* [ ] Fix failing tests; since tests are running as a CLI command (and not as an Android app), we could use `libuv`
* [ ] Adjust Jenkins file
* [x] Use the old Android toolchain - or upgrade to CMake 3.7 (testing with CMake 3.4.1)


CMake/RealmCore.cmake
```diff
@@ -20,6 +20,8 @@ include(ExternalProject)
 include(ProcessorCount)
 
 find_package(Threads)
+find_package(PkgConfig)
```
emanuelez: This is done twice. I think this one can be removed.

CMake/RealmCore.cmake
```diff
@@ -390,8 +392,16 @@ macro(build_realm_sync)
     set_property(TARGET realm-sync-server PROPERTY IMPORTED_LOCATION_RELEASE ${sync_server_library_release})
     set_property(TARGET realm-sync-server PROPERTY IMPORTED_LOCATION ${sync_server_library_release})
 
-    find_package(PkgConfig)
-    pkg_check_modules(YAML QUIET yaml-cpp)
+    if(REALM_PLATFORM STREQUAL "Android")
+      ExternalProject_Add(YAML
+        GIT_REPOSITORY "git@github.com:jbeder/yaml-cpp.git"
+        CMAKE_ARGS "-DCMAKE_SYSTEM_NAME=Android -DCMAKE_ANDROID_ARCH_ABI=${ANDROID_ABI} -DCMAKE_ANDROID_NDK=${ANDROID_NDK}"
```
emanuelez: If you do it this way (and not using a toolchain) then you have to bump the minimum required version to 3.7.
I think a toolchain would still be the way to go at this point.

CMake/RealmCore.cmake
```diff
@@ -390,8 +392,16 @@ macro(build_realm_sync)
...
+        )
+      set(YAML_LDFLAGS "-L${YAML_BINARY_DIR}-lyaml")
```
emanuelez: missing space?

CMake/RealmCore.cmake
kneth: True. 3.7 is so much nicer but we should probably keep using the old toolchain.

CMake/RealmCore.cmake
kneth: I removed the other one.

CMake/RealmCore.cmake
beeender: The cmake shipped with Android Studio gets stuck with 3.6. Although it may not be a problem for compiling object store.

emanuelez: @beeender Can Android Studio be set to use another cmake installed on the system?

beeender: I don't think so since AS is even shipping a modified version of cmake. But i think it is fine for object store to use 3.7, since will never build this alone in the AS.

kneth: I have changed it, and tested with CMake 3.4.1.

CMake/RealmCore.cmake
```diff
@@ -390,8 +390,21 @@ macro(build_realm_sync)
     set_property(TARGET realm-sync-server PROPERTY IMPORTED_LOCATION_RELEASE ${sync_server_library_release})
     set_property(TARGET realm-sync-server PROPERTY IMPORTED_LOCATION ${sync_server_library_release})
 
-    find_package(PkgConfig)
-    pkg_check_modules(YAML QUIET yaml-cpp)
+    if(REALM_PLATFORM STREQUAL "Android")
+      find_host_package(Git)
+      ExternalProject_Add(YAML
```
beeender: `YAML` -> `yaml`? We always use lower cases for external project name in this file.

PR #429, opened 2017-04-14
Add some simple convenience APIs to `PermissionResults`

austinzheng:


src/sync/sync_permission.hpp
```diff
@@ -93,12 +93,24 @@ class PermissionResults {
...
     }
 
+    // Create a new instance by sorting this instance.
+    PermissionResults sort(SortDescriptor&& s) const
+    {
+        return PermissionResults(m_results.sort(std::move(s)));
+    }
+
+    // Get the results.
+    Results& results()
```
alazier: who is going to use/abuse this?

austinzheng: I use it a few times in [this commit](https://github.com/realm/realm-cocoa/pull/4799/commits/4543d5365edd4e509c5d01ac4fd1ea52b03f01d2) (search for `_results->results()`). Much of the Cocoa machinery to assemble sort descriptors and other thing assumes you have easy access to the raw results value.

alazier: From that pr it looks like all you really need is the `realm` and `objectSchema`, but not direct access to the actual results. I'm also not a fan of looking up column indexes with core apis if the value is already cached in the objectSchema (unless this isn't actually the case?)

alazier: That being said I think this is probably fine, but makes it easy to choose a dangerous path!

tgoyne: Getting the column number from the object schema is both slower and more code than getting it from core.

austinzheng: You bring up good points. I'll go back to the Cocoa code and see if I can do it a different way first.

alazier: @tgoyne performance isn't really an issue here and even if it were the difference seems inconsequential. Also seems like it is about the same amount of code, ie ~1 line. In any case I'm more concerned with the fact that we are accessing something which is meant to be private/hidden behind an abstraction.

tgoyne: Results does not attempt to abstract away the use of Table for constructing SortDescriptors.

PR #432, opened 2017-04-18
Some link's field changes are not reported

beeender:
This is reported by https://github.com/realm/realm-java/issues/4474
The root cause is the nested links is not handled well. In some cases
those indices have been saved to the m_not_modified accidentally.

### DIFF
--- a/src/impl/collection_notifier.cpp
+++ b/src/impl/collection_notifier.cpp
@@ -94,7 +94,7 @@ bool DeepChangeChecker::check_outgoing_links(size_t table_ndx,
             if (p->table == table_ndx && p->row == row_ndx && p->col == col)
                 return true;
         }
-        m_current_path[depth] = {table_ndx, row_ndx, col, false};
+        m_current_path[depth] = {table_ndx, row_ndx, col};
         return false;
     };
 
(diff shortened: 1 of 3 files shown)


beeender:
@tgoyne I am not quite sure if this is the right fix. If it is not, please just help me to fix it. thanks!

tgoyne:
This also disables the optimization in a lot of perfectly valid cases.

tgoyne:
Changing `already_checking` to the following fixes this while preserving the optimization:
```
 auto already_checking = [&](size_t col) {
        auto end = m_current_path.begin() + depth;
        auto match = std::find_if(m_current_path.begin(), end, [&](auto& p) {
            return p.table == table_ndx && p.row == row_ndx && p.col == col;
        });
        if (match != end) {
            for (; match < end; ++match)
                match->depth_exceeded = true;
                return true;
        }
        m_current_path[depth] = {table_ndx, row_ndx, col, false};
        return false;
    };
```

The key detail is that exiting early due to hitting a circular link needs to be handled the same as exiting early due to hitting the max recursion depth, but we still want to mark intermediate links as unmodified if we actually did fully walk the tree from there.

beeender:
@tgoyne thanks! and PR updated.

PR #433, opened 2017-04-18
Fix delivery of notifications for current version when there are later versions

tgoyne:
advance_to_ready() may be called even if the notifications are for the current version, so it needs to handle that case.

@alazier should fix the GN issue.


alazier:
This fixes the gn issues 👍

PR #434, opened 2017-04-21
Stable ID support

simonask:
Note: How to merge this is still TBD, as support is needed from core libraries.


src/object_accessor.hpp
```diff
@@ -246,23 +250,38 @@ Object Object::create(ContextType ctx, SharedRealm realm, const ObjectSchema &ob
...
                 auto value = Accessor::to_string(ctx, primary_value);
+#if REALM_ENABLE_SYNC
+                row_index = sync::create_object_with_primary_key(realm->read_group(), *table, value);
+#else
+                row_index = table->add_empty_row();
                 table->set_string_unique(primary_prop->table_column, row_index, value);
+#endif // REALM_ENABLE_SYNC
+            }
+            else {
+                REALM_TERMINATE("Unsupported primary key type.");
```
tgoyne: This is unreachable because there's validation of the primary key type earlier.

simonask: `REALM_UNREACHABLE` is mainly intended to silence compiler warnings, I think — even though it does happen to abort in our version. I'm using `REALM_TERMINATE` as a substitute for `REALM_ASSERT(false)` — isn't it appropriate with the equivalent of an assertion here?

simonask:
> All the stuff that has to vary depending on whether or not sync is enabled should be hidden behind a wrapper rather than slathering ifdefs everywhere.

I agree, but I guess I'm questioning why we keep those code paths around. Is there any use case for building Object Store without Sync support?

I'm happy to wrap the calls, but I do think the ugly ifdefs serve as a useful reminder that there is an architectural problem that should be addressed soon-ish.

cmelchior:
Java builds a version of Object Store without Sync. Right now we do a lot of manual work, manipulating which files are being included. It is rather annoying: https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/cpp/CMakeLists.txt#L189

kristiandupont:
It's a product decision but I think it's worth challenging whether we want to keep supporting the non-sync binaries? I get that it's nice for those who don't use sync but those are per definition not our paying customers and I don't know if it's worth the overhead keeping them happy? It adds some complexity in .NET and it would in JS if we supported it (which we don't yet).

cmelchior:
Binary size is an important product parameter on Android, and the question keeps coming up on Android again and again (which is also why we track both method count + library size: https://greatscott-pinheads-70.c.influxdb.com:3000/dashboard/db/realm-java)

Only supporting a synced variant would seriously harm our adoption funnel on Android. I tend to agree with Simon that it is more an architectural problem than anything else.

simonask:
@cmelchior Hm, interesting point. We could also make an effort to reduce the binary size of the Sync binary.

tgoyne:
> I'm happy to wrap the calls, but I do think the ugly ifdefs serve as a useful reminder that there is an architectural problem that should be addressed soon-ish.

I don't think that making the object store code uglier will remind the correct people about that.

bdash:
> src/object_store.cpp:32:33: fatal error: realm/sync/object.hpp: No such file or directory

I assume this is because dependencies.list hasn't been updated to a version of sync with the necessary changes?

simonask:
> I assume this is because dependencies.list hasn't been updated to a version of sync with the necessary changes?

Yes. There also is no such version (yet). :-)

simonask:
> I don't think that making the object store code uglier will remind the correct people about that.

Let's discuss this through other channels. I'll change it. My opinion is that the distinction between Object Store and the Sync library has to be fundamentally rethought.

simonask:
I'm moving this back into "In Active Progress" because upstream changes have required further changes to this code before it can be merged.

simonask:
Ok, this should be ready for review once again. The latest changes in Sync, which mandate that tables are born with information about their primary key, had a slightly larger impact than expected, and I had to make some changes to the migration handling. Essentially tables are now created in two phases: first, all tables are created, and second all non-PK columns are added. This is to ensure that link targets are created before link columns are added, which is necessary because the schema information for the link target wasn't available when the source table is being created.

I'm really not sure this is the right way to implement this - would love your input @tgoyne!

For a reason I could not figure out, it was also necessary to remove the implementation of `RemoveProperty` in `create_initial_tables`, otherwise the Object Store would try to generate schema-breaking changes, causing sync to assert. I have no idea why this is only happening now.

src/object_store.cpp
```diff
@@ -128,20 +137,41 @@ void replace_column(Group& group, Table& table, Property const& old_property, Pr
...
 
+void add_initial_columns(Group& group, ObjectSchema const& object_schema)
+{
+    auto name = ObjectStore::table_name_for_object_type(object_schema.name);
+    TableRef table = group.get_table(name);
+
+    for (auto const& prop : object_schema.persisted_properties) {
+#if REALM_ENABLE_SYNC
+        // The sync::create_table* functions create the PK column for us.
+        if (object_schema.primary_key.size() && prop.name == object_schema.primary_key) {
```
tgoyne: `if (prop.is_primary)`

src/object_store.cpp
```diff
@@ -242,13 +272,22 @@ StringData ObjectStore::get_primary_key_for_object(Group const& group, StringDat
...
 
-    // get row or create if new object and populate
     size_t row = table->find_first_string(c_primaryKeyObjectClassColumnIndex, object_type);
+
+#if REALM_ENABLE_SYNC
+    // sync::create_table* functions should have already updated the pk table.
+    if (primary_key.size() && sync::has_object_ids(group)) {
+      REALM_ASSERT(row != not_found);
+      REALM_ASSERT(table->get_string(c_primaryKeyPropertyNameColumnIndex, row) == primary_key);
+    }
```
bdash: The indentation of the body of this `if` isn't right.

simonask:
> I don't understand what AddInitialProperties is doing or why it needs to be split out from the table creation.

I tried to explain it in my comment, but basically it's because AddTable could add link columns to tables that didn't yet exist. The previous behavior was to simply add those tables as needed, but that can no longer work without some significant restructuring (or maybe not so significant? can't tell), because adding the table requires knowledge about which column is the primary key column.

tgoyne:
Hm, makes sense. Making two passes over the changes to create the tables in the first pass and then populate them in the second would be a better fit for the design I was going for with the schema change stuff, but I'm not sure that the end result would actually be better.

simonask:
Yeah, doing two passes would be better in my opinion, but I'm not entirely sure how easy it would be to do with the current application of the visitor pattern. Maybe not so hard actually.

beeender:
I am curious about what will happen for the future migration - if user wants to add a pk column, how to connect the newly added pk with the existing object id? Actually I don't know how does adding a PK column work for the additive mode migration.

cmelchior:
I believe that is not allowed under Additive mode, since such an operation do not converge

simonask:
Yes, Primary Keys are considered part of the schema, so changing a table's primary key column after it has been populated with rows is considered invalid. This restriction may be lifted in the future.

src/object_accessor.hpp
```diff
@@ -246,23 +250,38 @@ Object Object::create(ContextType ctx, SharedRealm realm, const ObjectSchema &ob
         row_index = get_for_primary_key_impl(ctx, *table, *primary_prop, primary_value);
 
         if (row_index == realm::not_found) {
-            row_index = table->add_empty_row();
             created = true;
-            if (primary_prop->type == PropertyType::Int)
+            if (primary_prop->type == PropertyType::Int) {
+#if REALM_ENABLE_SYNC
+                row_index = sync::create_object_with_primary_key(realm->read_group(), *table, Accessor::to_long(ctx, primary_value));
```
beeender: @simonask So if user is using the sync version software, but he also need a non-sync Realm storage at the same time. What will happen? The non-sync version realm file he created cannot be read by the core without sync?

simonask: Good question — I should document this!

`sync::create_object_with_primary_key()` and `sync::create_object()` should work on non-sync'ed Realms. They check dynamically if the Realm is sync'ed or not. :)

src/object_store.cpp
```diff
@@ -128,20 +137,42 @@ void replace_column(Group& group, Table& table, Property const& old_property, Pr
...
 
+void add_initial_columns(Group& group, ObjectSchema const& object_schema)
+{
+    auto name = ObjectStore::table_name_for_object_type(object_schema.name);
+    TableRef table = group.get_table(name);
+
+    for (auto const& prop : object_schema.persisted_properties) {
+#if REALM_ENABLE_SYNC
+        // The sync::create_table* functions create the PK column for us.
+        if (object_schema.primary_key.size() && prop.is_primary()) {
```
nirinchev: `is_primary` is not a function

src/object_store.cpp
```diff
@@ -242,13 +273,22 @@ StringData ObjectStore::get_primary_key_for_object(Group const& group, StringDat
...
     TableRef table = group.get_table(c_primaryKeyTableName);
 
-    // get row or create if new object and populate
     size_t row = table->find_first_string(c_primaryKeyObjectClassColumnIndex, object_type);
+
+#if REALM_ENABLE_SYNC
+    // sync::create_table* functions should have already updated the pk table.
+    if (primary_key.size() && sync::has_object_ids(group)) {
+        REALM_ASSERT(row != not_found);
+        REALM_ASSERT(table->get_string(c_primaryKeyPropertyNameColumnIndex, row) == primary_key);
```
beeender: @simonask So user cannot create a table first then add a primary key later?

simonask: Correct.

bdash:
@simonask, it doesn't appear that sync provides a variant of `sync::create_object_with_primary_key` that supports creating an object with null for the primary key value. It appears that such an overload would be necessary to accommodate changes to the accessor logic made on master to support Cocoa's use of it (see [here](https://github.com/realm/realm-object-store/blob/6decdfbb88494908c8116b64b21c8715175fc2f3/src/object_accessor.hpp#L194-L199)).

simonask:
@bdash Support for nullable primary keys was added last week. `sync::create_object_with_primary_key()` was changed to accept a `util::Optional<int64_t>` or a `StringData` with the intrinsic null value. :-)

bdash:
Which sync branch has that change? I don't see it in [`su/stable-ids`](https://github.com/realm/realm-sync/blob/su/stable-ids/src/realm/sync/object.hpp#L159-L160)?

simonask:
@bdash Oh, sorry, work has moved to `develop-2.0` :)

bdash:
It'd be beneficial to get this reviewed and merged now that it builds with both sync v1.x and 2.x.

src/object_store.cpp
```diff
@@ -242,13 +274,22 @@ StringData ObjectStore::get_primary_key_for_object(Group const& group, StringDat
...
 
-    // get row or create if new object and populate
     size_t row = table->find_first_string(c_primaryKeyObjectClassColumnIndex, object_type);
+
+#if REALM_HAVE_SYNC_STABLE_IDS
+    // sync::create_table* functions should have already updated the pk table.
+    if (primary_key.size() && sync::has_object_ids(group)) {
+        REALM_ASSERT(row != not_found);
+        REALM_ASSERT(table->get_string(c_primaryKeyPropertyNameColumnIndex, row) == primary_key);
+    }
```
tgoyne: // sync::create_table* functions should have already updated the pk table.
    if (sync::has_object_ids(group)) {
        if (primary_key.empty())
            REALM_ASSERT(row == not_found);
        else {
             REALM_ASSERT(row != not_found);
             REALM_ASSERT(table->get_string(c_primaryKeyPropertyNameColumnIndex, row) == primary_key);
        }
        return;
    }

Both cases seem worth checking, and no need to continue on to do a redundant check in the case where there is a PK.

src/object_store.cpp
```diff
@@ -492,14 +544,15 @@ static void create_initial_tables(Group& group, std::vector<SchemaChange> const&
...
+        void operator()(AddInitialProperties op) { add_initial_columns(group, *op.object); }
 
         // Note that in normal operation none of these will be hit, as if we're
         // creating the initial tables there shouldn't be anything to update.
         // Implementing these makes us better able to handle weird
         // not-quite-correct files produced by other things and has no obvious
         // downside.
         void operator()(AddProperty op) { add_column(group, table(op.object), *op.property); }
-        void operator()(RemoveProperty op) { table(op.object).remove_column(op.property->table_column); }
+        void operator()(RemoveProperty) { }
```
tgoyne: This change (presumably) being required is concerning as it shouldn't be possible to hit this in the first place and almost certainly points to a bug somewhere.

bdash:
The following new unit test fails with this branch:
```diff
diff --git a/tests/migrations.cpp b/tests/migrations.cpp
index abe7dec..497e985 100644
--- a/tests/migrations.cpp
+++ b/tests/migrations.cpp
@@ -352,6 +352,25 @@ TEST_CASE("migration: Automatic") {
             };
             REQUIRE_MIGRATION_NEEDED(*realm, schema, set_primary_key(schema, "object", ""));
         }
+
+        SECTION("adding column and table in same migration doesn't add duplicate columns") {
+            auto realm = Realm::get_shared_realm(config);
+
+            Schema schema1 = {
+                {"object", {
+                    {"col1", PropertyType::Int, "", "", false, false, false},
+                }},
+            };
+            auto schema2 = add_table(add_property(schema1, "object",
+                                        {"col2", PropertyType::Int, "", "", false, false, false}),
+                                     {"object2", {
+                                         {"value", PropertyType::Int, "", "", false, false, false}}});
+            REQUIRE_UPDATE_SUCCEEDS(*realm, schema1, 0);
+            REQUIRE_UPDATE_SUCCEEDS(*realm, schema2, 1);
+
+            auto& table = *get_table(realm, "object2");
+            REQUIRE(table.get_column_count() == 1);
+        }
     }
 
     SECTION("migration block invocations") {
```

The table for `object2` ends up with two columns named `value`. The underlying problem is that `AddInitialProperties` ends up being applied as part of both the pre- and post-migration schema modifications when no migration function is specified. Since it's not an idempotent operation, we end up creating the same columns at both points. It looks like `AddInitialProperties` needs to be skipped in the post-migration schema modifications if no migration function was specified (if a migration function was specified, we re-read the schema and compute a new set of changes prior to performing the post-migration modifications, so `AddInitialProperties` will only be present in the set of changes if it's really needed).

src/object_store.cpp
bdash: The tests still pass after reverting this change, so if it's truly necessary for correctness we're missing a test that demonstrates why.

bdash:
Fixed the above issue in c21bb9c9f8ae65fbc5bdeb1a77c89c4e21354ff6.

dependencies.list
```diff
@@ -1,2 +1,2 @@
-REALM_CORE_VERSION=2.8.1
-REALM_SYNC_VERSION=1.10.0
+REALM_CORE_VERSION=2.8.4
+REALM_SYNC_VERSION=2.0.0-rc9
```
beeender: If we merge this to master, then the sync version should be kept as the current stable (`1.x.x`) so CI can test with the minor release i think?

bdash: Ideally we'd have CI test with both 1.x and 2.x, but we don't have a good way to express that via dependencies.list. While we're testing only one of the two options, I don't think there's a good reason to prefer testing 1.x to 2.x.

simonask:
@bdash THANK YOU for taking this on, this saves us several days! Is anything preventing us from merging at this point?

bdash:
If you're happy with the changes I made, I don't think there's anything we need to wait on to merge this.

PR #435, opened 2017-04-21
Optionally prolong the lifetime of a sync session until a completion …

austinzheng:
…handler is called

Currently we don’t do anything to prolong the lifetime of a sync session upon which a “wait for *” handler has been registered, causing surprising results when a binding attempts to register a handler but doesn’t do anything to keep the session alive.

To do:
- [ ] Tests (?)

### DIFF
--- a/src/sync/sync_session.hpp
+++ b/src/sync/sync_session.hpp
@@ -75,14 +75,20 @@ public:
     // chooses to run it on. The method returns immediately with true if the callback was
     // successfully registered, false otherwise. If the method returns false the callback will
     // never be run.
+    //
     // This method will return true if the completion handler was registered, either immediately
     // or placed in a queue. If it returns true the completion handler will always be called
     // at least once, except in the case where a logged-out session is never logged back in.
-    bool wait_for_upload_completion(std::function<void(std::error_code)> callback);
+    //
+    // If the prolong_lifetime flag is set to true, the session will guarantee that, if and
+    // only if it reaches the point where the waiter is registered with the sync client, that
+    // the session will remain alive until the callback is called. The flag will not stop an
+    // inactive session with a queued waiter from being destroyed.
+    bool wait_for_upload_completion(std::function<void(std::error_code)> callback, bool prolong_lifetime=false);
 
     // Register a callback that will be called when all pending downloads have been completed.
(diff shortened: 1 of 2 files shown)


bdash:
What's the notification for this when the binding can keep the session alive by simply capturing it in the callback?

Was it intentional that this was marked as needing review while still labeled as In Progress?

src/sync/sync_session.hpp
```diff
@@ -75,14 +75,20 @@ class SyncSession : public std::enable_shared_from_this<SyncSession> {
...
+    // If the prolong_lifetime flag is set to true, the session will guarantee that, if and
+    // only if it reaches the point where the waiter is registered with the sync client, that
+    // the session will remain alive until the callback is called. The flag will not stop an
+    // inactive session with a queued waiter from being destroyed.
+    bool wait_for_upload_completion(std::function<void(std::error_code)> callback, bool prolong_lifetime=false);
 
     // Register a callback that will be called when all pending downloads have been completed.
     // Works the same way as `wait_for_upload_completion()`.
-    bool wait_for_download_completion(std::function<void(std::error_code)> callback);
+    bool wait_for_download_completion(std::function<void(std::error_code)> callback, bool prolong_lifetime=false);
```
bdash: This sort of parameter results in literal booleans being used at the caller where it's completely non-obvious what they mean.

austinzheng:
> What's the notification for this when the binding can keep the session alive by simply capturing it in the callback?

We could do that (and we do), but when https://github.com/realm/realm-cocoa/pull/4883 was being reviewed it was considered surprising that the session needed to be captured explicitly by the binding.

> Was it intentional that this was marked as needing review while still labeled as In Progress?

I decided to add the reviewers when creating the review, but perhaps it doesn't make sense to actually add them until the PR moves into the "ready to review" phase.

bdash:
> We could do that (and we do), but when realm/realm-cocoa#4883 was being reviewed it was considered surprising that the session needed to be captured explicitly by the binding.

Ok, so why make it conditional whether it keeps it alive?

austinzheng:
There's no known use case right now. I can make it unconditional if you prefer.

austinzheng:
Doesn't look like the other bindings actually need this, so closing it.

PR #439, opened 2017-04-27
Allow Android to validate SSL certificate (using Sync 1.7.0)

nhachicha:
This PR will help Android use SSL correctly in order to validate ROS certificate. Details [here](https://github.com/realm/realm-sync/issues/1270)

This has been added in this [PR]( https://github.com/realm/realm-sync/pull/1306)

`Client::Config` has lost the two properties that relate to verifying a server’s SSL certificate. The functionality has been moved to `Session::bind()`.


jpsim:
The macOS compilation failures in CI appear legitimate:

```
tests/sync/session/session.cpp:144:20: error: no matching constructor for initialization of 'realm::SyncConfig'
        SyncConfig config;
                   ^
src/sync/sync_config.hpp:101:8: note: candidate constructor (the implicit move constructor) not viable: requires 1 argument, but 0 were provided
struct SyncConfig {
       ^
src/sync/sync_config.hpp:101:8: note: candidate constructor (the implicit copy constructor) not viable: requires 1 argument, but 0 were provided
src/sync/sync_config.hpp:113:5: note: candidate constructor not viable: requires at least 4 arguments, but 0 were provided
    SyncConfig(std::shared_ptr<SyncUser> user, std::string realm_url, SyncSessionStopPolicy stop_policy,
    ^
```

jpsim:
Also, since several bindings have already shipped public APIs that expose `SyncManager.disableSSLValidation`, I suggest we keep that API in the object store too. It could be implemented by iterating over the existing sync configurations and updating all their values. New sync configs could look up this value on creation if not specified.

bdash:
I don't think there's any reason to keep that setting around on `SyncManager` at the object store level. Bindings that wish to continue to expose it on their `SyncManager` can do so by using the specified value as the default for sync configurations that have not had SSL validation explicitly configured.

nhachicha:
@jpsim fixed the `macOS` compilation issue. 
This was caused by the `#if __GNUC__ < 5`. apparently the compiler used for the `macos_sync` flavour (i.e `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc`) also defines the  `__GNUC__`. By adding the explicit ctor, the implicit ctor will not be generated. Which will prevent creating `SynConfig` without any parameters.
```c++
SyncConfig conf;
```

bdash:
> By adding the explicit ctor, the implicit ctor will not be generated. Which will prevent creating `SynConfig` without any parameters.
>```c++
>SyncConfig conf;
>```

That breaks the ability to build the tests with GCC 4.9.

austinzheng:
> can you weigh in on the right way to handle this new error code?

I looked at the source, and it doesn't seem like that error is being used anywhere. I find it a little strange that sync has defined an HTTP-related error, unless this is laying the groundwork for future feature development.

It seems the best way to handle it in object store is just to add it to the list of cases in the switch statement, perhaps with a `FIXME` to revisit it whenever sync actually does something with it.

nhachicha:
@bdash added the default ctor

bdash:
This looks good, but the warning I mentioned earlier needs to be addressed before it can be merged.

PR #440, opened 2017-04-28
Add two helper methods, so all bindings can use them

austinzheng:
In support of https://github.com/realm/realm-cocoa/pull/4799

### DIFF
--- a/src/sync/sync_permission.cpp
+++ b/src/sync/sync_permission.cpp
@@ -78,16 +78,17 @@ bool Permission::paths_are_equivalent(std::string path_1, std::string path_2,
                                       std::string user_id_1, std::string user_id_2)
 {
     if (path_1 == path_2) {
-        return true;
+        // If both paths have `/~/`, the user IDs must match.
+        return (path_1.find("~") == npos) ? true : (user_id_1 == user_id_2);
     }
     size_t index = path_1.find("~");
     if (index != npos) {
         // Substitute in the user ID for the `/~/` portion of the path, if applicable.
-        return path_1.replace(index, 1, user_id_2) == path_2;
+        return path_1.replace(index, 1, user_id_1) == path_2;
     }
     index = path_2.find("~");
     if (index != npos) {
-        return path_2.replace(index, 1, user_id_1) == path_1;
+        return path_2.replace(index, 1, user_id_2) == path_1;
     }
     return false;
 }
--- a/src/sync/sync_permission.cpp
+++ b/src/sync/sync_permission.cpp
@@ -61,6 +61,37 @@ Permission::AccessLevel extract_access_level(Object& permission, CppContext& con
 
 }
 
+// MARK: - Permission
+
+std::string Permission::description_for_access_level(AccessLevel level)
+{
+    switch (level) {
+        case AccessLevel::None: return "none";
+        case AccessLevel::Read: return "read";
+        case AccessLevel::Write: return "write";
+        case AccessLevel::Admin: return "admin";
+    }
+    REALM_UNREACHABLE();
+}
+
+bool Permission::paths_are_equivalent(std::string path_1, std::string path_2,
+                                      std::string user_id_1, std::string user_id_2)
+{
+    if (path_1 == path_2) {
+        return true;
+    }
+    size_t index = path_1.find("~");
+    if (index != npos) {
+        // Substitute in the user ID for the `/~/` portion of the path, if applicable.
+        return path_1.replace(index, 1, user_id_2) == path_2;
+    }
+    index = path_2.find("~");
+    if (index != npos) {
+        return path_2.replace(index, 1, user_id_1) == path_1;
+    }
+    return false;
+}
+
 // MARK: - PermissionResults
 
 Permission PermissionResults::get(size_t index)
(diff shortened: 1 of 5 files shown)


src/sync/sync_permission.cpp
```diff
@@ -61,6 +61,37 @@ Permission::AccessLevel extract_access_level(Object& permission, CppContext& con
...
+        case AccessLevel::Admin: return "admin";
+    }
+    REALM_UNREACHABLE();
+}
+
+bool Permission::paths_are_equivalent(std::string path_1, std::string path_2,
+                                      std::string user_id_1, std::string user_id_2)
+{
+    if (path_1 == path_2) {
+        return true;
```
bdash: Is this correct if `user_id_1` and `user_id_2` are different?

austinzheng: Oh, that's a good point. In the original Objective-C code we checked user IDs before calling this method, but we shouldn't assume the same will be true for this public C++ API.

src/sync/sync_permission.cpp
```diff
@@ -61,6 +61,37 @@ Permission::AccessLevel extract_access_level(Object& permission, CppContext& con
...
+    switch (level) {
+        case AccessLevel::None: return "none";
+        case AccessLevel::Read: return "read";
+        case AccessLevel::Write: return "write";
+        case AccessLevel::Admin: return "admin";
+    }
+    REALM_UNREACHABLE();
+}
+
+bool Permission::paths_are_equivalent(std::string path_1, std::string path_2,
```
bdash: This feels like the kind of function that could really benefit from unit tests.

austinzheng:
I added tests and made some updates.

src/sync/sync_permission.cpp
```diff
@@ -61,6 +61,38 @@ Permission::AccessLevel extract_access_level(Object& permission, CppContext& con
...
+                                      std::string user_id_1, std::string user_id_2)
+{
+    if (path_1 == path_2) {
+        // If both paths have `/~/`, the user IDs must match.
+        return (path_1.find("~") == npos) ? true : (user_id_1 == user_id_2);
```
bdash: I think this comparison should be against `std::string::npos` rather than `realm::npos` as it currently.

src/sync/sync_permission.cpp
```diff
@@ -61,6 +61,38 @@ Permission::AccessLevel extract_access_level(Object& permission, CppContext& con
...
+    if (index != npos) {
+        // Substitute in the user ID for the `/~/` portion of the path, if applicable.
+        return path_1.replace(index, 1, user_id_1) == path_2;
+    }
+    index = path_2.find("~");
+    if (index != npos) {
+        return path_2.replace(index, 1, user_id_2) == path_1;
+    }
+    return false;
+}
```
bdash: I think this logic is equivalent to:
```c++
size_t index = path_1.find("~");
if (index != std::string::npos)
    path_1.replace(index, 1, user_id_1);

index = path_2.find("~");
if (index != std::string::npos)
    path_2.replace(index, 1, user_id_2);

return path_1 == path_2;
```

That's easier to reason about for me.

src/sync/sync_permission.cpp
```diff
@@ -61,6 +61,38 @@ Permission::AccessLevel extract_access_level(Object& permission, CppContext& con
...
+        case AccessLevel::None: return "none";
+        case AccessLevel::Read: return "read";
+        case AccessLevel::Write: return "write";
+        case AccessLevel::Admin: return "admin";
+    }
+    REALM_UNREACHABLE();
+}
+
+bool Permission::paths_are_equivalent(std::string path_1, std::string path_2,
+                                      std::string user_id_1, std::string user_id_2)
```
bdash: User IDs can be const reference since we never modify them.

src/sync/sync_permission.cpp
tgoyne: `? true :` is `||`

austinzheng:
I believe all outstanding feedback has been addressed.

src/sync/sync_permission.cpp
```diff
@@ -61,6 +61,38 @@ Permission::AccessLevel extract_access_level(Object& permission, CppContext& con
...
+    switch (level) {
+        case AccessLevel::None: return "none";
+        case AccessLevel::Read: return "read";
+        case AccessLevel::Write: return "write";
+        case AccessLevel::Admin: return "admin";
+    }
+    REALM_UNREACHABLE();
+}
+
+bool Permission::paths_are_equivalent(std::string path_1, std::string path_2,
```
jpsim: Is it legal for `~` to appear anywhere other than at the beginning with `/~/`? For example, can I access a Realm at URL `/rootRealmWithUserID_~`? Where the trailing `~` would be substituted with the user ID?

This function replaces _any_ `~` in the path, not just leading `/~/`.

austinzheng: I'm 99% sure we don't support tilde-substitution anywhere besides in a leading `/~/`. Your other concern is definitely valid. Not sure if we support `~` in a Realm name.

austinzheng: I briefly chatted with @kspangsege and updated the code to reflect what we believe are the correct semantics. (I can also make the check more restrictive, to only support a leading `/~/`, if you prefer.)

austinzheng:
I hope this is a bit closer to approval-acceptability now.

PR #441, opened 2017-04-28
Add support for `updatedAt` property

austinzheng:
Changes:
- Permission change timestamps are now set properly
- `realm::Permission` values now expose an `updated_at` property

Note that this PR is based on top of #440 so that I can continue working on the binding, but will be changed to be on top of master if that previous PR is accepted.

### DIFF
--- a/src/sync/sync_permission.cpp
+++ b/src/sync/sync_permission.cpp
@@ -33,6 +33,7 @@
 #include <realm/util/to_string.hpp>
 
 using namespace realm;
+using namespace std::chrono;
 
 using PrimaryKey = Property::PrimaryKey;
 using Indexed = Property::Indexed;
@@ -57,6 +58,16 @@ Permission::AccessLevel extract_access_level(Object& permission, CppContext& con
         return Permission::AccessLevel::Read;
 
     return Permission::AccessLevel::None;
+}
+
+/// Turn a system time point value into the 64-bit integer representing ns since the Unix epoch.
+int64_t ns_since_unix_epoch(const system_clock::time_point& point)
+{
+    tm unix_epoch = {0};
+    unix_epoch.tm_year = 70;
+    time_t epoch_time = mktime(&unix_epoch);
+    auto epoch_duration = system_clock::from_time_t(epoch_time).time_since_epoch();
+    return duration_cast<nanoseconds>(point.time_since_epoch() + epoch_duration).count();
 }
 
 }
@@ -171,13 +182,12 @@ void Permissions::set_permission(std::shared_ptr<SyncUser> user,
                                  PermissionChangeCallback callback,
                                  const ConfigMaker& make_config)
 {
-    using namespace std::chrono;
     const auto realm_url = user->server_url() + permission.path;
     auto realm = Permissions::management_realm(std::move(user), make_config);
     CppContext context;
 
     // Get the current time.
-    int64_t ns_since_epoch = duration_cast<nanoseconds>(system_clock::now().time_since_epoch()).count();
+    int64_t ns_since_epoch = ns_since_unix_epoch(system_clock::now());
     int64_t s_arg = ns_since_epoch / (int64_t)Timestamp::nanoseconds_per_second;
     int32_t ns_arg = ns_since_epoch % Timestamp::nanoseconds_per_second;
 
--- a/src/sync/sync_permission.cpp
+++ b/src/sync/sync_permission.cpp
@@ -63,11 +63,11 @@ Permission::AccessLevel extract_access_level(Object& permission, CppContext& con
 /// Turn a system time point value into the 64-bit integer representing ns since the Unix epoch.
 int64_t ns_since_unix_epoch(const system_clock::time_point& point)
 {
-    tm unix_epoch = {0};
+    tm unix_epoch{};
     unix_epoch.tm_year = 70;
     time_t epoch_time = mktime(&unix_epoch);
-    auto epoch_duration = system_clock::from_time_t(epoch_time).time_since_epoch();
-    return duration_cast<nanoseconds>(point.time_since_epoch() + epoch_duration).count();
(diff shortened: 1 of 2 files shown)


austinzheng:
This PR can be reviewed now. It's pretty small.

src/sync/sync_permission.cpp
```diff
@@ -170,16 +171,22 @@ void Permissions::set_permission(std::shared_ptr<SyncUser> user,
                                  PermissionChangeCallback callback,
                                  const ConfigMaker& make_config)
 {
+    using namespace std::chrono;
     const auto realm_url = user->server_url() + permission.path;
     auto realm = Permissions::management_realm(std::move(user), make_config);
     CppContext context;
 
+    // Get the current time.
+    int64_t ns_since_epoch = duration_cast<nanoseconds>(system_clock::now().time_since_epoch()).count();
```
bdash: This code assumes that the system clock's epoch is the UNIX epoch that `Timestamp` expects. I don't think there's any guarantee that this is the case.

src/sync/sync_permission.cpp
```diff
@@ -170,16 +171,22 @@ void Permissions::set_permission(std::shared_ptr<SyncUser> user,
...
+    int32_t ns_arg = ns_since_epoch % Timestamp::nanoseconds_per_second;
+
     // Write the permission object.
     realm->begin_transaction();
     auto raw = Object::create<util::Any>(&context, realm, *realm->schema().find("PermissionChange"), AnyDict{
         { "id", util::uuid_string() },
-        { "createdAt", Timestamp(0, 0) },
-        { "updatedAt", Timestamp(0, 0) },
+        { "createdAt", Timestamp(s_arg, ns_arg) },
+        { "updatedAt", Timestamp(s_arg, ns_arg) },
```
bdash: Is it possible for `createdAt` and `updatedAt` to have different values? How would that come about? If it's not, why do we have both of them?

austinzheng: I assume that the server may change `updatedAt`. None of our existing binding-level permission code changes `updatedAt` separately from `createdAt`.

src/sync/sync_permission.cpp
austinzheng: Since there's apparently no portable standardized C++ way to get the epoch for the system clock, should I just add an extension point that can call out to the binding for the time since UNIX epoch?

bdash: Since what we care about is the time since the UNIX epoch, I'd think we could populate a `struct tm` with the UNIX epoch, convert that to a `time_t`, then use `std::chrono::system_clock::from_time_t` to convert that to a `time_point` that we can then subtract from the current time point.

alazier:
Is this the time the server executes the changes or the time where the requests was made?

austinzheng:
@alazier That's a good question, and I don't think it's documented anywhere. I based this PR off how the existing Cocoa-level permissions stuff works today.

src/sync/sync_permission.cpp
austinzheng: Oh, that's clever. I'll do that then.

austinzheng:
I updated this PR.

src/sync/sync_permission.cpp
```diff
@@ -59,6 +60,16 @@ Permission::AccessLevel extract_access_level(Object& permission, CppContext& con
...
 }
 
+/// Turn a system time point value into the 64-bit integer representing ns since the Unix epoch.
+int64_t ns_since_unix_epoch(const system_clock::time_point& point)
+{
+    tm unix_epoch = {0};
+    unix_epoch.tm_year = 70;
+    time_t epoch_time = mktime(&unix_epoch);
+    auto epoch_duration = system_clock::from_time_t(epoch_time).time_since_epoch();
+    return duration_cast<nanoseconds>(point.time_since_epoch() + epoch_duration).count();
```
bdash: Shouldn't you be subtracting the UNIX epoch from the current time rather than adding it?

I don't think it's necessary to call `time_since_epoch()` either, since there's an `operator-` defined for two `std::time_point`s that gives you the duration between them.

src/sync/sync_permission.cpp
```diff
@@ -59,6 +60,16 @@ Permission::AccessLevel extract_access_level(Object& permission, CppContext& con
     return Permission::AccessLevel::None;
 }
 
+/// Turn a system time point value into the 64-bit integer representing ns since the Unix epoch.
+int64_t ns_since_unix_epoch(const system_clock::time_point& point)
+{
+    tm unix_epoch = {0};
```
bdash: This could be `tm unix_epoch{};`.

jpsim:
I can confirm that the ROS is responsible for updating `updatedAt`. See https://github.com/realm/realm-object-server/blob/v1.4.2/object-server/src/node/services/management/listeners/baseListener.js#L101

src/sync/sync_permission.cpp
austinzheng: Yeah, you're right. I confused myself, sorry.

bdash:
What about `createdAt`? If ROS is responsible for setting `updatedAt`, is there any reason for us to set it when creating the object?

austinzheng:
I assume `createdAt` is for the sake of the app developer, so they can look at the objects in the management Realm and see when they changed permissions. It doesn't appear to be touched by ROS.

bdash:
Oh hey, I found [some documentation in the JS source](https://github.com/realm/realm-object-server/blob/9e8f775701b17715bcfb179412c711251f7f42c0/object-server/src/node/services/management/models/permissionBaseObject.js#L43-L58) that lays out how these properties are expected to be used. Short version: `createdAt` populated by client, never changed. `updatedAt` populated by client, then updated by the server in the future.

austinzheng:
I'm a little miffed that GitHub code search (which is useless in more ways than one) didn't pick up on those. Oh well.

PR #442, opened 2017-05-01
Set default value for SSL validation option

austinzheng:
Changes:
- Set default option so SSL validation is enabled by default
- Code formatting

AFAIK without this change the SSL validation option only defaults to `true` if we use an old version of GCC, which enables the explicitly defined initializer.


austinzheng:
I asked Mark about this. Apple platforms already have the right behavior (since LLVM uses the explicit constructor), but this will fix our behavior for Windows/newer GCC versions.

PR #443, opened 2017-05-02
Update object accessor functionality for Cocoa and arrays of primitives

tgoyne:
Depends on https://github.com/realm/realm-core/pull/2571 so this isn't mergeable yet but the code changes here are ready for review.

The design of the native accessor is changed to a stateful object that is passed around rather than static functions defined by specializing a templated type plus a context pointer, which makes attaching metadata beyond that tracked by the object store code much less awkward.

Some hooks have been added to support Cocoa's pre-existing functionality (such as will_change/did_change callbacks for KVO).

The actual type conversion functions have been renamed to box/unbox rather than functions named after what type they're converting for the sake of arrays of primitives, which need to be able to convert generic types to and from the binding's type. List enumeration has been changed to a function taking a callback rather than iterating by index to make it possible to use obj-c fast enumeration.

Implementations using this: https://github.com/realm/realm-cocoa/commit/2dda22e319fb83f95acd471b076eb8661a3011de https://github.com/realm/realm-js/commit/faa876839f7742a7ae7e5e388b3189da51f3493a


src/object_accessor.hpp
```diff
@@ -268,75 +216,68 @@ Object Object::create(ContextType ctx, SharedRealm realm, const ObjectSchema &ob
...
         }
+        if (v)
+            object.set_property_value_impl(ctx, prop, *v, try_update, is_default);
     }
     return object;
 }
 
 template<typename ValueType, typename ContextType>
-Object Object::get_for_primary_key(ContextType ctx, SharedRealm realm, const ObjectSchema &object_schema, ValueType primary_value)
+Object Object::get_for_primary_key(ContextType& ctx, SharedRealm realm,
```
bdash: Since we're changing the signature of this method, we should consider changing these methods to take the Realm by (const) reference since there's no reason for `shared_ptr` to be involved here.

tests/object.cpp
```diff
@@ -215,10 +254,10 @@ TEST_CASE("object") {
         }
     }
 
-    CppContext context;
+    TestContext d(r);
```
bdash: `context` wasn't a good name, but I at least have a general idea of what it meant. `d` on the other hand…

src/object_accessor.hpp
```diff
@@ -36,64 +36,8 @@
 #include <string>
 
 namespace realm {
-//
-// Value converters - template specializations must be implemented for each platform in order to call templated methods on Object
-//
-template<typename ValueType, typename ContextType>
-class NativeAccessor {
```
tgoyne: Any ideas on how to do that? I guess I could just add a comment pointing at CppContext as a reference implementation of it. The previous approach of having a declaration that was never actually used had the problem of that there was nothing verifying that it was correct or complete (and it wasn't).

bdash: A comment pointing at `CppContext` is the best idea I have (until concepts come along).

PR #444, opened 2017-05-02
Expose Apple reachability observer

austinzheng:
Changes:
- Reachability observer is now exposed for binding use
- Reachability observer changed to be singleton
- Reachability observer can now register multiple callbacks, and unregister them
- Updated sync client code

### DIFF
--- a/src/sync/impl/apple/network_reachability_observer.cpp
+++ b/src/sync/impl/apple/network_reachability_observer.cpp
@@ -52,7 +52,6 @@ NetworkReachabilityStatus reachability_status_for_flags(SCNetworkReachabilityFla
 
 NetworkReachabilityObserver& NetworkReachabilityObserver::shared(util::Optional<util::Logger&> logger_ref)
 {
-    // FIXME: is this static property going to screw up our tests?
     static NetworkReachabilityObserver shared;
     if (!shared.start_observing() && logger_ref)
         logger_ref->error("Failed to set up network reachability observer.");
--- a/src/sync/impl/apple/network_reachability_observer.cpp
+++ b/src/sync/impl/apple/network_reachability_observer.cpp
@@ -18,6 +18,8 @@
 
 #include "sync/impl/apple/network_reachability_observer.hpp"
 
+#include <realm/util/logger.hpp>
+
 #if NETWORK_REACHABILITY_AVAILABLE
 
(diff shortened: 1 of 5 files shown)


austinzheng:
I think I may need to add a dummy version for platforms that don't support reachability (i.e. watchOS), similar to how our other platform-specific implementations work.

bdash:
Why bother with this when Cocoa could just use the SystemConfiguration reachability API directly?

austinzheng:
Don't you have to wrap that API anyways, per the famous [Apple sample project](https://developer.apple.com/library/content/samplecode/Reachability/Introduction/Intro.html)? If you prefer just using `SystemConfiguration` directly, I can do that too.

src/sync/impl/apple/network_reachability_observer.cpp
```diff
@@ -48,10 +50,17 @@ NetworkReachabilityStatus reachability_status_for_flags(SCNetworkReachabilityFla
 
 } // (anonymous namespace)
 
-NetworkReachabilityObserver::NetworkReachabilityObserver(util::Optional<std::string> hostname,
-                                                         std::function<void (const NetworkReachabilityStatus)> handler)
+NetworkReachabilityObserver& NetworkReachabilityObserver::shared(util::Optional<util::Logger&> logger_ref)
```
bdash: Why turn this into a singleton? This makes it impossible to track reachability for different hostnames.

austinzheng: I figured we weren't using or planning to use the hostnames functionality. Maybe it makes more sense to not have singletons, though, and just have each instance of an observer associated with a single callback (like before).

bdash: Doing reachability using hostnames is preferable because it can give much more accurate results.

bdash: Even ignoring that aspect, the singleton appears to make things more complex to use rather than easier.

austinzheng: Oh, that's good to know.

austinzheng:
I think the best course of action is to just build out whatever I need for Reachability in the binding itself, and leave the object store stuff alone.

PR #445, opened 2017-05-03
Slightly modify function signatures for support API

austinzheng:
If I don't do this, attempting to call the affected API from the Cocoa binding results in a linker error (since an `NSObject` is being passed into the method).

I'm also wondering if I need to move the header file out of an `impl` directory in the file system, or if we can live with a single Cocoa source file importing a single implementation header.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -115,8 +115,9 @@ if(REALM_ENABLE_SYNC)
         sync/impl/sync_metadata.cpp)
     if(APPLE)
         list(APPEND SOURCES
-            sync/impl/apple/network_reachability_observer.cpp
-            sync/impl/apple/system_configuration.cpp)
+            sync/apple/system_configuration.cpp
+            sync/impl/apple/network_reachability_observer.cpp        
+        )
     endif()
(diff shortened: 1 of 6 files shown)


bdash:
Can you share more details about the problem this is fixing? I assume it's related to dispatch objects being declared as Objective-C types when building with an Obj-C compiler and declared as plain structs for non-Obj-C?

If these files are being used from outside of the object store, they should be moved out of the `impl` directory.

austinzheng:
Yes, that's pretty much it. This is the linker failure I'm talking about. Note how the second argument is treated as an `NSObject`.

```
Undefined symbols for architecture x86_64:
  "realm::_impl::SystemConfiguration::network_reachability_set_dispatch_queue(__SCNetworkReachability const*, NSObject<OS_dispatch_queue>*)", referenced from:
      -[RLMReachability start] in RLMReachability.o
      -[RLMReachability stop] in RLMReachability.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

austinzheng:
This isn't useful anymore; #454 handles what this PR was partially meant to address.

PR #446, opened 2017-05-09
Expose the compact api on Win32

nirinchev:
The .NET assembly isn't aware if we're running on Windows or other platform, so there's no elegant way to prevent it from calling into those missing methods. Furthermore, the other compact methods are not behind and `#ifndef`, so for consistency's sake, we shouldn't hide these either.

### DIFF
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -161,7 +161,6 @@ void Realm::open_with_config(const Config& config,
             };
             shared_group = std::make_unique<SharedGroup>(*history, options);
 
-#ifndef _WIN32
             if (config.should_compact_on_launch_function) {
                 size_t free_space = -1;
                 size_t used_space = -1;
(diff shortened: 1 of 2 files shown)


jpsim:
https://github.com/realm/realm-core/blob/v2.6.0/src/realm/group_shared.hpp#L413-L414

> WARNING / FIXME: compact() should NOT be exposed publicly on Windows
> because it's not crash safe! It may corrupt your database if something fails

tgoyne:
I'm not really sure why `compact()` isn't crash-safe on Windows. It uses `copy()` rather than `move()` which results in the file being corrupted if the process crashes in the middle of the copy, but it's not obvious why, as Windows does have atomic renaming.

nirinchev:
Agreed. However, ObjectStore isn't the public API. We're throwing when compacting on Windows - https://github.com/realm/realm-dotnet/pull/1389/files#diff-8c3f214c5d8da72534fc6d685d3305a9R220, but hiding those methods means our structs would have different layout, which would cause all sorts of problems.

nirinchev:
If nobody objects, I'm going to go ahead and merge that.

jpsim:
If you merge this, you should at least restate the risks of calling into this on Windows.

PR #451, opened 2017-05-16
Change reconnect logic so sessions are revived upon reconnect

austinzheng:
~WIP pending integration with Cocoa changes and manual verification.~ It works

Changes:
- When the reachability system detects network connectivity, the sync manager's `reconnect()` method is now called, rather than directly calling the client cancel delay method.
- The sync manager's `reconnect()` method now asks each session to revive itself if needed.
- If a session is revived in the 'waiting for token' state, the session will now run the login callback. For the Cocoa binding, this simply means that any existing refresh handle will be cancelled and a new one will be created, causing a login request to be sent immediately.

### DIFF
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -184,7 +184,7 @@ struct sync_session_states::WaitingForAccessToken : public SyncSession::State {
     bool revive_if_needed(std::unique_lock<std::mutex>&, SyncSession& session) const override
     {
         session.m_deferred_close = false;
-        return false;
+        return true;
     }
 
     void nonsync_transact_notify(std::unique_lock<std::mutex>&,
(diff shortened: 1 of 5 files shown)


austinzheng:
I changed the way this works because `revive_if_needed()` wasn't what we wanted.

`revive_if_needed()` is intended for taking a session which was inactive or slated to be destroyed and indicating that someone does want to use it further.

The new `handle_reconnect()` method has a different meaning. If the session is already active or intended to be active, it'll trigger a token refresh. Otherwise, it does nothing.

src/sync/sync_session.cpp
```diff
@@ -184,7 +187,15 @@ struct sync_session_states::WaitingForAccessToken : public SyncSession::State {
     bool revive_if_needed(std::unique_lock<std::mutex>&, SyncSession& session) const override
     {
         session.m_deferred_close = false;
-        return false;
+        return true;
```
bdash: Why is this necessary?

austinzheng: That should be `false`. I thought I changed it back...

PR #452, opened 2017-05-17
Use sync 1.9.0

radu-tutueanu:
Update objectStore to sync 1.9.0 and core 2.7.0


austinzheng:
Is there something preventing this from being merged?

radu-tutueanu:
No, I just waited for a ROS that supports the protocol version to be out.

PR #453, opened 2017-05-18
Expose stable IDs for testing

simonask:
This is a crude way to expose stable IDs until an API is finalized.

It turns out that it was almost impossible to test the data connector without this, but the way it is exposed is probably suboptimal.

Note: This is also based on @radu-tutueanu's branch #452.


kristiandupont:
Some things like exposing `compact` on Windows seem unrelated?

bmunkholm:
@simonask Can this be closed?

PR #454, opened 2017-05-23
Make more changes to reconnect APIs

austinzheng:
Changes:
- `SyncSession` now exposes the underlying `Session::cancel_reconnect_delay()` method
- `SyncManager::reconnect()` no longer automatically cancels the reconnect delay

Notes: there's a bug in sync that manifests if you hit `Client::cancel_reconnect_delay()` more than once in rapid succession. Also, it makes more sense to have sessions cancel their reconnect delays on a per-session basis, rather than hitting the client method over and over again.


src/sync/sync_manager.hpp
```diff
@@ -85,7 +85,10 @@ friend class SyncSession;
     void set_client_should_reconnect_immediately(bool reconnect_immediately);
     bool client_should_reconnect_immediately() const noexcept;
 
-    /// Force sync client to reconnect immediately if the connection was lost.
+    /// Ask all valid sync sessions to perform whatever tasks might be necessary to
```
cmelchior: Is this a permanent change? IMO it makes the semantics of this method really muddled, so I would much rather see this getting fixed in Sync (Which is the source of this, right? )

austinzheng: I'd prefer to keep it this way, at least until we remove the responsibility of the bindings to fetch and refresh tokens. Would changing the name of the method help? You can always call `SyncManager::cancel_reconnect_delay()` if it's the underlying client reconnect delay that you care about.

cmelchior: Ok, I cannot really come up with a better name so I guess keeping it as `reconnect` is fine

bdash:
Can you clarify which of the changes here are necessary to work around the underlying sync issue (is there a GitHub issue tracking this problem?), vs which are just changes in design? For those that are simply changes in design, can you please explain the motivation for them?

src/sync/sync_manager.cpp
```diff
@@ -261,17 +261,13 @@ bool SyncManager::client_should_reconnect_immediately() const noexcept
     return m_client_reconnect_mode == ReconnectMode::immediate;
 }
 
-void SyncManager::reconnect()
+void SyncManager::request_that_sessions_reconnect()
```
bdash: Is this really an improvement? The new name is effectively "please ask the sessions to reconnect". I could maybe see using `reconnect_sessions` if there were some ambiguity about what was being reconnected, but `reconnect` by itself doesn't really seem unclear to me?

austinzheng:
There are two different code paths, and two slightly different issues that we are attempting to solve at the same time.

First of all, there exists a `Client::cancel_reconnect_delay()`, and a `Session::cancel_reconnect_delay()`. The former calls the latter for every session that the client currently manages.

Upon re-establishing network connectivity, the old (Cocoa + object store) behavior was that the client would be asked to cancel its reconnect delay. Each session would then be asked to refresh its token. Upon refreshing the token, each session would also then ask the client to cancel its reconnect delay.

Upon awakening from a prolonged period of sleep, the old behavior was that sync would receive a 202 error and ask the binding to fetch a new token for each session. The binding would refresh the token. Upon refreshing the token, each session would ask the client to cancel its reconnect delay. (Otherwise the client's back-off, from the initial failed connection attempt, would be active and it would take ~10 seconds to re-establish connectivity).

There seems to be a problem with sync where calling the `Client::cancel_reconnect_delay()` more than once in rapid succession causes an assert to be failed. Upon looking at the code Morten thought that the assert should be removed and the condition the assert was checking again should be handled explicitly. The ticket tracking this issue is here: https://github.com/realm/realm-sync/issues/1365.

At the same time, it's also wasteful for each session to call `Client::cancel_reconnect_delay()`, since that method calls the `cancel_reconnect_delay()` method for all the sessions each time it's called.

So, the new behavior is this:

Upon re-establishing network connectivity, each session will be asked to refresh its token, and upon refreshing the token that session's `cancel_reconnect_delay()` method will be called.

Upon awakening from a prolonged period of sleep, sync will get its 202 error, ask each session to refresh its token, and upon refreshing the token the session's `cancel_reconnect_delay()` method will be called.

In neither case will the _client's_ `cancel_reconnect_delay()` method be called. It doesn't seem to make sense to call this method before every session has a new token anyways.

This new behavior is both more efficient, and it gets around the sync bug described above.

Let me know if this helps.

src/sync/sync_manager.cpp
austinzheng: Alright, I'm convinced. I'll change it back.

bdash:
I'm a little confused as to why the binding ever needs to make a distinction between calling `reconnect` and `cancel_reconnect_delay`. If the object store knows that network connectivity has been reestablished, can't it drive the rest of the process itself (request an updated access token if the current one has expired, then cancel the sync session's reconnect delay)?

austinzheng:
Right now nothing in the Cocoa binding calls `SyncManager::cancel_reconnect_delay()`. `SyncManager::reconnect()` is only called by the Apple platform reachability handler. The methods I described in my previous comment belong to the sync classes `Session` and `Client`. (`SyncManager`'s `cancel_reconnect_delay()` simply calls into its `Client`'s `cancel_reconnect_delay()`.)

Instead of the binding cancelling the reconnect delay after passing in the updated token, we could have the `SyncSession` handle that.

bdash:
Two questions:

1. What's the difference between `SyncManager::cancel_reconnect_delay` and `SyncManager::reconnect`. When would a binding choose to use one vs the other? Isn't the latter always desirable, due to the work it does related to refreshing tokens prior to triggering the sync-level reconnect?

2. Why should the Cocoa binding need to add calls to `SyncSession::cancel_reconnect_delay` (realm/realm-cocoa#4957)? Can't the object store call that automatically as a result of `SyncManager::reconnect` having been called (automatically by the reachability hook in the case of Cocoa, and presumably by the binding in other cases)?

austinzheng:
@bdash 

1. It seems to me that `SyncManager::cancel_reconnect_delay()` is going to always be preferable. However, other bindings may want to use that method to call into `Client::cancel_reconnect_delay()`? If nobody objects I can just kill it completely.

2. I think you are right, and am about to update the PRs to make calling `Session::cancel_reconnect_delay()` the object store's responsibility.

Note that, all the above does imply the following requirement: a login callback for a given session provided for a binding must fetch a new access token for the session every time it is called, not just the first time.

austinzheng:
I did what I could to simplify things. I can't completely remove the Reachability's hook into `handle_reconnect()`, since we need to be able to restart a stalled request for an initial access token, and sync can't drive that.

src/sync/sync_manager.cpp
```diff
@@ -263,15 +263,11 @@ bool SyncManager::client_should_reconnect_immediately() const noexcept
...
-    {
-        std::lock_guard<std::mutex> lock(m_mutex);
-        if (m_sync_client) {
-            m_sync_client->cancel_reconnect_delay();
-        } else {
-            return;
-        }
-    }
     // Ask all sessions to process reconnection.
+    // The binding is responsible for cancelling the reconnect delay.
```
bdash: Is this still true?

src/sync/sync_manager.hpp
```diff
@@ -85,7 +85,13 @@ friend class SyncSession;
...
+    /// re-establish connectivity with the Realm Object Server. It is presumed that
+    /// the caller knows that network connectivity has been restored.
+    ///
+    /// Note that calling this method does not imply cancelling the reconnect delay.
```
bdash: Is this still true?

bdash:
> I did what I could to simplify things. I can't completely remove the Reachability's hook into handle_reconnect(), since we need to be able to restart a stalled request for an initial access token, and sync can't drive that.

To make sure I understand you correctly: you're referring to the fact that `SyncManager::reconnect` still calls `SyncSession::handle_reconnect`, which is necessary when the object store is waiting on the binding to provide an access token? That seems reasonable to me.

austinzheng:
Yep, that is correct (and is, right now, the only way `handle_reconnect()` is used).

austinzheng:
Linux build failed last time CI ran. Trying to figure out why...

```
[linux_sync] ../realm/noinst/client_file_access_cache.hpp:139: [realm-core-2.7.0] Assertion failed: !m_first_open_file
[linux_sync] IMPORTANT: if you see this error, please send this log to help@realm.io.
[linux_sync] Thread name: tests
[linux_sync] 
[linux_sync] ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[linux_sync] tests is a Catch v1.7.0 host application.
[linux_sync] Run with -? for options
[linux_sync] 
[linux_sync] -------------------------------------------------------------------------------
[linux_sync] SyncSession: management by SyncUser
[linux_sync] -------------------------------------------------------------------------------
[linux_sync] /home/jenkins/workspace/m_realm-object-store_PR-454-FM54CFS4UI2XYSW247PGYIFM54QZBEWQZJJIG4ONE5PQLN6VDDGQ@2/tests/sync/session/session.cpp:40
[linux_sync] ...............................................................................
[linux_sync] 
[linux_sync] /home/jenkins/workspace/m_realm-object-store_PR-454-FM54CFS4UI2XYSW247PGYIFM54QZBEWQZJJIG4ONE5PQLN6VDDGQ@2/tests/sync/session/session.cpp:140: FAILED:
[linux_sync]   {Unknown expression after the reported line}
[linux_sync] due to a fatal error condition:
[linux_sync]   SIGABRT - Abort (abnormal termination) signal
[linux_sync] 
[linux_sync] ===============================================================================
[linux_sync] test cases:   78 |   77 passed | 1 failed
[linux_sync] assertions: 6731 | 6730 passed | 1 failed
[linux_sync] 
[linux_sync] tests/CMakeFiles/generate-coverage-cobertura.dir/build.make:57: recipe for target 'tests/CMakeFiles/generate-coverage-cobertura' failed
[linux_sync] make[3]: *** [tests/CMakeFiles/generate-coverage-cobertura] Aborted (core dumped)
```

austinzheng:
Not surprised at all that I can't reproduce the linux-sync failure locally. Trying some other stuff...

austinzheng:
Alright. I've fixed the issue that caused the test to fail, and filed a bug (linking this PR). It's ready for final review. (I'll squash before merging.)

austinzheng:
The latest failure is the known ping assert issue, please don't hold it against my poor PR!!!

src/sync/sync_manager.hpp
```diff
@@ -85,7 +85,12 @@ friend class SyncSession;
...
-    /// Force sync client to reconnect immediately if the connection was lost.
+    /// Ask all valid sync sessions to perform whatever tasks might be necessary to
+    /// re-establish connectivity with the Realm Object Server. It is presumed that
+    /// the caller knows that network connectivity has been restored.
+    ///
+    /// Refer to `SyncSession::handle_reconnect()` to see what sort of work is done
+    /// on a per-session basis.
     void reconnect();
 
     /// Ask sync client to cancel its reconnection delay. This does not ask the sessions to
```
bdash: When we talked on Slack I got the impression we were intending to remove this method, since `reconnect` is a better choice in all situations.

austinzheng: Yeah, I plan to remove this before I merge. But I don't think JP will sign off on merging this PR until the next release of sync is cut.

austinzheng:
I looked at sync's git tree, and it looks like we need to wait until 1.9.3 at the earliest for a version containing the 1379 fix mentioned above.

austinzheng:
This branch now uses sync 1.10.0, but I'm told the object store unit tests might have to be adjusted to compensate for an unrelated change.

bdash:
`SyncSession::register_progress_notifier`'s header documentation mentions:
> The notifier will always be called immediately during the function, to provide the caller with an initial assessment of the state of synchronization.

Is this no longer true after updating to the new version of sync? What impact, if any, does that have on the semantics the bindings are documented as providing from progress notifications?

austinzheng:
The object store docstring is wrong, the Cocoa docstring for the public API that consumes this method is correct. I will update the object store docstring.

PR #456, opened 2017-05-26
Let binding be able to handle sync client error

beeender:
This is mostly due to the fact that we don't have too much information
when native exception thrown on Android.
With this at least Android has a chance to print the exception message
when error happens on the sync client thread.

### DIFF
--- a/src/sync/impl/sync_client.hpp
+++ b/src/sync/impl/sync_client.hpp
@@ -41,21 +41,27 @@ using ReconnectMode = sync::Client::ReconnectMode;
 struct SyncClient {
     sync::Client client;
 
-    SyncClient(std::unique_ptr<util::Logger> logger,
-               ReconnectMode reconnect_mode = ReconnectMode::normal)
-    : client(make_client(*logger, reconnect_mode)) // Throws
-    , m_logger(std::move(logger))
-    , m_thread([this] {
-        if (g_binding_callback_thread_observer)
-            g_binding_callback_thread_observer->did_create_thread();
-
-        auto will_destroy_thread = util::make_scope_exit([&]() noexcept {
-            if (g_binding_callback_thread_observer)
-                g_binding_callback_thread_observer->will_destroy_thread();
-        });
-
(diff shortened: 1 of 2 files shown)


src/sync/impl/sync_client.hpp
```diff
@@ -41,21 +41,26 @@ using ReconnectMode = sync::Client::ReconnectMode;
...
-
-        auto will_destroy_thread = util::make_scope_exit([&]() noexcept {
-            if (g_binding_callback_thread_observer)
-                g_binding_callback_thread_observer->will_destroy_thread();
-        });
-
-        client.run(); // Throws
-    }) // Throws
+    SyncClient(std::unique_ptr<util::Logger> logger, ReconnectMode reconnect_mode = ReconnectMode::normal)
+        : client(make_client(*logger, reconnect_mode)) // Throws
```
bdash: It's a bit unfortunate that you've chosen to re-reformat / re-indent the constructor. It completely obscures your change, while also deviating from the dominant style used in the object store.

src/sync/impl/sync_client.hpp
```diff
@@ -41,21 +41,26 @@ using ReconnectMode = sync::Client::ReconnectMode;
...
+        , m_logger(std::move(logger))
+        , m_thread([this] {
+            if (g_binding_callback_thread_observer) {
```
bdash: This is getting to the point where it may be clearer to provide a default observer that does the appropriate actions so we don't have to worry about these conditionals. Maybe that's not worth it though since it's currently only used in this one place.

src/sync/impl/sync_client.hpp
beeender: @bdash Are we using the same `.clang-format` file from core for object-store? https://github.com/realm/realm-core/blob/master/.clang-format . If not, can we have one in the object-store?

bdash: We're not currently using clang-format with the object store.

PR #463, opened 2017-05-31
Add a user info field to `SyncUser` for binding use

austinzheng:
Required for https://github.com/realm/realm-cocoa/pull/4964

### DIFF
--- a/src/sync/sync_user.cpp
+++ b/src/sync/sync_user.cpp
@@ -24,7 +24,8 @@
 
 namespace realm {
 
-SyncUserContextFactory* SyncUser::s_binding_context_factory;
+util::Optional<SyncUserContextFactory> SyncUser::s_binding_context_factory;
+std::mutex SyncUser::s_binding_context_factory_mutex;
 
 SyncUser::SyncUser(std::string refresh_token, std::string identity,
                    util::Optional<std::string> server_url, TokenType token_type)
@@ -35,7 +36,7 @@ SyncUser::SyncUser(std::string refresh_token, std::string identity,
 , m_identity(std::move(identity))
 {
     if (s_binding_context_factory) {
-        binding_context = s_binding_context_factory->make_context();
+        m_binding_context = s_binding_context_factory.value()();
     }
     if (token_type == TokenType::Normal) {
         SyncManager::shared().perform_metadata_update([this, server_url=std::move(server_url)](const auto& manager) {
@@ -218,6 +219,12 @@ void SyncUser::register_session(std::shared_ptr<SyncSession> session)
         case State::Error:
             break;
     }
+}
+
+void SyncUser::set_binding_context_factory(SyncUserContextFactory factory)
+{
+    std::lock_guard<std::mutex> lock(s_binding_context_factory_mutex);
+    s_binding_context_factory = factory;
 }
 
 void SyncUser::register_management_session(const std::string& path)
--- a/src/sync/sync_user.cpp
+++ b/src/sync/sync_user.cpp
@@ -24,7 +24,7 @@
 
 namespace realm {
 
-util::Optional<SyncUserContextFactory> SyncUser::s_binding_context_factory;
+SyncUserContextFactory SyncUser::s_binding_context_factory;
 std::mutex SyncUser::s_binding_context_factory_mutex;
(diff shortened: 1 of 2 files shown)


src/sync/sync_user.hpp
```diff
@@ -100,6 +107,8 @@ friend class SyncSession;
     std::string refresh_token() const;
     State state() const;
 
+    std::unique_ptr<SyncUserContext> context;
```
austinzheng: This really should be synchronized.

austinzheng:
I've updated this and the corresponding Cocoa PR to use our `AtomicSharedPtr` instead.

src/sync/sync_user.hpp
```diff
@@ -25,12 +25,27 @@
 #include <vector>
 #include <mutex>
 
+#include "../util/atomic_shared_ptr.hpp"
```
bdash: `../` should be unnecessary as object store's `src` directory should be in the include path?

src/sync/sync_user.hpp
```diff
@@ -100,19 +115,32 @@ friend class SyncSession;
...
     void register_management_session(const std::string&);
     void register_permission_session(const std::string&);
 
 private:
     State m_state;
 
+    static SyncUserContextFactory* s_binding_context_factory;
+
+    mutable std::mutex m_context_mutex;
+    std::unique_ptr<SyncUserContext> m_context;
```
bdash: These two members appear to be unused.

src/sync/sync_user.hpp
```diff
@@ -25,12 +25,27 @@
...
 
+// A superclass that bindings can inherit from in order to store information
+// upon a `SyncUser` object.
+class SyncUserContext {
+public:
+    virtual ~SyncUserContext() = default;
+};
+
+// A superclass that represents a way for a sync user to build binding contexts.
+class SyncUserContextFactory {
```
bdash: Any reason to use a base class for this rather than `std::function<std::shared_ptr<SyncUserContext>()>`?

src/sync/sync_user.hpp
```diff
@@ -100,19 +115,32 @@ friend class SyncSession;
...
     // Register a session to this user.
     // A registered session will be bound at the earliest opportunity: either
     // immediately, or upon the user becoming Active.
     // Note that this is called by the SyncManager, and should not be directly called.
     void register_session(std::shared_ptr<SyncSession>);
 
+    // Optionally set a context factory. If so, must be set before any sessions are created.
+    static void set_binding_context_factory(SyncUserContextFactory& factory)
+    {
+        s_binding_context_factory = &factory;
```
bdash: This likely requires some form of synchronization since it may be set from a different thread than on which it is later read.

src/sync/sync_user.hpp
```diff
@@ -100,19 +115,32 @@ friend class SyncSession;
     std::string refresh_token() const;
     State state() const;
 
+    util::AtomicSharedPtr<SyncUserContext> binding_context;
```
bdash: Exposing a getter rather than a public member would be more consistent with the existing interface of this class.

src/sync/sync_user.hpp
austinzheng: No reason, except that I was following what we do for loggers. I don't think it's necessary in this case, and will change it.

src/sync/sync_user.cpp
```diff
@@ -24,6 +24,9 @@
 
 namespace realm {
 
+util::Optional<SyncUserContextFactory> SyncUser::s_binding_context_factory;
```
bdash: `util::Optional` is redundant here, since `std::function` has an empty state.

src/sync/sync_user.cpp
```diff
@@ -32,6 +35,9 @@ SyncUser::SyncUser(std::string refresh_token, std::string identity,
 , m_refresh_token(std::move(refresh_token))
 , m_identity(std::move(identity))
 {
+    if (s_binding_context_factory) {
```
bdash: You'd need to hold the mutex when checking / accessing this value, otherwise there's no synchronization between these accesses and `set_binding_context_factory`.

PR #464, opened 2017-06-01
Adjust how property types are defined to support arrays of all types

tgoyne:
More prep work to support arrays of primitives that should have no functional impact.

The primary change here is to turn `PropertyType::Array` into a bitflag that can be added to any of the other property types, rather than its own type which implies `Object`. Since this makes it so that there's no longer a 1:1 correspondence between `PropertyType` and the core data types, it also goes ahead and renumbers the enum values and also makes the nullability another flag rather than a separate boolean on the `Property`.

This also changes how `is_indexed` and `is_primary` are set on `Property` to use a tagged bool type that makes it a compile-time error to pass them in the wrong order. This change isn't directly related, but untangling it from all the changes to the schema definitions in the tests would have taken a while so it's lumped in.

### DIFF
--- a/src/object_accessor.hpp
+++ b/src/object_accessor.hpp
@@ -86,6 +86,8 @@ void Object::set_property_value_impl(ContextType& ctx, const Property &property,
     }
 
     if (is_array(property.type)) {
+        if (property.type == PropertyType::LinkingObjects)
+            throw ReadOnlyPropertyException(m_object_schema->name, property.name);
         REALM_ASSERT(property.type == PropertyType::Object);
 
         List list(m_realm, m_row.get_linklist(col));
@@ -130,8 +132,6 @@ void Object::set_property_value_impl(ContextType& ctx, const Property &property,
             table.set_link(col, row, link.get_index(), is_default);
             break;
         }
-        case PropertyType::LinkingObjects:
-            throw ReadOnlyPropertyException(m_object_schema->name, property.name);
         default:
             REALM_COMPILER_HINT_UNREACHABLE();
     }
--- a/src/object_accessor.hpp
+++ b/src/object_accessor.hpp
@@ -148,7 +148,7 @@ ValueType Object::get_property_value_impl(ContextType& ctx, const Property &prop
         return ctx.null_value();
     }
 
-    if (is_array(property.type)) {
+    if (is_array(property.type) && property.type != PropertyType::LinkingObjects) {
         REALM_ASSERT(property.type == PropertyType::Object);
         return ctx.box(List(m_realm, m_row.get_linklist(column)));
     }
--- a/src/object_accessor.hpp
+++ b/src/object_accessor.hpp
@@ -72,7 +72,7 @@ void Object::set_property_value_impl(ContextType& ctx, const Property &property,
     auto& table = *m_row.get_table();
     size_t col = property.table_column;
     size_t row = m_row.get_index();
-    if (property.is_nullable && ctx.is_null(value)) {
+    if (is_nullable(property.type) && ctx.is_null(value)) {
         if (property.type == PropertyType::Object) {
             if (!is_default)
                 table.nullify_link(col, row);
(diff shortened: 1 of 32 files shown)


src/object_schema.cpp
```diff
@@ -66,6 +53,25 @@ ObjectSchema::ObjectSchema(std::string name, std::initializer_list<Property> per
...
+        case type_Double:    return PropertyType::Double | optional;
+        case type_Bool:      return PropertyType::Bool | optional;
+        case type_String:    return PropertyType::String | optional;
+        case type_Binary:    return PropertyType::Data | optional;
+        case type_Timestamp: return PropertyType::Date | optional;
+        case type_Mixed:     return PropertyType::Any | optional;
+        case type_Link:      return PropertyType::Object | PropertyType::Nullable;
+        case type_LinkList:  return PropertyType::Object | PropertyType::Array;
+        case type_Table:     return from_core_type(*table.get_subdescriptor(col), 0) | PropertyType::Array;
+        default: REALM_UNREACHABLE(); // FIXME
```
bdash: What remains to be fixed here?

src/object_store.cpp
```diff
@@ -96,19 +96,35 @@ void add_index(Table& table, size_t col)
...
+    switch (type & ~PropertyType::Flags) {
+        case PropertyType::Int:    return type_Int;
+        case PropertyType::Bool:   return type_Bool;
+        case PropertyType::Float:  return type_Float;
+        case PropertyType::Double: return type_Double;
+        case PropertyType::String: return type_String;
+        case PropertyType::Date:   return type_Timestamp;
+        case PropertyType::Data:   return type_Binary;
+        case PropertyType::Object: return type_Link;
+        default: REALM_TERMINATE("unknown type");
```
bdash: Is the omission of `Any` intentional?

Is always mapping `Object` to `type_Link` intentional given that it should be `type_LinkList` when the array flag is set?

src/property.hpp
```diff
@@ -19,145 +19,218 @@
...
+enum class PropertyType : unsigned char {
+    Int    = 0,
+    Bool   = 1,
+    String = 2,
+    Data   = 3,
+    Date   = 4,
+    Float  = 5,
+    Double = 6,
+    Object = 7,
+    LinkingObjects = 8, // Implies Array
```
bdash: What do you mean by "Implies Array" here? Is a linking objects property always be represented as `LinkingObjects | Array`, or is `LinkingObjects` by itself always treated as an array? The implementation of `is_array` suggests the former.

src/object_store.cpp
tgoyne: Mapping to the core type is only needed for creating columns, and it doesn't allow creating Mixed columns. Maybe should explicitly assert that type isn't `Any`.

bdash: On that same vein it seems like omitting `Object` may be sensible since the column creation logic for link columns is different than for other column types.

tgoyne:
I've now addressed all of the review feedback, rebased this onto master, and added some more tests for schema validation logic (this adds some more illegal states to check, and there were a bunch of existing ones that weren't tested).

bdash:
It looks like there's a build failure here due to #434 introducing new references to `Property::is_nullable`.

src/object_accessor.hpp
```diff
@@ -85,7 +85,24 @@ void Object::set_property_value_impl(ContextType& ctx, const Property &property,
...
     }
 
-    switch (property.type) {
+    if (is_array(property.type)) {
+        if (property.type == PropertyType::LinkingObjects)
+            throw ReadOnlyPropertyException(m_object_schema->name, property.name);
+        REALM_ASSERT(property.type == PropertyType::Object);
+
+        List list(m_realm, m_row.get_linklist(col));
+        list.remove_all();
```
bdash: This appears to suffer from the same issue that you fixed in realm/realm-cocoa#5127. It looks like a preexisting issue, so it's not necessary to fix it prior to merging unless it's trivial to do so.

tgoyne: I'll make another PR to fix that in a bit.

PR #466, opened 2017-06-02
Add basic handling for subtable columns

tgoyne:
This makes it so that the schema handling and notification code doesn't break entirely when given a file with subtable columns. It doesn't make it possible to actually interact with them, but it should be enough to be able to open a file containing them without any issues.

### DIFF
--- a/tests/migrations.cpp
+++ b/tests/migrations.cpp
@@ -559,6 +559,32 @@ TEST_CASE("migration: Automatic") {
                 ObjectStore::delete_data_for_object(realm->read_group(), "foo");
             }));
         }
+
+        SECTION("subtables columns are not modified by unrelated changes") {
+            config.in_memory = false;
+            Schema schema = {
+                {"object", {
+                    {"value", PropertyType::Int},
+                }},
+            };
+
+            {
+                auto realm = Realm::get_shared_realm(config);
+                realm->update_schema(schema, 1);
+                realm->begin_transaction();
+                realm->read_group().get_table("class_object")->add_column(type_Table, "subtable");
+                realm->commit_transaction();
+            }
+            // close and reopen the Realm tu ensure it rereads the schema from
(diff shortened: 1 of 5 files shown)


tests/transaction_log_parsing.cpp
```diff
@@ -1727,6 +1758,31 @@ TEST_CASE("Transaction log parsing: changeset calcuation") {
...
+            REQUIRE_FALSE(changes.modified(0, 0));
+            REQUIRE_FALSE(changes.modified(0, 1));
+        }
+
+        SECTION("modifying a subtable in an observed table does not interfere with notifications for other columns") {
+            Row r = target->get(0);
+            auto changes = observe({r}, [&] {
+                target->insert_column(0, type_Table, "subtable");
+                auto subdesc = target->get_subdescriptor(0);
+                subdesc->insert_column(0, type_Int, "value");
```
kneth: No need to `Table::insert_done()` before accessing the subtable?

tgoyne: That function doesn't exist?

kneth: Sorry, my mistake (that method was removed in core 0.92.0 when the `insert_*` family was removed.

tests/migrations.cpp
```diff
@@ -559,6 +559,32 @@ TEST_CASE("migration: Automatic") {
...
+            };
+
+            {
+                auto realm = Realm::get_shared_realm(config);
+                realm->update_schema(schema, 1);
+                realm->begin_transaction();
+                realm->read_group().get_table("class_object")->add_column(type_Table, "subtable");
+                realm->commit_transaction();
+            }
+            // close and reopen the Realm tu ensure it rereads the schema from
```
bdash: `s/tu/to/`

PR #467, opened 2017-06-05
Upgrade to core's new sort and distinct behaviour (core v3.0.0)

### LINKED ISSUE #266: Further refinement of `Results` with `distinct()`
The current implementation of `distinct()` is limited. It is not possible to call it multiple time to do refinements of `Results`. According to Realm Core, the limitation is:

```
Each time you call distinct() it will first fetch the full original TableView contents and then apply
distinct() on that. So it distinct() does not filter the result of the previous distinct().
```

See also https://github.com/realm/realm-core/issues/2332

ironage:
This PR will apply for a version of core that includes https://github.com/realm/realm-core/pull/2644

It changes queries so the order that sort/distinct is now remembered, and chaining distinct calls is supported.

Hopefully this will allow cocoa to include distinct? If not, now is probably the time to discuss what is still lacking at this level.

Fixes #266.


src/results.cpp
```diff
@@ -520,24 +516,21 @@ TableView Results::get_tableview()
 
 Results Results::sort(realm::SortDescriptor&& sort) const
 {
-    return Results(m_realm, get_query(), std::move(sort), m_distinct);
+    DescriptorOrdering new_order = m_descriptor_ordering;
+    new_order.emplace_sort(std::move(sort));
+    return Results(m_realm, get_query(), new_order);
```
tgoyne: Move `new_order`?

src/results.cpp
```diff
@@ -53,21 +52,22 @@ Results::Results(SharedRealm r, Table& table)
...
 , m_link_view(lv)
-, m_sort(std::move(s))
 , m_mode(Mode::LinkView)
 {
     m_table.reset(&lv->get_target_table());
     if (q) {
         m_query = std::move(*q);
         m_mode = Mode::Query;
     }
+    if (s.is_valid()) {
```
ironage: True, that's not necessary.

ironage:
It is marked as a breaking change in the core change notes. This is because unless Java intends to emulate the old semantics, some queries that combine sort and distinct can produce different results after the upgrade. @cmelchior @kneth is Java able to release the breaking change as it is in a major release? Otherwise, we would probably want to add support for clearing all previously applied distincts every time distinct is called to emulate the old behaviour.

kneth:
As #266 says, the current behaviour is limited (some would say, it is wrong), and I think we will adapt the new semantic. It is a breaking change, so we need to have the timing right (we don't want to have too many major releases).

bdash:
This is now waiting on realm/realm-core#2704.

src/impl/transact_log_handler.cpp
```diff
@@ -497,6 +499,13 @@ class TransactLogObserver : public TransactLogValidationMixin, public MarkDirtyM
         return true;
     }
 
+    bool add_row_with_key(size_t row_ndx, size_t prior_num_rows, size_t key_col, int64_t)
+    {
+        insert_empty_rows(row_ndx, 1, prior_num_rows, false);
+        m_active_table->modify(row_ndx, key_col);
```
tgoyne: This shouldn't be marking the row as modified (and there should be tests for the new instruction).

bdash: @tgoyne, my thought process for marking the row as modified is that `add_row_with_key` is semantically the same as a `add_empty_row` / `set_int` pair, which results in the row as being marked modified. Can you elaborate on why you feel it shouldn't be marked as modified?

tgoyne: The modification from set_int() on a newly-inserted row is spurious and eventually gets filtered out. SetUnique and SetDefault skip marking the row as modified because they can only ever be called on newly inserted rows, but checking if a row is newly inserted in normal set() ends up being a pessimization.

bdash: Ok. Addressed (and test added) in ab716a63b83abbcb8d9a250a387ab238ba572729.

bdash:
Yeah, I figured that would be broken. There's probably other build system combinations that aren't quite right either. I'll take a crack at getting them resolved next.

PR #468, opened 2017-06-06
pass the correct property name in the exception

blagoev:


tgoyne:
The code throwing the exception should be fixed. Trying to munge the property name here is gross and can't work when setting nested objects.

blagoev:
@tgoyne  Sorry didn't get that. Can you elaborate?

tgoyne:
I'm not sure what is unclear. Rather than hacking around the fact that something is throwing an exception with a bad message you should modify that code to just throw an exception with the correct message in the first place.

blagoev:
@tgoyne Would love to do that. But the code churn will be huge in my opinion. 
or in other words this is easier said than done.

In short the js binding calls object sotre which enumerates the properties and then calls back the js binding for to get the value for the property without specifying the property name. The code that converts the property fails and knows only what the type should be. It is the objects-store that knows what the property name is hence it catches the exception and throws a new one with the correct property name and type.
It is by "accident" that these exceptions are with the same type. To my opinion there is nothing wrong of some layer to catch an exception from a layer below and re-throw a different one. (Yeah In this case the same)

 Up until now the layer that converts the value was hard coding 'Property' as property name. Now it is the responsibility of the caller to handle the erroneous conversion and fill in the gap for the name.  

Also this fix was done in compliance with this fix https://github.com/realm/realm-js/commit/bd8a27b963e1b12755959abe6fa711800d762dfb#diff-1add330f13f633f44905c0e38f02e439R110
Which to me look pretty fine. This is not a performance critical path.

When I started fixing this I looked into ways to achieve what you suggested but unfortunately it requires too much code to be changed. And really I don't see an issue to catch and re-throw an exception. If the property name was not const (https://github.com/realm/realm-object-store/blob/4330f13eedfad4f34bcecdca25f71fb949fbc4b2/src/object.hpp#L110) I would have re-thrown the same exception actually.

tgoyne:
The Cocoa code has no trouble throwing an exception with the correct property name, so you could look at the obj-c implementation of the accessor context if you are unsure as to how to pass the property name to the relevant place.

> To my opinion there is nothing wrong of some layer to catch an exception from a layer below and re-throw a different one. (Yeah In this case the same)

That is not what you're doing. The layering here is core -> object store -> realm-js. Translating exceptions as they move upwards in the layers is perfectly reasonable. Translating an exception here which is thrown by realm-js is inverting the layering.

blagoev:
The code path is realm-js -> object store -> realm-js which throws the type convertion exception. Object store calls back into realm-js to get the value but then this conversion fails and the calling layer (object-store) knows why and how to report to the upper layer which in this case is realm-js create object call. I simply used the common exception between those two layers (since object-store does not know about the TypeErrorException)

Maybe in cocoa you can infer the property name at the place you are trying to convert the value. In Js this is not possible. If it was I would have loved to take the property name from there at the point of failure.

bdash:
I think the suggestion is that if the `realm::js::NativeAccessor` instance is aware of the property it is being used for, it could then include information about that property in the exception it throws when the types don't match.

blagoev:
I am not sure, maybe I need to look again. But the thing is the exception originates at a place in realm-js code which needs to convert the value. At that point the property name is not known. The only place where this gets enough information is in object store code that cycles through all properties and calls unbox. 

The other way of passing all information down so the error path has all info to construct the correct exception is what I tried to do but quickly realized too much code needs to change to support this. And then I looked into rethrowing an exception which seems ok for the job. Also there is such fix for the object write error message.  Maybe we can talk about introducing a more specific exception type for that case but than this will not improve much. 

I really don't see much value into trying to optimize this code to throw single exception with the correct error message from the point of failure that propagates back to user code without intermediate handling on the way. (but I agree that would be an ideal case). I tried to make it readable enough by introducing the local variables with descriptive names instead of passing them onto the constructors directly.

bdash:
See https://github.com/realm/realm-js/compare/mar/property-names-in-exceptions for a change to realm-js that attempts to follow Cocoa's general approach to validating the types when setting properties, which removes the need for this change.

blagoev:
Nice! Guys you really put me into my place. I should be ashamed. I never said it is not possible it's just bigger code churn than I wanted to introduce. I will review the changes more. If you find more value in optimizing this path in the other way I am ok with closing this PR and using yours.

bdash:
@blagoev, I wasn't trying to put you in your place or embarrass you. I was merely demonstrating an alternative approach that seems like it might be a better fit for the design of object store's accessor class, and doesn't suffer from the layering issue that Thomas mentioned. I'd be happy to complete the change if you're ok with JS using that approach instead.

blagoev:
I guess what I wanted to say is that you have more deeper knowledge than me. I think we can continue your way. To me it seems like making a bigger change than required for this particular case , but I agree this will deliver better result and code structure.

tgoyne:
Avoiding needless churn in the code is certainly a good thing, but I don't see fixing the code that's doing something wrong as ever being that. Refactorings done for their own sake are unpleasant when you have multiple people working on a code base, but refactorings and reorganizations done to better solve functional problems are key for keeping a code base understandable in the long term.

blagoev:
@tgoyne you are absolutely right

bdash:
@blagoev, can we close this PR?

blagoev:
closing this

PR #476, opened 2017-06-20
Add support for deleting Realm upon permission error

austinzheng:
Changes:
- 'Permission denied' error now triggers the Realm file to be deleted upon next launch
- Refactored helper method for deleting Realm files
- Added a method for deleting a metadata action directly
- Renamed enum case 'HandleRealmForClientReset' to 'BackUpThenDeleteRealm' to better reflect what it does

Note: the corresponding binding work will introduce a system similar to client reset, allowing users to immediately delete or to keep the Realm file permanently.

### DIFF
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -413,6 +413,32 @@ std::string SyncSession::get_recovery_file_path()
                                           util::create_timestamped_template("recovered_realm"));
 }
 
+void SyncSession::update_error_and_mark_file_for_deletion(SyncError& error, ShouldBackup should_backup)
+{
+    // Add a SyncFileActionMetadata marking the Realm as needing to be deleted.
+    std::string recovery_path;
+    auto original_path = path();
+    error.user_info[SyncError::c_original_file_path_key] = original_path;
+    if (should_backup == ShouldBackup::yes) {
+        recovery_path = get_recovery_file_path();
+        error.user_info[SyncError::c_recovery_file_path_key] = recovery_path;
+    }
+    auto action = ((should_backup == ShouldBackup::yes)
+                   ? SyncFileActionMetadata::Action::BackUpThenDeleteRealm
+                   : SyncFileActionMetadata::Action::DeleteRealm);
+    SyncManager::shared().perform_metadata_update([this,
+                                                   action,
+                                                   original_path=std::move(original_path),
+                                                   recovery_path=std::move(recovery_path)](const auto& manager) {
+        SyncFileActionMetadata(manager,
+                               action,
+                               original_path,
+                               m_config.realm_url,
+                               m_config.user->identity(),
+                               util::Optional<std::string>(std::move(recovery_path)));
+    });
+}
+
 // This method should only be called from within the error handler callback registered upon the underlying `m_session`.
 void SyncSession::handle_error(SyncError error)
 {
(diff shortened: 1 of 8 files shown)


src/sync/sync_session.cpp
```diff
@@ -474,30 +500,18 @@ void SyncSession::handle_error(SyncError error)
             }
             case ProtocolError::illegal_realm_path:
             case ProtocolError::no_such_realm:
+                break;
             case ProtocolError::permission_denied:
+                update_error_and_mark_file_for_deletion(error, ShouldBackup::no);
```
austinzheng: This really, really, really needs to be checked to ensure it's not a permission or management Realm.

austinzheng: No it doesn't.

src/sync/sync_session.cpp
```diff
@@ -413,10 +412,36 @@ std::string SyncSession::get_recovery_file_path()
...
+                               m_config.user->identity(),
+                               util::Optional<std::string>(std::move(recovery_path)));
+    });
+}
+
 // This method should only be called from within the error handler callback registered upon the underlying `m_session`.
 void SyncSession::handle_error(SyncError error)
 {
-    bool should_invalidate_session = error.is_fatal;
+    auto next_state = error.is_fatal ? NextStateAfterError::error : NextStateAfterError::none;
```
bdash: Given that `NextStateAfterError` is only used within this function, why not declare it within the function as well?

src/sync/sync_session.cpp
```diff
@@ -413,10 +412,36 @@ std::string SyncSession::get_recovery_file_path()
...
+        error.user_info[SyncError::c_recovery_file_path_key] = recovery_path;
+    }
+    auto action = ((should_backup == ShouldBackup::yes)
+                   ? SyncFileActionMetadata::Action::BackUpThenDeleteRealm
+                   : SyncFileActionMetadata::Action::DeleteRealm);
+    SyncManager::shared().perform_metadata_update([this,
+                                                   action,
+                                                   original_path=std::move(original_path),
+                                                   recovery_path=std::move(recovery_path)](const auto& manager) {
+        SyncFileActionMetadata(manager,
```
bdash: It's surprising to me that creating a C++ object like this modifies a Realm file on disk. It took me a bit of reading to figure out what was going on here.

src/sync/sync_session.cpp
```diff
@@ -413,10 +412,36 @@ std::string SyncSession::get_recovery_file_path()
...
+    std::string recovery_path;
+    auto original_path = path();
+    error.user_info[SyncError::c_original_file_path_key] = original_path;
+    if (should_backup == ShouldBackup::yes) {
+        recovery_path = get_recovery_file_path();
+        error.user_info[SyncError::c_recovery_file_path_key] = recovery_path;
+    }
+    auto action = ((should_backup == ShouldBackup::yes)
+                   ? SyncFileActionMetadata::Action::BackUpThenDeleteRealm
+                   : SyncFileActionMetadata::Action::DeleteRealm);
```
bdash: Perhaps:
```c++
using Action = SyncFileActionMetadata::Action;
auto action = should_backup == ShouldBackup::yes ? Action::BackUpThenDeleteRealm : Action::DeleteRealm;
```

src/sync/sync_session.cpp
austinzheng: Yes, this is one of the things I really want to change in the near future.

src/sync/sync_session.cpp
austinzheng: Huh. I totally forgot you could do that.

PR #478, opened 2017-06-27
Better message to show the mismatched schema

beeender:
Property 'CatOwner.cats' has been changed from `array` to
`array`."
->
Property 'CatOwner.cats' has been changed from `array<Dog>` to
`array<Cat>`."


bdash:
I believe this may be fixed incidentally as part of Thomas's changes in #464.

tgoyne:
#464 improves it a little in some cases, but looking at how it's used I think all uses of `string_for_property_type()` should be changed to `type_string()` (or perhaps `string_for_property_type()` should be changed to be more like `type_string()`, since we don't always have a `Property`).

beeender:
I will wait for #464 merged then, and check if there are other places need to be changed as well.

Zhuinden:
#464 is merged

beeender:
@tgoyne @bdash updated.

PR #484, opened 2017-06-30
Add support for sorting Lists and Results on keypaths

tgoyne:
Mostly just a port of the SortDescriptor building code from Cocoa, since the existing obj-c code doesn't work for primitive arrays and JS (and maybe other bindings) could use this too.

### DIFF
--- a/src/results.cpp
+++ b/src/results.cpp
@@ -528,14 +528,13 @@ static std::vector<size_t> parse_keypath(StringData keypath, Schema const& schem
                                                      keypath, util::format(fmt, args...)));
         }
     };
-    auto sortable_type = [](PropertyType type) {
+    auto is_sortable_type = [](PropertyType type) {
         return !is_array(type) && type != PropertyType::LinkingObjects && type != PropertyType::Data;
     };
 
     const char* begin = keypath.data();
     const char* end = keypath.data() + keypath.size();
     check(begin != end, "missing property name");
-    check(std::find(begin, end, '@') == end, "sorting on collection operators is not implemented");
 
     std::vector<size_t> indices;
     while (begin != end) {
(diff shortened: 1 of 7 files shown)


src/results.cpp
```diff
@@ -518,6 +518,65 @@ TableView Results::get_tableview()
...
+        }
+    };
+    auto sortable_type = [](PropertyType type) {
+        return !is_array(type) && type != PropertyType::LinkingObjects && type != PropertyType::Data;
+    };
+
+    const char* begin = keypath.data();
+    const char* end = keypath.data() + keypath.size();
+    check(begin != end, "missing property name");
+    check(std::find(begin, end, '@') == end, "sorting on collection operators is not implemented");
```
bdash: It feels a little strange to have knowledge of Cocoa's collection operators at this level.

src/results.cpp
```diff
@@ -518,6 +518,65 @@ TableView Results::get_tableview()
...
 
+static std::vector<size_t> parse_keypath(StringData keypath, Schema const& schema, const ObjectSchema *object_schema)
+{
+    auto check = [&](bool condition, const char* fmt, auto... args) {
+        if (!condition) {
+            throw std::invalid_argument(util::format("Cannot sort on key path '%1': %2.",
+                                                     keypath, util::format(fmt, args...)));
+        }
+    };
+    auto sortable_type = [](PropertyType type) {
```
bdash: `is_sortable_type` would be slightly clearer.

src/results.cpp
```diff
@@ -518,6 +518,65 @@ TableView Results::get_tableview()
...
+
+        auto prop = object_schema->property_for_name(key);
+        check(prop, "property '%1.%2' does not exist", object_schema->name, key);
+        check(sortable_type(prop->type), "property '%1.%2' is of unsupported type '%3'",
+              object_schema->name, key, string_for_property_type(prop->type));
+        if (prop->type == PropertyType::Object)
+            check(begin != end, "property '%1.%2' of type 'object' cannot be the final property in the key path",
+                  object_schema->name, key);
+        else
+            check(begin == end, "property '%1.%2' of type '%3' must be the final property in the key path",
```
bdash: "must be the final property in the key path" reads strangely to me. Perhaps "may only be the final property in the key path"?

src/results.cpp
tgoyne: I suppose there's no benefit to doing the check here rather than in the obj-c code calling it since it's just a contains check and not something that requires looking at the parsed form.

PR #485, opened 2017-07-04
Support for initialization function

beeender:
This is a feature for realm-java which has been implemented in java side
before, see https://github.com/realm/realm-java/issues/2254

This gives user possibility to create some initial data only when the
Realm file is created at the first time.

By implementing this in Object Store:
- realm-java can use ResetFile schema mode for
  RealmConfigration.deleteIfMigrationNeeded()
- realm-java doesn't have to start a transaction anymore when calling
  update_schema().
- in_transaction param of update_schema() can be finally removed from
  Object Store.

### DIFF
--- a/src/impl/realm_coordinator.cpp
+++ b/src/impl/realm_coordinator.cpp
@@ -120,7 +120,7 @@ void RealmCoordinator::set_config(const Realm::Config& config)
     if (config.schema_mode == SchemaMode::ReadOnly && config.migration_function)
         throw std::logic_error("Realms opened in read-only mode do not use a migration function");
     if (config.schema_mode == SchemaMode::ReadOnly && config.initialization_function)
-        throw std::logic_error("Realms opened in read-only mode do not use a initialization function");
+        throw std::logic_error("Realms opened in read-only mode do not use an initialization function");
     if (config.schema && config.schema_version == ObjectStore::NotVersioned)
         throw std::logic_error("A schema version must be specified when the schema is specified");
     if (!config.realm_data.is_null() && (!config.read_only() || !config.in_memory))
--- a/src/impl/realm_coordinator.cpp
+++ b/src/impl/realm_coordinator.cpp
@@ -119,6 +119,8 @@ void RealmCoordinator::set_config(const Realm::Config& config)
         throw std::logic_error("Realms opened in Additive-only schema mode do not use a migration function");
     if (config.schema_mode == SchemaMode::ReadOnly && config.migration_function)
         throw std::logic_error("Realms opened in read-only mode do not use a migration function");
+    if (config.schema_mode == SchemaMode::ReadOnly && config.initialization_function)
+        throw std::logic_error("Realms opened in read-only mode do not use a initialization function");
     if (config.schema && config.schema_version == ObjectStore::NotVersioned)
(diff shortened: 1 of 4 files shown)


src/impl/realm_coordinator.cpp
```diff
@@ -119,6 +119,8 @@ void RealmCoordinator::set_config(const Realm::Config& config)
         throw std::logic_error("Realms opened in Additive-only schema mode do not use a migration function");
     if (config.schema_mode == SchemaMode::ReadOnly && config.migration_function)
         throw std::logic_error("Realms opened in read-only mode do not use a migration function");
+    if (config.schema_mode == SchemaMode::ReadOnly && config.initialization_function)
+        throw std::logic_error("Realms opened in read-only mode do not use a initialization function");
```
kneth: "a" -> "an"

src/shared_realm.hpp
```diff
@@ -127,6 +127,13 @@ class Realm : public std::enable_shared_from_this<Realm> {
     // functions which take a Schema from within the migration function.
     using MigrationFunction = std::function<void (SharedRealm old_realm, SharedRealm realm, Schema&)>;
 
+    // A callback function to be called after the first time when schema is created.
```
kneth: remove "after"
"when" -> "a"

beeender:
@kneth I will remove the `in_transaction` later after java side PR merged.

src/shared_realm.hpp
```diff
@@ -127,6 +127,13 @@ class Realm : public std::enable_shared_from_this<Realm> {
     // functions which take a Schema from within the migration function.
     using MigrationFunction = std::function<void (SharedRealm old_realm, SharedRealm realm, Schema&)>;
 
+    // A callback function to be called the first time when a schema is created.
+    // It is passed a SharedRealm with in-transaction state at the version after schema
```
tgoyne: It is passed a SharedRealm which is in a write transaction with the schema initialized.

tests/realm.cpp
```diff
@@ -1136,3 +1142,105 @@ TEST_CASE("SharedRealm: compact on launch") {
...
+    };
+    bool initialization_function_called = false;
+    Realm::DataInitializationFunction initialization_function = [&initialization_function_called](auto shared_realm) {
+        REQUIRE(shared_realm->is_in_transaction());
+        initialization_function_called = true;
+    };
+
+    SECTION("Initialization function should be called for unversioned realm") {
+        config.initialization_function = initialization_function;
+        SECTION("Automatic mode")
```
tgoyne: Structuring the tests like this with the actual checks being done after the leaf section results in test failures not actually reporting which section failed.

tests/realm.cpp
```diff
@@ -1136,3 +1142,105 @@ TEST_CASE("SharedRealm: compact on launch") {
...
+            should_be_called = true;
+        }
+
+        Realm::get_shared_realm(config);
+        REQUIRE(initialization_function_called == should_be_called);
+    }
+
+    SECTION("Call initialization function directly by update_schema") {
+        config.schema_version = ObjectStore::NotVersioned;
+        auto schema = config.schema.value();
```
tgoyne: `auto schema = *config.schema` is the normal way to write this.

beeender:
@tgoyne comments addressed.

PR #487, opened 2017-07-06
Always close the uv_async handle on the loop thread

fealebenpae:
The EventLoopSignal destructor might be invoked on a thread other than the loop thread so we need to marshal the uv_close call to it.

### DIFF
--- a/src/util/uv/event_loop_signal.hpp
+++ b/src/util/uv/event_loop_signal.hpp
@@ -29,16 +29,22 @@ public:
 
         // This assumes that only one thread matters: the main thread (default loop).
         uv_async_init(uv_default_loop(), m_handle, [](uv_async_t* handle) {
-            (*static_cast<Callback*>(handle->data))();
+            if (handle->data == handle) {
+                uv_close(reinterpret_cast<uv_handle_t*>(handle), [](uv_handle_t* handle) {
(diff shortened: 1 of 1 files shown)


tgoyne:
This doesn't look at all thread-safe to me. Is there something that guarantees that `~EventLoopSignal()` won't be called while the signal callback is being invoked on the main thread?

fealebenpae:
Would you suggest creating a new struct to hold the callback and a mutex as well as a close_requested flag or some such?

tgoyne:
Something like the following might work:

    template<typename Callback>
    class EventLoopSignal {
    public:
        struct Data {
            Callback callback;
            std::atomic<bool> delete_requested;
        };

        EventLoopSignal(Callback&& callback)
        {
            m_handle->data = new Data{std::move(callback), {false}};

            // This assumes that only one thread matters: the main thread (default loop).
            uv_async_init(uv_default_loop(), m_handle, [](uv_async_t* handle) {
                auto& data = *static_cast<Data*>(handle->data);
                if (data.delete_requested) {
                    delete &data;
                    delete handle;
                }
                else {
                    data.callback();
                }
            });
        }

        ~EventLoopSignal()
        {
            static_cast<Data*>(m_handle->data)->deleted_requested = true;
            notify();
        }

        // ...
    };

(completely untested)

PR #493, opened 2017-07-10
Fix for sync SSL related API changes

beeender:
ssl related options have been moved to Seesion::Config.
See https://github.com/realm/realm-sync/pull/1435


beeender:
@bdash I am not quite sure how should we deal with the different APIs in the sync 2.0/1.0.

The Object Store master cannot be built with sync `1.10.5` because of this change and the change of `SortDescriptor`.

Shall we use compile options for this and the `SortDescriptor`?

bdash:
@beeender, I'm not quite sure what you mean about `SortDescriptor`. That change was made in such a way that it works with both core v2.8.x and the new v3.x. branch. The stable ID changes were also made in such a way that they work with both sync v1.x and v2.x.

Did a point release of v1.10.x really include a breaking API change?… Whatever happened to the use of semantic versioning.

beeender:
@bdash the `SortDesriptor` PR broke the build with core `2.8.6` because of the `DistinctDescriptor` and `DescriptorOrdering` are not defined in core master https://github.com/realm/realm-core/blob/4860d5b771c9bde4dacfcc3909430acba96be35e/src/realm/views.hpp#L36

Should https://github.com/realm/realm-core/pull/2644 be applied to the core master as well @ironage

bdash:
They're provided by [descriptor_ordering.hpp](https://github.com/realm/realm-object-store/blob/master/src/descriptor_ordering.hpp) for older versions of core.

beeender:
@bdash Uh! OK! thanks. Then what we need is release sync RC12 with https://github.com/realm/realm-sync/pull/1435 to get it build with Object Store?

bdash:
What problem would that solve?

beeender:
The different SSL related API in sync `1.10.5` vs `2.0.0-RC11` -- what this PR try to fix. Or shall I just add compile options in this PR?

bdash:
Oh, I see what you mean. I guess we'll need to `#if` things for 2.0 and pre-2.0 for now.

PR #494, opened 2017-07-10
Change object store to allow sync user auth URL to be modified

austinzheng:
With this PR, getting a user with a different auth URL than the one it was originally given updates the URL, instead of throwing an exception.

This fixes the issue (commonly encountered during development) when a user is opened with one URL, persisted, the server is moved or SSL is enabled, and an attempt to log in the user with the new address is made.

This does not fix the corner case where two different ROSes have users with the same user ID, and the user wishes to be logged into both simultaneously, but this case wasn't supported before, nor is it supported by any of the other subsystem code.

### DIFF
--- a/src/sync/sync_manager.cpp
+++ b/src/sync/sync_manager.cpp
@@ -299,8 +299,9 @@ std::shared_ptr<SyncUser> SyncManager::get_user(const std::string& identity,
         return new_user;
     } else {
         auto user = it->second;
+        util::Optional<std::string> new_server_url;
         if (auth_server_url && *auth_server_url != user->server_url()) {
-            throw std::invalid_argument("Cannot retrieve an existing user specifying a different auth server.");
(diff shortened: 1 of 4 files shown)


bdash:
Did we get clarity on whether `identity` is expected to be unique across different servers? If that's not guaranteed, is this change safe?

austinzheng:
It's complicated.

The default implementation of `identity` is a UUID. However, there's the possibility of plugging a module into ROS that allows you to specify arbitrary identities instead of using the built-in system, so a conflict is theoretically possible. Even if it is, though, using Realms across multiple servers concurrently is something we intentionally de-emphasized when designing v1 of ROS, so I don't think it's a supported use case anyways.

bdash:
It's something that Realm Browser inherently needs to support.

austinzheng:
Back to the drawing board, then.

PR #495, opened 2017-07-11
Add a CI job to test using core 2.x/sync 1.x

tgoyne:


CMake/RealmCore.cmake
```diff
@@ -179,7 +180,11 @@ function(download_realm_core core_version)
         download_realm_tarball(${url_release} ${core_directory_release} ${core_library_release})
     else()
         if(APPLE)
-            set(basename "realm-core-cocoa")
+            if(REALM_CORE_PACKAGING STREQUAL "1")
```
bdash: You could probably use `if(${core_version} VERSION_GREATER_EQUAL "3.0.0")` instead of adding the extra flag.

tgoyne: My idea was to use a separate flag to make it easier to support a future 2.x release using the updated packaging, but maybe that isn't actually useful (since once such a release is made we'd no longer need to support the old packaging).

bdash: I don't feel strongly about it either way. You've done the flag, so we may as well stick with it until we drop support for the old packaging.

Jenkinsfile
```diff
@@ -171,8 +171,9 @@ stage('unit-tests') {
     linux: doDockerBuild('linux', true, false),
     linux_sync: doDockerBuild('linux', true, true),
     android: doAndroidDockerBuild(),
-    macos: doBuild('osx', 'macOS', false),
-    macos_sync: doBuild('osx', 'macOS', true),
+    macos: doBuild('osx', 'macOS', false, ''),
+    macos_1_x: doBuild('osx', 'macOS', false, '-1.x'),
```
bdash: Should we also test with sync enabled, as the PR title mentions?

tgoyne: I started out trying to do that but it turned out to be more involved since the sync-enabled tests build core/sync from source and v2.8.6 doesn't have the cmake build system.

bdash: Oh, right.

PR #496, opened 2017-07-11
avoid calling should_compact_on_launch_function

jpsim:
when a Realm is already open at that path. Fixes Cocoa #5010 (https://github.com/realm/realm-cocoa/issues/5010).

### DIFF
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -56,6 +56,18 @@ Realm::Realm(Config config, std::shared_ptr<_impl::RealmCoordinator> coordinator
         m_schema = ObjectStore::schema_from_group(*m_group);
     }
     else if (!coordinator || !coordinator->get_cached_schema(m_schema, m_schema_version, m_schema_transaction_version)) {
+        if (m_config.should_compact_on_launch_function) {
+            size_t free_space = -1;
+            size_t used_space = -1;
+            // getting stats requires committing a write transaction beforehand.
+            Group* group = nullptr;
+            if (m_shared_group->try_begin_write(group)) {
+                m_shared_group->commit();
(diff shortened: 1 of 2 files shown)


bdash:
Please don't use "Fixes org/repo#xyz" across repositories. It'd result in the original issue being closed as soon as the PR is merged into this repository, while further work is still required for the fix to be made available to Cocoa.

jpsim:
Alright. I reworded the comment to still convey what I meant without making GitHub think it actually fixes it.

jpsim:
@tgoyne think you could review this one?

src/shared_realm.cpp
```diff
@@ -161,7 +161,9 @@ void Realm::open_with_config(const Config& config,
             };
             shared_group = std::make_unique<SharedGroup>(*history, options);
 
-            if (realm && config.should_compact_on_launch_function) {
+            auto coordinator = RealmCoordinator::get_existing_coordinator(config.path);
+
+            if (realm && config.should_compact_on_launch_function && !coordinator->has_initialized_realms()) {
```
tgoyne: I think this block makes more sense in Realm's constructor than here. We don't want to do the compact when we're calling `open_with_config()` from anywhere but the constructor (it's used for things like opening the helper SGs for the notification worker), and there we already have a pointer to the coordinator.

jpsim:
Thanks for the tip, that's much better,

bdash:
@jpsim, not sure if you noticed but there are a number of test failures on CI that need to be addressed before this can be merged.

jpsim:
CI is now passing, but I'm still writing a regression test, porting it over from the cocoa PR.

jpsim:
I've added a regression test, so this is now ready for review again.

PR #497, opened 2017-07-12
Refactor sync user subsystem to support multiple servers

### LINKED ISSUE #296: Cannot retrieve an existing user specifying a different auth server
`SyncManager` throws an uncaught exception (https://github.com/realm/realm-browser-osx-private/issues/147) when logging in the user with the same credentials but with a different server URL (when the same instance of the server is launched on another hostname).

austinzheng:
Changes:
- Users are no longer keyed exclusively by user identity, but rather identity and auth URL
- Local copies of synced Realms are now stored in a directory based on a locally assigned UUID, rather than the user identity
- Upgraded tests

TODO:
- [ ] Make sure API is okay
- [x] Write better migration tests
- [x] Write tests to check the text file we write to user directories
- [x] Test sync manager's `path_for_realm()` API with metadata enabled

Fixes #296.

### DIFF
--- a/src/sync/impl/sync_file.cpp
+++ b/src/sync/impl/sync_file.cpp
@@ -254,14 +254,15 @@ std::string SyncFileManager::user_directory(const std::string& local_identity,
                                             util::Optional<SyncUserIdentifier> user_info) const
 {
     REALM_ASSERT(local_identity.length() > 0);
-    if (filename_is_reserved(local_identity))
+    std::string escaped = util::make_percent_encoded_string(local_identity);
+    if (filename_is_reserved(escaped))
         throw std::invalid_argument("A user can't have an identifier reserved by the filesystem.");
 
     auto user_path = file_path_by_appending_component(get_base_sync_directory(),
-                                                      local_identity,
+                                                      escaped,
                                                       util::FilePathType::Directory);
     bool dir_created = util::try_make_dir(user_path);
-    if (dir_created && user_info && user_info->user_id != local_identity) {
+    if (dir_created && user_info) {
         // Add a text file in the user directory containing the user identity, for backup purposes.
         // Only do this the first time the directory is created.
         auto info_path = util::file_path_by_appending_component(user_path, c_user_info_file);
@@ -282,9 +283,28 @@ void SyncFileManager::remove_user_directory(const std::string& local_identity) c
         throw std::invalid_argument("A user can't have an identifier reserved by the filesystem.");
 
     auto user_path = file_path_by_appending_component(get_base_sync_directory(),
-                                                      local_identity,
+                                                      util::make_percent_encoded_string(local_identity),
                                                       util::FilePathType::Directory);
     util::remove_nonempty_dir(user_path);
+}
+
+bool SyncFileManager::try_rename_user_directory(const std::string& old_name, const std::string& new_name) const
+{
+    REALM_ASSERT_DEBUG(old_name.length() > 0 && new_name.length() > 0);
+    const std::string& base = get_base_sync_directory();
+    auto old_path = file_path_by_appending_component(base,
+                                                     util::make_percent_encoded_string(old_name),
+                                                     util::FilePathType::Directory);
+    auto new_path = file_path_by_appending_component(base,
+                                                     util::make_percent_encoded_string(new_name),
+                                                     util::FilePathType::Directory);
+
+    try {
+        File::move(old_path, new_path);
+    } catch (File::NotFound const&) {
+        return false;
+    }
+    return true;
 }
 
 bool SyncFileManager::remove_realm(const std::string& absolute_path) const
@@ -331,9 +351,9 @@ bool SyncFileManager::remove_realm(const std::string& local_identity, const std:
 {
     REALM_ASSERT(local_identity.length() > 0);
     REALM_ASSERT(raw_realm_path.length() > 0);
-    if (filename_is_reserved(local_identity) || filename_is_reserved(raw_realm_path)) {
+    if (filename_is_reserved(local_identity) || filename_is_reserved(raw_realm_path))
         throw std::invalid_argument("A user or Realm can't have an identifier reserved by the filesystem.");
-    }
+
     auto escaped = util::make_percent_encoded_string(raw_realm_path);
     auto realm_path = util::file_path_by_appending_component(user_directory(local_identity), escaped);
     return remove_realm(realm_path);
@@ -344,9 +364,9 @@ std::string SyncFileManager::path(const std::string& local_identity, const std::
 {
     REALM_ASSERT(local_identity.length() > 0);
     REALM_ASSERT(raw_realm_path.length() > 0);
-    if (filename_is_reserved(local_identity) || filename_is_reserved(raw_realm_path)) {
+    if (filename_is_reserved(local_identity) || filename_is_reserved(raw_realm_path))
         throw std::invalid_argument("A user or Realm can't have an identifier reserved by the filesystem.");
-    }
+
     auto escaped = util::make_percent_encoded_string(raw_realm_path);
     auto realm_path = util::file_path_by_appending_component(user_directory(local_identity, user_info), escaped);
     return realm_path;
--- a/src/sync/impl/sync_file.cpp
+++ b/src/sync/impl/sync_file.cpp
@@ -267,8 +267,10 @@ std::string SyncFileManager::user_directory(const std::string& local_identity,
         auto info_path = util::file_path_by_appending_component(user_path, c_user_info_file);
         std::ofstream info_file;
         info_file.open(info_path.c_str());
-        info_file << std::get<0>(*user_info) << "\n" << std::get<1>(*user_info) << "\n";
-        info_file.close();
+        if (info_file.is_open()) {
+            info_file << std::get<0>(*user_info) << "\n" << std::get<1>(*user_info) << "\n";
+            info_file.close();
+        }
     }
     return user_path;
 }
--- a/src/sync/impl/sync_file.cpp
+++ b/src/sync/impl/sync_file.cpp
@@ -26,6 +26,7 @@
 #include <iomanip>
 #include <sstream>
 #include <system_error>
+#include <fstream>
 
 #ifdef _WIN32
 #include <io.h>
@@ -229,6 +230,7 @@ constexpr const char SyncFileManager::c_utility_directory[];
 constexpr const char SyncFileManager::c_recovery_directory[];
 constexpr const char SyncFileManager::c_metadata_directory[];
 constexpr const char SyncFileManager::c_metadata_realm[];
+constexpr const char SyncFileManager::c_user_info_file[];
 
 std::string SyncFileManager::get_special_directory(std::string directory_name) const
 {
@@ -248,27 +250,37 @@ std::string SyncFileManager::get_base_sync_directory() const
     return sync_path;
 }
 
-std::string SyncFileManager::user_directory(const std::string& user_identity) const
+std::string SyncFileManager::user_directory(const std::string& local_identity,
+                                            util::Optional<StringPair> user_info) const
 {
-    REALM_ASSERT(user_identity.length() > 0);
-    if (filename_is_reserved(user_identity)) {
+    REALM_ASSERT(local_identity.length() > 0);
+    if (filename_is_reserved(local_identity))
         throw std::invalid_argument("A user can't have an identifier reserved by the filesystem.");
-    }
+
     auto user_path = file_path_by_appending_component(get_base_sync_directory(),
-                                                      user_identity,
+                                                      local_identity,
                                                       util::FilePathType::Directory);
-    util::try_make_dir(user_path);
+    bool success = util::try_make_dir(user_path);
+
+    if (success && user_info && std::get<0>(*user_info) != local_identity) {
+        // Add a text file in the user directory containing the user identity, for backup purposes.
+        auto info_path = util::file_path_by_appending_component(user_path, c_user_info_file);
+        std::ofstream info_file;
+        info_file.open(info_path.c_str());
+        info_file << std::get<0>(*user_info) << "\n" << std::get<1>(*user_info) << "\n";
+        info_file.close();
+    }
     return user_path;
 }
 
-void SyncFileManager::remove_user_directory(const std::string& user_identity) const
+void SyncFileManager::remove_user_directory(const std::string& local_identity) const
 {
-    REALM_ASSERT(user_identity.length() > 0);
-    if (filename_is_reserved(user_identity)) {
+    REALM_ASSERT(local_identity.length() > 0);
+    if (filename_is_reserved(local_identity))
         throw std::invalid_argument("A user can't have an identifier reserved by the filesystem.");
-    }
+
     auto user_path = file_path_by_appending_component(get_base_sync_directory(),
-                                                      user_identity,
+                                                      local_identity,
                                                       util::FilePathType::Directory);
     util::remove_nonempty_dir(user_path);
 }
@@ -313,27 +325,28 @@ bool SyncFileManager::copy_realm_file(const std::string& old_path, const std::st
     return true;
 }
 
-bool SyncFileManager::remove_realm(const std::string& user_identity, const std::string& raw_realm_path) const
+bool SyncFileManager::remove_realm(const std::string& local_identity, const std::string& raw_realm_path) const
 {
-    REALM_ASSERT(user_identity.length() > 0);
+    REALM_ASSERT(local_identity.length() > 0);
     REALM_ASSERT(raw_realm_path.length() > 0);
-    if (filename_is_reserved(user_identity) || filename_is_reserved(raw_realm_path)) {
+    if (filename_is_reserved(local_identity) || filename_is_reserved(raw_realm_path)) {
         throw std::invalid_argument("A user or Realm can't have an identifier reserved by the filesystem.");
     }
     auto escaped = util::make_percent_encoded_string(raw_realm_path);
-    auto realm_path = util::file_path_by_appending_component(user_directory(user_identity), escaped);
+    auto realm_path = util::file_path_by_appending_component(user_directory(local_identity), escaped);
     return remove_realm(realm_path);
 }
 
-std::string SyncFileManager::path(const std::string& user_identity, const std::string& raw_realm_path) const
+std::string SyncFileManager::path(const std::string& local_identity, const std::string& raw_realm_path,
+                                  util::Optional<StringPair> user_info) const
 {
-    REALM_ASSERT(user_identity.length() > 0);
+    REALM_ASSERT(local_identity.length() > 0);
     REALM_ASSERT(raw_realm_path.length() > 0);
-    if (filename_is_reserved(user_identity) || filename_is_reserved(raw_realm_path)) {
+    if (filename_is_reserved(local_identity) || filename_is_reserved(raw_realm_path)) {
         throw std::invalid_argument("A user or Realm can't have an identifier reserved by the filesystem.");
     }
     auto escaped = util::make_percent_encoded_string(raw_realm_path);
-    auto realm_path = util::file_path_by_appending_component(user_directory(user_identity), escaped);
+    auto realm_path = util::file_path_by_appending_component(user_directory(local_identity, user_info), escaped);
     return realm_path;
 }
 
--- a/src/sync/impl/sync_file.cpp
+++ b/src/sync/impl/sync_file.cpp
@@ -251,7 +251,7 @@ std::string SyncFileManager::get_base_sync_directory() const
 }
 
 std::string SyncFileManager::user_directory(const std::string& local_identity,
-                                            util::Optional<StringPair> user_info) const
+                                            util::Optional<SyncUserIdentifier> user_info) const
 {
     REALM_ASSERT(local_identity.length() > 0);
     if (filename_is_reserved(local_identity))
@@ -262,13 +262,13 @@ std::string SyncFileManager::user_directory(const std::string& local_identity,
                                                       util::FilePathType::Directory);
     bool success = util::try_make_dir(user_path);
 
-    if (success && user_info && std::get<0>(*user_info) != local_identity) {
+    if (success && user_info && user_info->user_id != local_identity) {
         // Add a text file in the user directory containing the user identity, for backup purposes.
         auto info_path = util::file_path_by_appending_component(user_path, c_user_info_file);
         std::ofstream info_file;
         info_file.open(info_path.c_str());
         if (info_file.is_open()) {
-            info_file << std::get<0>(*user_info) << "\n" << std::get<1>(*user_info) << "\n";
+            info_file << user_info->user_id << "\n" << user_info->auth_server_url << "\n";
             info_file.close();
         }
     }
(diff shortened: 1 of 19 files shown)


src/sync/sync_manager.cpp
```diff
@@ -286,25 +287,21 @@ bool SyncManager::perform_metadata_update(std::function<void(const SyncMetadataM
 }
 
 std::shared_ptr<SyncUser> SyncManager::get_user(const std::string& identity,
```
austinzheng: Note that this method signature change might be perilous, since it doesn't break existing bindings (but the arguments they pass in will not be in the right order anymore). I may change this.

austinzheng:
This is ready for review. Please feel free to ping me with any questions.

src/sync/impl/sync_file.hpp
```diff
@@ -59,19 +61,21 @@ void remove_nonempty_dir(const std::string& path);
 
 class SyncFileManager {
 public:
+    using StringPair = std::tuple<std::string, std::string>;
```
bdash: Is there a reason you opted for tuple? It'd be clearer if this were a struct with named fields.

src/sync/impl/sync_metadata.cpp
```diff
@@ -110,6 +112,32 @@ SyncMetadataManager::SyncMetadataManager(std::string path,
...
+            // Get all the SyncUserMetadata objects.
+            Results results(old_realm, *old_table);
+
+            // Column indices.
+            size_t old_idx_identity = old_table->get_column_index(c_sync_identity);
+            size_t old_idx_url = old_table->get_column_index(c_sync_auth_server_url);
+            size_t idx_local_uuid = table->get_column_index(c_sync_local_uuid);
+            size_t idx_url = table->get_column_index(c_sync_auth_server_url);
+
+            for (size_t i=0; i<results.size(); i++) {
```
bdash: Whitespace around the `=` and `<` operators, please.

src/sync/impl/sync_metadata.cpp
```diff
@@ -110,6 +112,32 @@ SyncMetadataManager::SyncMetadataManager(std::string path,
...
+                RowExpr entry = results.get(i);
+                // Set the UUID equal to the user identity for existing users.
+                auto identity = entry.get_string(old_idx_identity);
+                table->set_string(idx_local_uuid, entry.get_index(), identity);
+                // Migrate the auth server URLs to a non-nullable property.
+                auto url = entry.get_string(old_idx_url);
+                table->set_string(idx_url, entry.get_index(), url.is_null() ? "" : url);
```
bdash: So the server URL is non-null now, but empty strings are meaningful? Why not leave them as null?

src/sync/impl/sync_metadata.cpp
```diff
@@ -189,110 +212,159 @@ bool SyncMetadataManager::delete_metadata_action(const std::string& original_nam
...
+            table->set_string(schema.idx_identity, row_idx, identity);
+            table->set_string(schema.idx_auth_server_url, row_idx, url);
+            table->set_string(schema.idx_local_uuid, row_idx, uuid);
+            table->set_bool(schema.idx_user_is_admin, row_idx, false);
+            table->set_bool(schema.idx_marked_for_removal, row_idx, false);
+            realm->commit_transaction();
         } else {
             // Someone beat us to adding this user.
-            m_realm->cancel_transaction();
+            row_idx = maybe_row_expr->get_index();
```
bdash: Why get a row index from the `RowExpr` only to then retrieve a `Row` for it, rather than going directly to a `Row`?

src/sync/impl/sync_metadata.cpp
```diff
@@ -189,110 +212,159 @@ bool SyncMetadataManager::delete_metadata_action(const std::string& original_nam
...
-        table->set_bool(m_schema.idx_marked_for_removal, row_idx, false);
-        m_realm->commit_transaction();
-        m_invalid = false;
     } else {
-        m_invalid = m_row.get_bool(m_schema.idx_marked_for_removal);
+        row_idx = maybe_row_expr->get_index();
+    }
+    REALM_ASSERT(row_idx != not_found);
+    auto row = table->get(row_idx);
+    if (make_if_absent || !row.get_bool(schema.idx_marked_for_removal)) {
```
bdash: I'm having a hard time understanding this logic. It seems to read like this: if `make_if_absent` is true, or the user is not marked for removal… then ensure the user is not marked for removal, and return the the user. Is that correct? What's the purpose of the unconditional write transaction?

src/sync/sync_manager.cpp
bdash: Yeah… why change the order?

src/sync/sync_manager.cpp
```diff
@@ -344,20 +350,27 @@ std::shared_ptr<SyncUser> SyncManager::get_current_user() const
...
 {
+    using StringPair = std::tuple<std::string, std::string>;
     std::lock_guard<std::mutex> lock(m_file_system_mutex);
     REALM_ASSERT(m_file_manager);
-    return m_file_manager->path(user_identity, raw_realm_url);
+    const auto user_identity = user.identity();
+    const auto user_local_identity = user.local_identity();
+    std::string local_identity = user_local_identity.value_or(user_identity);
+    util::Optional<StringPair> user_info = (user_local_identity
+                                            ? util::Optional<StringPair>(StringPair(user_identity, user.server_url()))
```
bdash: This'd probably be easier to read if you used an `if` instead of the ternary expression.

```c++
util::Optional<StringPair> user_info;
if (user_local_identity)
    user_info = StringPair(user_identity, user.server_url());
```

src/sync/sync_manager.cpp
```diff
@@ -381,18 +394,15 @@ std::shared_ptr<SyncSession> SyncManager::get_existing_session_locked(const std:
 {
     REALM_ASSERT(!m_session_mutex.try_lock());
     auto it = m_sessions.find(path);
-    if (it == m_sessions.end()) {
-        return nullptr;
-    }
-    return it->second;
+    return (it == m_sessions.end() ? nullptr : it->second);
```
bdash: Parens aren't necessary here.

src/sync/sync_manager.hpp
```diff
@@ -105,18 +105,19 @@ friend class SyncSession;
...
-                                       std::string refresh_token,
-                                       util::Optional<std::string> auth_server_url=none,
-                                       SyncUser::TokenType token_type=SyncUser::TokenType::Normal);
-    // Get an existing user for a given identity, if one exists and is logged in.
-    std::shared_ptr<SyncUser> get_existing_logged_in_user(const std::string& identity) const;
+                                       const std::string& auth_server_url,
+                                       std::string refresh_token);
+
+    // Get an admin token user for the given identifier. 
+    std::shared_ptr<SyncUser> get_admin_token_user(const std::string&, std::string);
```
bdash: The parameters need names since otherwise it's impossible to tell what should be passed.

src/sync/sync_manager.hpp
```diff
@@ -155,8 +156,30 @@ friend class SyncSession;
     // Protects m_users
     mutable std::mutex m_user_mutex;
 
-    // A map of user identities to (shared pointers to) SyncUser objects.
-    std::unordered_map<std::string, std::shared_ptr<SyncUser>> m_users;
+    // A custom key for the user map system. It's basically a hashable tuple.
+    struct UserMapKey {
```
bdash: Does this represent the same information as `StringPair` above? If so, they should probably just be the same type.

src/sync/sync_manager.hpp
```diff
@@ -155,8 +156,30 @@ friend class SyncSession;
...
+        std::string user_id;
+        std::string auth_server_url;
+
+        bool operator==(const UserMapKey& other) const
+        {
+            return user_id == other.user_id && auth_server_url == other.auth_server_url;
+        }
+
+        struct Hasher {
```
bdash: I'd suggest specializing `std::hash` instead, so the type works as-is without requiring the hash function be explicitly specified when using the type with `std::unordered_map`. You can see list.hpp for an example of how we do this for `List`.

tests/sync/file.cpp
```diff
@@ -35,6 +37,8 @@ static void prepare_sync_manager_test(void) {
     util::make_dir(manager_path);
 }
 
+// TODO: test writing the backup file
```
bdash: It looks like you wrote a test for this?

src/sync/impl/sync_file.hpp
austinzheng: No, you're right: a struct makes more sense. I'm still stuck in a Swift mindset.

tests/sync/file.cpp
austinzheng: Yeah, I forgot to remove this comment though.

src/sync/impl/sync_metadata.cpp
austinzheng: My thinking is that it's always an error to omit the URL, and if a URL was missing before that reflects an internal error. (The reason the URL was optional in the first place was to work around some issues with state being stored in `RLMSyncUser` or other bindings' equivalents, issues that no longer exist.)

src/sync/impl/sync_metadata.cpp
austinzheng: Good point. I clearly need to spend more time familiarizing myself with Core's API.

src/sync/impl/sync_metadata.cpp
austinzheng: I rewrote the logic for this function to make things clearer.

src/sync/sync_manager.cpp
austinzheng: I changed this method signature to use my new identifier struct type, so hopefully this won't be an issue anymore.

src/sync/impl/sync_metadata.cpp
bmunkholm: Can I suggest OS to adopt the same style and autoformatting that is being used in core? Then we only need to run the tool and arguments are over :-)

cmelchior:
How does this affect existing Realms?

austinzheng:
This is a purely additive change. Right now, trying to register two users with the same ROS identity but different servers will throw an exception.

src/sync/impl/sync_file.cpp
```diff
@@ -248,27 +250,39 @@ std::string SyncFileManager::get_base_sync_directory() const
...
+
     auto user_path = file_path_by_appending_component(get_base_sync_directory(),
-                                                      user_identity,
+                                                      local_identity,
                                                       util::FilePathType::Directory);
-    util::try_make_dir(user_path);
+    bool success = util::try_make_dir(user_path);
+
+    if (success && user_info && user_info->user_id != local_identity) {
+        // Add a text file in the user directory containing the user identity, for backup purposes.
```
bdash: It might be worth stating that we're intentionally ignoring errors (if it is in fact intentional that we're doing so).

src/sync/impl/sync_metadata.hpp
```diff
@@ -188,12 +207,9 @@ friend class SyncFileActionMetadata;
 
 private:
     SyncUserMetadataResults get_users(bool marked) const;
-
     Realm::Config m_metadata_config;
-
     SyncUserMetadata::Schema m_user_schema;
     SyncFileActionMetadata::Schema m_file_action_schema;
-    mutable std::mutex m_metadata_lock;
```
bdash: Why was this lock necessary in the past? Why is it no longer necessary?

src/sync/sync_manager.cpp
```diff
@@ -344,20 +348,27 @@ std::shared_ptr<SyncUser> SyncManager::get_current_user() const
...
 
-std::string SyncManager::path_for_realm(const std::string& user_identity, const std::string& raw_realm_url) const
+std::string SyncManager::path_for_realm(const SyncUser& user, const std::string& raw_realm_url) const
 {
     std::lock_guard<std::mutex> lock(m_file_system_mutex);
     REALM_ASSERT(m_file_manager);
-    return m_file_manager->path(user_identity, raw_realm_url);
+    const auto user_identity = user.identity();
+    const auto user_local_identity = user.local_identity();
+    std::string local_identity = user_local_identity.value_or(user_identity);
```
bdash: `user_identity`, `user_local_identity` and `local_identity` should all be references to avoid unnecessary copying.

src/sync/sync_manager.hpp
```diff
@@ -104,19 +104,18 @@ friend class SyncSession;
 
     // Get a sync user for a given identity, or create one if none exists yet, and set its token.
     // If a logged-out user exists, it will marked as logged back in.
-    std::shared_ptr<SyncUser> get_user(const std::string& identity,
-                                       std::string refresh_token,
-                                       util::Optional<std::string> auth_server_url=none,
-                                       SyncUser::TokenType token_type=SyncUser::TokenType::Normal);
-    // Get an existing user for a given identity, if one exists and is logged in.
-    std::shared_ptr<SyncUser> get_existing_logged_in_user(const std::string& identity) const;
+    std::shared_ptr<SyncUser> get_user(const SyncUserIdentifier& identifier, std::string refresh_token);
```
bdash: What's the replacement for `get_existing_logged_in_user` for the bindings that use it?

austinzheng: There's no replacement, but if bindings are using it it would be trivial to add it back in (and the tests), and make the few adjustments necessary to work.

@cmelchior, @nirinchev, @nhachicha are any of you using `get_existing_logged_in_user`? Sorry, I should have asked first.

bdash: [They are](https://github.com/search?q=org%3Arealm+get_existing_logged_in_user&type=Code).

austinzheng: That's what I get for being a Cocoa chauvinist.

src/sync/impl/sync_metadata.hpp
austinzheng: I actually don't think this was necessary in the first place, since the config was created in the constructor and only ever read by external consumers. It's definitely not necessary anymore, since the config is created in the constructor, never changed, and only used by code internal to this class.

austinzheng:
I've addressed this next round of feedback.

src/sync/sync_manager.cpp
```diff
@@ -313,28 +308,37 @@ std::shared_ptr<SyncUser> SyncManager::get_user(const std::string& identity,
...
+    users.reserve(m_users.size() + m_admin_token_users.size());
     for (auto& it : m_users) {
         auto user = it.second;
         if (user->state() == SyncUser::State::Active) {
             users.emplace_back(std::move(user));
         }
     }
+    for (auto& it : m_admin_token_users) {
+        users.emplace_back(std::move(it.second));
+    }
```
bdash: Why are we now adding admin token users into this list? Is that in some way related to the general theme of this change?

src/sync/sync_manager.cpp
```diff
@@ -344,20 +348,38 @@ std::shared_ptr<SyncUser> SyncManager::get_current_user() const
...
 
-std::string SyncManager::path_for_realm(const std::string& user_identity, const std::string& raw_realm_url) const
+std::shared_ptr<SyncUser> SyncManager::get_existing_logged_in_user(const SyncUserIdentifier& identifier) const
+{
+    std::lock_guard<std::mutex> lock(m_user_mutex);
+    auto it = m_users.find(identifier);
+    if (it == m_users.end())
+        return nullptr;
+
+    auto ptr = it->second;
```
bdash: `user` would be a better name than `ptr`.

src/sync/sync_manager.cpp
```diff
@@ -344,20 +348,38 @@ std::shared_ptr<SyncUser> SyncManager::get_current_user() const
...
+    return (ptr->state() == SyncUser::State::Active ? ptr : nullptr);
```
bdash: Unnecessary parens? 😃

src/sync/sync_manager.cpp
bdash: Oh, I see. Previously `m_users` contained all users while now admin token users are stored separately. Makes sense.

austinzheng:
@bdash and I discussed privately and there are a few minor changes that should be made to allow auth-token users to work with certain features (like admin Realms).

austinzheng:
I've addressed the feedback. Here are the main changes to admin token users.

- Admin token users will no longer show up either in `all_logged_in_users` or as the `current_user`.
- Admin token users continue to be keyed exclusively upon the client-provided unique ID, but they can be defined with an optional auth server URL.
- When admin token users are assigned a sync user directory, the name of that directory is now prefixed.

The intent is to separate admin token users from normal users as much as possible, since they are only intended as a debug/development tool, while still allowing them to support additional features in the future.

fealebenpae:
> since they are only intended as a debug/development tool

admin token users are essential for ROS to use Realm internally - at the level ROS uses realms, the auth service isn't up yet so it needs to use admin tokens. They are also the only way to use the Global Notifier feature so for realm-js they are very much a supported feature in production. I would ask that we don't change their behavior unless we really need to as that would be a customer-affecting breaking change in realm-js.

austinzheng:
Can you define exactly what you want admin token users to be able to do? There's no specification and everything we've been doing has been guessing at what people need.

fealebenpae:
I'm out of the office this week and am only browsing GitHub casually so I haven't actually been able to look into this change in depth, but I suggest consulting with @alazier and @mbalex99 on use cases for the admin token user in realm-js.

My one wish is to handle the case of opening `/~/foo` realms with the admin token user (which we've had people requesting in realm-js so that they can use the management and permission realms with an admin token user). I've had to implement it manually in https://github.com/realm/realm-js/pull/1111/commits/6168df8b271f6da2aca16eece786fad852ddd11b. Also, based on discussions with @mrackwitz it seems that we can hard-code the admin token user to have the `__auth` identity instead of a random string that bindings have to provide as is now.

austinzheng:
Okay, I have some ideas. I'm going to push another change and then describe what I did.

austinzheng:
I made more changes. Admin token users...

- ...are now part of `current_user` and `all_logged_in_users()` again.
- ...always have their identity as "__auth"
- ...are uniqued and identified based on their auth server URL (which is now mandatory)
- ...have a user directory named by prepending a special prefix to the escaped version of their auth server URL (as opposed to normal users, which use the locally generated UUID identifier)

bdash:
I spoke with Austin yesterday afternoon and expressed my concern that the current requirement that auth server URL be mandatory for admin token users forces an API breaking change on all bindings. While that requirement is desirable longer term, we need to be able to ship the bug fix aspect of this change in our stable branch to address #296. We should ensure that all bindings are aware of the desire to make the auth server URL mandatory, and look to make that change in their APIs for our upcoming releases in support of sync v2.0.

austinzheng:
I updated my PR just now based on what Mark and I discussed yesterday. This new design should require no breaking changes for now while making it as easy as possible to "migrate" bindings to the new semantics (URLs as identifiers and mandatory) once we are able to make a breaking change.

src/sync/sync_manager.cpp
```diff
@@ -313,28 +311,82 @@ std::shared_ptr<SyncUser> SyncManager::get_user(const std::string& identity,
...
+
+            auto new_user = std::make_shared<SyncUser>(std::move(*token),
+                                                       c_admin_identity,
+                                                       server_url,
+                                                       c_admin_identity + server_url,
+                                                       SyncUser::TokenType::Admin);
+            m_admin_token_users.insert({ server_url, new_user });
+            user = std::move(new_user);
+        } else {
+            user = it->second;
```
bdash: If the user already exists, we probably don't want to be renaming the user directory out from under it since there might already be active sync sessions.

My suggestion would be to invert the `if` condition and turn the "user found" case into an early return in the true branch, then handle the "user not found" case after the `if` statement.

src/sync/sync_manager.cpp
```diff
@@ -313,28 +311,82 @@ std::shared_ptr<SyncUser> SyncManager::get_user(const std::string& identity,
...
+    {
+        std::lock_guard<std::mutex> lock(m_user_mutex);
+        // Look up the user based off the server URL.
+        auto it = m_admin_token_users.find(server_url);
+        if (it == m_admin_token_users.end()) {
+            // No existing user.
+            if (!token)
+                throw std::invalid_argument("User did not exist, but token was not provided.");
+
+            auto new_user = std::make_shared<SyncUser>(std::move(*token),
```
bdash: Why bother with `new_user` rather than just assigning to `user` directly?

src/sync/sync_manager.hpp
```diff
@@ -104,19 +104,34 @@ friend class SyncSession;
...
+
+    // Get or create an admin token user based on the given identity.
+    // Please note: a future version will remove this method and deprecate the
+    // use of identities for admin users completely.
+    std::shared_ptr<SyncUser> get_admin_token_user_from_identity(const std::string& identity,
+                                                                 util::Optional<std::string> server_url,
+                                                                 util::Optional<std::string> token);
```
bdash: As we discussed on Slack, I don't think there's a good reason for `token` to be optional.

src/sync/sync_user.cpp
```diff
@@ -42,11 +45,21 @@ SyncUser::SyncUser(std::string refresh_token, std::string identity,
...
+            auto metadata = manager.get_or_make_user_metadata(m_identity, m_server_url);
+            metadata->set_user_token(m_refresh_token);
+            m_is_admin = metadata->is_admin();
+            m_local_identity = metadata->local_uuid();
         });
+        if (!updated)
+            m_local_identity = m_identity;
+    } else {
+        // Admin token users. The local identity serves as the directory path.
+        if (!local_identity)
```
bdash: Does the absence of a local identity at this point indicate a problem with data that the user has provided, or a failure elsewhere within our code to enforce our preconditions? If the latter, this should be an assertion rather than a exception.

src/sync/sync_manager.cpp
austinzheng: Yes, this is a good point.

On a related note, I think it should be considered an error to get a token user using _identity_ and specifying a URL, and then later try getting that same token user using _identity_ but not specifying a URL. This is already a pretty inexplicable scenario and not one that is difficult to avoid. (The reason I say so is because I don't want an "old-style" directory being created after an existing old-style directory is renamed to use the new-style name.)

src/sync/sync_user.cpp
austinzheng: This should indeed be an assertion, given that it's a private API and we should know exactly what we're passing in.

src/sync/impl/sync_file.cpp
```diff
@@ -248,31 +250,63 @@ std::string SyncFileManager::get_base_sync_directory() const
...
     return user_path;
 }
 
-void SyncFileManager::remove_user_directory(const std::string& user_identity) const
+void SyncFileManager::remove_user_directory(const std::string& local_identity) const
 {
-    REALM_ASSERT(user_identity.length() > 0);
-    if (filename_is_reserved(user_identity)) {
+    REALM_ASSERT(local_identity.length() > 0);
+    if (filename_is_reserved(local_identity))
```
bdash: In `user_directory` we percent-encode the identity before checking if the filename is reserved. Here we do it after. This seems like something we should be consistent about.

src/sync/impl/sync_file.cpp
```diff
@@ -248,31 +250,63 @@ std::string SyncFileManager::get_base_sync_directory() const
...
+                                                      util::make_percent_encoded_string(local_identity),
                                                       util::FilePathType::Directory);
     util::remove_nonempty_dir(user_path);
 }
 
+bool SyncFileManager::try_rename_user_directory(const std::string& old_name, const std::string& new_name) const
+{
+    REALM_ASSERT_DEBUG(old_name.length() > 0 && new_name.length() > 0);
+    const std::string& base = get_base_sync_directory();
+    auto old_path = file_path_by_appending_component(base,
```
bdash: Do we need to check whether `old_name` and `new_name` are reserved filenames?

PR #499, opened 2017-07-18
Add support for type-templated test cases

tgoyne:
Original motivation for this was for writing tests for primitive lists/results, but it also turned out to be useful for refactoring some tests to both be shorter and involve fewer macros.

### DIFF
--- a/tests/util/templated_test_case.hpp
+++ b/tests/util/templated_test_case.hpp
@@ -16,11 +16,15 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
+// Work around a msvc preprocessor bug that considers __VA_ARGS__ a single token
+// by forcing another round of macro expansion whereever it's used.
+#define REALM_EXPAND(x) x
+
 // Define a Catch test case templated on up to 20 types
 // The current type is exposed inside the test as `TestType`
 #define TEMPLATE_TEST_CASE(name, ...) \
-    REALM_TEMPLATE_TEST_CASE(name, INTERNAL_CATCH_UNIQUE_NAME(REALM_TEMPLATE_TEST_), \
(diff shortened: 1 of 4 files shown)


tests/util/templated_test_case.hpp
```diff
@@ -0,0 +1,148 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+// Define a Catch test case templated on up to 20 types
+// The current type is exposed inside the test as `TestType`
+#define TEMPLATE_TEST_CASE(name, ...) \
```
bdash: Just to make sure I'm following this correctly:

`TEMPLATE_TEST_CASE` expands into a function template with a unique name, whose implementation is the body of the block following the `TEMPLATE_TEST_CASE`, and a regular `TEST_CASE` with one `SECTION` per type, with each section containing a call to the function template with the type explicitly specified?

tgoyne: Yes.

An example of the test failure message (for `TEMPLATE_TEST_CASE("primitive list", BoxedOptional<::Int>) { SECTION("swap()") { ... } }`) is:

```
primitive list
  BoxedOptional<::Int>
  swap()
/src/realm-object-store/tests/primitive_list.cpp:345

/src/realm-object-store/tests/primitive_list.cpp:351: FAILED:
  REQUIRE( list.get<T>(0) == values[1] )
with expansion:
  some(2) == some(1)
```

bdash:
It looks like this needs some adjustments to work with MSVC.

PR #501, opened 2017-07-21
Optimize in a write transaction

ironage:
As discussed, we should be using a write transaction for this code to be effective.

### DIFF
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -625,11 +625,12 @@ bool Realm::compact()
         throw InvalidTransactionException("Can't compact a Realm within a write transaction");
     }
 
-    Group& group = read_group();
+    verify_open();
+    Group& group = m_shared_group->begin_write();
     for (auto &object_schema : m_schema) {
         ObjectStore::table_for_object_type(group, object_schema.name)->optimize();
     }
-    m_shared_group->end_read();
+    m_shared_group->commit();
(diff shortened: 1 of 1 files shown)


bdash:
Do we have any concerns about the fact that this will cause us to use a bunch of core codepaths that apparently haven't seen much real-world use in recent times?

ironage:
In the core meeting today @finnschiermer and @rrrlasse  discussed possibly delaying this change so we can test a bit longer. I will be increasing test coverage of `StringEnumColumn` this week in core. Since I don't think there is any rush on this, can we wait at least a week?

bdash:
That's fine with me.

ironage:
This should be good to go once https://github.com/realm/realm-core/pull/2780 is released.

ironage:
https://github.com/realm/realm-core/pull/2780 was released in core version 4.0.0. So this can go in anytime.

ironage:
Please hold this, libfuzzer just found another problem moving string enum columns around - I plan to make another fix next week.

ironage:
Please wait for https://github.com/realm/realm-core/pull/2924 to be released.

Zhuinden:
https://github.com/realm/realm-core/pull/2926 is merged?

tgoyne:
No longer applicable as we don't optimize at all and enum columns will work differently in core 6.

PR #503, opened 2017-07-24
Add schema change callback

beeender:
There are some difficulties for java binding to hold a pointer to
Realm::schema() and access group though column indices in the Schema
instance, mostly because of performance concerns.
This commit enable a possibility for bindings to know when the Realm
instance's schema gets changed.
    
- Add schema_changed_function to Realm config.
- schema_changed_function will be called when the schema changes
  delivered to the Realm instance to notify binding to refresh the
  schema information.

### DIFF
--- a/src/shared_realm.hpp
+++ b/src/shared_realm.hpp
@@ -213,17 +213,7 @@ public:
     // encryption key will raise an exception.
     static SharedRealm get_shared_realm(Config config);
 
-    // This is called by RealmCoordinator to update a Realm to a given schema, using
-    // the Realm's pre-set schema mode.
-    // By setting notify to false, schema changed function won't be called to ensure
-    // the notification will only be sent after Realm::get_shared_realm() returns.
-    void update_schema_notify(Schema schema, uint64_t version,
-                       MigrationFunction migration_function,
-                       DataInitializationFunction initialization_function,
-                       bool in_transaction, bool notify);
-
     // Updates a Realm to a given schema, using the Realm's pre-set schema mode.
-    // Schema changed function will be called if it exists in the config.
     void update_schema(Schema schema, uint64_t version=0,
                        MigrationFunction migration_function=nullptr,
                        DataInitializationFunction initialization_function=nullptr,
--- a/src/shared_realm.hpp
+++ b/src/shared_realm.hpp
@@ -145,6 +145,12 @@ public:
     // because it's not crash safe! It may corrupt your database if something fails
     using ShouldCompactOnLaunchFunction = std::function<bool (uint64_t total_bytes, uint64_t used_bytes)>;
 
+    // A callback function which is called when this Realm's schema is changed through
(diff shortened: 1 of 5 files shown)


tgoyne:
This is something that needs to be done by `Realm`, not by `Schema`.

While the standard does not strictly require it, `swap()` and move operations must be `noexcept`, unable to fail, and have no side effects for them to actually be usable in practice.

src/shared_realm.hpp
```diff
@@ -158,10 +164,12 @@ class Realm : public std::enable_shared_from_this<Realm> {
         // Optional schema for the file.
         // If the schema and schema version are supplied, update_schema() is
         // called with the supplied schema, version and migration function when
-        // the Realm is actually opened and not just retreived from the cache
+        // the Realm is actually opened and not just retrieved from the cache
         util::Optional<Schema> schema;
         uint64_t schema_version = -1;
         MigrationFunction migration_function;
+        // Called when the schema changes are delivered to this Realm instance.
+        SchemaChangedFunction schema_changed_function;
```
tgoyne: I think this would make more sense as a member function on `BindingContext`. Based on the doc comment for `update_schema_notify()` it would remove the need for the `notify` parameter, and currently things are split up such that user-provided callbacks are supplied on the `Config` object, while callbacks from the object store to binding are on the `BindingContext` (which is not a split I ever considered before now, but now that I've noticed it seems sensible?).

tgoyne: After thinking about how we'd use this in Cocoa a bit, it occurred to me that we actually need to notify both the binding and the end user to properly support a truly dynamic schema, since the Browser also needs to know when the schema has changed. I think this strengthens my view that notifying the binding should be done via the context.

beeender:
@tgoyne updated

src/shared_realm.hpp
```diff
@@ -213,7 +213,17 @@ class Realm : public std::enable_shared_from_this<Realm> {
     // encryption key will raise an exception.
     static SharedRealm get_shared_realm(Config config);
 
+    // This is called by RealmCoordinator to update a Realm to a given schema, using
+    // the Realm's pre-set schema mode.
+    // By setting notify to false, schema changed function won't be called to ensure
+    // the notification will only be sent after Realm::get_shared_realm() returns.
+    void update_schema_notify(Schema schema, uint64_t version,
```
bdash: It looks like this declaration wasn't removed when the implementation and calls to it were removed.

src/shared_realm.hpp
```diff
@@ -213,7 +213,17 @@ class Realm : public std::enable_shared_from_this<Realm> {
...
+                       MigrationFunction migration_function,
+                       DataInitializationFunction initialization_function,
+                       bool in_transaction, bool notify);
+
     // Updates a Realm to a given schema, using the Realm's pre-set schema mode.
+    // Schema changed function will be called if it exists in the config.
```
bdash: This comment is no longer accurate.

PR #505, opened 2017-07-25
Add 'Using Docker' section to readme

jpsim:


README.md
```diff
@@ -108,6 +108,26 @@ adb push tests/tests /data/local/tmp
...
+
+The `Dockerfile` included in this repo will provision a Docker image suitable
+for building and running tests for both the Linux and Android platforms.
+
+```
+# Build Docker image from Dockerfile
+docker build -t "objectstore" .
+# Run bash interactively from the built Docker image,
+# mounting the current directory
+docker run -it -v `pwd`:`pwd` -w `pwd` objectstore bash
```
emanuelez: I'd prefer something like `docker run --rm -it -v $(pwd):/tmp -w /tmp objectstore bash`

README.md
```diff
@@ -108,6 +108,26 @@ adb push tests/tests /data/local/tmp
...
+# Build the object store for Linux and run tests
+> cmake .
```
emanuelez: Let's suggest a more idiomatic use of CMake

```
mkdir build
cd build
cmake ..
make
make run-tests
```

bdash: Elsewhere in the object store README we assume that you're building in the root of the source directory. Using a different directory also works, but makes the instructions more convoluted than they need to be.

README.md
jpsim: Done in e67d4d0.

README.md
jpsim: I stuck with the instructions already in the readme. If you think they can be improved, I suggest you make a PR changing both the main instructions along with their use in this section, so we stay consistent.

jpsim:
Thanks for the feedback y'all!

PR #507, opened 2017-07-28
Allow `realm::Permission` to be constructed from `realm::Results`

austinzheng:
Allow `realm::Permission` to be used without involving `realm::PermissionResults` at all.

Prerequisite work for https://github.com/realm/realm-cocoa/pull/5168.

### DIFF
--- a/src/sync/sync_permission.cpp
+++ b/src/sync/sync_permission.cpp
@@ -67,6 +67,25 @@ int64_t ns_since_unix_epoch(const system_clock::time_point& point)
 
 // MARK: - Permission
 
+Permission::Permission(realm::Results& results, size_t index)
+{
+    Object permission(results.get_realm(), results.get_object_schema(), results.get(index));
+    CppContext context;
+    path = any_cast<std::string>(permission.get_property_value<util::Any>(context, "path"));
+    access = extract_access_level(permission, context);
+    condition = Condition(any_cast<std::string>(permission.get_property_value<util::Any>(context, "userId")));
+    updated_at = any_cast<Timestamp>(permission.get_property_value<util::Any>(context, "updatedAt"));
+}
+
+Permission::Permission(std::string path, AccessLevel access, Condition condition, util::Optional<Timestamp> updated_at)
+{
+    this->path = std::move(path);
+    this->access = access;
+    this->condition = std::move(condition);
+    if (updated_at)
+        this->updated_at = std::move(*updated_at);
+}
+
 std::string Permission::description_for_access_level(AccessLevel level)
 {
     switch (level) {
@@ -138,7 +157,22 @@ private:
 
 void Permissions::get_permissions(std::shared_ptr<SyncUser> user,
                                   PermissionResultsCallback callback,
-                                  const ConfigMaker& make_config)
+                                  const ConfigMaker& config)
+{
+    // callback(std::make_unique<PermissionResults>(results->filter(std::move(query))), nullptr);
+    auto cb = [callback=std::move(callback)](Results results, std::exception_ptr ex) {
+        if (ex) {
+            callback(nullptr, ex);
+        } else {
+            callback(std::make_unique<PermissionResults>(std::move(results)), nullptr);
+        }
+    };
+    get_raw_permissions(std::move(user), std::move(cb), config);
+}
+
+void Permissions::get_raw_permissions(std::shared_ptr<SyncUser> user,
+                                      RawPermissionResultsCallback callback,
+                                      const ConfigMaker& make_config)
 {
     auto realm = Permissions::permission_realm(user, make_config);
     auto table = ObjectStore::table_for_object_type(realm->read_group(), "Permission");
@@ -150,7 +184,7 @@ void Permissions::get_permissions(std::shared_ptr<SyncUser> user,
     // unregistered by nulling out the `results_wrapper` container.
     auto async = [results, callback=std::move(callback)](CollectionChangeSet, std::exception_ptr ex) mutable {
         if (ex) {
-            callback(nullptr, ex);
+            callback(Results(), ex);
             results.reset();
             return;
         }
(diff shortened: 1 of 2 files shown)


src/sync/sync_permission.cpp
```diff
@@ -67,6 +67,25 @@ int64_t ns_since_unix_epoch(const system_clock::time_point& point)
...
+    CppContext context;
+    path = any_cast<std::string>(permission.get_property_value<util::Any>(context, "path"));
+    access = extract_access_level(permission, context);
+    condition = Condition(any_cast<std::string>(permission.get_property_value<util::Any>(context, "userId")));
+    updated_at = any_cast<Timestamp>(permission.get_property_value<util::Any>(context, "updatedAt"));
+}
+
+Permission::Permission(std::string path, AccessLevel access, Condition condition, util::Optional<Timestamp> updated_at)
+{
+    this->path = std::move(path);
```
bdash: Initialize these members in the member initializer list?

src/sync/sync_permission.cpp
```diff
@@ -67,6 +67,25 @@ int64_t ns_since_unix_epoch(const system_clock::time_point& point)
 
 // MARK: - Permission
 
+Permission::Permission(realm::Results& results, size_t index)
```
bdash: Why take a `Results` and an index rather than an `Object`?

src/sync/sync_permission.hpp
```diff
@@ -150,11 +156,19 @@ class Permissions {
     using ConfigMaker = std::function<Realm::Config(std::shared_ptr<SyncUser>, std::string url)>;
 
     // Callback used to asynchronously vend a `PermissionResults` object.
+    // DEPRECATED. Will be replaced by RawPermissionResultsCallback in the future.
     using PermissionResultsCallback = std::function<void(std::unique_ptr<PermissionResults>, std::exception_ptr)>;
 
+    // Callback used to asynchronously vend permissions results.
+    using RawPermissionResultsCallback = std::function<void(Results, std::exception_ptr)>;
```
bdash: Can you elaborate on why the new type alias / function is necessary given what this PR claims it is about?

src/sync/sync_permission.cpp
austinzheng: To save a bit of typing, but thinking about it taking an `Object` is cleaner and makes more sense.

src/sync/sync_permission.hpp
austinzheng: Our existing `get_permissions()` static API gets the raw `realm::Results`, wraps it in a `PermissionResults`, and then returns it in that callback.

I don't want that wrapping step anymore; I just want the raw `realm::Results`, but I don't want to break the other bindings, hence this new callback and new static method that work the same as the existing ones.

src/sync/sync_permission.cpp
```diff
@@ -67,6 +67,22 @@ int64_t ns_since_unix_epoch(const system_clock::time_point& point)
 
 // MARK: - Permission
 
+Permission::Permission(Object& permission)
```
bdash: Const?

src/sync/sync_permission.cpp
```diff
@@ -67,6 +67,22 @@ int64_t ns_since_unix_epoch(const system_clock::time_point& point)
...
+{
+    CppContext context;
+    path = any_cast<std::string>(permission.get_property_value<util::Any>(context, "path"));
+    access = extract_access_level(permission, context);
+    condition = Condition(any_cast<std::string>(permission.get_property_value<util::Any>(context, "userId")));
+    updated_at = any_cast<Timestamp>(permission.get_property_value<util::Any>(context, "updatedAt"));
+}
+
+Permission::Permission(std::string path, AccessLevel access, Condition condition, util::Optional<Timestamp> updated_at)
```
bdash: Taking an `Optional<Timestamp>` is a little weird given that `Timestamp` itself has a null state. Can this just take a `Timestamp`?

src/sync/sync_permission.cpp
```diff
@@ -138,7 +154,22 @@ struct NotificationWrapper : public T {
 
 void Permissions::get_permissions(std::shared_ptr<SyncUser> user,
                                   PermissionResultsCallback callback,
-                                  const ConfigMaker& make_config)
+                                  const ConfigMaker& config)
```
bdash: `make_config` seems clearer, given the type.

src/sync/sync_permission.cpp
```diff
@@ -138,7 +154,22 @@ struct NotificationWrapper : public T {
...
+{
+    // callback(std::make_unique<PermissionResults>(results->filter(std::move(query))), nullptr);
```
bdash: This should be removed.

src/sync/sync_permission.hpp
```diff
@@ -88,6 +88,12 @@ struct Permission {
     Condition condition;
 
     Timestamp updated_at;
+
+    /// Create a Permission value from a raw Object.
```
bdash: I'm not sure what "raw" means in this context.

src/sync/sync_permission.cpp
austinzheng: I tried `const` but the implementation of `Object` won't allow it; there are a whole bunch of things on that class that would have to be audited and annotated :(.

src/sync/sync_permission.hpp
nirinchev: I don't think there are any other bindings using this API, so there would be nothing to break.

austinzheng: It's my understanding that JS uses this, although that might have changed since we first wrote this code.

bdash: JS never merged their PR that uses this: realm/realm-js#1054

austinzheng: :pained_grimace:

PR #508, opened 2017-08-01
Update schema before initialization callback

beeender:
Binding needs to see the latest schema in the initialization callback.

### DIFF
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -451,6 +451,14 @@ void Realm::update_schema(Schema schema, uint64_t version, MigrationFunction mig
     }
 
     if (initialization_function && old_schema_version == ObjectStore::NotVersioned) {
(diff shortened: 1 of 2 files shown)


src/shared_realm.cpp
```diff
@@ -451,6 +451,14 @@ void Realm::update_schema(Schema schema, uint64_t version, MigrationFunction mig
     }
 
     if (initialization_function && old_schema_version == ObjectStore::NotVersioned) {
+        // Initialization function needs to see the latest schema
+        uint64_t temp_version = ObjectStore::get_schema_version(read_group());
```
tgoyne: `temp_version` here should always be `NotVersioned`, so this seems sort of redundant.

beeender: it is not always `NotVersioned`, it will be set to the version when the schema & version has been initialized.

PR #512, opened 2017-08-10
Add support for arrays of all supported data types

tgoyne:
Depends on https://github.com/realm/realm-core/pull/2785, so this will fail on CI for now.

The only API changes here are that `List` and `Results` now have templated versions of `get()`/`set()`/etc. so that they can be used with non-object types; everything else is just updating the innards of everything to work properly when non-Object array types are used.

### DIFF
--- a/src/impl/transact_log_handler.cpp
+++ b/src/impl/transact_log_handler.cpp
@@ -47,8 +47,9 @@ private:
 
     struct ListInfo {
         BindingContext::ObserverState* observer;
-        size_t col;
         _impl::CollectionChangeBuilder builder;
+        size_t col;
+        size_t initial_size;
     };
     std::vector<ListInfo> m_lists;
     VersionID m_version;
@@ -77,9 +78,10 @@ KVOAdapter::KVOAdapter(std::vector<BindingContext::ObserverState>& observers, Bi
         auto table = group.get_table(observer.table_ndx);
         for (size_t i = 0, count = table->get_column_count(); i < count; ++i) {
             auto type = table->get_column_type(i);
-            if (type != type_LinkList && type != type_Table)
-                continue;
-            m_lists.push_back({&observer, i, {}});
+            if (type == type_LinkList)
+                m_lists.push_back({&observer, {}, i, size_t(-1)});
+            else if (type == type_Table)
+                m_lists.push_back({&observer, {}, i, table->get_subtable_size(i, observer.row_ndx)});
         }
     }
 
@@ -142,8 +144,14 @@ void KVOAdapter::before(SharedGroup& sg)
     }
 
     for (auto& list : m_lists) {
-        if (list.builder.empty())
+        if (list.builder.empty()) {
+            // We may have pre-emptively marked the column as modified if the
+            // LinkList was selected but the actual changes made ended up being
+            // a no-op
+            if (list.col < list.observer->changes.size())
+                list.observer->changes[list.col].kind = BindingContext::ColumnInfo::Kind::None;
             continue;
+        }
         // If the containing row was deleted then changes will be empty
         if (list.observer->changes.empty()) {
             REALM_ASSERT_DEBUG(tables[new_table_ndx(list.observer->table_ndx)].deletions.contains(list.observer->row_ndx));
@@ -154,17 +162,34 @@ void KVOAdapter::before(SharedGroup& sg)
         auto& builder = list.builder;
         auto& changes = list.observer->changes[list.col];
 
+        builder.modifications.remove(builder.insertions);
+
         // KVO can't express moves (becaue NSArray doesn't have them), so
         // transform them into a series of sets on each affected index when possible
-        if (!builder.insertions.empty() && builder.insertions.count() == builder.moves.size()) {
+        if (!builder.moves.empty() && builder.insertions.count() == builder.moves.size() && builder.deletions.count() == builder.moves.size()) {
             changes.kind = BindingContext::ColumnInfo::Kind::Set;
             changes.indices = builder.modifications;
+            changes.indices.add(builder.deletions);
 
-            size_t start = std::min(builder.insertions.begin()->first, builder.deletions.begin()->first);
-            size_t end = std::max(std::prev(builder.insertions.end())->second,
-                                  std::prev(builder.deletions.end())->second);
-            for (size_t i = start; i < end; ++i)
-                changes.indices.add(i);
+            auto in_range = [](auto& it, auto end, size_t i) {
+                if (it != end && i >= it->second)
+                    ++it;
+                return it != end && i >= it->first && i < it->second;
+            };
+
+            auto del_it = builder.deletions.begin(), del_end = builder.deletions.end();
+            auto ins_it = builder.insertions.begin(), ins_end = builder.insertions.end();
+            size_t start = std::min(ins_it->first, del_it->first);
+            size_t end = std::max(std::prev(ins_end)->second, std::prev(del_end)->second);
+            ptrdiff_t shift = 0;
+            for (size_t i = start; i < end; ++i) {
+                if (in_range(del_it, del_end, i))
+                    --shift;
+                else if (in_range(ins_it, ins_end, i + shift))
+                    ++shift;
+                if (shift != 0)
+                    changes.indices.add(i);
+            }
         }
         // KVO can't express multiple types of changes at once
         else if (builder.insertions.empty() + builder.modifications.empty() + builder.deletions.empty() < 2) {
@@ -181,7 +206,12 @@ void KVOAdapter::before(SharedGroup& sg)
         else {
             REALM_ASSERT(!builder.deletions.empty());
             changes.kind = BindingContext::ColumnInfo::Kind::Remove;
-            changes.indices = builder.deletions;
+            // Table clears don't come with the size, so we need to fix up the
+            // notification to make it just delete all rows that actually existed
+            if (std::prev(builder.deletions.end())->second > list.initial_size)
+                changes.indices.set(list.initial_size);
+            else
+                changes.indices = builder.deletions;
         }
     }
     m_context->will_change(m_observers, m_invalidated);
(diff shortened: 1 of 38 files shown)


src/impl/transact_log_handler.cpp
```diff
@@ -279,7 +310,7 @@ class TransactLogValidationMixin {
     bool insert_empty_rows(size_t, size_t, size_t, bool) { return true; }
     bool erase_rows(size_t, size_t, size_t, bool) { return true; }
     bool swap_rows(size_t, size_t) { return true; }
-    bool clear_table() noexcept { return true; }
+    bool clear_table(size_t) noexcept { return true; }
```
bdash: It looks like your core PR that changes this method signature is based on `develop`, meaning it'll only be available in core v3.x. Can we maintain compatibility with core v2.x without too much extra hassle?

src/object_store.cpp
```diff
@@ -116,6 +117,12 @@ void insert_column(Group& group, Table& table, Property const& property, size_t
         table.insert_column_link(col_ndx, is_array(property.type) ? type_LinkList : type_Link,
                                  property.name, *link_table);
     }
+    else if (is_array(property.type)) {
+        DescriptorRef desc;
+        table.insert_column(col_ndx, type_Table, property.name, &desc);
+        desc->add_column(to_core_type(property.type & ~PropertyType::Flags), "!ARRAY_VALUE",
```
bdash: Since we're using this magic column name in a few different places, it'd be worth pulling it into a named constant.

src/results.cpp
```diff
@@ -590,17 +618,37 @@ Results Results::filter(Query&& q) const
...
     return Results(m_realm, get_query(), std::move(new_order));
 }
 
+Results Results::distinct(std::vector<std::string> const& keypaths) const
+{
+    if (keypaths.empty())
+        return *this;
+    if (get_type() != PropertyType::Object) {
+        if (keypaths.size() != 1)
+            throw std::invalid_argument(util::format("Cannot sort array of '%1' on more than one key path",
```
bdash: s/sort/distinct/ (though that reads a bit awkwardly).

src/util/compiler.hpp
```diff
@@ -39,4 +39,11 @@
 #define REALM_COMPILER_HINT_UNREACHABLE abort
 #endif
 
+#if __GNUC__ < 5 && !defined(__clang__)
+#define REALM_NO_BRACED_INIT 1
+#else
+#define REALM_NO_BRACED_INIT 0
```
bdash: This does not appear to be used?

src/results.cpp
tgoyne: I kinda feel that using distinct as a verb is awkward enough that it's not worth making `parse_keypath()` use the correct word, and then concluded that these might as well be consistent with that. I don't feel strongly that this is the right thing to do, though.

bdash: Yeah, I'm not sure how to use distinct in the error and still have it be readable. I'm fine with leaving it as-is.

src/list.cpp
```diff
@@ -254,39 +401,66 @@ util::Optional<Mixed> List::min(size_t column)
     return as_results().min(column);
 }
 
-util::Optional<Mixed> List::sum(size_t column)
+Mixed List::sum(size_t column)
 {
-    return as_results().sum(column);
+    // Results::sum() returns none only for Empty results
```
bdash: To make sure I understand correctly: you're referring to `Mode::Empty` here, rather than a results that happens to contain 0 items, and since we can never have a results in `Mode::Empty` here we can discount the possibility of encountering `none`?

Why does `Results::sum` return different results if it's in `Mode::Empty` vs containing 0 items? Aren't the two intended to provide the same behavior?

src/sync/sync_session.cpp
```diff
@@ -461,7 +461,7 @@ void SyncSession::handle_error(SyncError error)
             // Connection level errors
             case ProtocolError::connection_closed:
             case ProtocolError::other_error:
-            case ProtocolError::pong_timeout:
+//            case ProtocolError::pong_timeout:
```
bdash: I assume these changes were made in order to build against whichever version of sync you happened to be testing with? We should be sure to revert these and determine the right way to accommodate sync's API changes.

tgoyne: Yes, that entire commit is just to make it compile with a not-yet-released version of sync, and adapting to the changes there once it's actually released should be done separately from this PR.

src/list.cpp
tgoyne: In Mode::Empty it doesn't know what type of value to return since it gets the type from the table, and it can't get it from the ObjectSchema because it looks that up from the Table name. This is all a pre-existing design issue that should probably be fixed at some point.

bdash: Ah yes, I remember discussing that in the past.

src/impl/transact_log_handler.cpp
```diff
@@ -153,17 +162,34 @@ void KVOAdapter::before(SharedGroup& sg)
...
+            size_t end = std::max(std::prev(ins_end)->second, std::prev(del_end)->second);
+            ptrdiff_t shift = 0;
+            for (size_t i = start; i < end; ++i) {
+                if (in_range(del_it, del_end, i))
+                    --shift;
+                else if (in_range(ins_it, ins_end, i + shift))
+                    ++shift;
+                if (shift != 0)
+                    changes.indices.add(i);
+            }
```
tgoyne: The KVO tests for move() and swap() all rely on this.

Previously move() just marked the full potentially-modified range as modified, and swap() marked the two relevant indices as modified (and didn't actually record that the rows were moved). However, this logic doesn't work for subtables, as Table::move_row() is emulated via insert/swap/erase and so for KVO it'd decay to just SetAll (because there's multiple kinds of changes). As a result I adjusted it so that swap() is actually handled as a pair of moves, which made the insert/swap/erase dance work, but also made it so that swap() would mark all the rows in between the swapped rows as modified, even though they didn't change. This code now enumerates over all of the rows which might have been shifted by moves and calculates if it actually has been shifted, or if the combination of moves has resulted in the row being in the same place as it started.

tests/primitive_list.cpp
```diff
@@ -0,0 +1,799 @@
...
+using namespace realm;
+
+template<PropertyType prop_type, typename T>
+struct Base {
+    using Type = T;
+    using Wrapped = T;
+    using Boxed = T;
+
+    static PropertyType property_type() { return prop_type; }
+    static std::vector<int64_t> values() { return {3, 1, 2}; }
```
bdash: It looks like the expectation is that each subclass provides an implementation of `values()`. This base class version is unused, and wouldn't work for most subclasses anyway. Probably best to just omit it?

tests/primitive_list.cpp
```diff
@@ -0,0 +1,799 @@
...
+struct StringifyingContext {
+    template<typename T>
+    std::string box(T value)
+    {
+        std::stringstream ss;
+        ss << value;
+        return ss.str();
+    }
+
+    std::string box(RowExpr row) { return std::to_string(row.get_index()); }
```
bdash: IIRC we need to use `util::to_string` due to Android lacking `std::to_string`.

src/impl/transact_log_handler.cpp
bdash: Thanks for the explanation.

src/list.cpp
```diff
@@ -117,126 +144,246 @@ void List::verify_in_transaction() const
...
+size_t List::find(T const& value) const
 {
-    return m_link_view->get(row_ndx).get_index();
+    verify_attached();
+    return m_table->find_first(0, value);
 }
 
-size_t List::find(ConstRow const& row) const
+template<>
+size_t List::find(RowExpr const& row) const
```
bdash: [GCC 5.4 appears to take issue with this](https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-object-store/detail/PR-512/4/pipeline/35):
```
/src/list.cpp:193:39: error: specialization of 'template<class T> size_t realm::List::find(const T&) const' in different namespace [-fpermissive]
 size_t List::find(RowExpr const& row) const
                                       ^
In file included from src/list.cpp:19:0:
src/list.hpp:85:12: error:   from definition of 'template<class T> size_t realm::List::find(const T&) const' [-fpermissive]
     size_t find(T const& value) const;
```

I guess this is because the file has `using namespace realm;` rather than its contents being enclosed in `namespace realm { … }`?

bdash:
This looks good to go once the compilation issues with GCC are addressed.

tgoyne:
Remaining CI failure is probably due to https://github.com/realm/realm-core/pull/2807.

nirinchev:
I have a notification related test case that is deadlocking, though I'm not sure if it's an OS bug, or something in our binding.

The minimal repro test adds an object with a list with few elements (element type doesn't seem to make a difference). Then, we call `List::add_notification_callback`, and finally `List::remove_all`. After that, upon refresh, we get a `CollectionChangeSet` that contains a single Chunk, whose count is reported as the maximum value a `size_t` can hold.

Am I reading the debug information correctly (don't have too much experience with native debug tools) and is this something that's tested by OS unit tests?

tgoyne:
It reporting -1 rather than the previous size is unsurprising (but shouldn't be hard to fix), but that shouldn't result in a deadlock.

nirinchev:
No, the "deadlock" is in our code - we're trying to create a vector of the same size (uint64.max), which never completes. Sorry for the confusing description.

nirinchev:
When building for Android, we get the following compilation errors:
```
jni/../src/object-store/src/list.cpp:189: error: undefined reference to 'unsigned int realm::Table::find_first<realm::BinaryData>(unsigned int, realm::BinaryData) const'
jni/../src/object-store/src/results.cpp:329: error: undefined reference to 'unsigned int realm::Table::find_first<realm::BinaryData>(unsigned int, realm::BinaryData) const'
```

Could those be because the `Table::find_first<BinaryData>` generic ccase hasn't been explicitly instantiated like the others in Core (https://github.com/realm/realm-core/blob/master/src/realm/table.cpp#L3971)?

PR #516, opened 2017-08-18
Update to core v3.0.0-rc5 and sync v2.0.0-rc15

bdash:


tgoyne:
The windows build needs the runtime switched to the static runtime to match the change made in core.

PR #521, opened 2017-08-21
Store sync metadata keychain info on per-app basis

austinzheng:
It's not clear how to test this change in isolation, since it touches Apple's Keychain stuff. The binding-level integration tests continue to work unhindered.

### DIFF
--- a/src/impl/apple/keychain_helper.cpp
+++ b/src/impl/apple/keychain_helper.cpp
@@ -111,34 +111,31 @@ void set_key(const std::vector<char>& key, CFStringRef account, CFStringRef serv
 std::vector<char> metadata_realm_encryption_key(bool check_legacy_service)
 {
     CFStringRef account = CFSTR("metadata");
-    CFStringRef default_service = CFSTR("io.realm.sync.keychain");
+    CFStringRef legacy_service = CFSTR("io.realm.sync.keychain");
 
-    CFPtr<CFStringRef> managed_service;
+    CFPtr<CFStringRef> service;
     if (auto bundle_id = retainCF(CFBundleGetIdentifier(CFBundleGetMainBundle()))) {
-        managed_service = adoptCF(CFStringCreateWithFormat(NULL, NULL,
-                                                           CFSTR("%@ - Realm Sync Metadata Key"),
-                                                           bundle_id.get()));
+        service = adoptCF(CFStringCreateWithFormat(NULL, NULL, CFSTR("%@ - Realm Sync Metadata Key"), bundle_id.get()));
     } else {
-        managed_service = retainCF(default_service);
+        service = retainCF(legacy_service);
         check_legacy_service = false;
     }
 
     // Try retrieving the key.
-    CFStringRef service = managed_service.get();
-    if (auto existing_key = get_key(account, service)) {
+    if (auto existing_key = get_key(account, service.get())) {
         return *existing_key;
     } else if (check_legacy_service) {
-        // See if there's a key in the old shared keychain.
-        if (auto existing_legacy_key = get_key(account, default_service)) {
-            // If so, copy it to the per-app keychain before returning it.
-            set_key(*existing_legacy_key, account, service);
+        // See if there's a key stored using the legacy shared keychain item.
+        if (auto existing_legacy_key = get_key(account, legacy_service)) {
+            // If so, copy it to the per-app keychain item before returning it.
+            set_key(*existing_legacy_key, account, service.get());
             return *existing_legacy_key;
         }
     }
     // Make a completely new key.
     std::vector<char> key(key_size);
     arc4random_buf(key.data(), key_size);
-    set_key(key, account, service);
+    set_key(key, account, service.get());
     return key;
 }
 
--- a/src/impl/apple/keychain_helper.cpp
+++ b/src/impl/apple/keychain_helper.cpp
@@ -38,10 +38,10 @@ KeychainAccessException::KeychainAccessException(int32_t error_code)
 
 namespace {
 
+using CFStringPtr = CFPtr<CFStringRef>;
 constexpr size_t key_size = 64;
-const std::string account = "metadata";
 
-CFPtr<CFStringRef> convert_string(const std::string& string)
+CFStringPtr convert_string(const std::string& string)
 {
     auto result = adoptCF(CFStringCreateWithBytes(nullptr, reinterpret_cast<const UInt8*>(string.data()),
                                                   string.size(), kCFStringEncodingASCII, false));
@@ -51,7 +51,8 @@ CFPtr<CFStringRef> convert_string(const std::string& string)
     return result;
 }
 
-CFPtr<CFMutableDictionaryRef> build_search_dictionary(const std::string& account, const std::string& service,
+CFPtr<CFMutableDictionaryRef> build_search_dictionary(const CFStringPtr& account,
+                                                      const CFStringPtr& service,
                                                       __unused util::Optional<std::string> group)
 {
     auto d = adoptCF(CFDictionaryCreateMutable(nullptr, 0, &kCFTypeDictionaryKeyCallBacks,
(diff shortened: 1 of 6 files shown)


austinzheng:
This is ready for review. Apologies for the spam earlier.

src/impl/apple/keychain_helper.cpp
```diff
@@ -50,66 +51,91 @@ CFPtr<CFStringRef> convert_string(const std::string& string)
     return result;
 }
 
-CFPtr<CFMutableDictionaryRef> build_search_dictionary(const std::string& account, const std::string& service,
+CFPtr<CFMutableDictionaryRef> build_search_dictionary(const CFStringPtr& account,
```
bdash: Taking smart pointers by const reference is a little weird. Typically you'd take a raw pointer to indicate that you're merely borrowing a reference, rather than doing anything with the ownership of the object. For that reason I'd suggest just using `CFStringRef` for the string arguments.

src/impl/apple/keychain_helper.cpp
```diff
@@ -50,66 +51,91 @@ CFPtr<CFStringRef> convert_string(const std::string& string)
...
+    if (group)
         CFDictionaryAddValue(d.get(), kSecAttrAccessGroup, convert_string(*group).get());
-    }
 #endif
     return d;
 }
 
-std::vector<char> metadata_realm_encryption_key()
+/// Get the encryption key for a given service.
+util::Optional<std::vector<char>> try_getting_key(const CFStringPtr& account, const CFStringPtr& service)
```
bdash: Can this just be `get_key`?

src/impl/apple/keychain_helper.cpp
```diff
@@ -50,66 +51,91 @@ CFPtr<CFStringRef> convert_string(const std::string& string)
...
+        throw KeychainAccessException(status);
 }
+
+}   // anonymous namespace
+
+std::vector<char> metadata_realm_encryption_key(bool check_legacy_service)
+{
+    const CFStringPtr account = convert_string("metadata");
+    CFStringPtr default_service = convert_string("io.realm.sync.keychain");
+    auto service = adoptCF(CFBundleGetIdentifier(CFBundleGetMainBundle()));
```
bdash: I think we want to use something beyond just the bare bundle identifier as the keychain item name to avoid conflicting with anything created by the app itself. Maybe something like "com.example.app - Realm Sync Metadata Key"?

`adoptCF` isn't appropriate here. `CFBundleGetIdentifier` does not return an owned object, only a borrowed object. Adopting the returned reference will result in a use after free. You want `retainCF` instead.

src/impl/apple/keychain_helper.cpp
```diff
@@ -50,66 +51,91 @@ CFPtr<CFStringRef> convert_string(const std::string& string)
...
+    CFDictionaryAddValue(search_dictionary.get(), kSecValueData, key_data.get());
+    if (OSStatus status = SecItemAdd(search_dictionary.get(), nullptr))
+        throw KeychainAccessException(status);
 }
+
+}   // anonymous namespace
+
+std::vector<char> metadata_realm_encryption_key(bool check_legacy_service)
+{
+    const CFStringPtr account = convert_string("metadata");
```
bdash: These constants should use the `CFSTR` macro instead, so they end up in the `__cfstring` section of the binary rather than being allocated on the heap at runtime. There'd be no reason to use `CFPtr` with the resulting strings, either, as they'd have an eternal lifetime.

src/impl/apple/keychain_helper.cpp
austinzheng: I totally forgot about this pattern.

src/impl/apple/keychain_helper.cpp
austinzheng: Good call. I need to review the CF ownership conventions before writing any more CF code.

bdash: The [CF documentation has a section on ownership](https://developer.apple.com/library/content/documentation/CoreFoundation/Conceptual/CFMemoryMgmt/Concepts/Ownership.html#//apple_ref/doc/uid/20001148-SW1) that's relatively readable. Many CF APIs describe their return values as using one of the two rules described within (create vs get).

src/impl/apple/keychain_helper.cpp
```diff
@@ -50,66 +51,96 @@ CFPtr<CFStringRef> convert_string(const std::string& string)
...
+    if (OSStatus status = SecItemAdd(search_dictionary.get(), nullptr))
+        throw KeychainAccessException(status);
 }
+
+}   // anonymous namespace
+
+std::vector<char> metadata_realm_encryption_key(bool check_legacy_service)
+{
+    CFStringRef account = CFSTR("metadata");
+    CFStringRef default_service = CFSTR("io.realm.sync.keychain");
```
bdash: I think `legacy_service` would better reflect how the relationship between this and the `check_legacy_service` argument.

src/impl/apple/keychain_helper.cpp
```diff
@@ -50,66 +51,96 @@ CFPtr<CFStringRef> convert_string(const std::string& string)
...
+        managed_service = adoptCF(CFStringCreateWithFormat(NULL, NULL,
+                                                           CFSTR("%@ - Realm Sync Metadata Key"),
+                                                           bundle_id.get()));
+    } else {
+        managed_service = retainCF(default_service);
+        check_legacy_service = false;
+    }
+
+    // Try retrieving the key.
+    CFStringRef service = managed_service.get();
```
bdash: I'd be inclined to ditch this local and just add the `.get()` calls in the handful of places it is necessary. It's effectively free to call, and would avoid the need for the slightly odd `managed_` prefix on the variable.

src/impl/apple/keychain_helper.cpp
```diff
@@ -50,66 +51,96 @@ CFPtr<CFStringRef> convert_string(const std::string& string)
...
+    if (auto existing_key = get_key(account, service)) {
+        return *existing_key;
+    } else if (check_legacy_service) {
+        // See if there's a key in the old shared keychain.
```
bdash: "See if there's a key stored using the legacy shared keychain item"?

src/impl/apple/keychain_helper.cpp
```diff
@@ -50,66 +51,96 @@ CFPtr<CFStringRef> convert_string(const std::string& string)
...
+        if (auto existing_legacy_key = get_key(account, default_service)) {
+            // If so, copy it to the per-app keychain before returning it.
```
bdash: The keychain isn't per application, the item name is.

austinzheng:
Thank you for the review! I will hold off on merging this until the Cocoa PR has been updated to compile and the integration tests have been run successfully.

austinzheng:
There was an issue that caused the tests to fail. I created https://github.com/realm/realm-core/issues/2821 and worked around it in my latest change.

src/impl/apple/keychain_helper.cpp
```diff
@@ -50,66 +51,97 @@ CFPtr<CFStringRef> convert_string(const std::string& string)
...
+    CFPtr<CFStringRef> service;
+    CFStringRef bundle_id = CFBundleGetIdentifier(CFBundleGetMainBundle());
+    if (!bundle_id) {
+        service = retainCF(legacy_service);
+        check_legacy_service = false;
+    } else {
+        // FIXME: Get rid of this manual retain once we fix CFPtr to accept null.
+        CFRetain(bundle_id);
+        service = adoptCF(CFStringCreateWithFormat(NULL, NULL, CFSTR("%@ - Realm Sync Metadata Key"), bundle_id));
+        CFRelease(bundle_id);
```
bdash: There's no reason to retain / release `bundle_id` here.

src/impl/apple/keychain_helper.cpp
```diff
@@ -119,10 +119,8 @@ std::vector<char> metadata_realm_encryption_key(bool check_legacy_service)
         service = retainCF(legacy_service);
         check_legacy_service = false;
     } else {
-        // FIXME: Get rid of this manual retain once we fix CFPtr to accept null.
-        CFRetain(bundle_id);
+        // FIXME: Collapse the null check on bundle_id once CFPtr is fixed to accept null pointers.
```
bdash: I'm not 100% sure what you mean by "collapse the null check" here. We could write this like so, though:

```c++
if (CFStringRef bundle_id = CFBundleGetIdentifier(CFBundleGetMainBundle()))
    service = adoptCF(CFStringCreateWithFormat(NULL, NULL, CFSTR("%@ - Realm Sync Metadata Key"), bundle_id));
else {
    service = retainCF(legacy_service);
    check_legacy_service = false;
}
```

austinzheng: We don't actually need to use `retainCF` for the temporary `bundle_id`, so your suggestion addresses the problem and there isn't anything we need to do in the future.

bdash:
👍

PR #524, opened 2017-08-23
Treat admin token users like regular users, require URL

austinzheng:
Changes:
- Admin token users are now treated like regular users: they can be persisted in the metadata Realm, logged out and back in, etc
- Added a flag when configuring file system to allow admin token users to opt out of being persisted (they also will not show up in `allSessions` or `currentSession` if so)
- Admin token users now require a URL to initialize
- Updated tests

Shouldn't be merged in until Sync 2 is ready to ship.


austinzheng:
We can't do this for the ROS 2.0 timeframe, so I'm closing it. We can revisit in the future.

PR #527, opened 2017-08-28
add support to ssl_verify_callback option

nhachicha:
following the work on Sync repo https://github.com/realm/realm-sync/pull/1312 to allow the binding to pass a callback to validate the SSL certificate, this PR expose this option to the binding so we can set the callback when building the `SyncConf`

Will be used mainly for Java to allow Android trust manager to validate  the cert https://github.com/realm/realm-java/issues/4759

### DIFF
--- a/src/sync/sync_config.hpp
+++ b/src/sync/sync_config.hpp
@@ -108,6 +108,7 @@ struct SyncConfig {
     util::Optional<std::array<char, 64>> realm_encryption_key;
     bool client_validate_ssl = true;
     util::Optional<std::string> ssl_trust_certificate_path;
+    std::function<sync::Session::SSLVerifyCallback> ssl_verify_callback;
 #if __GNUC__ < 5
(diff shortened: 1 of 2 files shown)


src/sync/sync_config.hpp
```diff
@@ -108,14 +108,15 @@ struct SyncConfig {
     util::Optional<std::array<char, 64>> realm_encryption_key;
     bool client_validate_ssl = true;
     util::Optional<std::string> ssl_trust_certificate_path;
+    std::function<sync::Session::SSLVerifyCallback> ssl_verify_callback;
```
austinzheng: What's the definition of `SSLVerifyCallback`? If that's a function type (as the name implies), wrapping it in `std::function` seems unnecessary.

austinzheng: Never mind. It's probably a typedef for an actual function type, which does need to be wrapped in a std::function.

PR #534, opened 2017-09-02
Batch update for Results

cmelchior:
This PR contains first stab at making a batch update API for the `Results` class. A similar API should come for `List` in another PR once this is complete.

The PR is work in progress, but it would be nice to know if this is the route to take. If yes, I will add support for the remaining types, if not, better change it now :)

My only concern with current implementation is that I'm calling `get()` in a loop which carries quite a lot of unnecessary overhead as it calls down to `try_get`, but the logic for both `size()` and `get()` seem fairly complex, so I'm not sure it would be a good idea to copy it. The current implementation does solve the main problem for Java though, which is allocations on the Java side for each object.

I have not done any sort of benchmarking.

What do you think @bdash? and others ( @kneth , @tgoyne ? )


kneth:
@cmelchior If performance comes up as an issue later on, we can address it. But moving the loop and object creation away from Java is a good, first optimization 😄 .

tgoyne:
This should reuse the existing object accessor code as much as possible rather than duplicating it. If constructing an `Object` for each row has too much overhead, it shouldn't be too hard to extract the actual logic into free functions which the `Object` functions just call.

cmelchior:
After implementing the Java accessor, the changes required for this PR made a lot more sense to me 🙄.... Anyway, ready for review.

src/results.cpp
```diff
@@ -813,8 +812,15 @@ Results::UnsupportedColumnTypeException::UnsupportedColumnTypeException(size_t c
 {
 }
 
+Results::InvalidPropertyException::InvalidPropertyException(const std::string& object_type, const std::string& property_name)
+        : std::logic_error(util::format("Property '%1.%2' does not exist", object_type, property_name))
```
tgoyne: 8 spaces?

src/results.hpp
```diff
@@ -219,11 +230,17 @@ class Results {
...
+    // Throws an exception if the value does not match the type for given prop_name
+    template<typename ValueType, typename ContextType>
+    void set_property_value(ContextType& ctx, StringData prop_name, ValueType value);
+
     // Execute the query immediately if needed. When the relevant query is slow, size()
     // may cost similar time compared with creating the tableview. Use this function to
     // avoid running the query twice for size() and other accessors.
     void evaluate_query_if_needed(bool wants_notifications = true);
-
+
```
tgoyne: trailing whitespace

src/results.hpp
```diff
@@ -313,6 +330,30 @@ size_t Results::index_of(Context& ctx, T value)
...
+    }
+    if (prop->is_primary && !m_realm->is_in_migration()) {
+        throw ModifyPrimaryKeyException(m_object_schema->name, prop->name);
+    }
+
+    // Update all objects in this ResultSets
+    size_t size = this->size();
+    for (size_t i = 0; i < size; ++i) {
+        const RowExpr row = this->get(i);
+        Object obj = realm::Object(m_realm, *m_object_schema, row);
```
tgoyne: `Object obj(m_realm, *m_object_schema, get(i));`

PR #535, opened 2017-09-05
Only look for libuv when building for Node

fealebenpae:


bdash:
What's the justification for this change? It looks like this will result in tests failing out of the box on Linux unless you configure with `REALM_PLATFORM` set to Node. We were intentionally look for `libuv` in that case to make things work without extra configuration.

fealebenpae:
> What's the motivation for this change?

In dotnet we reuse parts of the Object Store CMake infrastructure (`CompilerFlags.cmake` being an instance) and when we recently spread our build pipeline across more macOS nodes, we noticed that some of them had `libuv` installed for whatever reason, which resulted in our binaries linking against it when built on those nodes.

> this will result in tests failing out of the box on Linux

That's a fair point. What if we always look for `libuv` on Linux, regardless of whether `REALM_PLATFORM` is configured? This way Linux builds should work more or less out of the box just as they have before, but we won't be picking up stray libraries on other platforms when we don't need them.

bdash:
Perhaps the CMake files should be checking for `REALM_PLATFORM=DotNet` (or similar) and using that to determine whether or not `libuv` should be used? Then we could continue to detect libraries by their existence when `REALM_PLATFORM` is unset.

fealebenpae:
I can move the search for `libuv` in an `if(NOT REALM_PLATFORM OR REALM_PLATFORM STREQUAL "Node")` block - how does that sound?

bdash:
Seems fine!

PR #536, opened 2017-09-06
Fix multiple exceptions seen in sync session unit tests with sync v2.0.0-rc17

bdash:
Sync now enforces that rows are added using one of `sync::create_object` / `sync::create_object_with_primary_key`. Failing to do this results in an exception. It also now verifies that the object ID column is not modified at any point. Our sync stop policy behavior tests violated both of these restrictions.


austinzheng:
I wonder why the only place this issue showed up was in this one test case. Did I implement it improperly?

tgoyne:
We don't have very many object store tests which use sync and actually create objects.

PR #537, opened 2017-09-06
Perform an "offline" client reset when opening a Realm with a sync 1.x history when using sync 2.x

### LINKED ISSUE #474: Offline Client Reset
Add support for resetting the Realm file when an `IncompatibleHistories` ("Incompatible histories. Nonupgradable history schema.") exception is thrown as a result of an non-migratable upgrade.

/cc @beeender @bdash

bdash:
The Realm file is moved aside, and the client is given the opportunity to open it as a read-only, non-synced Realm to recover any data that may remain in the file. They can then attempt to reopen the original sync URL.

Fixes #474.


bdash:
I'll fix the Windows build issues tomorrow.

src/impl/realm_coordinator.cpp
```diff
@@ -241,6 +237,9 @@ std::shared_ptr<Realm> RealmCoordinator::get_realm(Realm::Config config)
         m_weak_realm_notifiers.emplace_back(realm, m_config.cache);
     }
 
+    if (realm->config().sync_config)
```
beeender: missed `#if REALM_ENABLE_SYNC` here

bdash: The `sync_config` member of `Realm::Config` is available even when sync is not enabled, so AFAICT no `#if` is needed here?

beeender: but `create_sync_session();` is not there if sync is not enabled, right?

bdash: It is: https://github.com/realm/realm-object-store/blob/90d7e73ad4605618c414e1d517bb83a4229e0538/src/impl/realm_coordinator.cpp#L78

beeender: uh, sorry!! I was thinking a wrong thing `SyncSession::create_sync_session()`. 

👍

src/sync/sync_session.cpp
```diff
@@ -403,7 +403,26 @@ SyncSession::SyncSession(SyncClient& client, std::string realm_path, SyncConfig
 : m_state(&State::inactive)
 , m_config(std::move(config))
 , m_realm_path(std::move(realm_path))
-, m_client(client) { }
+, m_client(client)
+{
+#if REALM_HAVE_SYNC_STABLE_IDS
+   if (m_config.validate_sync_history) {
```
austinzheng: Worth adding a comment explaining what's going on here for future reference?

tgoyne:
I don't understand the `validate_sync_history` changes. Opening a Realm inside `SyncSesssion::get_session()` when it's called directly by the binding seems to defeat the entire purpose of exposing that to the binding.

src/shared_realm.cpp
```diff
@@ -181,6 +191,24 @@ void Realm::open_with_config(const Config& config,
...
+    catch (IncompatibleHistories const& ex) {
+        if (!server_synchronization_mode || !is_nonupgradable_history(ex))
+            translate_file_exception(config.path, config.immutable()); // Throws
+
+        // Move the Realm file into the recovery directory.
+        std::string recovery_directory = SyncManager::shared().recovery_directory_path();
+        std::string new_realm_path = util::reserve_unique_file_name(recovery_directory, "synced-realm-XXXXXXX");
+        util::File::move(config.path, new_realm_path);
+
+        std::string message = util::format("The local copy of this synced Realm was created with an incompatible "
```
tgoyne: This is worded as a message a user writing code in an interactive environment. It doesn't really work as either an error displayed to the user in a app, or to a developer in a crash log after they failed to handle it.

bdash: The nature of this error is that the developer needs to be proactively accommodate it when they update their app from sync v1.x to 2.x. It’s not intended to be displayed to users, and if it crashes their app in production its too late for them to do anything about it.

bdash:
> I don't understand the validate_sync_history changes. Opening a Realm inside SyncSesssion::get_session() when it's called directly by the binding seems to defeat the entire purpose of exposing that to the binding.

I’m not sure what you mean by this.

This logic exists within the `SyncSession` constructor so that a direct call to `get_session` results in the incompatible history schema being thrown out of `get_session` rather than when sync eventually happens to open the Realm on its background worker thread. The reason it is conditional is so that we don’t do this work redundantly when `RealmCoordinator` creates the sync session, as that would a) be unnecessary, and b) deadlock due to the relevant locks already being held.

src/shared_realm.cpp
tgoyne: Yes, that's my point. It's worded as a message to an end-user rather than something telling a developer that they need to write code to handle this case. Things like "you can now retry doing X" make sense when you're telling to user to click on the button they just clicked on again, but less so when they actually need to write code to handle this error case.

tgoyne:
My understanding is that the only reason that `get_session()` exists as part of the public API is for some hypothetical use-case where the binding needs to create a session without opening the Realm file (there doesn't appear to actually be any calls to it in any of the bindings or in any non-test code at all except for the single one in RealmCoordinator). If it has to open the Realm file anyway, they might as well just do that first and avoid the perf hit of opening and closing it.

bdash:
The intent of this change was to provide the desired semantics while requiring only minimal changes from bindings. It may be desirable to get rid of `SyncManager::get_session` in favor of having the bindings open `Realm` instances in those cases, but that's more invasive and harder for bindings to accommodate in the limited time we have available.

.NET currently uses `get_session()`. Java [provides API that looks like it would need to use `get_session()`](https://realm.io/docs/java/latest/api/io/realm/SyncManager.html#getSession-io.realm.SyncConfiguration-), but it no longer appears to call it so I'm not sure how they're even providing the functionality that method claims to provide.

cmelchior:
> My understanding is that the only reason that `get_session()` exists as part of the public API is for some hypothetical use-case where the binding needs to create a session without opening the Realm file (there doesn't appear to actually be any calls to it in any of the bindings or in any non-test code at all except for the single one in RealmCoordinator)

It is not hypothetical, we had good use cases for it in the Java session, and using the current auto-managed model is actually what has caused us the most pain. E.g. with an explicit Sync session and start/stop method you can trigger sync based on certain system events like only syncing when connecting to 3G or better or while charging.

nirinchev:
How does the backup realm config look like? I imagine the binding needs to set `path`, `schema_mode`, `encryption_key`, and `force_sync_history`? Anything else?

bdash:
The binding needs to set `path`, `schema_mode` and `encryption_key`. It should not set `force_sync_history`. A schema should likely also be provided.

nirinchev:
What happens if the schema doesn't match the original file? E.g. if the developer has added or removed properties from their models?

bdash:
I think they'd need to use dynamic schema mode in that case.

src/shared_realm.cpp
bdash: @tgoyne, would you mind suggesting an edit to this error message? I'm struggling to come up with something that's an improvement, and this is the only part of your feedback that I've not yet addressed. I'd like to get this merged ASAP.

austinzheng: I was thinking something like this:

```
That the local copy of this synced Realm was created with an incompatible version of Realm. The existing file has been moved to '%1'. The application developer is responsible for catching this error, opening the backup copy of the Realm to recover any pending changes, and opening the synchronized Realm again.
```

tgoyne: "The local copy of this sync Realm was created with an incompatible version of Realm. It has been moved aside, and the Realm will be re-downloaded the next time it is opened. You will need to write a handler for this error which either uses the attached configuration to open the Realm in read-only mode to recover any pending changes or deletes the old Realm file. See http://realm.io/???? for more information about this migration."

I'm not terribly happy with this wording either, but the idea behind it is that the message is trying to tell the developer what code they need to write. I'm assuming that we will have some sort of news post about this which should be linked. The path is not included in the error because it's hard to imagine a situation where it'd be useful, while it's easy to imagine situations where people do bad things with it (like trying to extract the path from the error message rather than getting it from the config) because it's the most obvious approach.

Also now I'm wondering if we should do something more like how migration blocks work since writing code which handles a Realm instance failing to open, doing stuff, and then retrying creating the Realm instance is probably going to be really awkward for people that don't have a wrapper factory function for getting Realm instances, but maybe making that sufficiently flexible would get too complicated.

bdash: I've updated to use your wording with a few minor tweaks. I've omitted the link since there's nothing to link to for now. Bindings can append their own at that level if they choose.

I think the number of people that'll need to deal with this is relatively small, so I'm ok with the API being a little clunky (particularly given the time we have available and the need for it to be implemented in each binding).

PR #538, opened 2017-09-08
Upgrade to 2.0.0-RC18

cmelchior:


src/sync/sync_session.cpp
```diff
@@ -469,6 +469,11 @@ void SyncSession::handle_error(SyncError error)
             case ProtocolError::reuse_of_session_ident:
             case ProtocolError::bound_in_other_session:
             case ProtocolError::bad_message_order:
+            case ProtocolError::bad_changeset_header_syntax:
```
bdash: These need to be #if'd to avoid breaking use with sync 1.x.

src/sync/sync_session.cpp
```diff
@@ -469,6 +469,13 @@ void SyncSession::handle_error(SyncError error)
...
             case ProtocolError::bound_in_other_session:
             case ProtocolError::bad_message_order:
+#if REALM_SYNC_VER_MAJOR > 1
+            case ProtocolError::bad_changeset_header_syntax:
+            case ProtocolError::bad_changeset_size:
+            case ProtocolError::bad_changesets:
+            case ProtocolError::bad_decompression:
+#endif
+
 #if REALM_SYNC_VER_MAJOR == 1
```
bdash: This could be an `#else` rather than `#endif` / `#if`.

cmelchior: That would change the semantics slightly if REALM_SYNC_VER_MAJOR was ever 0. I was actually wondering why it checks explicitly for `== 1`, because I would assume that was the default, so couldn't we remove that one completely?

bdash: I'm not sure I understand your concern. Sync's major version number can never go backwards, so it'll always be >= 1 at this point.

cmelchior: It isn't really a concern. It is mostly me musing about readability. Having `x > 1` and then `x == 1` implies that `x == 0` is different for some reason, Having `if (x > 1) { .. } else { ... }` also implies that the 2nd else has a different scope than the rest of the code (which it doesn't from what I know?)

But I'm not sure what kind of patterns you use elsewhere in OS, so if you do `ifdef/else` I'll just do that here as well.

PR #543, opened 2017-09-13
Add ObjectStore::property_for_column_index

beeender:
To return a Property based on the given table and column index dynamically.
This will be used by realm-java's dynamic API. Currently java has many
duplicated logic in JNI code.


src/object_schema.cpp
```diff
@@ -89,33 +89,10 @@ ObjectSchema::ObjectSchema(Group const& group, StringData name, size_t index) :
...
-        property.type = from_core_type(*table->get_descriptor(), col);
-        property.is_indexed = table->has_search_index(col);
-        property.table_column = col;
-
-        if (property.type == PropertyType::Object) {
-            // set link type for objects and arrays
-            ConstTableRef linkTable = table->get_link_target(col);
-            property.object_type = ObjectStore::object_type_for_table_name(linkTable->get_name().data());
+        auto property = ObjectStore::property_for_column_index(table, col);
+        if (property) {
```
bdash: `if (auto property = …)`

Object store coding style typically omits braces around single-line bodies of if statements.

src/object_store.cpp
```diff
@@ -774,6 +774,37 @@ Schema ObjectStore::schema_from_group(Group const& group) {
     return schema;
 }
 
+util::Optional<Property> ObjectStore::property_for_column_index(ConstTableRef& table, size_t column_index)
+{
+    StringData column_name = table->get_column_name(column_index);
+
+#if REALM_HAVE_SYNC_STABLE_IDS
+    // The object ID column is an implementation detail, and is omitted from the schema.
+    // FIXME: Consider filtering out all column names starting with `__`.
```
bdash: Would you mind updating this comment to say column names starting with `!`, since `sync::object_id_column_name` was changed to `!OID` at some point.

tests/object_store.cpp
```diff
@@ -30,3 +37,102 @@ TEST_CASE("ObjectStore: table_name_for_object_type()") {
...
+                REQUIRE(table->get_column_name(col) == sync::object_id_column_name);
+#else
+                FAIL();
+#endif
+                continue;
+            }
+
+            REQUIRE(property->name == table->get_column_name(col));
+            if (property->name == "int") {
+                REQUIRE(property->type == PropertyType::Int);
```
bdash: Rather than hard-coding these various tests, could we instead look up the expected `Property` in `schema` and use the overload of `operator==` to compare expected vs actual?

beeender: Yes! much less code

PR #545, opened 2017-09-14
Describe true and false expressions

ironage:
This is required to compile with core versions >= 3.2.0. It adds descriptions for query expressions which is part of the metrics feature. `description()` is a pure virtual method so if these are not present there will be a compile time error.

### DIFF
--- a/src/parser/query_builder.cpp
+++ b/src/parser/query_builder.cpp
@@ -63,6 +63,7 @@ struct TrueExpression : realm::Expression {
     void set_base_table(const Table*) override {}
     const Table* get_base_table() const override { return nullptr; }
     void verify_column() const override {}
+    std::string description() const override { return "true"; }
     std::unique_ptr<Expression> clone(QueryNodeHandoverPatches*) const override
     {
         return std::unique_ptr<Expression>(new TrueExpression(*this));
@@ -73,6 +74,7 @@ struct FalseExpression : realm::Expression {
     size_t find_first(size_t, size_t) const override { return realm::not_found; }
     void set_base_table(const Table*) override {}
     void verify_column() const override {}
+    std::string description() const override { return "false"; }
     const Table* get_base_table() const override { return nullptr; }
     std::unique_ptr<Expression> clone(QueryNodeHandoverPatches*) const override
     {


bdash:
Shouldn't their descriptions correspond to the form that they're represented in the query strings (`TRUEPREDICATE` / `FALSEPREDICATE`)?

bdash:
It looks like these types have now been duplicated in sync, object store, and Cocoa… would you be interested in moving them down to core at some point so we can eliminate this redundancy?

ironage:
Sure, this is as good a time as any to move them down. I'll close this PR and make a new one removing them after the core changes.
I think simply "true" and "false" will read better when the query is printed out, but I'll add a core test using them to check this.

bdash:
I'm not sure what you mean by "read better". Shouldn't the goal be to have the queries be displayed to the user in the same form that they'd write them? Or at least as close as possible?

ironage:
Yes that is the goal. But I don't think users want to know the details about our implementation in "TrueExpression" Here's an example:
```
Query q0 = person->column<int64_t>(0) == 0;
Query and_true = q0.and_query(std::unique_ptr<realm::Expression>(new TrueExpression));
and_true.find_all();
...
// retrieving metrics about this query will produce the following description:
// queries->at(0).get_description() == "person.age is equal to 0 and true"
```
Could be debatable but I would prefer "true" to "TRUEPREDICATE" at the end there. 
Reading it out like that either way makes the construction of a query containing a `TrueExpression` or `FalseExpression` seems a bit redundant to me and may not be necessary at all.

bdash:
`TRUEPREDICATE` is the form that's used in queries as consumed by object store's query parser, Cocoa's `NSPredicate` queries, and sync's recently-introduced parser.

For the query you demonstrate, I think the user would much prefer to see `person.age == 0 AND TRUEPREDICATE`, because that's likely to be much closer to the query they wrote. Generating human-readable text like "is equal to" instead of the obvious `==` feels out of place.

ironage:
Ok, good point. I'll use `TRUEPREDICATE`. 
Regarding "is equal to" there are probably other operator descriptions that need to change so I'll leave that to another time when the requirements for query descriptions are clear. I'm not familiar with how the bindings construct user queries and we would have to reach some readable compromise that works for each binding.

PR #547, opened 2017-09-15
Add partial sync support

austinzheng:
This PR implements partial sync support in object store.

The user-facing aspects of this are a new API on `realm::Realm` for making partial sync queries, and new properties on `SyncConfig` for defining a partial sync Realm.

The partial sync identifier is implemented by using the existing user local identifier (which is currently used to create the directories in which a user's synced Realms are saved on disk). This identifier is guaranteed to be unique per user, does not change for a given user, and will never contain backslashes.

**Important**: bindings should stop accessing `SyncConfig.realm_url` directly, and should instead use the new helper method.


src/sync/sync_manager.cpp
```diff
@@ -406,7 +406,7 @@ std::shared_ptr<SyncUser> SyncManager::get_existing_logged_in_user(const SyncUse
     return user->state() == SyncUser::State::Active ? user : nullptr;
 }
 
-std::string SyncManager::path_for_realm(const SyncUser& user, const std::string& raw_realm_url) const
+std::string SyncManager::path_for_realm(const SyncUser& user, const std::string& raw_realm_url, bool partial_sync) const
```
austinzheng: This is wrong, fix it.

austinzheng:
There's still a ton of work to get this into reviewable shape, but the test passes!

austinzheng:
Figures the only test suite passing on CI is the environment I develop on locally. I will fix everything tomorrow.

src/sync/sync_manager.cpp
```diff
@@ -415,7 +415,11 @@ std::string SyncManager::path_for_realm(const SyncUser& user, const std::string&
     if (user.token_type() == SyncUser::TokenType::Normal)
         user_info = SyncUserIdentifier{ user.identity(), user.server_url() };
 
-    return m_file_manager->path(user_local_identity, raw_realm_url, std::move(user_info));
+    // FIXME: check for forward slash
+    // FIXME: how to derive unique identifier?
```
simonask: A random string (like a UUID or something) should do the trick, if persisted and reused.

Ideally it would be possible to override it in the Config too, since there are use cases for that (shared partial views).

src/sync/impl/partial_sync_helper.hpp
```diff
@@ -0,0 +1,71 @@
...
+#include <functional>
+#include <string>
+#include <unordered_map>
+
+namespace realm {
+    
+class Realm;
+class List;
+
+using PartialSyncResultCallback = void(List results, std::exception_ptr error);
```
simonask: Recommend using `std::error_code` to pass errors by value rather than `std::exception_ptr`.

bdash: It looks like this is used in conjunction with our existing object notification callbacks, which already use `std::execption_ptr` for this purpose. Trying to change it here would require being able to translate an `std::exception_ptr` to a `std::error_code`, which seems impractical.

src/sync/sync_manager.cpp
bdash: My thought for the unique identifier was to use a UUID that we then store in the sync metadata Realm.

Making the unique identifier configurable seems like something we can punt on for now in the interests of minimizing the amount of work required from each binding. It shouldn't be hard to add after the fact.

austinzheng: If we can use a per-user UUID (e.g. each {user, Realm} tuple only has one partially-synced Realm aside from custom shared partial views), we already generate a local UUID for this purpose and could use that.

austinzheng:
CI is really slow, but the builds don't seem to be broken anymore. This can be reviewed whenever.

src/sync/partial_sync.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+    Object raw_object;
+    {
+        realm->begin_transaction();
+        auto cleanup = util::make_scope_exit([&]() noexcept {
+            if (realm->is_in_transaction())
+                realm->cancel_transaction();
+        });
+
+        // FIXME: Is it ok for us to modify the schema of the Realm like this?
```
bdash: @tgoyne, can you please weigh in on the question asked in this FIXME?

tgoyne: Seems likely to be a problem. IMO `__ResultsSet` should not be exposed in the Realm's schema (even for dynamic realms), and the schema reading code should be modified to skip it.

A drycoded idea:
```
    auto& group = realm.read_group();
    ObjectSchema result_sets_schema(result_sets_type_name, {
        {"matches_property", PropertyType::String},
        {"query", PropertyType::String},
        {"status", PropertyType::Int},
        {"error_message", PropertyType::String},
        {matches_property, PropertyType::Object|PropertyType::Array, object_class}
    });
    auto required_changes = Schema{{ObjectSchema{group, result_sets_type_name}}}.compare(Schema{{results_set_schema}})
    if (!required_changes.empty())
        apply_additive_changes(group, changes_required);
```

bdash: Thanks for the suggestion. I'll give it a shot.

> IMO __ResultsSet should not be exposed in the Realm's schema (even for dynamic realms), and the schema reading code should be modified to skip it.

I agree with this, but some people have expressed a desire for it to be visible in the browser.

bdash:
Attempting to update this PR to sync-v2.0.0-rc21 results in the test for error handling waiting indefinitely for the change notification. I've pinged @kspangsege on Slack about this.

tests/sync/partial_sync.cpp
```diff
@@ -0,0 +1,240 @@
...
+            table->set_int(number_prop.table_column, row_idx, current.number);
+            table->set_string(first_string_prop.table_column, row_idx, current.first_string);
+            table->set_string(second_string_prop.table_column, row_idx, current.second_string);
+        }
+    }
+    r->commit_transaction();
+    // Wait for uploads
+    std::atomic<bool> upload_done(false);
+    auto session = SyncManager::shared().get_existing_active_session(config.path);
+    session->wait_for_download_completion([&](auto) { upload_done = true; });
```
austinzheng: Not sure this matters in practice, but I made a mistake here: should be `wait_for_upload_completion`.

src/impl/notification_wrapper.hpp
```diff
@@ -0,0 +1,51 @@
...
+// A wrapper that stores a value and an associated notification token.
+// The point of this type is to keep the notification token alive
+// until the value can be properly processed or handled.
+template<typename T>
+struct NotificationWrapper : public T {
+    using T::T;
+
+    NotificationWrapper(T&& object)
+    : T(object)
+    { }
```
simonask: Out of curiosity, why is the move constructor necessary when the base class' constructors have already been included per the `using` statement above?

src/sync/partial_sync.cpp
```diff
@@ -0,0 +1,124 @@
...
+#include "thread_safe_reference.hpp"
+
+#include <realm/util/scope_exit.hpp>
+
+namespace realm {
+namespace partial_sync {
+
+namespace {
+
+constexpr const char* result_sets_type_name = "ResultSets";
```
bmunkholm: This changed to "__ResultSets" in recent release of sync.
https://github.com/realm/realm-sync/commit/d7991124c9f5285e9b603fde1424fe90e4bcf5c2

kspangsege:
> This identifier is guaranteed to be unique per user, does not change for a given user, and will never contain backslashes.

@austinzheng I assume this identifier will be different for the same user on two different devices. I has to be.

kspangsege:
> This identifier is guaranteed to be unique per user, does not change for a given user, and will never contain backslashes.

Do you mean "slashes" rather than "backslashes". The partial sync spec requires that it has no "slashes", but does not care about "backslashes".

austinzheng:
@kspangsege It's not. If it has to be we need to generate a separate ID and store it apart from the user ID.

src/sync/partial_sync.cpp
kspangsege: Yes, specifically @bigfish24 has.

tests/sync/partial_sync.cpp
```diff
@@ -0,0 +1,240 @@
...
+    EventLoop::main().run_until([&] { return upload_done.load(); });
```
kspangsege: Hmm, isn't this effectively a busy loop?

austinzheng:
@kspangsege Actually, the local identity is both device and user-unique in most cases, not if the user is part of a very old ROS deployment and logged into both devices before about a few months ago. We can still add in a new string field into the metadata if you'd rather be completely sure (and I need to look at how admin token users do it).

There are neither slashes nor backslashes in this identifier, or any UUID we generate.

kspangsege:
> Attempting to update this PR to sync v2.0.0-rc21 (i.e., merging in object store master) results in the test for error handling waiting indefinitely for the change notification. I've pinged @kspangsege on Slack about this.

@bdash Hopefully [this PR](https://github.com/realm/realm-sync/pull/1674) provides a work-around.

tests/sync/partial_sync.cpp
austinzheng: This is unit test code intended to stop execution of the test until all the changes are ready. It won’t affect the bindings in any way.

src/impl/notification_wrapper.hpp
austinzheng: We wrote this wrapper originally for a different feature, and I believe that gcc complained when we used the constructor unless we explicitly declared it here. So it’s a workaround, basically.

tests/sync/partial_sync.cpp
kspangsege: I understand, it just seems wasteful. If there is no easy way to avoid it, no big problem.

tests/sync/partial_sync.cpp
kspangsege: It seems to me that this matters. If you want to make some local changes, and have the server react to them, and see the result locally, then you want to force a complete round-trip. This is done by first waiting for upload completion, then waiting for download completion (needs to be two separate steps).

kspangsege: On the other hand, it seems like your `run_query()` is going to force a full round trip anyway, however, that only holds as long as you follow `populate_realm()` with `run_query()`. I'd put a wait for upload -> download completion step in this function just to increase its utility.

tests/sync/partial_sync.cpp
bdash: We currently have only a very minimal runloop abstraction available in the object store tests, so this is the best we can do.

src/sync/partial_sync.cpp
bigfish24: I think it is valuable right now during the tech preview but maybe at GA we are more confident that we can hide it. I think it would be better to debug with a customer by having it available for the time being.

bdash:
I rebased this on `master` and updated it to sync v2.0.0-rc22 to pick up the fix from realm/realm-sync#1674.

kspangsege:
> I rebased this on master and updated it to sync v2.0.0-rc22 to pick up the fix from realm/realm-sync#1674.

@bdash Where is the code that registers as a listener on `query_parse_counter`? Will that be in the bindings?

bdash:
@kspangsege, https://github.com/realm/realm-object-store/blob/cb8f0277269c0d4c4a910bf4da954f9bcf8d1816/src/sync/partial_sync.cpp#L96-L122

kspangsege:
Alright, so you aren't listening specifically to `query_parse_counter`, but that property ensures that your are notified at appropriate times, and in the callback you check the value of status. Makes sense.

austinzheng:
Yes, exactly. We just need _some_ change to occur that triggers the object store change notifications, and then we check for whatever we're looking for.

src/sync/impl/sync_metadata.cpp
```diff
@@ -121,6 +123,27 @@ SyncMetadataManager::SyncMetadataManager(std::string path,
...
+        if (old_realm->schema_version() < 3) {
+            TableRef table = ObjectStore::table_for_object_type(realm->read_group(), c_sync_userMetadata);
+            size_t idx_identity = table->get_column_index(c_sync_identity);
+            size_t idx_local_uuid = table->get_column_index(c_sync_local_uuid);
+            size_t idx_device_unique_uuid = table->get_column_index(c_sync_device_unique_uuid);
+
+            // Get all the SyncUserMetadata objects.
+            Results results(realm, *table);
+            for (size_t i = 0; i < results.size(); i++) {
+                // Set the device local UUID, but only if the local UUID and the identity are identical.
```
bdash: My brain's feeling a little mushy right now, but this doesn't seem to describe what the code below is actually doing?

src/sync/impl/sync_metadata.hpp
```diff
@@ -41,6 +41,10 @@ class SyncUserMetadata {
         size_t idx_identity;
         // A locally issued UUID for the user. This is used to generate the on-disk user directory.
         size_t idx_local_uuid;
+        // Another UUID that is guaranteed (as far as UUIDs go) to be unique to this user and device.
+        // Used for partial sync. (Note that `idx_local_uuid` does NOT make this guarantee.)
+        // This property is identical to `idx_local_uuid` if that property does suffice.
```
bdash: I think it'd help to expand the comment for `idx_local_uuid` to make clear that it may not be a UUID and why (coz we use the user identity when updating from older versions of Realm for compatibility with the existing file system structure?), and then contrast this field with it.

src/sync/sync_config.hpp
```diff
@@ -110,14 +112,40 @@ struct SyncConfig {
     util::Optional<std::string> ssl_trust_certificate_path;
     std::function<sync::Session::SSLVerifyCallback> ssl_verify_callback;
     bool validate_sync_history = true;
+    bool is_partial = false;
+    util::Optional<std::string> custom_partial_sync_identifier;
+
+    std::string resolved_realm_url() const
```
bdash: I'm a little torn on this. There are multiple aspects of the Realm URL that could, in theory, be resolved ('~` -> user ID, and partial sync identifier), but this only handles one aspect of it.

It's also not clear to me in what circumstances we should be calling this function vs accessing `realm_url` directly.

src/sync/impl/sync_metadata.cpp
austinzheng: Yes, you're right. That comment does not describe the implementation I actually went with; I'll fix it :/.

src/sync/sync_user.cpp
```diff
@@ -51,12 +53,16 @@ SyncUser::SyncUser(std::string refresh_token,
...
         });
-        if (!updated)
+        if (!updated) {
             m_local_identity = m_identity;
+            m_device_unique_uuid = util::uuid_string();
+        }
     } else {
         // Admin token users. The local identity serves as the directory path.
         REALM_ASSERT(local_identity);
+        m_device_unique_uuid = *local_identity;
```
bdash: What guarantees that `local_identity` is unique across devices here?

src/sync/sync_config.hpp
austinzheng: The naming needs work, but I figured that it would be better to "construct" the final Realm URL intended to be passed into sync in one place, rather than having partial sync URLs be constructed ad-hoc in multiple places.

If it were up to me accessing `realm_url` directly would be forbidden, but it didn't seem practical to make that drastic a change.

src/sync/sync_config.hpp
```diff
@@ -110,14 +112,40 @@ struct SyncConfig {
...
+    {
+        REALM_ASSERT(realm_url.length() > 0);
+        REALM_ASSERT(user);
+        std::string base_url = realm_url;
+        if (base_url.back() == '/')
```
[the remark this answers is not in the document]
austinzheng: I haven't tried a Realm URL with a trailing slash, but I think they should be rejected (since such a path would indicate a directory path). It may make more sense to throw an exception or take a similar action if the path has a trailing slash, rather than silently stripping it. (On the other hand, I don't want a path to be constructed with two consecutive slashes and then have it fail much later in the sync pipeline.)

src/sync/sync_user.cpp
austinzheng: This requirement puts me in a bit of a pickle, since we don't have persistent storage for admin users. I will revisit this.

PR #548, opened 2017-09-18
Update to latest Core and sync

fealebenpae:
```
REALM_CORE_VERSION=3.2.1
REALM_SYNC_VERSION=2.0.0-rc21
```


src/parser/query_builder.cpp
```diff
@@ -51,7 +52,8 @@ T stot(std::string const& s) {
 // of the function checking its preconditions
 #define precondition(condition, message) if (!REALM_LIKELY(condition)) { throw std::logic_error(message); }
 
-// FIXME: TrueExpression and FalseExpression should be supported by core in some way
+// realm-core comes with TrueExpression and FalseExpression as of version 3.2.1
+#if REALM_VERSION_MAJOR < 3
```
bdash: Why not check for v3.2.1 to avoid introducing an incompatibility with v3.2.0 unnecessarily?

fealebenpae: v3.2.0 is already incompatible because of the new `description` pure virtual method on `realm::Expression`. We could try to support v3.0.x and v3.1.x and only break on v3.2.0 precisely, but we won't be able to target those versions as soon as Sync hits 2.0. I can extend the check if we want to support v3.0.x and v3.1.x explicitly, though - I just didn't feel it was necessary.

bdash: Given the existing incompatibility you mentioned, I think this is fine as-is.

PR #550, opened 2017-09-19
Use constexpr for NotVersioned

beeender:
- Use constexpr for NotVersioned which is easier for java to do
  static_assert check.
- Remove FIXME for ObjectStore::set_schema_version. It is still useful
  for realm-java after switching to use Object Store schema:
  - It is useful for constructing schema related test cases.
  - It is used to create meta tables for realm-java's dynamic Realm.

### DIFF
--- a/src/object_store.cpp
+++ b/src/object_store.cpp
@@ -38,7 +38,6 @@
 
 using namespace realm;
 
-// Make linker happy
 constexpr uint64_t ObjectStore::NotVersioned;
 
 namespace {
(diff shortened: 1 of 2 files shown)


src/object_store.cpp
```diff
@@ -38,7 +38,8 @@
 
 using namespace realm;
 
-const uint64_t ObjectStore::NotVersioned = std::numeric_limits<uint64_t>::max();
+// Make linker happy
```
bdash: I don't think this comment adds anything.

PR #554, opened 2017-09-26
Sync on Windows

fealebenpae:


CMakeSettings.json
```diff
@@ -0,0 +1,135 @@
+{
+  // See https://go.microsoft.com//fwlink//?linkid=834763 for more information about this file.
```
bdash: Does this file really need to exist at the root of the repository?

fealebenpae: Yes, it's a requirement for the Visual Studio IDE experience.

tests/util/test_file.cpp
```diff
@@ -54,33 +52,53 @@ using namespace realm;
...
+			success = true;
+		}
+		catch (const util::File::AccessError& e) {
+			attempts++;
+			std::this_thread::sleep_for(std::chrono::milliseconds(15));
+		}
+	} while (!success && attempts < 3);
+
+	if (!success)
+		printf("Failed to delete realm file at %s\n", path.c_str());
```
kneth: `printf()`? Isn't `std::cerr` better?

fealebenpae: That’s diagnostic code I’ll need to delete before merging. It helped me diagnose that cleaning up test realms on Windows doesn’t work because they’re still in use by sync when the `TestFile` destructor is invoked.

bdash:
What's the status of this PR? Is there a reason we're not pushing to get it merged?

Thomas made some Windows-related fixes in #588, which appear to have introduced some conflicts with this PR.

nirinchev:
Sorry about that. Yavor has been very focused on the native portion of GN for .NET and this has been on a back-burner for the past week. As the GN is on a very aggressive timetable, I don't think he'll be able to get this PR mergeable this week. If you or someone else is available and willing to take it over, that'll be amazing, otherwise Yavor will hopefully be able to get back at it early next week (considering there are no GN-related surprises).

As for the PR, as far as I remember, the major issue is around the way tests are structured. On unix-based systems, deleting a file in use is not a problem but on Windows is an error which causes tests that depend on ROS to fail because the in-memory server instance is closed after the file deletion is attempted. If someone is willing to look into it, I'm sure Yavor will be more than happy to hand it over and explain things in more detail.

bmunkholm:
@fealebenpae Should this be put "On Hold" or in the Backlog?

PR #556, opened 2017-09-28
Added use/handling of move_row

finnschiermer:
This PR attempts to align to Core 4.0 by
 * setting Core 4.0 as a dependency
 * Having list::move use Cores Table::move_row instead of swap_rows
 * Adding handling of move_row to the transaction log observer
 * Adding trivial handling of move_row to the transaction log validation mixin
I'm unsure about the changes to the transaction log observer.
The existing unit-test does not appear to trigger execution of the new code.


src/impl/transact_log_handler.cpp
```diff
@@ -609,6 +610,14 @@ class TransactLogObserver : public TransactLogValidationMixin, public MarkDirtyM
         return true;
     }
 
+    bool move_row(size_t from_ndx, size_t to_ndx) {
+        if (m_active_table)
+            m_active_table->move(from_ndx, to_ndx);
+        // Move row is not supported for top level tables:
+	REALM_ASSERT(m_is_top_level_table);
```
tgoyne: This assertion is backwards (and indentation is wrong).

bmunkholm:
@tgoyne @bdash Can you take this one over and merge asap?

PR #558, opened 2017-10-02
Add model definitions for PermissionOffer and PermissionOfferResponse

austinzheng:
This accomplishes a few things:
- Prevents ROS from complaining
- Allows us to leverage the object store level permissions code in the future to implement permission offer/response at this layer

### DIFF
--- a/src/sync/sync_permission.hpp
+++ b/src/sync/sync_permission.hpp
@@ -121,10 +121,6 @@ public:
 private:
     static SharedRealm management_realm(std::shared_ptr<SyncUser>, const ConfigMaker&);
     static SharedRealm permission_realm(std::shared_ptr<SyncUser>, const ConfigMaker&);
-
-    static ObjectSchema permission_change_schema();
-    static ObjectSchema permission_offer_schema();
-    static ObjectSchema permission_offer_response_schema();
 };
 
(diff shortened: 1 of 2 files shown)


bdash:
Ugh, I thought this was supposed to be fixed by realm/ros#420, but apparently that never made it into a release…

src/sync/sync_permission.hpp
```diff
@@ -121,6 +121,10 @@ class Permissions {
 private:
     static SharedRealm management_realm(std::shared_ptr<SyncUser>, const ConfigMaker&);
     static SharedRealm permission_realm(std::shared_ptr<SyncUser>, const ConfigMaker&);
+
+    static ObjectSchema permission_change_schema();
```
bdash: Is there any reason for these to be static methods? Having them be inline in `management_realm()` seems fine to me, but if you feel having them as separate methods is preferable then they could just as easily be in static functions in `sync_permission.cpp` rather than static methods on the `SyncPermission` type.

austinzheng: Hmm...I think inline probably works for us. We never refer back to the `ObjectSchema` objects directly; our API uses lookup by the object name, so breaking them out into separate methods isn't buying us anything.

PR #559, opened 2017-10-03
Add support for permission offer/response

austinzheng:
Changes:
- Added APIs for creating permission offers and accepting them
- Refactored some code to facilitate reuse
- Renamed a few types to reflect their new, expanded usage

Rationale: right now, most bindings (Cocoa included) use the object store APIs to get and change permissions, but binding-level APIs to create and accept permission offers. This API allows the existing object store machinery to be used to perform the latter task as well.

More detailed description.
1. I broke most of the existing method for changing permissions into a new method, `perform_async_operation()`. This method basically creates an object, writes it to the management Realm, and observes it until its status code indicates either success or failure. Then it calls a callback.
2. Changing permissions, creating offers, and accepting offers now call into `perform_async_operation()` with customization options (e.g. what object to create, what additional properties to set on it).

### DIFF
--- a/src/sync/sync_permission.cpp
+++ b/src/sync/sync_permission.cpp
@@ -38,21 +38,21 @@ using namespace std::chrono;
 // MARK: - Utility
 
 namespace {
-Permission::AccessLevel extract_access_level(Object& permission, CppContext& context)
+AccessLevel extract_access_level(Object& permission, CppContext& context)
 {
     auto may_manage = permission.get_property_value<util::Any>(context, "mayManage");
     if (may_manage.has_value() && any_cast<bool>(may_manage))
-        return Permission::AccessLevel::Admin;
+        return AccessLevel::Admin;
 
     auto may_write = permission.get_property_value<util::Any>(context, "mayWrite");
     if (may_write.has_value() && any_cast<bool>(may_write))
-        return Permission::AccessLevel::Write;
+        return AccessLevel::Write;
 
     auto may_read = permission.get_property_value<util::Any>(context, "mayRead");
     if (may_read.has_value() && any_cast<bool>(may_read))
-        return Permission::AccessLevel::Read;
+        return AccessLevel::Read;
 
-    return Permission::AccessLevel::None;
+    return AccessLevel::None;
 }
 
 /// Turn a system time point value into the 64-bit integer representing ns since the Unix epoch.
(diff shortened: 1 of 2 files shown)


src/sync/sync_permission.cpp
```diff
@@ -157,7 +157,91 @@ void Permissions::set_permission(std::shared_ptr<SyncUser> user,
...
 {
-    const auto realm_url = user->server_url() + permission.path;
+    auto props = AnyDict{
+        {"userId", permission.condition.user_id},
+        {"realmUrl", user->server_url() + permission.path},
+        {"mayRead", permission.access != AccessLevel::None},
+        {"mayWrite", permission.access == AccessLevel::Write || permission.access == AccessLevel::Admin},
+        {"mayManage", permission.access == AccessLevel::Admin},
+    };
+    switch (permission.condition.type) {
```
tgoyne: A one-case switch is sort of silly.

src/sync/sync_permission.cpp
```diff
@@ -157,7 +157,91 @@ void Permissions::set_permission(std::shared_ptr<SyncUser> user,
...
+        case Permission::Condition::Type::KeyValue:
+            props.insert({"metadataKey", permission.condition.key_value.first});
+            props.insert({"metadataValue", permission.condition.key_value.second});
+            break;
+        default:
+            break;
+    }
+    auto cb = [callback=std::move(callback)](Object*, std::exception_ptr exception) {
+        if (exception) {
```
tgoyne: This if/else is redundant.

src/sync/sync_permission.cpp
```diff
@@ -157,7 +157,91 @@ void Permissions::set_permission(std::shared_ptr<SyncUser> user,
...
+{
+    auto props = AnyDict{
+        {"token", token},
+    };
+    auto cb = [callback=std::move(callback)](Object* object, std::exception_ptr exception) {
+        if (exception) {
+            callback(none, exception);
+        } else {
+            CppContext context;
+            auto token = any_cast<std::string>(object->get_property_value<util::Any>(context, "realmUrl"));
```
tgoyne: The string here is the only thing that differs from the above, so it's probably worth extracting this all to a function.

src/sync/sync_permission.cpp
```diff
@@ -166,38 +250,27 @@ void Permissions::set_permission(std::shared_ptr<SyncUser> user,
...
-    // Set condition properties based on type
-    switch (permission.condition.type) {
-        case Permission::Condition::Type::KeyValue:
-            raw.set_property_value<util::Any>(context, "metadataKey", permission.condition.key_value.first, false);
-            raw.set_property_value<util::Any>(context, "metadataValue", permission.condition.key_value.second, false);
-            break;
-        default:
-            break;
+    };
+    for (auto&& prop : additional_props) {
```
tgoyne: `props.insert(additional_props.begin(), additional_props.end());`

src/sync/sync_permission.cpp
```diff
@@ -261,23 +325,23 @@ SharedRealm Permissions::management_realm(std::shared_ptr<SyncUser> user, const
...
             Property{"updatedAt",         PropertyType::Date},
-            Property{"expiresAt",         PropertyType::Date},
             Property{"statusCode",        PropertyType::Int|PropertyType::Nullable},
             Property{"statusMessage",     PropertyType::String|PropertyType::Nullable},
             Property{"token",             PropertyType::String|PropertyType::Nullable},
             Property{"realmUrl",          PropertyType::String},
             Property{"mayRead",           PropertyType::Bool},
             Property{"mayWrite",          PropertyType::Bool},
             Property{"mayManage",         PropertyType::Bool},
+            Property{"expiresAt",         PropertyType::Date|PropertyType::Nullable},
```
tgoyne: Why is this moved to the end?

src/sync/sync_permission.hpp
```diff
@@ -22,28 +22,31 @@
 #include "results.hpp"
 #include "shared_realm.hpp"
 
+#include "util/any.hpp"
```
tgoyne: A forward declaration of `Any` suffices here.

src/sync/sync_permission.cpp
austinzheng: I was trying to troubleshoot something; forgot to move it back. Thank you for catching this.

src/sync/sync_permission.cpp
```diff
@@ -213,28 +278,19 @@ void Permissions::set_permission(std::shared_ptr<SyncUser> user,
...
-                                  : util::format("Error code: %1", code);
-            callback(std::make_exception_ptr(PermissionChangeException(error_str, code)));
+            std::string error_str = (status.has_value()
+                                     ? any_cast<std::string>(status)
+                                     : util::format("Error code: %1", code));
+            handler(nullptr, std::make_exception_ptr(PermissionActionException(error_str, code)));
         }
         else {
-            callback({});
+            handler(&*object, nullptr);
```
bdash: Isn't `&*` here equivalent to `object.get()`?

src/sync/sync_permission.cpp
```diff
@@ -157,7 +157,85 @@ void Permissions::set_permission(std::shared_ptr<SyncUser> user,
...
+                               const ConfigMaker& make_config)
+{
+    perform_async_operation("PermissionOfferResponse",
+                            std::move(user),
+                            make_handler_extracting_property("realmUrl", std::move(callback)),
+                            AnyDict{ {"token", token} },
+                            make_config);
+}
+
+Permissions::AsyncOperationHandler Permissions::make_handler_extracting_property(std::string property,
```
bdash: Is there any reason for this to be a static member function rather than just a static function in the .cpp file?

src/sync/sync_permission.cpp
austinzheng: Ah, that's it.

austinzheng:
@bdash you are correct about the tests. We can't test anything dependent on ROS at the object store level currently.

nirinchev:
> Rationale: right now, most bindings (Cocoa included) use the object store APIs to get and change permissions, but binding-level APIs to create and accept permission offers. 

To clarify, Cocoa is the only binding that uses the object store API for anything. .NET, Java, and js have convenience wrappers around the binding-level objects - none of them touches OS.

PR #560, opened 2017-10-05
Fix SyncConfig initialization to be consistent between Clang and GCC >= 5

bdash:
Always use the `SyncConfig` constructor. It removes the potential for inconsistent initialization between the constructor and braced initializers. In addition, the default constructor is removed as it doesn't result in a meaningful `SyncConfig`.

At some point soon we should reevaluate how `SyncConfig` is used. It was initially intended to contain only two or three values, but has grown to over ten ☹️

### DIFF
--- a/src/sync/sync_config.hpp
+++ b/src/sync/sync_config.hpp
@@ -112,10 +112,10 @@ struct SyncConfig {
     std::function<SyncSessionErrorHandler> error_handler;
     std::shared_ptr<ChangesetTransformer> transformer;
     util::Optional<std::array<char, 64>> realm_encryption_key;
-    bool client_validate_ssl = true;
+    bool client_validate_ssl;
     util::Optional<std::string> ssl_trust_certificate_path;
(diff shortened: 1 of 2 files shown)


src/sync/sync_config.hpp
```diff
@@ -115,10 +115,11 @@ struct SyncConfig {
     bool client_validate_ssl = true;
     util::Optional<std::string> ssl_trust_certificate_path;
     std::function<sync::Session::SSLVerifyCallback> ssl_verify_callback;
-    bool validate_sync_history = true;
     bool is_partial = false;
     util::Optional<std::string> custom_partial_sync_identifier;
 
+    bool validate_sync_history = true;
+
```
kspangsege: As a short term solution, I would suggest committing to having the constructor. This introduces a fortunate decoupling from the actual member layout.

As a longer term solution, I propose that user sites declare default instances of `SyncConfig`, and then assign explicitly to its members.

Anyway, I agree that your fix works.

PR #561, opened 2017-10-06
Unregister sync_session_states::Error session when it's closed

nhachicha:
- According to the [documentation](https://github.com/realm/realm-object-store/blob/7906c22a4837657243161e5f17dce5147ac50c28/src/sync/sync_session.cpp#L84-L88) the error state is terminal, and the session should not transition to another state.

when a client reset happens and we close the Realm, [SyncSession::State#close](https://github.com/realm/realm-object-store/blob/7906c22a4837657243161e5f17dce5147ac50c28/src/sync/sync_session.cpp#L119) will be called on the session but since it's a no op (in the case of ` sync_session_states::Error`) it will remain in the `sync_manager.hpp` [m_session](https://github.com/realm/realm-object-store/blob/7906c22a4837657243161e5f17dce5147ac50c28/src/sync/sync_manager.hpp#L198) map.

This will cause an error in Java binding when trying to open the Realm again, and call `session->wait_for_download_completion` for instance.

Java interface with the OS using the `session_path` (`SyncManager::get_existing_session(const std::string& path)`), that will return the remaining error session (which is unusable).
Actually unregistering the session when close is called on error state will help prevent this use case

- I also introduced `SyncManager::get_existing_session` in this [PR] (https://github.com/realm/realm-object-store/pull/398) when I integrated SyncManager with Java. I removed this method in the OS and replaced the call from Java to `get_existing_active_session` (I think we were the only binding using the deleted method)

### DIFF
--- a/src/sync/sync_manager.cpp
+++ b/src/sync/sync_manager.cpp
@@ -442,15 +442,6 @@ std::shared_ptr<SyncSession> SyncManager::get_existing_session_locked(const std:
     return it == m_sessions.end() ? nullptr : it->second;
 }
 
-std::shared_ptr<SyncSession> SyncManager::get_existing_session(const std::string& path) const
-{
-    std::lock_guard<std::mutex> lock(m_session_mutex);
-    if (auto session = get_existing_session_locked(path))
-        return session->external_reference();
-
-    return nullptr;
-}
-
 std::shared_ptr<SyncSession> SyncManager::get_session(const std::string& path, const SyncConfig& sync_config)
 {
     auto& client = get_sync_client(); // Throws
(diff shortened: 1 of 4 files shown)


austinzheng:
I think this is actually a bug in the `sync_session` code.

In `SyncSession::handle_error()`, we have a `next_state` variable that determines what state the session transitions into (if any). Normally, this next state is `error`, but upon a bad auth error the next state is set to `inactive` instead. If you look at line 544, which handles client reset errors, you will note I forgot to set the next state to `inactive` there as well. I think that is the proper fix, because it preserves the `error` state invariant you described above.

Let me know what you think.

nhachicha:
@austinzheng sounds good, reverted the change to set the `next_state` instead.

bdash:
Why did you leave the `get_existing_session` change in place? It appears that it's not really related to the issue being fixed, and it looks like it'll result in the two Realms in question being null in cases where they weren't before (it'll no longer reactivate an inactive session, but would instead return null).

nhachicha:
@bdash reverted the changes

PR #562, opened 2017-10-06
Exclude 'none' permissions from permission results

austinzheng:
ROS may indicate the absence of a permission in two ways: the absence of a permission object, or the setting of the flags on the permission object all to false. This change filters out the latter type of permission, since it's functionally identical to a non-existent permission.


bdash:
Is hiding these at the object store level the best option? One alternative would be for ROS to remove objects that don't represent any permissions.

austinzheng:
I was thinking that as well. I'll talk to the ROS folks and see if they have any opinion either way.

austinzheng:
It makes more sense to do this at the server level, and I just verified that we can.

PR #565, opened 2017-10-10
Add debug information for internal session states

cmelchior:
First step in making the inner workings of Sessions available to external debugging.

This adds a name to each state that can be used during `advance_state` to log the state transition, which is really helpful for debugging lifecycle issues.

Unfortunately, there doesn't seem to be a logger component in Object Store currently (or I couldn't find it), so for now this is up to bindings to do this manually.

Java e.g. can do this:

```
    std::ostringstream out;
    out << "SyncSession: " << path() << " (old:" << m_state->name() << " -> " << state.name() << ")";
    __android_log_print(ANDROID_LOG_ERROR, "Session state change", out.str().c_str(), 1);
```

### DIFF
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -130,6 +130,9 @@ struct SyncSession::State {
         return false;
     }
 
+    // Returns the name of this state. Only useful for debugging
+    virtual std::string name() const { return ""; };
+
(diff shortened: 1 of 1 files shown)


cmelchior:
Cool, I didn't knew that was possible. No reason for this PR then. Closing

PR #567, opened 2017-10-11
Make ExternalCommitHelper robust on Windows

fealebenpae:
Starting with Core 3.0 we have a robust interprocess mutex and condition variable implementation on Windows we can use to reliably notify interested parties of commits at the Object Store level.

The implementation heavily borrows from `SharedGroup::wait_for_change()`. In fact, I believe the generic ExternalCommitHelper implementation should perform equally well on Windows - we’d be using it instead of this one if it only worked with synced realms.

### DIFF
--- a/src/impl/windows/external_commit_helper.cpp
+++ b/src/impl/windows/external_commit_helper.cpp
@@ -17,69 +17,65 @@
 ////////////////////////////////////////////////////////////////////////////
 
 #include "impl/external_commit_helper.hpp"
-
 #include "impl/realm_coordinator.hpp"
 
 #include <algorithm>
-#include <codecvt>
 
 using namespace realm;
 using namespace realm::_impl;
 
-static HANDLE CreateNotificationEvent(std::string realm_path)
-{
-    // replace backslashes because they're significant in object namespace names
-    std::replace(realm_path.begin(), realm_path.end(), '\\', '/');
-
-    std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
-    std::wstring path(L"Local\\" + converter.from_bytes(realm_path));
-
-    HANDLE event = CreateEventEx(nullptr, path.c_str(), CREATE_EVENT_MANUAL_RESET, SYNCHRONIZE | EVENT_MODIFY_STATE);
-    if (event == nullptr) {
-        throw std::system_error(GetLastError(), std::system_category());
-    }
-
-    return event;
-}
-
 ExternalCommitHelper::ExternalCommitHelper(RealmCoordinator& parent)
 : m_parent(parent)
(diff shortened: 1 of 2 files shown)


src/impl/windows/external_commit_helper.cpp
```diff
@@ -17,69 +17,65 @@
...
-, m_event(CreateNotificationEvent(parent.get_path()))
-, m_close_mutex(CreateMutexEx(nullptr, nullptr, CREATE_MUTEX_INITIAL_OWNER, SYNCHRONIZE | MUTEX_MODIFY_STATE))
 {
+    std::string path = parent.get_path();
+    std::replace(path.begin(), path.end(), '\\', '/');
+    std::wstring shared_memory_name = L"Local\\Realm_ObjectStore_ExternalCommitHelper_SharedCondVar_" + std::wstring(path.begin(), path.end());
+#if REALM_WINDOWS
+    m_shared_memory = CreateFileMappingW(INVALID_HANDLE_VALUE, nullptr, PAGE_READWRITE, 0, sizeof(*m_condvar_shared), shared_memory_name.c_str());
+    auto error = GetLastError();
+    m_condvar_shared = reinterpret_cast<InterprocessCondVar::SharedPart*>(MapViewOfFile(m_shared_memory, FILE_MAP_ALL_ACCESS, 0, 0, sizeof(*m_condvar_shared)));
```
tgoyne: Won't this crash if `CreateFileMappingW` failed? I would expect the error checking to need to appear before code using the result.

src/impl/windows/external_commit_helper.cpp
```diff
@@ -17,69 +17,65 @@
...
+    auto error = GetLastError();
+    m_condvar_shared = reinterpret_cast<InterprocessCondVar::SharedPart*>(MapViewOfFileFromApp(m_shared_memory, FILE_MAP_ALL_ACCESS, 0, sizeof(*m_condvar_shared)));
+#endif
+    if (error == 0) {
+        InterprocessCondVar::init_shared_part(*m_condvar_shared);
+    } else if (error != ERROR_ALREADY_EXISTS) {
+        throw std::system_error(error, std::system_category());
+    }
+
+    m_mutex.set_shared_part(InterprocessMutex::SharedPart(), parent.get_path(), "ExternalCommitHelper_ControlMutex");
```
tgoyne: If any of this code throws the file mapping will be leaked. This can be fixed by using something like the `FdHolder` type used in other helper implementations to do the closing rather than having cleanup logic in the destructor (as a general rule of thumb if you have cleanup logic in a destructor in C++ and that isn't the only thing the class does there's a good chance you'll leak memory when exceptions are thrown).

rrrlasse:
@finnschiermer I think you could review this better

PR #568, opened 2017-10-11
Add the ability to override the server that's used by a given SyncSession

### LINKED ISSUE #563: Add the ability to force a SyncSession reconnect to a different endpoint
In ROS, using realm-js, we need to be able to reconfigure the location of a Realm while the session is open.  

It was previously discussed that it would be nice to be able to just change the URL on an existing session, but that may have larger implications with the internal state machine in object store.

Another proposal is to add an optional `resolver` callback to the Sync configuration, which is called when opening or reopening a connection to a sync server.  The resolver callback, when invoked with the original URL by the binding, should return a (in JS terms) Promise that will resolve the host part of the URL to an IP address and port. This IP address and port should then be used by the sync session to (re)open the underlying connection.

This way, the underlying session and URL never change, only the actual destination of the connection.

//cc @bigfish24 @bdash

bdash:
This is for internal use by ROS as part of its support for failover. If the `SyncSession` is active, we ask the underlying `sync::Session` to disconnect and reconnect to the new server. We also stash away the overridden server's details so it can be used again we later disconnect and reconnect. If the session is not yet active, we simply stash away the details to be used when establishing the connection.

Fixes #563.

Blocked on getting the necessary sync API added (https://github.com/realm/realm-sync/commit/1f08d1534cc340c3e4303b24d14b0be6c7de906e, but there's no corresponding PR at this time).

Still needs:
- [x] Update dependencies.list to a version that includes the new sync API.
- [ ] Tests
- [x] Documentation


bdash:
Testing this is a little bit complicated as we need to fake having a second server with the same server-side state as the first server. We can likely simulate this by just listening on a second port and proxying all of the data received over to the real server. I've tested it manually by doing the proxying using `nc`, and have started work on adding unit tests, but they're much more involved than the code change itself was.

kneth:
@bdash I am merging now, and I hope you will have time to finish the tests.

src/sync/sync_session.hpp
```diff
@@ -149,7 +149,9 @@ class SyncSession : public std::enable_shared_from_this<SyncSession> {
     //
     // NOTE: This is intended for use only in very specific circumstances. Please check with the
     // object store team before using it.
+#if (REALM_VERSION_MAJOR >= 4)
```
kvap: It should probably be guarded by the version of `realm-sync`, because the function is from there.

kneth: Yes, a better check is `REALM_SYNC_VER_MAJOR > 1`.

kneth: I will do it in a separate PR later.

PR #572, opened 2017-10-12
Repair bad metadata key migration

### LINKED ISSUE #569: Metadata key migration contains a bug.
While investigating an issue https://github.com/realm/realm-cocoa/issues/5391 bug was found.

https://github.com/realm/realm-object-store/blob/master/src/sync/impl/sync_metadata.cpp#L90-L98

First you move path variable, and then you check if file exists at path. Sure it's not. And legacy key is not checked. Beep.

austinzheng:
TODO:
- [ ] Manual testing

### DIFF
--- a/src/impl/apple/keychain_helper.cpp
+++ b/src/impl/apple/keychain_helper.cpp
@@ -107,6 +107,23 @@ void set_key(const std::vector<char>& key, CFStringRef account, CFStringRef serv
         throw KeychainAccessException(status);
 }
 
+void update_key(const std::vector<char>& key, CFStringRef account, CFStringRef service)
+{
+    auto search_dictionary = build_search_dictionary(account, service, none);
+    auto key_data = adoptCF(CFDataCreate(nullptr, reinterpret_cast<const UInt8 *>(key.data()), key_size));
+    if (!key_data)
+        throw std::bad_alloc();
+
+    const CFStringRef keys[1] = { kSecValueData };
+    const CFDataRef values[1] = { key_data.get() };
+    auto update_dictionary = adoptCF(CFDictionaryCreate(NULL,
+                                                        (const void**)keys, (const void**)values, 1,
+                                                        &kCFTypeDictionaryKeyCallBacks,
+                                                        &kCFTypeDictionaryValueCallBacks));
+    if (OSStatus status = SecItemUpdate(search_dictionary.get(), update_dictionary.get()))
+        throw KeychainAccessException(status);
+}
+
 }   // anonymous namespace
 
 std::vector<char> metadata_realm_encryption_key(bool check_legacy_service)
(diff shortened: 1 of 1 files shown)


src/impl/apple/keychain_helper.cpp
```diff
@@ -109,21 +109,35 @@ void set_key(const std::vector<char>& key, CFStringRef account, CFStringRef serv
...
     else {
         service = retainCF(legacy_service);
+        skip_legacy_recheck = true;
         check_legacy_service = false;
     }
 
     // Try retrieving the key.
     if (auto existing_key = get_key(account, service.get())) {
+        if (recheck_legacy_service && !skip_legacy_recheck) {
+            // 2.10.0 introduced a bug where a new key might be generated even though
```
bdash: I think it'd be preferable to reference a commit or PR here rather than a Cocoa version number.

src/sync/impl/sync_metadata.hpp
```diff
@@ -214,7 +214,8 @@ friend class SyncFileActionMetadata;
     /// exception will be thrown.
     SyncMetadataManager(std::string path,
                         bool should_encrypt,
-                        util::Optional<std::vector<char>> encryption_key=none);
+                        util::Optional<std::vector<char>> encryption_key=none,
+                        bool should_recheck_keychain=false);
```
bdash: `SyncMetadataManager` is internal to object store, and it doesn't look like any callers were updated to pass this argument. How is this code path ever activated?

src/sync/impl/sync_metadata.cpp
```diff
@@ -82,18 +82,19 @@ namespace realm {
 
 SyncMetadataManager::SyncMetadataManager(std::string path,
                                          bool should_encrypt,
-                                         util::Optional<std::vector<char>> encryption_key)
+                                         util::Optional<std::vector<char>> encryption_key,
+                                         __unused bool should_recheck_keychain)
```
tgoyne: `__unused` is GCC/clang-specific. The pattern core uses for parameters only used on some platforms is a `static_cast<void>(param)` in the `#else` block.

austinzheng: Oh, that makes sense. Thanks!

src/impl/apple/keychain_helper.cpp
```diff
@@ -109,21 +109,36 @@ void set_key(const std::vector<char>& key, CFStringRef account, CFStringRef serv
 
 }   // anonymous namespace
 
-std::vector<char> metadata_realm_encryption_key(bool check_legacy_service)
+std::vector<char> metadata_realm_encryption_key(bool check_legacy_service, bool recheck_legacy_service)
```
bdash: Why did you leave this argument in place? The only caller always passes a constant value for it.

austinzheng: I was thinking we might want to make it conditional in the future, or even remove it altogether. But maybe it makes more sense to just not have it at all.

bdash: When we determine what to do we'll have to make a code change, so we can make it conditional at that point. As it stands, having it look like it's conditional when it's not just makes the code harder to follow (IMO).

austinzheng: Yeah, I agree with you. I'll change it.

src/impl/apple/keychain_helper.cpp
```diff
@@ -95,16 +97,32 @@ util::Optional<std::vector<char>> get_key(CFStringRef account, CFStringRef servi
     return std::vector<char>(key_bytes, key_bytes + key_size);
 }
 
-void set_key(const std::vector<char>& key, CFStringRef account, CFStringRef service)
+void set_key(const std::vector<char>& key, CFStringRef account, CFStringRef service, SetKeyMode mode)
```
bdash: This is starting to look like it'd make more sense as two separate functions, `add_key` and `update_key`, rather than a single function with a mode flag.

bdash:
For sake of expediency, how about we land the _fix_ for the original issue (i.e., removing the `std::move`), and pull out the recovery into a separate PR where we can spend a little more time getting it right?

molind:
When app is launched with 2.10.2 it's rewrites old metadata.realm with new one. Because old couldn't be read. And when you have two keys in keychain it's not always legacy key is the one you should use. Best fix would be to check what key could decrypt metadata.realm and store it as new one. Or always try new key, then catch an exception and if it's encryption - it should try to read metadata with the legacy key.

austinzheng:
Thanks for bringing this to our attention.

- If the metadata Realm is deleted because of an invalid key error (as you stated), a new Realm will be generated. It will use the newly added keychain item as its encryption key.
- Upon all subsequent launches the metadata subsystem will find the new key, use it properly, and ignore the old key.
- We don't make any provision for deleting the old key; you may do so yourself if you desire. That being said, once the new key is added there is no code path for removing the old key. The new key is added either as a newly generated random value or as a copy of the old key, and is exclusively used from this point on. The bug in question skipped the copying of the old key and required the Realm to be deleted and recreated with the newly generated key as a result.

As a result this fix is unnecessary and users who have already experienced this bug, while having lost their metadata, should now have properly functioning metadata.

PR #573, opened 2017-10-12
Defer delivery of the initial notification when a callback is added from within a callback on the same collection

tgoyne:
The previous behavior of sending the initial notification immediately would result in notifications being lost if the background worker ran at exactly the wrong time. Fixing this turned out to be complicated and sending the initial notification immediately was inconsistent with the normal behavior, so just don't.

The newly added test is a simplification of what the java permissions code is doing that results in https://github.com/realm/realm-java/issues/5413, and without these changes is called once with `table->get_int(0, 0) == 0`, and then is not called again when the realm is advanced to the version with it set to 5.

### DIFF
--- a/src/impl/collection_notifier.cpp
+++ b/src/impl/collection_notifier.cpp
@@ -213,8 +213,10 @@ void CollectionNotifier::remove_callback(size_t token)
         }
 
         size_t idx = distance(begin(m_callbacks), it);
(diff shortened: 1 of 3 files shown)


cmelchior:
While the unit test looks correct, I cannot yet confirm it working on our integration test. I'm trying to get more information

PR #574, opened 2017-10-13
Fix improper migration of sync metadata keychain item

### LINKED ISSUE #569: Metadata key migration contains a bug.

austinzheng:
#572 will provide a fix for users who have already encountered this bug.

Fixes #569


bdash:
Thanks for splitting this out, Austin.

PR #577, opened 2017-10-13
Fix some use-after-move errors

austinzheng:
`clang-tidy` uncovered a few issues.


austinzheng:
That's a relief, good to know.

PR #578, opened 2017-10-13
Fix two use-after-move issues flagged in sync-related code

austinzheng:
It seems these are somewhat more serious than those from #577. Without these fixes, a user's internal permissions and management Realms will continue syncing even if the user logs out.

### DIFF
--- a/src/sync/sync_permission.cpp
+++ b/src/sync/sync_permission.cpp
@@ -341,9 +341,8 @@ SharedRealm Permissions::management_realm(std::shared_ptr<SyncUser> user, const 
         }},
     };
     config.schema_version = 0;
-    const std::string path = config.path;
     auto shared_realm = Realm::get_shared_realm(std::move(config));
-    user->register_management_session(path);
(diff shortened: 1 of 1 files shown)


src/sync/sync_permission.cpp
```diff
@@ -341,8 +341,9 @@ SharedRealm Permissions::management_realm(std::shared_ptr<SyncUser> user, const
         }},
     };
     config.schema_version = 0;
+    const std::string path = config.path;
     auto shared_realm = Realm::get_shared_realm(std::move(config));
-    user->register_management_session(config.path);
+    user->register_management_session(path);
```
tgoyne: Could do shared_realm->config().path to avoid the extra local copy of the path, but it's not very important.

tgoyne: On second thought, while the perf difference is whatever, adopting the general pattern of reading the values from the realm whenever possible seems like it'd increase the chance of accidentally writing correct code in cases where it appears to be safe to use either (but actually isn't).

PR #584, opened 2017-10-23
rename update_tableview to evaluate_query_if_needed and expose it to public

beeender:
The orignal purpose of not creating TableView in size() is to make the
call faster if the query has not been executed before. But in the real
usecases, users always try to call accessors after the size() which
makes the query gets executed twice. See java issue
https://github.com/realm/realm-java/issues/5328
https://github.com/realm/realm-java/issues/5387

With this change, there would be no difference if the query has been
evaluated before.

It will be faster if:
- Calling Results::size() multiple times in the same event loop.
- Calling any accessors after Results::size().

It will be slower if:
- Calling Results::size() at the first time.

### DIFF
--- a/src/results.cpp
+++ b/src/results.cpp
@@ -138,7 +138,7 @@ size_t Results::size()
                 return m_query.count();
             REALM_FALLTHROUGH;
         case Mode::TableView:
-            update_tableview();
+            evaluate_query_if_needed();
             return m_table_view.size();
     }
     REALM_COMPILER_HINT_UNREACHABLE();
@@ -201,7 +201,7 @@ util::Optional<T> Results::try_get(size_t row_ndx)
             REALM_FALLTHROUGH;
         case Mode::Query:
         case Mode::TableView:
-            update_tableview();
+            evaluate_query_if_needed();
             if (row_ndx >= m_table_view.size())
                 break;
             if (m_update_policy == UpdatePolicy::Never && !m_table_view.is_row_attached(row_ndx))
@@ -230,7 +230,7 @@ util::Optional<T> Results::last()
 {
     validate_read();
     if (m_mode == Mode::Query)
-        update_tableview(); // avoid running the query twice (for size() and for get())
+        evaluate_query_if_needed(); // avoid running the query twice (for size() and for get())
     return try_get<T>(size() - 1);
 }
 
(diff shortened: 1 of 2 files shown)


tgoyne:
:-1:

The original reason why we did this (to provide a fast way to get the number of objects matching the  query) is still fully relevant, and in the cases where they do actually go on to do something that requires a tableview, running the query for the count is typically not a significant portion of the runtime.

beeender:
From my testing, when the data set is big, `Query::count()` and `Query::find_all()` take similar time to run. I will do some more benchmark to get some numbers.

bmunkholm:
@beeender The difference is likely insignificant in big datasets with a low number of matches. So try with big datasets with high number of matches (i.e. you have to create large TableView)

beeender:
Made some benchmark for this https://gist.github.com/beeender/a2e5d02257df61eadf70909583358ec8

results:
```
## data set: 3000 loops: 1000
### No condition query
#### size
Time cost: 5

## data set: 3000 loops: 1000
### No condition query
#### get(0)
Time cost: 1502

## data set: 3000 loops: 1000
### Conditions query, no index
#### size
Time cost: 21

## data set: 3000 loops: 1000
### Conditions query, no index
#### get(0)
Time cost: 510

## data set: 3000 loops: 1000
### Conditions query, indexed
#### size
Time cost: 23

## data set: 3000 loops: 1000
### Conditions query, indexed
#### get(0)
Time cost: 548

## data set: 3000 loops: 1000
### Conditions string query
#### size
Time cost: 892

## data set: 3000 loops: 1000
### Conditions string query
#### get(0)
Time cost: 919
```

The `results.size()` will be much faster than updating the table view when the query doesn't involve string queries.

With string queries, the `size()` and first `get()` calls take similar time. And that is the situation java users met.

I don't have too much ideas than exposing the `update_table_view()` as a public API. Java differentiate the sync & async queries in public APIs because of the historical reasons. So it would be acceptable for user to immediately run the query, get the table view if he is using the synced query API.

@tgoyne any suggestions?

tgoyne:
In tests we call `.last()` when we want to force it to evaluate the query, but of course that's terribly hacky. Exposing `update_tableview()` sounds reasonable to me, although it should have a different name (`evaluate_query_if_needed()`?).

beeender:
@tgoyne PR updated.

PR #587, opened 2017-10-27
Expose an option to enable sync multiplexing

tgoyne:
This is disabled by default as it doesn't work with load balancing, but the GN really needs it.

Since the option to disable multiplexing was originally just there for tests the API for this is pretty awkward, but we can sort that out later.

I'm not really sure how to usefully test this at the objectstore layer, but I have verified that it achieves the desired end result (by calling this function the GN is able to open >10k realms rather than hitting the open file limit at 1500).

### DIFF
--- a/src/sync/sync_manager.hpp
+++ b/src/sync/sync_manager.hpp
@@ -78,11 +78,12 @@ public:
     // The metadata and file management subsystems must also have already been configured.
     bool immediately_run_file_actions(const std::string& original_name);
 
-    // Use a single connection for all sync sessions rather than one per session.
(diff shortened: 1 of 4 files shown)


src/sync/sync_manager.hpp
```diff
@@ -78,6 +78,12 @@ friend class SyncSession;
     // The metadata and file management subsystems must also have already been configured.
     bool immediately_run_file_actions(const std::string& original_name);
 
+    // Use a single connection for all sync sessions rather than one per session.
```
bdash: A single connection for all sync sessions that share a host / port, presumably?

PR #588, opened 2017-10-30
Fix sync-enabled Windows compilation

tgoyne:
realm-js is now testing this on CI and it turns out it never worked.

### DIFF
--- a/src/sync/impl/sync_file.cpp
+++ b/src/sync/impl/sync_file.cpp
@@ -83,35 +83,6 @@ char decoded_char_for(const std::string& percent_encoding, size_t index)
 
 namespace util {
 
-void remove_nonempty_dir(const std::string& path)
-{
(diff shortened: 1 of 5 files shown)


src/sync/impl/sync_file.cpp
```diff
@@ -83,35 +83,6 @@ char decoded_char_for(const std::string& percent_encoding, size_t index)
 
 namespace util {
 
-void remove_nonempty_dir(const std::string& path)
```
bdash: There should be a corresponding removal of this from sync_file.hpp. The two callers in our tests will need to be updated to use `util::try_remove_dir_recursive` as well.

tgoyne: Whoops, missed that this wasn't a private helper function.

PR #590, opened 2017-10-31
Improve sync config

### LINKED ISSUE #579: Prevent addition of 2nd `/__partial/...` suffix to Realm path as a result of `partial: true`
Currently, if one specifies a sync URL with a path that explicitly contains the `__partial/<unique ident>` suffix, and then also enables partial sync through `partial` property on sync config object, then the path that is sent to the server, will end up containing the substring `/__partial/` twice.

This pushes the server into undefined territory, and can easily lead to a server-side crash.

The right solution seems to be, to prevent object store from adding the partial suffix to a path that already has a partial suffix.

From the point of view of the server, a path has a partial suffix if, and only if the path matches the regular expression, `/__partial/[^/]*$`.

In my opinion, and for extra safety, we should actually go one step further, and disallow the partial suffix on any application specified sync URL path, regardless of whether `partial: true` is specified or not. A consequence of this, of course, is that it will no longer be possible to bypass the partial sync API, but I think that is the desirable, as it also prevents accidental activation of partial sync mode.

cc @bdash 

For some context, see https://github.com/realm/ros/issues/529#issuecomment-336603233.

austinzheng:
Changes:
- SyncConfig's constructor only takes the two required arguments now; all others must be explicitly set
- SyncConfig now throws an exception if the URL passed in is invalid
- Fixed a broken test case

Fixes #579

### DIFF
--- a/tests/util/test_file.hpp
+++ b/tests/util/test_file.hpp
@@ -106,8 +106,8 @@ struct SyncTestFile : TestFile {
     template<typename BindHandler, typename ErrorHandler>
     SyncTestFile(const realm::SyncConfig& sync_config, 
         realm::SyncSessionStopPolicy stop_policy, 
-        BindHandler bind_handler, 
-        ErrorHandler error_handler)
+        BindHandler&& bind_handler, 
+        ErrorHandler&& error_handler)
     {
         this->sync_config = std::make_shared<realm::SyncConfig>(sync_config);
         this->sync_config->stop_policy = stop_policy;
--- a/tests/util/test_file.hpp
+++ b/tests/util/test_file.hpp
@@ -26,12 +26,15 @@
 #include <realm/util/optional.hpp>
 
 #if REALM_ENABLE_SYNC
+#include "sync/sync_config.hpp"
+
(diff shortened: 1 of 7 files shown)


tests/util/test_file.hpp
```diff
@@ -100,7 +103,19 @@ class SyncServer {
...
-    SyncTestFile(const realm::SyncConfig&);
+    template<typename BindHandler, typename ErrorHandler>
+    SyncTestFile(const realm::SyncConfig& sync_config, 
+        realm::SyncSessionStopPolicy stop_policy, 
+        BindHandler bind_handler, 
+        ErrorHandler error_handler)
+    {
+        this->sync_config = std::make_shared<realm::SyncConfig>(sync_config);
+        this->sync_config->stop_policy = stop_policy;
+        this->sync_config->bind_session_handler = std::forward<BindHandler>(bind_handler);
```
tgoyne: `std::forward` is only meaningful when the argument is declared as `BindHandler&&`

austinzheng:
`linux-sync` job ran on a machine that had no more disk space left; `win32` job is just outright broken.

PR #596, opened 2017-11-08
Don't ignore unrecognized sync errors

austinzheng:
The sync error handler was silently ignoring unrecognized errors. This change flags such errors as unrecognized and passes them to bindings, giving bindings the option to handle them further (e.g. present them to users).

### DIFF
--- a/src/sync/sync_config.hpp
+++ b/src/sync/sync_config.hpp
@@ -59,6 +59,10 @@ struct SyncError {
     std::string message;
     bool is_fatal;
     std::unordered_map<std::string, std::string> user_info;
+    /// The sync server may send down an error that the client does not recognize,
+    /// whether because of a version mismatch or an oversight. It is still valuable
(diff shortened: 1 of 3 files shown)


src/sync/sync_config.hpp
```diff
@@ -59,6 +59,10 @@ struct SyncError {
     std::string message;
     bool is_fatal;
     std::unordered_map<std::string, std::string> user_info;
+    /// The sync server may send down an error that the client does not recognize,
+    /// whether because of a version mismatch or an oversight. It is still valuable
+    /// to expose these errors so that users can do something about them.
+    bool is_unrecognized_by_client = false;
```
tgoyne: What is the intended use of this field? A different error message built by the binding?

austinzheng: Yes, making the unknown error-ness explicit to the binding so they can report it as such without checking for everything else first.

PR #597, opened 2017-11-08
Implement sync_session_state::Dying::bind_with_admin_token()

tgoyne:
This is needed to support reviving an admin token user's session, since this is called instead of revive_if_needed().

Fixes https://github.com/realm/realm-js-private/issues/375.

### DIFF
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -355,6 +355,14 @@ struct sync_session_states::Dying : public SyncSession::State {
         return false;
     }
 
+    void bind_with_admin_token(std::unique_lock<std::mutex>& lock, SyncSession& session,
+                               const std::string& admin_token,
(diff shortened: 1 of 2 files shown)


tgoyne:
I just assumed that it'd be something that isn't realistically testable with the tools we have available, but if there's already tests that cover similar things I'll take a look at them.

bdash:
The tests I'm thinking of are at https://github.com/realm/realm-object-store/blob/65279c531980f25e68a93aef7f2369c409571770/tests/sync/session/session.cpp#L402-L409 I've not thought too hard about if the same approach will work for testing your change.

tgoyne:
It looks like I just need to update those tests to try with both admin token and regular users.

tgoyne:
I've now added some tests which hit the new code.

PR #601, opened 2017-11-20
Test case to reproduce java issue 5507

beeender:
This test case shows part of the java issue
https://github.com/realm/realm-java/issues/5507

There are some other issues related with the java issue, but this one
can be isolated.

### DIFF
--- a/tests/list.cpp
+++ b/tests/list.cpp
@@ -472,6 +472,27 @@ TEST_CASE("list") {
             advance_and_notify(*r);
             REQUIRE_INDICES(change.deletions, 5);
         }
+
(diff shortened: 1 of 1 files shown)


beeender:
@tgoyne would you please help to take a look of this?

tgoyne:
Oof. The handling of notifiers which have all their callbacks removed and then re-added is wrong in a way that I thought didn't matter in practice, but I guess it only doesn't matter for Results and does matter for List.

I'll take a stab at getting a fix in for this before I leave for Thanksgiving, but it may end up waiting until next week.

PR #602, opened 2017-11-21
Differnt change set behavior for list and results

beeender:
If there is no actual changes in the list like remove all elements from
list and added them again in the same transaction, the Results built
from List can detect there is no change actually happened thus no
notifications will be triggered. But in the same case, notification will
be triggered for the List with a change set of deletions and insertions.

Not sure if this is by design.

### DIFF
--- a/tests/list.cpp
+++ b/tests/list.cpp
@@ -472,6 +472,49 @@ TEST_CASE("list") {
             advance_and_notify(*r);
             REQUIRE_INDICES(change.deletions, 5);
         }
+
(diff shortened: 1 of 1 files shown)


beeender:
@tgoyne

tgoyne:
The behavior is different because for Results it diffs the old and new result sets, while for List it just reports the changes from the transaction log. This means that List can report moves and is dramatically faster, but Results is better at only reporting things which resulted in actual changes.

beeender:
Good to know it is by design

PR #603, opened 2017-11-25
Improve the CMake scripts

emanuelez:
This PR leverages the work done on core and sync.


bmunkholm:
@jedelbo Please assess if anything from here is useful, or close the PR if not.

fealebenpae:
This will become obsolete with the monorepo. We can salvage things piecemeal as needed.

PR #604, opened 2017-11-27
 Don't reuse list notifiers which had all callbacks removed

tgoyne:
If a transaction is processed while a notifier has no callbacks, no work is done for that notifier and it gets into an inconsistent state. Adding a new callback to that notifier thus results in incorrect results from the first notification sent due to that the cached state is not properly reinitialized. Work around this for now by just not reusing list notifiers.

This problem does not apply to Results notifiers due to that the changeset in the Initial notification for those is not meaningful.

Fixes #601.

### DIFF
--- a/src/list.cpp
+++ b/src/list.cpp
@@ -439,6 +439,14 @@ bool List::operator==(List const& rgt) const noexcept
 NotificationToken List::add_notification_callback(CollectionChangeCallback cb) &
 {
     verify_attached();
+    // Adding a new callback to a notifier which had all of its callbacks
+    // removed does not properly reinitialize the notifier. Work around this by
+    // recreating it instead.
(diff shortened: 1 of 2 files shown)


src/list.cpp
```diff
@@ -439,6 +439,14 @@ bool List::operator==(List const& rgt) const noexcept
...
 {
     verify_attached();
+    // Adding a new callback to a notifier which had all of its callbacks
+    // removed does not properly reinitialize the notifier. Work around this by
+    // recreating it instead.
+    // FIXME: The notifier lifecycle here is dumb (when all callbacks are removed
+    // from a notifier a zombie is left sitting around uselessly) and should be
+    // cleaned up.
+    if (m_notifier && !m_notifier->have_callbacks())
+        m_notifier.reset();
```
bdash: This is a great, pragmatic solution for this bug!

PR #608, opened 2017-12-04
Cocoa package

fealebenpae:


bdash:
How would one produce such a package locally? This, or something like it, seems necessary in order to test object store changes against the Cocoa binding.

fealebenpae:
I was hoping to avoid that, but that’s a good reason to move the logic in a shell script.

bdash:
It really does need to be possible to build the entire stack (core / sync / object store) and test it in a binding, without having to create a release package via CI. Ever since core switched to CMake that has been challenging, and it's made it significantly harder to debug issues that surface in the Cocoa binding. If that difficulty spreads to debugging issues in the object store, we'll be in a bad place.

fealebenpae:
That will be possible in a super-project - create a `CMakeLists.txt` that includes the repos for core, sync, and object-store as subdirectories and everything will build from the local checkouts. I think @simonask has a repository like that set up, or was meaning to set up one.

bdash:
Will the build artifacts that are produced via that approach be structured in the same way as the package produced by this PR? If not, how would the Cocoa binding consume them?

fealebenpae:
@emanuelez has taken an alternative approach where building Core, Sync, or Object Store for any of the Cocoa platforms will append platform suffixes to the build products, which renders this PR's approach of manually renaming files obsolete.

@bdash with the latest code in #603 one can create a super-project including the three repositories, configure CMake with the toolchain file for the desired platform, build, and install. The installation prefix will match the contents of the aggregate archive produced by CI. It should be possible to further automate the process in https://github.com/realm/integration with a shell script, though I'm not sure what the Cocoa team's workflow right now is and if you would prefer this level of automation.

PR #609, opened 2017-12-04
Exclude `__perm` internal Realm from permission results

austinzheng:
Necessary to upgrade Cocoa dependency to 2.1.0 without adding test workarounds. (This third implementation detail Realm was added at some point between 2.0.13 and 2.1.0, and needs to be excluded in order to hide extraneous results.)


tgoyne:
__perm and __permissions contain the same data in different formats, and __permissions will go away once everything has stopped using it.

PR #610, opened 2017-12-08
Extend notification system to support partially synced results

cmelchior:
Very much WIP (does not compile):

General concept: Extend the `CollectionChangeSet` with information about Partial Sync status and errors and expose them in the general notification system.

 TODO
- [x] Expose Partial Sync error message and state as part of CollectionChange events.
- [ ] Finish refactor moving all partial sync methods to `partial_sync.cpp`
- [ ] Add support for explicitly named results.
- [ ] Add support for updating the schema with *_matches columns for new queries
- [ ] Consider adding support for Realm in collection_notifier.cpp instead of m_sg as it otherwise requires a lot of low-level logic that have unclear rules.
- [ ] Consider adding a status enum instead of int8.
- [x] Find where "old" state is read. We need to read status flag here.
- [ ] Trigger notifications when only partial sync error message is triggered
- [ ] Question: Is it safe to do a write transaction in the notifier?
- [ ] Unit tess

Open questions:
- Tracking initial callback is currently done in bindings, can this move to object store?

### DIFF
--- a/src/collection_notifications.hpp
+++ b/src/collection_notifications.hpp
@@ -88,10 +88,24 @@ struct CollectionChangeSet {
     // Per-column version of `modifications`
     std::vector<IndexSet> columns;
 
+    // Any potential error message from partial sync for the new collection
+    // This is only set if partial_sync_new_status_code is -1
+    std::string partial_sync_error_message;
+
+    /// Partial sync status codes:
+    // int8_t::min() -> Undefined
+    // -2            -> Not a partial Realm
+    // -1            -> Error occurred while processing the subscription.
+    // 0             -> Subscription created locally, but not processed by the server yet.
+    // 1             -> Subscription successfully processed by the server.
+    int8_t partial_sync_old_status_code;
+    int8_t partial_sync_new_status_code;
+
     bool empty() const noexcept
     {
         return deletions.empty() && insertions.empty() && modifications.empty()
-            && modifications_new.empty() && moves.empty();
+            && modifications_new.empty() && moves.empty()
+            && partial_sync_old_status_code == partial_sync_new_status_code;
     }
 };
 
--- a/src/collection_notifications.hpp
+++ b/src/collection_notifications.hpp
@@ -88,10 +88,6 @@ struct CollectionChangeSet {
     // Per-column version of `modifications`
     std::vector<IndexSet> columns;
 
-    // Any potential error message from partial sync for the new collection
-    // This is only set if partial_sync_new_status_code is -1
-    std::string partial_sync_error_message;
-
     /// Partial sync status codes:
     // int8_t::min() -> Undefined
     // -2            -> Not a partial Realm
@@ -100,6 +96,10 @@ struct CollectionChangeSet {
     // 1             -> Subscription successfully processed by the server.
     int8_t partial_sync_old_status_code;
     int8_t partial_sync_new_status_code;
+
+    // Any potential error message from partial sync for the new collection
+    // This is only set if partial_sync_new_status_code is -1
+    std::string partial_sync_error_message;
 
     bool empty() const noexcept
     {
--- a/src/collection_notifications.hpp
+++ b/src/collection_notifications.hpp
@@ -101,6 +101,8 @@ struct CollectionChangeSet {
     // This is only set if partial_sync_new_status_code is -1
     std::string partial_sync_error_message;
 
+    // Returns `true` iff this change set is "empty", i.e. no changes of interest
+    // can be reported.
     bool empty() const noexcept
     {
         return deletions.empty() && insertions.empty() && modifications.empty()
--- a/src/impl/collection_change_builder.cpp
+++ b/src/impl/collection_change_builder.cpp
@@ -504,6 +504,21 @@ void CollectionChangeBuilder::move_column(size_t from, size_t to)
         std::rotate(begin(columns) + to, begin(columns) + from, begin(columns) + from + 1);
 }
 
+void CollectionChangeBuilder::old_partial_sync_status_code(int8_t code)
+{
+    this->partial_sync_old_status_code = code;
+}
+
+void CollectionChangeBuilder::new_partial_sync_status_code(int8_t code)
+{
+    this->partial_sync_new_status_code = code;
+}
+
+void CollectionChangeBuilder::new_partial_sync_error_message(std::string error)
+{
+    this->partial_sync_error_message = error;
+}
+
 namespace {
 struct RowInfo {
     size_t row_index;
--- a/src/impl/collection_change_builder.cpp
+++ b/src/impl/collection_change_builder.cpp
@@ -58,6 +58,11 @@ void CollectionChangeBuilder::merge(CollectionChangeBuilder&& c)
         return;
     }
 
+    // Always override partial sync codes/status
+    m_partial_sync_old_status_code = c.partial_sync_old_status_code;
+    m_partial_sync_new_status_code = c.partial_sync_new_status_code;
+    m_partial_sync_error_message = c.partial_sync_error_message;
+
     verify();
     c.verify();
 
@@ -517,17 +522,17 @@ void CollectionChangeBuilder::move_column(size_t from, size_t to)
 
 void CollectionChangeBuilder::old_partial_sync_status_code(int8_t code)
 {
-    this->partial_sync_old_status_code = code;
+    m_partial_sync_old_status_code = code;
 }
 
 void CollectionChangeBuilder::new_partial_sync_status_code(int8_t code)
 {
-    this->partial_sync_new_status_code = code;
+    m_partial_sync_new_status_code = code;
 }
 
-void CollectionChangeBuilder::new_partial_sync_error_message(std::string error)
+void CollectionChangeBuilder::partial_sync_error_message(std::string error)
 {
-    this->partial_sync_error_message = error;
+    m_partial_sync_error_message = error;
 }
 
 namespace {
@@ -909,8 +914,8 @@ CollectionChangeSet CollectionChangeBuilder::finalize() &&
         std::move(modifications),
         std::move(moves),
         std::move(columns),
-        std::move(partial_sync_old_status_code),
-        std::move(partial_sync_new_status_code),
-        std::move(partial_sync_error_message)
+        std::move(m_partial_sync_old_status_code),
+        std::move(m_partial_sync_new_status_code),
+        std::move(m_partial_sync_error_message)
     };
 }
--- a/src/impl/collection_change_builder.cpp
+++ b/src/impl/collection_change_builder.cpp
@@ -29,8 +29,19 @@ using namespace realm::_impl;
 CollectionChangeBuilder::CollectionChangeBuilder(IndexSet deletions,
                                                  IndexSet insertions,
                                                  IndexSet modifications,
-                                                 std::vector<Move> moves)
-: CollectionChangeSet({std::move(deletions), std::move(insertions), std::move(modifications), {}, std::move(moves)})
+                                                 std::vector<Move> moves,
+                                                 int8_t partial_sync_old_status_code,
+                                                 int8_t partial_sync_new_status_code,
+                                                 std::string partial_sync_error_message)
+: CollectionChangeSet({std::move(deletions),
+                       std::move(insertions),
+                       std::move(modifications),
+                       {},
+                       std::move(moves),
+                       {},
+                       partial_sync_old_status_code,
+                       partial_sync_new_status_code,
+                       partial_sync_error_message})
 {
     for (auto&& move : this->moves) {
         this->deletions.add(move.from);
@@ -897,6 +908,9 @@ CollectionChangeSet CollectionChangeBuilder::finalize() &&
         std::move(modifications_in_old),
         std::move(modifications),
         std::move(moves),
-        std::move(columns)
+        std::move(columns),
+        std::move(partial_sync_old_status_code),
+        std::move(partial_sync_new_status_code),
+        std::move(partial_sync_error_message)
     };
 }
--- a/src/impl/collection_change_builder.hpp
+++ b/src/impl/collection_change_builder.hpp
@@ -40,7 +40,7 @@ public:
                             std::vector<Move> moves = {},
                             int8_t partial_sync_old_status_code = -2,
                             int8_t partial_sync_new_status_code = -2,
-                            std::string partial_sync_error_message = nullptr);
+                            std::string partial_sync_error_message = "");
 
     // Calculate where rows need to be inserted or deleted from old_rows to turn
     // it into new_rows, and check all matching rows for modifications
@@ -54,6 +54,8 @@ public:
 
     // generic operations {
     CollectionChangeSet finalize() &&;
+
+    // Merge input change set into this one.
     void merge(CollectionChangeBuilder&&);
 
     void insert(size_t ndx, size_t count=1, bool track_moves=true);
@@ -80,13 +82,17 @@ public:
     void insert_column(size_t ndx);
     void move_column(size_t from, size_t to);
 
+    // partial sync
     void old_partial_sync_status_code(int8_t code);
     void new_partial_sync_status_code(int8_t code);
-    void new_partial_sync_error_message(std::string error);
+    void partial_sync_error_message(std::string error);
 
 private:
     std::unordered_map<size_t, size_t> m_move_mapping;
     bool m_track_columns = true;
+    int8_t m_partial_sync_old_status_code;
+    int8_t m_partial_sync_new_status_code;
+    std::string m_partial_sync_error_message;
 
     template<typename Func>
     void for_each_col(Func&& f);
--- a/src/impl/collection_change_builder.hpp
+++ b/src/impl/collection_change_builder.hpp
@@ -37,7 +37,10 @@ public:
     CollectionChangeBuilder(IndexSet deletions = {},
                             IndexSet insertions = {},
                             IndexSet modification = {},
-                            std::vector<Move> moves = {});
+                            std::vector<Move> moves = {},
+                            int8_t partial_sync_old_status_code = -2,
+                            int8_t partial_sync_new_status_code = -2,
+                            std::string partial_sync_error_message = nullptr);
 
     // Calculate where rows need to be inserted or deleted from old_rows to turn
     // it into new_rows, and check all matching rows for modifications
--- a/src/impl/collection_change_builder.hpp
+++ b/src/impl/collection_change_builder.hpp
@@ -77,6 +77,10 @@ public:
     void insert_column(size_t ndx);
     void move_column(size_t from, size_t to);
 
+    void old_partial_sync_status_code(int8_t code);
+    void new_partial_sync_status_code(int8_t code);
+    void new_partial_sync_error_message(std::string error);
+
 private:
     std::unordered_map<size_t, size_t> m_move_mapping;
     bool m_track_columns = true;
--- a/src/impl/collection_notifier.hpp
+++ b/src/impl/collection_notifier.hpp
@@ -194,8 +194,9 @@ protected:
     void set_table(Table const& table);
     std::unique_lock<std::mutex> lock_target();
     SharedGroup& source_shared_group();
-
     std::function<bool (size_t)> get_modification_checker(TransactionChangeInfo const&, Table const&);
+    bool m_partial_sync_enabled = true;
+    std::shared_ptr<Realm> m_realm;
 
 private:
     virtual void do_attach_to(SharedGroup&) = 0;
@@ -205,12 +206,10 @@ private:
     virtual bool prepare_to_deliver() { return true; }
 
     mutable std::mutex m_realm_mutex;
-    std::shared_ptr<Realm> m_realm;
 
     VersionID m_sg_version;
     SharedGroup* m_sg = nullptr;
 
-    bool m_partial_sync_enabled = true;
     bool m_has_run = false;
     bool m_error = false;
     std::vector<DeepChangeChecker::RelatedTable> m_related_tables;
@@ -222,8 +221,6 @@ private:
         uint64_t token;
         bool initial_delivered;
         bool skip_next;
-        int partial_sync_status_code;   // -2 means partial sync is not enabled, -1 error has happened, 0 not processed, 1 processed
-        int partial_sync_error_message; // error message, only relevant if status_code is -1
     };
 
     // Currently registered callbacks and a mutex which must always be held
--- a/src/impl/collection_notifier.hpp
+++ b/src/impl/collection_notifier.hpp
@@ -164,6 +164,8 @@ public:
     bool is_alive() const noexcept;
 
     // precondition: RealmCoordinator::m_notifier_mutex is locked *or* is called on worker thread
+    // Returns `true` if this is not the first time the notification is running, `false` if this
+    // is the initial notification.
     bool has_run() const noexcept { return m_has_run; }
 
     // Attach the handed-over query to `sg`. Must not be already attached to a SharedGroup.
@@ -197,6 +199,7 @@ protected:
     std::function<bool (size_t)> get_modification_checker(TransactionChangeInfo const&, Table const&);
     bool m_partial_sync_enabled = true;
     std::shared_ptr<Realm> m_realm;
+    SharedGroup* m_sg = nullptr;
 
 private:
     virtual void do_attach_to(SharedGroup&) = 0;
@@ -208,7 +211,6 @@ private:
     mutable std::mutex m_realm_mutex;
 
     VersionID m_sg_version;
-    SharedGroup* m_sg = nullptr;
 
     bool m_has_run = false;
     bool m_error = false;
--- a/src/impl/collection_notifier.hpp
+++ b/src/impl/collection_notifier.hpp
@@ -210,6 +210,7 @@ private:
     VersionID m_sg_version;
     SharedGroup* m_sg = nullptr;
 
+    bool m_partial_sync_enabled = true;
     bool m_has_run = false;
     bool m_error = false;
     std::vector<DeepChangeChecker::RelatedTable> m_related_tables;
@@ -221,6 +222,8 @@ private:
         uint64_t token;
         bool initial_delivered;
         bool skip_next;
+        int partial_sync_status_code;   // -2 means partial sync is not enabled, -1 error has happened, 0 not processed, 1 processed
+        int partial_sync_error_message; // error message, only relevant if status_code is -1
     };
 
     // Currently registered callbacks and a mutex which must always be held
@@ -239,7 +242,7 @@ private:
     size_t m_callback_index = -1;
     // The number of callbacks which were present when the notifier was packaged
     // for delivery which are still present.
-    // Updated by packaged_for_delivery and removd_callback(), and used in
+    // Updated by packaged_for_delivery and  removd_callback(), and used in
     // for_each_callback() to avoid calling callbacks registered during delivery.
     size_t m_callback_count = -1;
 
--- a/src/impl/realm_coordinator.hpp
+++ b/src/impl/realm_coordinator.hpp
@@ -178,6 +178,8 @@ private:
     void set_config(const Realm::Config&);
     void create_sync_session();
 
+    // Run CollectionNotifiers on the advanced Realm.
+    // This will trigger all registered collection callbacks.
     void run_async_notifiers();
     void open_helper_shared_group();
     void advance_helper_shared_group_to_latest();
--- a/src/impl/results_notifier.cpp
+++ b/src/impl/results_notifier.cpp
@@ -19,6 +19,7 @@
 #include "impl/results_notifier.hpp"
 
 #include "shared_realm.hpp"
+#include <sstream>
 
 using namespace realm;
 using namespace realm::_impl;
@@ -172,7 +173,69 @@ void ResultsNotifier::run()
     if (!need_to_run())
         return;
 
-    m_query->sync_view_if_needed();
+
+    // Potential optimization: Ship the write to another background thread to prevent
+    // blocking the notifiers?
+
+    // TODO Partial Sync is only supported on non-looper threads, but since it shouldn't be possible
+    // to register listeners on threads without run loops we should only enter this code if
+    // we actually support notifications and thus partial sync subscriptions.
+
+    // While this would trigger notifications it should be safe as we are on the background worker
+    // thread where no notifications exist.
+    int8_t partial_sync_status_code = (m_partial_sync_enabled) ? 0 : -2;
+    std::string partial_sync_error_message = nullptr;
+    if (m_partial_sync_enabled) {
+        bool cancel = false;
+        m_realm->begin_transaction();
+
+        // TODO: Determine how to create key. Right now we just used the serialized query (are there any
+        // disadvantages to that?). Would it make sense to hash it (to keep it short), but then we need
+        // to handle hash collisions as well.
+        std::string key = m_query.get_description();
+
+        // TODO: It might change how the query is serialized. See https://realmio.slack.com/archives/C80PLGQ8Z/p1511797410000188
+        std::string serialized_query = m_query.get_description();
+
+        // Check current state and create subscription if needed. Throw in an error is found:
+        TableRef table = ObjectStore::table_for_object_type(m_realm->read_group(), "__ResultSet");
+        size_t name_idx = table->get_descriptor()->get_column_index("name");
+        size_t query_idx = table->get_descriptor()->get_column_index("query");
+        size_t status_idx = table->get_descriptor()->get_column_index("status");
+        size_t error_idx = table->get_descriptor()->get_column_index("error_message");
+        auto results = Results(realm, *table).filter(table->column<StringData>(col_idx).equal(key));
+        if (results.size() > 0) {
+            // Subscription with that ID already exist. Verify that we are not trying to reuse an
+            // existing name for a different query.
+            REALM_ASSERT(results.size() == 1);
+            auto obj = results.get(0);
+            if (obj.get_string(query_idx) != serialized_query)) {
+                std::stringstream ss;
+                ss << "Subscription cannot be created as another subscription already exists with the same name. ";
+                ss << "Name: " << key << ". ";
+                ss << "Existing query: " << obj.get_string(query_idx) << ". ";
+                ss << "New query: " << serialized_query << ".";
+                m_realm->cancel_transaction();
+                throw new std::logic_error(ss.str());
+            };
+
+            partial_sync_status_code = (int8_t) obj->get_int(status_idx);
+            partial_sync_error_message = obj->get_string(error_idx);
+
+        } else {
+            auto props = AnyDict{
+           O     {"name", key},
+                {"query", serialized_query}
+            };
+            Object::create<util::Any>(context, m_realm, m_realm->schema().find("__ResultSet"), std::move(props), false);
+            partial_sync_status_code = 0;
+            partial_sync_error_message = "";
+        }
+
+        m_realm->commit_transaction();
+    }
+
+    // m_query->sync_view_if_needed(); // No longer needed, begin_transaction will do this
     m_tv = m_query->find_all();
 #if REALM_HAVE_COMPOSABLE_DISTINCT
     m_tv.apply_descriptor_ordering(m_descriptor_ordering);
@@ -186,6 +249,11 @@ void ResultsNotifier::run()
     m_last_seen_version = m_tv.sync_if_needed();
 
     calculate_changes();
+
+    // Set partial sync properties
+    // TODO SHould we do this inside calculate_changes()?
+    m_changes->new_partial_sync_status_code = partial_sync_status_code;
+    m_changes->new_partial_sync_error_message = partial_sync_error_message;
 }
 
 void ResultsNotifier::do_prepare_handover(SharedGroup& sg)
--- a/src/impl/results_notifier.cpp
+++ b/src/impl/results_notifier.cpp
@@ -21,10 +21,13 @@
 #include "object_store.hpp"
 #include "object_schema.hpp"
 #include "shared_realm.hpp"
-#include <realm/query_expression.hpp>
 #include "impl/object_accessor_impl.hpp"
 
 #include <sstream>
+
+#if REALM_ENABLE_SYNC
+#include "sync/partial_sync.hpp"
+#endif
 
 using namespace realm;
 using namespace realm::_impl;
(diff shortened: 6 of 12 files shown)


src/collection_notifications.hpp
```diff
@@ -88,10 +88,26 @@ struct CollectionChangeSet {
     // Per-column version of `modifications`
     std::vector<IndexSet> columns;
 
+    /// Partial sync status codes:
+    // int8_t::min() -> Undefined
```
tgoyne: This should be an enum.

src/impl/collection_change_builder.hpp
```diff
@@ -77,9 +82,17 @@ class CollectionChangeBuilder : public CollectionChangeSet {
...
 
+    // partial sync
+    void old_partial_sync_status_code(int8_t code);
+    void new_partial_sync_status_code(int8_t code);
+    void partial_sync_error_message(std::string error);
+
 private:
     std::unordered_map<size_t, size_t> m_move_mapping;
     bool m_track_columns = true;
+    int8_t m_partial_sync_old_status_code;
```
tgoyne: Why are these duplicated from the base class?

src/impl/collection_notifier.hpp
```diff
@@ -194,8 +196,10 @@ class CollectionNotifier {
     void set_table(Table const& table);
     std::unique_lock<std::mutex> lock_target();
     SharedGroup& source_shared_group();
-
     std::function<bool (size_t)> get_modification_checker(TransactionChangeInfo const&, Table const&);
+    bool m_partial_sync_enabled = true;
+    std::shared_ptr<Realm> m_realm;
```
tgoyne: Moving this declaration here is not valid (it needs appear after the mutex guarding it, so that it is destroyed before that mutex).

src/impl/realm_coordinator.hpp
```diff
@@ -178,6 +178,8 @@ class RealmCoordinator : public std::enable_shared_from_this<RealmCoordinator> {
     void set_config(const Realm::Config&);
     void create_sync_session();
 
+    // Run CollectionNotifiers on the advanced Realm.
+    // This will trigger all registered collection callbacks.
```
tgoyne: This function does not trigger callbacks.

src/impl/results_notifier.cpp
```diff
@@ -149,20 +158,31 @@ void ResultsNotifier::calculate_changes()
         m_changes = CollectionChangeBuilder::calculate(m_previous_rows, next_rows,
                                                        get_modification_checker(*m_info, *m_query->get_table()),
                                                        move_candidates);
-
         m_previous_rows = std::move(next_rows);
+
+#if REALM_ENABLE_SYNC
+        if (m_partial_sync_enabled)
+            m_previous_partial_sync_status_code = partial_sync::create_or_update_subscription(*m_sg, m_changes, *m_query, m_previous_partial_sync_status_code);
```
tgoyne: ResultsNotifier should not be managing partial sync subscriptions. That's pretty unrelated to calculating changesets for Results.

src/impl/results_notifier.cpp
```diff
@@ -149,20 +158,31 @@ void ResultsNotifier::calculate_changes()
...
+#endif
     }
 }
 
 void ResultsNotifier::run()
 {
     // Table's been deleted, so report all rows as deleted
     if (!m_query->get_table()->is_attached()) {
+        // FIXME: How to handle partial sync here? Question: Why is this before need_to_run()?
```
tgoyne: This is at the beginning so that all of the following code can assume that the Table still exists.

src/impl/results_notifier.cpp
```diff
@@ -207,7 +227,7 @@ void ResultsNotifier::do_prepare_handover(SharedGroup& sg)
     m_tv_handover = sg.export_for_handover(m_tv, MutableSourcePayload::Move);
 
     add_changes(std::move(m_changes));
-    REALM_ASSERT(m_changes.empty());
+    // REALM_ASSERT(m_changes.empty()); // FIXME What exactly is this assert checking?
```
tgoyne: It is asserting that add_changes() actually consumed m_changes.

bdash:
Why is this tied into collection notifications at all? I thought we'd agreed that subscribing to a partial sync query was something that would be done explicitly.

cmelchior:
We did, but with the refocus of partial sync described in  https://docs.google.com/document/d/1NuxWxBl3VNtgZDYeXZaWXCSet4LrlWQDXQeI-xFs3xs/edit both Alexander and Adam felt strongly that having two ways of doing essentially the same thing was not desirable. The end result of that discussion is described at the top of https://docs.google.com/document/d/1zmIROxKKzzSy6zacIwwDr40r-47ixL3qRjQ0w6hoCB4/edit

The reason this PR integrates Partial Sync into the notification system was that there was also concern about the UI could respond to errors happening after registering a subscription. A partial sync only error handler would have no local knowledge about the UI.

So it was decided that we should try to adopt the notification style that Swift currently uses as a way to propagate errors using the `.error` state. This has 3 big advantages: 1) It requires relatively few changes in bindings 2) It makes it possible to tackle any errors where you are normally updating the UI anyway and 3) It makes the bindings more similar in how the notifications are used.

Note, that extending the notification system to include Partial Sync states, does not directly have an impact on how they are actually registered. The current PR just reflects the easiest way to actually test it.

cmelchior:
While still some issues are outstanding, this should be ready for review @tgoyne @bdash

src/impl/collection_change_builder.cpp
```diff
@@ -504,6 +526,21 @@ void CollectionChangeBuilder::move_column(size_t from, size_t to)
         std::rotate(begin(columns) + to, begin(columns) + from, begin(columns) + from + 1);
 }
 
+void CollectionChangeBuilder::set_old_partial_sync_state(partial_sync::SubscriptionState state)
```
tgoyne: Why do these exist?

src/impl/collection_change_builder.hpp
```diff
@@ -51,6 +55,9 @@ class CollectionChangeBuilder : public CollectionChangeSet {
 
     // generic operations {
     CollectionChangeSet finalize() &&;
+
+    // Merge input change set into this one. Once merged the input builder should
+    // no longer be used.
```
tgoyne: This is not true. It is perfectly fine to reuse the input builder, and we do so.

src/sync/partial_sync.cpp
```diff
@@ -16,20 +16,109 @@
...
 namespace realm {
 namespace partial_sync {
 
+SubscriptionState create_or_update_subscription(Realm::Config config, SharedGroup &sg, realm::_impl::CollectionChangeBuilder &changes, Query &query, SubscriptionState previous_state) {
+
+    if (!config.sync_config || !config.sync_config->is_partial) {
+        return SubscriptionState::NOT_SUPPORTED;
+    }
+
+    // FIXME: Question: Is it problematic to do a write transaction here? Should we move it to a background thread`?
```
tgoyne: Yes, it is very problematic. This design does not work at all outside of the most trivial cases. The RealmCoordinator owns the SharedGroup and very carefully manages what version it is at. Changing that version will break everything.

src/sync/partial_sync.hpp
```diff
@@ -19,6 +19,14 @@
 #ifndef REALM_OS_PARTIAL_SYNC_HPP
 #define REALM_OS_PARTIAL_SYNC_HPP
 
+#include "shared_realm.hpp"
+#include "subscription_state.hpp"
+#include "impl/collection_change_builder.hpp"
+#include <realm/group_shared.hpp>
+#include <realm/lang_bind_helper.hpp>
+#include <realm/query.hpp>
+#include <realm/query_expression.hpp>
```
tgoyne: Most of these includes are not needed here.

src/subscription_state.hpp
```diff
@@ -0,0 +1,59 @@
...
+// These states are propagated using the standard collection notification system.
+enum class SubscriptionState {
+    UNDEFINED,       // Unknown which state Partial Sync is in.
+    NOT_SUPPORTED,   // Partial Sync not supported.
+    ERROR,           // An error was detect in Partial Sync.
+    UNINITIALIZED,   // The subscription was just created, but not handled by sync yet.
+    INITIALIZED      // The subscription have been initialized successfully and are syncing data to the device.
+};
+
+static inline SubscriptionState status_code_to_state(int status_code) {
```
tgoyne: You can just set the values of the enum states and then perform a cast between the enum and int.

src/sync/partial_sync.cpp
```diff
@@ -16,20 +16,109 @@
...
+    std::string key = "SELECT * FROM XXX"; // Awaiting a release with query.get_description();
+
+    // TODO: It might change how the query is serialized. See https://realmio.slack.com/archives/C80PLGQ8Z/p1511797410000188
+    std::string serialized_query = "SELECT * FROM XXX"; // Awaiting a release with query.get_description();
+
+    LangBindHelper::promote_to_write(sg);
+    Group& group = _impl::SharedGroupFriend::get_group(sg);
+    // Check current state and create subscription if needed. Throw in an error is found:
+    TableRef table = ObjectStore::table_for_object_type(group, "__ResultSets");
+    size_t name_idx = table->get_descriptor()->get_column_index("name");
```
tgoyne: Table::get_column_index() exists

src/object_store.cpp
```diff
@@ -75,6 +76,23 @@ void create_metadata_tables(Group& group) {
...
+#if REALM_ENABLE_SYNC
+    // Add backing support for partial sync in all synced Realm, even if they don't support
+    // Partial Sync right now.
+    if (!group.has_table(result_sets_type_name)) {
+        TableRef resultsets_table = sync::create_table(group, result_sets_type_name);
+        size_t indexable_column_idx = resultsets_table->add_column(type_String, "name");
+        resultsets_table->add_search_index(indexable_column_idx);
+        resultsets_table->add_column(type_String, "matches_property");
+        indexable_column_idx = resultsets_table->add_column(type_String, "query");
+        resultsets_table->add_search_index(indexable_column_idx);
```
tgoyne: This index does not appear to be used anywhere.

src/sync/partial_sync.cpp
```diff
@@ -16,20 +16,109 @@
...
+    size_t error_idx = table->get_descriptor()->get_column_index("error_message");
+    size_t matches_property_idx = table->get_descriptor()->get_column_index("matches_property");
+    TableView results = (key == table->column<StringData>(name_idx)).find_all(0);
+    if (results.size() > 0) {
+        // Subscription with that ID already exist. Verify that we are not trying to reuse an
+        // existing name for a different query.
+        REALM_ASSERT(results.size() == 1);
+        auto obj = results.get(0);
+        if (obj.get_string(query_idx) != serialized_query) {
+            std::stringstream ss;
```
tgoyne: Use util::format() rather than stringstream (it is both less verbose and generates significantly smaller compiled code).

src/impl/results_notifier.cpp
```diff
@@ -18,7 +18,17 @@
...
+#include "object_schema.hpp"
 #include "shared_realm.hpp"
+#include "subscription_state.hpp"
+#include "impl/object_accessor_impl.hpp"
+#if REALM_ENABLE_SYNC
+#include "sync/partial_sync.hpp"
+#include "sync/sync_session.hpp"
+#endif
+
+#include <sstream>
```
tgoyne: Most of these added includes are not used.

src/sync/partial_sync.cpp
cmelchior: Browsing through the code, it appeared the SharedGroup wasn't used anymore after this step in `calculate_changes`, but I could, of course, have missed something?

It should be possible to move this write to another thread, and then optimistically assume it was successful. It will introduce a race condition where you might get an initial callback followed by an error, but since partial sync errors are asynchronous anyway, that is probably acceptable. 

In Java, I would use something like a thread pool to prevent spawning threads all over the place, but I don't know if we have something similar here that can be used?

tgoyne: A single SharedGroup instance is used for every run of all of the notifiers for a given realm path.

I don't understand why the subscription is being created in ResultsNotifier::calculate_changes() in the first place. It makes sense for the subscription status to be read there, but creating it is entirely unrelated to anything that ResultsNotifier does. I would expect the subscription to be created when the notification is registered on the collection.

src/impl/collection_change_builder.cpp
cmelchior: I got some compiler errors when doing `old_partial_sync_state = c.old_partial_sync_state` in the `merge()` method. I was not able to resolve them any other way, but happy to change it if there is an easier way?

src/impl/collection_change_builder.hpp
cmelchior: Ups. I added this after running into the REALM_ASSET in `collection_notifier.add_changes()`, which does appear to consume the builder, but not sure why I added it here which is clearly a mistake. I'll move this comment.

src/impl/collection_change_builder.cpp
tgoyne: What were the errors?

src/impl/collection_change_builder.hpp
cmelchior: Now I remember why. It was because `add_changes` internally calls `merge` which was why I assumed this invariant also applied to `merge`, but I guess that is not the case?

src/impl/collection_change_builder.cpp
cmelchior: Something about not being able to convert between method calls and expressions, but trying to provoke the error again, it actually compiled fine.

During development, I had issues with a faulty CCache, so perhaps these errors came from there as well. 

It appears to work now by just setting them directly, so I'll remove these methods.

src/sync/partial_sync.cpp
cmelchior: Yes, that would be a more natural place to create it, the problem is that any errors should still come through the notification system, but perhaps I can record them somewhere and just throw an exception as soon as the notification trigger similar to what is being done if something is wrong when the Relm is being opened.

I'll look into it.

cmelchior: Actually, now I remember why. Originally I discarded the idea because it would imply writing to the Realm registering the callback which would advance it (which would be bad). But maybe we can work around that by opening a specific shared group. 

The downside of that approach is that it means that every time an observer is registered you would have to open a shared group and close it again, but that might be cheap enough?

cmelchior: So, talking to @finnschiermer, SharedGroups are pretty expensive and also have an impact on file descriptors, so if we have an extra SharedGroup for each thread that can register an observer it will be quite some overhead

I see two options here really (might be more):

1) Keep the write in the notifier thread and add an extra SharedGroup for writing only (cleaner code/abstraction)
2) Maintain an extra SharedGroup pr. thread that uses callbacks  in order to not accidentially advance the Realm when registering the callback

What are your thoughts @tgoyne ?

tgoyne: An entirely untested and unpolished example of the sort of design I have in mind: https://github.com/realm/realm-object-store/commit/12dc158be5efbc7e499f36f564fec1cb314e657b

tgoyne: In case it is not clear, I think trying to make writes of any kind from a notifier (no matter how indirectly) is a very very bad idea. The notifier lifecycle and functionality is already far too complicated even with them being purely read-only, and extending them to also make writes to the Realm will make them even more nightmarish.

cmelchior: Thanks for that commit 👍 
I did consider a similar approach but discarded it mostly because it introduces some rather annoying race conditions regarding error handling. 

Most noticeably that you optimistically have to assume that creating the subscription will succeed thus you risk having situations where the first callback returns a query and the next callback returns an error saying the subscription could not be created.

However, since Partial Sync query validation is async as well, checking for name collision that way might be acceptable. Also, we can still do a synchronous read to check before even notifying the partial sync worker thread, which would drastically limit the number of times we would have to use async error handling.

And I agree that the approach you outlined has a much cleaner architecture, so I don't see a problem going with that for a first implementation.

src/impl/realm_coordinator.cpp
```diff
@@ -465,6 +470,67 @@ void RealmCoordinator::pin_version(VersionID versionid)
...
+        throw std::logic_error("A partial sync query can only be registered in a partially synced Realm");
+
+#if REALM_ENABLE_SYNC
+    std::string object_class = query.get_table()->get_name().substr(6);
+
+    if (realm.schema().find(object_class) == realm.schema().end())
+        throw std::logic_error("A partial sync query can only be registered for a type that exists in the Realm's schema");
+
+    std::string key = subscription_name;
+    std::string serialized_query = "FIXME"; // query.get_description();
```
ironage: The `query.get_description()` method exists since core 3.2.0 so you could already use it here, just that the text it will output is not conforming to the parser's standard.

cmelchior: I'm not sure that has ever been released with Sync though, at least it didn't seem so last I tried. But I'll look into it.

src/subscription_state.hpp
```diff
@@ -0,0 +1,39 @@
...
+namespace partial_sync {
+
+// Enum describing the various states a partial sync subscription can have.
+// These states are propagated using the standard collection notification system.
+enum class SubscriptionState : int8_t {
+    Undefined = -3,         // Unknown which state Partial Sync is in.
+    NotSupported = - 2,     // Partial Sync not supported.
+    Error = -1,             // An error was detect in Partial Sync.
+    Uninitialized = 0,      // The subscription was just created, but not handled by sync yet.
+    Initialized = 1         // The subscription have been initialized successfully and are syncing data to the device.
```
kneth: "have" -> "has"

cmelchior:
Ready for review @tgoyne 

I had quite a few issues with how to approach error handling like conflicting subscription names, since these error messages could not really be persisted in the Realm like query errors, so  I introduced a local_error_message property, which the notification system should be able to pick up.

src/impl/realm_coordinator.cpp
```diff
@@ -462,6 +467,64 @@ void RealmCoordinator::pin_version(VersionID versionid)
...
+                                                  &partial_sync_queue = self.m_partial_sync_queue]() {
+            // Do not attempt to create a SharedRealm here or capture a RealmCoordinator. For some
+            // reason it either crashes or deadlocks depending on the timing.
+            std::unique_ptr<Replication> history;
+            std::unique_ptr<SharedGroup> sg;
+            std::unique_ptr<Group> read_only_group;
+            Realm::open_with_config(config, history, sg, read_only_group, nullptr);
+            sg->begin_read();
+            LangBindHelper::promote_to_write(*sg);
+            Group& group = _impl::SharedGroupFriend::get_group(*sg);
```
tgoyne: These three lines can be just `Group& group = sg->begin_write();`.

src/impl/realm_coordinator.hpp
```diff
@@ -20,10 +20,14 @@
 #define REALM_COORDINATOR_HPP
 
 #include "shared_realm.hpp"
+#include "results_notifier.hpp"
 
+#include <realm/query.hpp>
+#include <realm/table.hpp>
```
tgoyne: These includes can all be replaced with forward declarations.

src/sync/partial_sync.hpp
```diff
@@ -19,20 +19,35 @@
...
 #include <string>
 
 namespace realm {
-
 class Realm;
 class Results;
+class Group;
 
 namespace partial_sync {
+	enum class SubscriptionState : int8_t;
```
tgoyne: The bodies of namespaces are not indented.

src/sync/partial_sync.hpp
```diff
@@ -19,20 +19,35 @@
...
+
+	// Returns the default name for subscriptions. Used if a specific name isn't provided.
+	std::string get_default_name(Query query);
+
+	void register_query(std::shared_ptr<Realm>, const std::string &object_class,
+						const std::string &query,
```
tgoyne: Indentation is all wrong for these.

src/sync/partial_sync.cpp
```diff
@@ -58,6 +60,81 @@ void update_schema(Group& group, Property matches_property)
 
 } // unnamed namespace
 
+std::string get_default_name(Query query) {
```
tgoyne: This should not be copying the Query object.

tests/sync/partial_sync.cpp
```diff
@@ -231,11 +246,12 @@ TEST_CASE("Partial sync", "[sync]") {
         });
     }
 
-    SECTION("invalid queries") {
-        run_query("this isn't a valid query!", partial_config, PartialSyncTestObjects::A, [](Results, std::exception_ptr e) {
-            REQUIRE(e);
-        });
-    }
+//    TODO: Figure out how to test this.
```
bdash: It seems important to know whether error handling is working. Can we reinstate / expand on this test?

src/object_store.cpp
```diff
@@ -50,19 +50,24 @@ const size_t c_primaryKeyObjectClassColumnIndex =  0;
 const char * const c_primaryKeyPropertyNameColumnName = "pk_property";
 const size_t c_primaryKeyPropertyNameColumnIndex =  1;
 
+constexpr const char* result_sets_type_name = "class___ResultSets";
```
tgoyne: It's not clear why any of the changes to this file or object_store.hpp are needed. The partial sync code already handles creating the ResultsSet table.

tests/sync/partial_sync.cpp
```diff
@@ -231,11 +246,12 @@ TEST_CASE("Partial sync", "[sync]") {
         });
     }
```
bdash: It looks like we're missing  tests for new functionality that's been added: query names (both user-provided and our automatically created ones), removal of subscriptions (is this implemented in this PR?), etc.

src/sync/partial_sync.hpp
bdash: Namespace body shouldn't be indented like this.

Indentation, when needed, should use spaces rather than tabs.

src/sync/partial_sync.hpp
```diff
@@ -19,20 +19,35 @@
...
-
 class Realm;
 class Results;
+class Group;
 
 namespace partial_sync {
+	enum class SubscriptionState : int8_t;
+
+	// Returns the default name for subscriptions. Used if a specific name isn't provided.
+	std::string get_default_name(Query query);
```
bdash: This should not copy the `Query`, as that's quite an expensive operation. It should take the query by reference to avoid this.

src/impl/results_notifier.hpp
```diff
@@ -61,6 +64,15 @@ class ResultsNotifier : public CollectionNotifier {
     // The changeset calculated during run() and delivered in do_prepare_handover()
     CollectionChangeBuilder m_changes;
     TransactionChangeInfo* m_info = nullptr;
+    partial_sync::SubscriptionState m_previous_partial_sync_state = partial_sync::SubscriptionState::Undefined;
+
+    // Key used for this partial sync subscription
+    std::string m_partial_sync_name;
```
tgoyne: This design does not work for named subscriptions as there is not a 1:! correspondence between ResultsNotifier instances and subscription names.

src/sync/partial_sync.cpp
```diff
@@ -58,6 +60,81 @@ void update_schema(Group& group, Property matches_property)
...
+    return util::format("[%1] %2", object_class, query.get_description());
+}
+
+void register_query(Group& group,
+                    std::string const& key,
+                    std::string const& object_class,
+                    std::string const& serialized_query,
+                    _impl::ResultsNotifier& notifier)
+{
+    auto table = ObjectStore::table_for_object_type(group, "__ResultSets").get();
```
bdash: Is there a reason why you opted to reimplement the logic in this function rather than reusing the very similar existing logic in the existing `register_query` implementation?

src/impl/collection_change_builder.hpp
```diff
@@ -20,6 +20,8 @@
 #define REALM_COLLECTION_CHANGE_BUILDER_HPP
 
 #include "collection_notifications.hpp"
+#include "subscription_state.hpp"
+#include "../subscription_state.hpp"
```
bdash: Is the presence of both of these includes intentional?

src/subscription_state.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include <cstdint>
+
+namespace realm {
+namespace partial_sync {
+
+// Enum describing the various states a partial sync subscription can have.
+// These states are propagated using the standard collection notification system.
+enum class SubscriptionState : int8_t {
+    Undefined = -3,         // Unknown which state Partial Sync is in.
```
bdash: Is including this as a member of the `enum` the best option? It feels like this is only needed for an `Optional<SubscriptionState>`, where `nil` would make more sense for this state.

src/subscription_state.hpp
```diff
@@ -0,0 +1,39 @@
...
+    NotSupported = - 2,     // Partial Sync not supported.
```
bdash: Why is it necessary to separate this state from `Error`?

src/subscription_state.hpp
bdash: Are these values expected to exactly match those used by the sync server? If so, we should make very clear what these values need to be kept in sync with so that we don't renumber these and unintentionally break things.

src/sync/partial_sync.cpp
```diff
@@ -58,6 +60,81 @@ void update_schema(Group& group, Property matches_property)
 
 } // unnamed namespace
 
+std::string get_default_name(Query query) {
+    // Include table name as part of key to be able to disambiguate the same query on different
+    // tables.
+    std::string object_class = query.get_table()->get_name().substr(6);
```
bdash: This should use `ObjectStore::object_type_for_table_name`.

src/collection_notifications.hpp
```diff
@@ -88,10 +90,30 @@ struct CollectionChangeSet {
...
+    realm::partial_sync::SubscriptionState partial_sync_new_state;
+
+    // Any potential error message from partial sync for the new collection
+    // This is only set if partial_sync_new_state is ERROR
+    std::string partial_sync_error_message;
+
+    // Returns `true` if the modification lists are `empty`, either because
+    // no changes existed, or because they have have otherwise beeen
+    // consumed (By other parts of the Collection Notifier).
+    bool changes_moved() const noexcept {
```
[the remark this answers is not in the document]
tgoyne: It's also not an accurate description of what the single assertion that uses it is asserting. The whole reason that there assert there is because the collection passed to `merge()` *isn't* simply moved from.

src/sync/sync_session.hpp
```diff
@@ -198,6 +198,9 @@ class SyncSession : public std::enable_shared_from_this<SyncSession> {
     // not make the session reconnect.
     void set_multiplex_identifier(std::string multiplex_identity);
 
+    // Inform sync that a local transaction has been commited.
+    void nonsync_transact_notify(VersionID::version_type);
```
tgoyne: Rather than making this public the new call to it should continue to go through Internal::nonsync_transact_notify().

tests/sync/partial_sync.cpp
tgoyne: Actually testing all of the new convoluted logic for handling errors is very important.

src/impl/collection_change_builder.hpp
```diff
@@ -37,7 +39,10 @@ class CollectionChangeBuilder : public CollectionChangeSet {
     CollectionChangeBuilder(IndexSet deletions = {},
                             IndexSet insertions = {},
                             IndexSet modification = {},
-                            std::vector<Move> moves = {});
+                            std::vector<Move> moves = {},
+                            realm::partial_sync::SubscriptionState partial_sync_old_state = realm::partial_sync::SubscriptionState::Undefined,
```
tgoyne: redundant realm::

src/impl/results_notifier.cpp
```diff
@@ -157,6 +187,24 @@ void ResultsNotifier::calculate_changes()
...
+
+#if REALM_ENABLE_SYNC
+    if (!m_partial_sync_name.empty()) {
+        // If a local error was set, this will override anything partial sync might otherwise report.
+        // This will mean that .error might be triggered multiple times if other updates are coming in.
+        // Consider removing the callback after the first local was reported as there is no way to 
+        // recover besides creating a new query.
+        m_changes.partial_sync_old_state = m_previous_partial_sync_state;
+        if (!get_partial_sync_local_error_message().empty()) {
+            m_changes.partial_sync_new_state = realm::partial_sync::SubscriptionState::Error;
```
tgoyne: redundant realm::

tests/sync/partial_sync.cpp
```diff
@@ -118,16 +120,29 @@ void run_query(const std::string& query, const Realm::Config& partial_config, Pa
 {
     std::atomic<bool> partial_sync_done(false);
     auto r = Realm::get_shared_realm(partial_config);
-    Results results;
+    auto table_name = (type == PartialSyncTestObjects::A) ? "class_partial_sync_object_a" : "class_partial_sync_object_b";
+    auto table = r->read_group().get_table(table_name);
+    Query q = table->where();
+    realm::parser::Predicate p = realm::parser::parse(query);
```
tgoyne: redundant realm::

src/object_store.cpp
```diff
@@ -74,6 +79,22 @@ void create_metadata_tables(Group& group) {
         pk_table->insert_column(c_primaryKeyPropertyNameColumnIndex, type_String, c_primaryKeyPropertyNameColumnName);
     }
     pk_table->add_search_index(c_primaryKeyObjectClassColumnIndex);
+
+#if REALM_ENABLE_SYNC
+    // Only add __ResultSets if Realm is a partial Realm
+    if (partial_realm && !group.has_table(result_sets_type_name)) {
```
bdash: Can you explain the motivation behind performing this work here, rather than doing it when the query is registered where it's still necessary to make schema changes too (adding the matches column)?

src/results.cpp
```diff
@@ -690,9 +693,21 @@ void Results::prepare_async()
...
-NotificationToken Results::add_notification_callback(CollectionChangeCallback cb) &
+#if REALM_ENABLE_SYNC
+NotificationToken Results::add_notification_callback(CollectionChangeCallback cb, util::Optional<std::string> subscription_name) &
+#else 
+NotificationToken Results::add_notification_callback(CollectionChangeCallback cb, util::Optional<std::string>) &
+#endif
 {
     prepare_async();
+#if REALM_ENABLE_SYNC
+    if (m_realm->is_partial() && !m_have_subscribed) {
```
bdash: Subscribing to the query needs to be explicit, and separate from simply adding a notification callback, in order to support the API that Cocoa and .NET are exposing (maybe also JS?). It seems like having it be explicit at this level shouldn't be problematic for Java since you can simply call the `subscribe` method (or whatever it's called) at the same point you call `add_notification_callback`.

src/results.cpp
```diff
@@ -690,9 +693,21 @@ void Results::prepare_async()
...
+        std::string key = (subscription_name) ? subscription_name.value() : partial_sync::get_default_name(get_query());
```
bdash: `get_query()` does a non-trivial amount of work, including copying the `Query`. We should use a local variable for the result to avoid doing this work multiple times.

src/impl/realm_coordinator.cpp
```diff
@@ -462,6 +467,64 @@ void RealmCoordinator::pin_version(VersionID versionid)
...
+void RealmCoordinator::register_partial_sync_query(Realm& realm, ResultsNotifier& notifier, Query query, std::string subscription_name)
+#else 
+void RealmCoordinator::register_partial_sync_query(Realm& realm, ResultsNotifier&, Query, std::string)
+#endif
+{
+    if (!realm.is_partial())
+        throw std::logic_error("A partial sync query can only be registered in a partially synced Realm");
+
+#if REALM_ENABLE_SYNC
+    std::string object_class = query.get_table()->get_name().substr(6);
```
bdash: `ObjectStore::object_type_for_table_name`?

src/impl/realm_coordinator.cpp
```diff
@@ -462,6 +467,64 @@ void RealmCoordinator::pin_version(VersionID versionid)
...
+
+    if (realm.schema().find(object_class) == realm.schema().end())
+        throw std::logic_error("A partial sync query can only be registered for a type that exists in the Realm's schema");
+
+    auto& self = Realm::Internal::get_coordinator(realm);
+    std::lock_guard<std::mutex> l(self.m_partial_sync_queue_mutex);
+    std::string key = subscription_name;
+    std::string serialized_query = query.get_description();
```
bdash: What happens with a query that cannot be serialized?

src/impl/realm_coordinator.cpp
```diff
@@ -462,6 +467,64 @@ void RealmCoordinator::pin_version(VersionID versionid)
...
+    self.m_partial_sync_queue.push_back(std::tuple<std::string, std::string, std::string, ResultsNotifier&>(key, object_class, serialized_query, notifier));
```
bdash: I'd expect that you could write this as something like:
```c++
self.m_partial_sync_queue.emplace_back(key, object_class, serialized_query, notifier);
```

src/impl/realm_coordinator.hpp
```diff
@@ -172,6 +183,11 @@ class RealmCoordinator : public std::enable_shared_from_this<RealmCoordinator> {
 
     std::shared_ptr<SyncSession> m_sync_session;
 
+    std::mutex m_partial_sync_queue_mutex;
+    std::deque<std::tuple<std::string, std::string, std::string, ResultsNotifier&>> m_partial_sync_queue;
```
bdash: Rather than using a tuple containing three strings that could easily be confused. It'd be clearer to use a struct with named fields.

src/impl/realm_coordinator.hpp
```diff
@@ -172,6 +183,11 @@ class RealmCoordinator : public std::enable_shared_from_this<RealmCoordinator> {
...
+    std::mutex m_partial_sync_thread_mutex;
```
bdash: This does not appear to be used.

src/impl/results_notifier.hpp
cmelchior: Not sure I understand you, but are you saying that a given instance of a `results_notifier.hpp` is used by multiple `Results`? Which mean that these states should be moved to `Results` instead and then read or sent as they are needed?

src/results.cpp
cmelchior: Yes, we can easily do that.

bdash: It's worth noting that using the notification callback to deliver changes in partial sync state introduces the artificial limitation of it being impossible to subscribe to a partial sync query while inside a write transaction.

src/subscription_state.hpp
cmelchior: Since this state is sent as part of the `CollectionChange` object to notification callbacks. It basically means "Ignore me".

src/collection_notifications.hpp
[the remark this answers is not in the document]
cmelchior: Any suggestion on how to improve it then?

tests/sync/partial_sync.cpp
cmelchior: Removal of a subscription was not part of this PR since removing it is the equivalent of just an asynchronous write transaction, and since there are differences in how the bindings do async, it wasn't clear that adding it to Object Store would be helpful? 

Perhaps adding a synchronous method for doing it, and then letting it be up to bindings to make sure to call this method in whatever way the run asynchronous write transactions? At least the logic for finding and deleting the subscription will be in one place

src/results.hpp
```diff
@@ -188,10 +188,15 @@ class Results {
...
-    NotificationToken add_notification_callback(CollectionChangeCallback cb, util::Optional<std::string> subscription_name = util::Optional<std::string>()) &; // throws
+    NotificationToken add_notification_callback(CollectionChangeCallback cb) &;
+
+    // Register a partial sync subscription for this result set. If an identical query is already subscribed with the
+    // same name, this method does nothing.
+    // If no `subscription_name` is provided, an automatic name is generated. The same name will be generated for the
+    // same query, so if an anonymous query is subscribed to twice, the last subscribe call is ignored.
+    // This method will throw an exception if partial sync could not serialize the query.
+    // Any other errors will be reported through the notification system using `add_notification_callback()`.
+    void subscribe(util::Optional<std::string> subscription_name = util::Optional<std::string>());
```
kneth: I can imagine it would be nice if it returns a `Results` object so you can write:

```
results = realm.objects('Person').filtered('age > 5').subscribe().addListener(...);
```

instead of

```
results = realm.objects('Person').filtered('age > 5');
results.subscribe();
results.addListener(...);
```

bdash: It doesn't seem like that matters at this level. A binding can still provide the API you're after by simply returning `this` after calling `subscribe`.

tests/sync/partial_sync.cpp
bdash: The logic for finding the object to delete, error handling, etc. should really be in object store so bindings don't have to reinvent it.

src/collection_notifications.hpp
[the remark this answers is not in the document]
tgoyne: Continue to assert that the changeset is empty after merge() is called and don't keep state around in strange places?

src/impl/results_notifier.hpp
tgoyne: Results should support calling subscribe() on a single instance more than one time on a single instance with different names.

src/object_store.cpp
tgoyne: The code doesn't need to be moved; it can just be deleted since there's already code for doing this.

src/impl/results_notifier.hpp
cmelchior: That doesn't seem compatible with getting changes reported back through the notification system? As you would then get changes back from multiple subscriptions, which would be really confusing?

tgoyne: Why would subscribing to the same query multiple times (whether through a single Results or multiple) have any impact on what changes get reported?

bdash: Any notifications related to partial sync state changes would potentially come in multiple times as the server may initially see each of the subscriptions at different times. This seems unavoidable if we're trying to force partial sync state changes through the Results notification mechanism.

cmelchior: I guess you got a point since any errors at this point would only be along the line of "Server rejected the query" which all subscribers would need to know about.

src/impl/realm_coordinator.cpp
cmelchior: Currently, this method cannot crash. Core will adjust their semantics so a proper exception is thrown

bdash: And when the exception is thrown here, how will it be handled?

src/results.cpp
```diff
@@ -696,6 +699,20 @@ NotificationToken Results::add_notification_callback(CollectionChangeCallback cb
     return {m_notifier, m_notifier->add_callback(std::move(cb))};
 }
 
+void Results::subscribe(util::Optional<std::string> subscription_name)
```
nirinchev: It bothers me a bit that this is not asynchronous call, considering it's feasible that an error occurs and the subscription doesn't register. Also, it will be asymmetric with regard to unsubscribing, which triggers my OCD. Are there compelling arguments for keeping it synchronous?

tgoyne: It is async (register_partial_sync_query() does the actual work on a different thread).

nirinchev: Yeah, I meant that it doesn't surface in the public API as such - there's no way for the consumer to get a confirmation that their subscription has been "registered".

bdash: This is more explicit in #628. Subscribing returns a `Subscription` object that you can observe for state changes / errors.

cmelchior:
Replaced by #634

PR #611, opened 2017-12-11
[CMake-improve] Build Android with sync enabled on CI

fealebenpae:


tests/util/event_loop.cpp
```diff
@@ -204,6 +205,69 @@ void EventLoop::Impl::perform(std::function<void()> f)
...
+static void release_main_looper()
+{
+    ALooper_release(s_main_looper);
+}
+
+bool EventLoop::has_implementation() { return true; }
+
+std::unique_ptr<EventLoop::Impl> EventLoop::Impl::main()
+{
+    return std::unique_ptr<Impl>(new Impl(s_main_looper));
```
bdash: May as well use `std::make_unique` here.

tests/util/event_loop.cpp
```diff
@@ -59,10 +53,17 @@ struct EventLoop::Impl {
...
     Impl(util::CFPtr<CFRunLoopRef> loop) : m_loop(std::move(loop)) { }
 
     util::CFPtr<CFRunLoopRef> m_loop;
+#elif REALM_USE_ALOOPER
+    Impl(ALooper* looper);
+    void perform_work();
+
+    std::vector<std::function<void()>> m_pending_work;
+    std::mutex m_mutex;
+    std::shared_ptr<util::EventLoopSignal<std::function<void()>>> m_signal;
```
bdash: Why does this need shared ownership? It looks to be owned exclusively by this `Impl` type.

fealebenpae: The implementation of the Android `EventLoopSignal` relies on it being owned by a `std::shared_ptr`, as far as I can tell.

bdash: Ah, that explains it!

tests/util/event_loop.cpp
fealebenpae: I tried that, but the `Impl` constructor is private.

PR #612, opened 2017-12-12
Configure target aliases and exports

fealebenpae:
I removed the `Pegtl` target because it kept being included in the interface link libraries list even after I moved it to private linking. It's a header-only library that's only used inside `src/parser/parser.cpp`.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -143,12 +143,17 @@ target_include_directories(ObjectStore
        $<BUILD_INTERFACE:${RealmObjectStore_SOURCE_DIR}/src>
        $<INSTALL_INTERFACE:include> 
 )
+target_include_directories(ObjectStore
+    PRIVATE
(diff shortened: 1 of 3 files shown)


src/CMakeLists.txt
```diff
@@ -143,13 +143,18 @@ target_include_directories(ObjectStore
        $<BUILD_INTERFACE:${RealmObjectStore_SOURCE_DIR}/src>
        $<INSTALL_INTERFACE:include> 
 )
+target_include_directories(ObjectStore
+    PRIVATE
+        ${RealmObjectStore_SOURCE_DIR}/external/pegtl
```
kneth: It is now pulled by core, right?

fealebenpae: That is correct - at some point we’ll rebase the CMake branches and remove the parser from Object Store.

PR #613, opened 2017-12-12
Don't treat unknown system errors as being fatal to the sync session.

bdash:
Fixes https://github.com/realm/realm-sync/issues/1857.


austinzheng:
One quick question: if the error is unrecognized, should we still respect the `is_fatal` flag?

bdash:
Superseded by #614.

PR #615, opened 2017-12-12
Remove object store's query parser in favor of the core query parser

bdash:
A lot of changes in tests are necessary due to `move_column` / `move_table` being removed. For now I've just #if'd them to get things building.

I think this is going to be the point at which we drop support for older versions of core, since it'll be painful to handle core versions without the new query parser / serializer.

This is not yet ready for review.

Depends on https://github.com/realm/realm-core/pull/2947.


tgoyne:
All of the tests which no longer compile due to the move_table/column removal can simply be deleted. I looked into if any of them were still applicable as part of reviewing that PR, and none were.

bdash:
Thanks for confirming that, @tgoyne!

ironage:
I've migrated the `Any` tests and the macros from `util/compiler.hpp` to core so those can be removed as well.

bdash:
I think this is now ready to be reviewed. It'll need to wait for realm-sync v2.1.11 to be released, as that process failed when last attempted.

PR #616, opened 2017-12-19
Remove support for core versions prior to 4.x and sync 1.x.

bdash:
This is a precursor for #615, since we don't want to keep object store's query parser in the tree once we're using core's query parser.

The majority of this change was running `unifdef` over the source for each of the version and feature check macros.


CMake/RealmCore.cmake
```diff
@@ -180,23 +180,15 @@ function(download_realm_core core_version)
...
             set(basename "realm-core-android")
             set(compression "tar.gz")
             set(platform "-android-${ANDROID_ABI}")
         endif()
-        if(REALM_CORE_PACKAGING STREQUAL "1")
-            set(vprefix "")
-        else()
-            set(vprefix "v")
-        endif()
+        set(vprefix "v")
```
tgoyne: I think you can just remove this entirely.

src/object_accessor.hpp
```diff
@@ -34,9 +34,7 @@
 #include <realm/util/assert.hpp>
 #include <realm/table_view.hpp>
 
-#if REALM_HAVE_SYNC_STABLE_IDS
```
tgoyne: The things in non-sync-specific files guarded by REALM_HAVE_SYNC_STABLE_IDS need to be changed to REALM_ENABLE_SYNC rather than removed entirely to maintain support for building non-sync.

bdash: Will do

PR #620, opened 2018-01-06
Remove bind_with_admin_token() and use the normal refresh flow

tgoyne:
When session multiplexing is enabled we need to perform a token refresh even with admin token users so that ROS knows that we're accessing a new Realm (as it can't just automatically detect that from the fact that we've opened a new connection).

All bindings need to be updated to support this, but bindings which do not support multiplexing (i.e. everything but JS) can simply "refresh" the token on the session without talking to the server. Cocoa doesn't support admin token users at all, so it doesn't need any updates.

Object store part of the fix for https://github.com/realm/realm-object-server-private/issues/836.

### DIFF
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -87,9 +87,6 @@ struct SyncSession::State {
                                       SyncSession&, std::string,
                                       const util::Optional<std::string>&) const { }
 
-    virtual void bind_with_admin_token(std::unique_lock<std::mutex>&,
-                                       SyncSession&, const std::string&, const std::string&) const { }
-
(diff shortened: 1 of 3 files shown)


src/sync/sync_session.cpp
```diff
@@ -849,12 +830,6 @@ void SyncSession::refresh_access_token(std::string access_token, util::Optional<
     m_state->refresh_access_token(lock, *this, std::move(access_token), server_url);
 }
 
-void SyncSession::bind_with_admin_token(std::string admin_token, std::string server_url)
```
bdash: This presumably needs to be removed from the header as well.

PR #622, opened 2018-01-12
Add support for Weak/Strong links aka Cascading deletes

cmelchior:
This PR exposes cores concept of Weak/Strong  links in the Object Store schema using the `Relationship::Strong` parameter when creating link properties. This can be used to implement cascading deletes.

```
// New constructor 
Property(std::string name, PropertyType type, std::string object_type, Relationship relationship)
```

Semantics:
- Default (and current Linktype) is Weak.
- Changing the link type is a breaking change
- Modifying the link type does not clean up the Realm automatically, i.e. changing from weak -> Strong will not automatically remove all objects not connected. This cleanup only happens if a links is actually updated.
- It is possible to create child objects with no parents, these will not be deleted automatically. This only happens if the parent is created, the link is created, and then removed again.
- Cascading behaviour is immediate inside a transaction. Rolling back the transaction will undo all deletes.
- Some corner cases exist for cyclic object graphs, they are documented in core

Java PR using this is here: https://github.com/realm/realm-java/pull/5678


src/property.hpp
```diff
@@ -21,9 +21,11 @@
 
 #include "util/tagged_bool.hpp"
 
+#include <string>
```
tgoyne: This include was previously in the correct place.

src/property.hpp
```diff
@@ -85,6 +94,17 @@ struct Property {
 
     bool requires_index() const { return is_primary || is_indexed; }
 
+    // Return the underlying LinkType. Only useful for Object or Array properties.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wreturn-type"
```
tgoyne: Giving `Relationship` a backing type of `bool` eliminates the need for this.

src/property.hpp
```diff
@@ -204,6 +224,14 @@ inline Property::Property(std::string name, PropertyType type,
 {
 }
 
+inline Property::Property(std::string name, PropertyType type,
```
tgoyne: This constructor is redundant.

src/property.hpp
```diff
@@ -85,6 +94,17 @@ struct Property {
 
     bool requires_index() const { return is_primary || is_indexed; }
 
+    // Return the underlying LinkType. Only useful for Object or Array properties.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wreturn-type"
+    LinkType link_type() const {
+        switch(relationship) {
```
tgoyne: space between keyword and (

src/schema.cpp
```diff
@@ -136,6 +136,9 @@ static void compare(ObjectSchema const& existing_schema,
         else if (current_prop.requires_index()) {
             changes.emplace_back(schema_change::RemoveIndex{&existing_schema, &current_prop});
         }
+        if(target_prop->relationship != current_prop.relationship) {
```
tgoyne: space between keyword and (

tests/link_types.cpp
```diff
@@ -0,0 +1,303 @@
...
+#include "object_accessor.hpp"
+#include "impl/realm_coordinator.hpp"
+#include "impl/object_accessor_impl.hpp"
+#include "util/test_file.hpp"
+
+// This class contain tests for the interaction between Strong and Weak links.
+// See Descriptor::set_link_type() in Core (descriptor.hpp) for an extended explanation of the exact semantics.
+
+namespace realm {
+class TestHelper {
```
tgoyne: There's no reason to put these functions in a class.

tests/link_types.cpp
```diff
@@ -0,0 +1,303 @@
+////////////////////////////////////////////////////////////////////////////
```
tgoyne: This entire file appears to just be testing core functionality, and should be part of core's test suite instead.

tests/migrations.cpp
```diff
@@ -146,6 +146,11 @@ Schema set_target(Schema schema, StringData object_name, StringData property_nam
     return schema;
 }
 
+Schema set_relationship(Schema schema, StringData object_name, StringData property_name, Relationship new_relationship_type) {
```
tgoyne: newline before {

src/property.hpp
cmelchior: I'm not entirely sure why, but If this is not here I get: 

```
In file included from /Users/cm/Realm/realm-object-store/src/property.hpp:24:
/Users/cm/Realm/realm-object-store/CMakeFiles/realm-core-5.0.1/include/realm/data_type.hpp:27:9: error: unknown type name 'int64_t'
typedef int64_t Int;
```

tgoyne: data_types.hpp needs to include <cstdint>

src/property.hpp
cmelchior: It is being used by tests and not having it results in:
```
Undefined symbols for architecture x86_64:
  "realm::Property::Property(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, realm::PropertyType, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >)", referenced from:
      ____C_A_T_C_H____T_E_S_T____0() in list.cpp.o
      ____C_A_T_C_H____T_E_S_T____3() in migrations.cpp.o
      ____C_A_T_C_H____T_E_S_T____111() in migrations.cpp.o
      ____C_A_T_C_H____T_E_S_T____134() in migrations.cpp.o
      ____C_A_T_C_H____T_E_S_T____0() in object.cpp.o
      ____C_A_T_C_H____T_E_S_T____3() in object_store.cpp.o
      ____C_A_T_C_H____T_E_S_T____102() in results.cpp.o
```

tgoyne: It is functionally identical to the three-argument version with a default value for the third argument (i.e. what previously existed).

cmelchior: Those was changed as well, these constructors have no default values anymore as having two 3-argument constructors with different defaults is ambiguous. So there is now a constructor for creating backlinks (with a string default value) and one for creating normal links (with a relationship default value)

tgoyne: Having `Property(std::string name, PropertyType type, std::string object_type, std::string link_origin_property_name="");` and `Property(std::string name, PropertyType type, std::string object_type, Relationship relationship);` is not ambiguous.

cmelchior: Ah, like that. No, your right.

cmelchior: 2-arg constructor has been removed

cmelchior:
All comments addressed

cmelchior:
Currently, Sync will crash with `Unsupported operation` if you try to change the link type, so merging this PR is probably not a good idea until that has been fixed. Apparently, that will require a protocol upgrade.

Zhuinden:
Destructive changes always caused "bad changeset received", how is this any different?

tgoyne:
The problem is not that you cannot make existing links strong, but rather that sync does not support strong links at all.

Zhuinden:
...oooh. Ouch. Thanks for the info.

Although in that case, what is the *current* link type?

nirinchev:
Cascading deletes are now (somewhat) supported via embedded objects and it's unlikely sync will ever support strong links, so closing this.

Issue #581, opened 2017-10-18
register_progress_notifier and is_streaming = false


beeender:
The problem:
`SyncSession::register_progress_notifier` is disconnected from the Realm version on the thread, which makes it somehow useless in the real life and feels buggy.

Example:
```
// on thread A
Realm realm = getInstance(config); // Group version 1
writeData(realm); // WriteA GroupVersion -> 2
writeData(realm); // WriteB GroupVersion -> 3
realm.close();
SyncSession session = SyncManager.getSession(config);
session.register_progress_notifier(callback, upload, false);
```

The `callback` in above code actually might be called in various conditions
1. The callback is called after all the data after `WriteB` get uploaded to the server. And then callback won't be triggered anymore. This is what I expected as a user, since before adding the callback, everything i have committed locally already. I probably want to show a progress bar to say that everything has just be synchronized to the server.  In this case, the `snapshot_version` passed from sync should be `>= 3`.

2. The callback is called with after all data in `WriteA` has been uploaded, but not in `WriteB`. Here the `snapshot_version` should be `== 2`

3. The callback is called with no data committed by user. This could happen if the Realm file was first time created, and sync only uploaded the meta data to the server. `snapshot_version == 0`. Quite similar to 2), but still, it is not user expected.

With current ObjectStore implementation, all above could happen. 

Maybe we can solve it by:
```c++
uint64_t SyncSession::register_progress_notifier(std::function<SyncProgressNotifierCallback> notifier,
                                                 NotifierType direction, bool is_streaming, int64_t expected_version)
{
    auto wrapped_progress_handler = [this, weak_self](uint_fast64_t downloaded, uint_fast64_t downloadable,
                                                  uint_fast64_t uploaded, uint_fast64_t uploadable,
                                                  bool is_fresh, uint_fast64_t snapshot_version) {
        if (auto self = weak_self.lock() && snapshot_version >= expected_version) {
            handle_progress_update(downloaded, downloadable, uploaded, uploadable, is_fresh);
        }
    };
}
```

So bindings need to change the public API as well to decide which version of data is user want to see the progress from.

The `uploaded` and `uploadable` might be less than what user expected, but i guess it is an issue has to be solved in the sync side and it is kind of minor.

thoughts? @cmelchior @austinzheng @morten-krogh

beeender:
also the `wrapped_progress_handler` doesn't match the signature of sync's `ProgressHandler`

```
[...]

PR #624, opened 2018-01-16
Fix upload progress reporting

tgoyne:
When a non-streaming progress notification is added immediately after a local commit was made, that progress notification could capture the total bytes to upload before that value was actually updated to reflect the newly made transaction, resulting in the notifier completing before the upload actually happened. Fix this by tracking the last seen local transaction version and only capture the uploadable bytes value once the sync progress is for that version or later.

The functional changes are all in the second commit. The first commit extracts the callback management code to a helper class that can be unit tested directly rather than via test helper functions on SyncSession (none of the existing tests were integration tests that meaningfully tested it by doing things via sync).

Fixes https://github.com/realm/realm-sync/issues/1904.

### DIFF
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -599,9 +599,10 @@ void SyncSession::cancel_pending_waits()
 }
 
 void SyncSession::handle_progress_update(uint64_t downloaded, uint64_t downloadable,
-                                         uint64_t uploaded, uint64_t uploadable, bool is_fresh)
+                                         uint64_t uploaded, uint64_t uploadable,
+                                         uint64_t download_version, uint64_t snapshot_version)
 {
-    m_notifier.update(downloaded, downloadable, uploaded, uploadable, is_fresh);
+    m_notifier.update(downloaded, downloadable, uploaded, uploadable, download_version, snapshot_version);
 }
 
 void SyncSession::create_sync_session()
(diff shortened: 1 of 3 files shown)


bdash:
It sounds like this also addresses https://github.com/realm/realm-object-store/issues/581?

tgoyne:
It appears so, but that issue makes it sound a lot more complicated so maybe I'm missing something?

src/sync/sync_session.cpp
```diff
@@ -899,3 +820,95 @@ void SyncSession::did_drop_external_reference()
 
     m_state->close(lock, *this);
 }
+
+uint64_t SyncProgressNotifier::register_callback(std::function<SyncProgressNotifierCallback> notifier,
+                                                 bool is_download, bool is_streaming)
```
bdash: Changing this argument from an enum to a boolean makes the call sites in the tests harder to decipher.

tests/sync/session/progress_notifications.cpp
```diff
@@ -295,16 +329,59 @@ TEST_CASE("progress notification", "[sync]") {
             callback_was_called = false;
             current_transferred = original_transferrable + 250;
             current_transferrable = 1228;
-            progress.update(current_transferred, current_transferrable, 199, 591);
+            progress.update(current_transferred, current_transferrable, 199, 591, 1, 1);
             CHECK(!callback_was_called);
         }
 
+        SECTION("download notifications are not sent until all a DOWNLOAD message has been received") {
```
bdash: Should "not sent until all a" be "not sent until a"?

src/sync/sync_session.cpp
tgoyne: Yeah, I think I'll revert that. I intended to translate at the public API boundary, but then I moved where the relevant API boundary was and didn't reconsider...

src/sync/sync_session.cpp
```diff
@@ -899,3 +823,96 @@ void SyncSession::did_drop_external_reference()
...
+        transferrable = *captured_transferrable;
+    }
+    else {
+        if (is_download)
+            captured_transferrable = current_progress.downloadable;
+        else {
+            // If the sync client has not yet processed all of the local
+            // transactions then the uploadable data is incorrect and we should
+            // not invoke the callback
+            if (snapshot_version > current_progress.snapshot_version)
```
morten-krogh: Yes. This is exactly the logic to check.

PR #627, opened 2018-01-29
Exposing get_number_of_versions

kneth:
We have a customer with large Realm files, and we need to understand why.

### DIFF
--- a/tests/realm.cpp
+++ b/tests/realm.cpp
@@ -1290,6 +1290,22 @@ TEST_CASE("SharedRealm: compact on launch") {
             REQUIRE(num_opens == 3);
         }).join();
     }
+
+    SECTION("multiple versions when multiple threads") {
+        r = Realm::get_shared_realm(config);
+        std::thread([&]{
+            auto r2 = Realm::get_shared_realm(config);
+            r2->begin_transaction();
+            auto table = r2->read_group().get_table("class_object");
+            size_t count = 1000;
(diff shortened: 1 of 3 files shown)


tests/realm.cpp
```diff
@@ -1290,6 +1290,22 @@ TEST_CASE("SharedRealm: compact on launch") {
             REQUIRE(num_opens == 3);
         }).join();
     }
+
+    SECTION("multiple versions when multiple threads") {
```
bmunkholm: I would argue this is the wrong place to test the core function. That must be assumed or validated that it's tested in core.

kneth: True, but I simply wish to be sure that I am able to get a value. Otherwise it feels a bit speculative.

tests/realm.cpp
```diff
@@ -1290,6 +1290,22 @@ TEST_CASE("SharedRealm: compact on launch") {
...
+        r = Realm::get_shared_realm(config);
+        std::thread([&]{
+            auto r2 = Realm::get_shared_realm(config);
+            r2->begin_transaction();
+            auto table = r2->read_group().get_table("class_object");
+            size_t count = 1000;
```
bmunkholm: why 1000? What's the least you can do with to verify the functionality?

kneth:
I don't think the method will provide any useful information - at least not for JS.

PR #628, opened 2018-01-30
Replace the partial sync API

bdash:
This replaces the partial sync API introduced in #547 with one that supports named queries, and that allows subscribing to an existing `Results` so bindings can build queries using their high-level APIs. It supersedes @cmelchior's PR in #610 based on discussions on Slack about the difficulty of adapting that implementation to Cocoa's API.

Not yet supported:
 - [x] Unsubscribing (this may come in a later PR?)

Remaining work needed here:
 - [x] Further tests around subscription naming.

Things to further evaluate:
 - [x] Is the locking in `Subscription::Notifier` correct? There's a comment on `ResultsNotifier::do_add_required_change_info` that outlines what's safe to access when, but I've not thought too hard about how that relates to this code.
 - [x] Should `Subscription::Notifier` be where the error / subscription completed state is stored, or should it be storing it on the `Subscription` as part of delivering the notification?

I'm flagging this for review despite their being some work remaining so I can start getting feedback.

### DIFF
--- a/src/sync/partial_sync.cpp
+++ b/src/sync/partial_sync.cpp
@@ -198,13 +198,13 @@ void async_register_query(Realm& realm, std::string object_type, std::string que
         auto matches_property = std::string(object_type) + "_matches";
 
         update_schema(realm->read_group(), Property(matches_property, PropertyType::Object|PropertyType::Array, object_type));
-        auto result_sets_schema = std::make_unique<ObjectSchema>(realm->read_group(), result_sets_type_name);
+        ObjectSchema result_sets_schema(realm->read_group(), result_sets_type_name);
 
         try {
             if (!validate_existing_subscription(realm, name, query,
-                                                matches_property, *result_sets_schema)) {
+                                                matches_property, result_sets_schema)) {
                 CppContext context;
-                auto object = Object::create<util::Any>(context, realm, *result_sets_schema,
+                auto object = Object::create<util::Any>(context, realm, result_sets_schema,
                                                         AnyDict{
                                                             {"matches_property", matches_property},
                                                             {"name", name},
@@ -403,7 +403,7 @@ void reset_for_testing()
 }
 
 Subscription::Subscription(std::string name, std::string object_type, std::shared_ptr<Realm> realm)
-: m_object_schema(std::make_unique<ObjectSchema>(realm->read_group(), result_sets_type_name))
+: m_object_schema(realm->read_group(), result_sets_type_name)
 {
     // FIXME: Why can't I do this in the initializer list?
     m_notifier = std::make_shared<Notifier>(realm);
@@ -413,8 +413,8 @@ Subscription::Subscription(std::string name, std::string object_type, std::share
 
     TableRef table = ObjectStore::table_for_object_type(realm->read_group(), result_sets_type_name);
     Query query = table->where();
-    query.equal(m_object_schema->property_for_name("name")->table_column, name);
-    query.equal(m_object_schema->property_for_name("matches_property")->table_column, matches_property);
+    query.equal(m_object_schema.property_for_name("name")->table_column, name);
+    query.equal(m_object_schema.property_for_name("matches_property")->table_column, matches_property);
     m_result_sets = Results(realm, std::move(query));
 }
 
@@ -437,7 +437,7 @@ util::Optional<Object> Subscription::result_set_object() const
 {
     if (m_notifier->subscription_completed()) {
         if (auto row = m_result_sets.first())
-            return Object(m_result_sets.get_realm(), *m_object_schema, *row);
+            return Object(m_result_sets.get_realm(), m_object_schema, *row);
     }
 
     return util::none;
--- a/src/sync/partial_sync.cpp
+++ b/src/sync/partial_sync.cpp
@@ -191,7 +191,7 @@ bool validate_existing_subscription(Table& table, ResultSetsColumns const& colum
     return true;
 }
 
-void async_register_query(Realm& realm, std::string object_type, std::string query, std::string name,
+void enqueue_registration(Realm& realm, std::string object_type, std::string query, std::string name,
                           std::function<void(std::exception_ptr)> callback)
 {
     auto config = realm.config();
@@ -234,7 +234,7 @@ void async_register_query(Realm& realm, std::string object_type, std::string que
     });
 }
 
-void async_unregister_query(Object result_set, std::function<void()> callback)
+void enqueue_unregistration(Object result_set, std::function<void()> callback)
 {
     auto realm = result_set.realm();
     auto config = realm->config();
@@ -448,7 +448,7 @@ Subscription subscribe(Results const& results, util::Optional<std::string> user_
 
     Subscription subscription(name, results.get_object_type(), realm);
     std::weak_ptr<Subscription::Notifier> weak_notifier = subscription.m_notifier;
-    async_register_query(*realm, results.get_object_type(), std::move(query), std::move(name),
+    enqueue_registration(*realm, results.get_object_type(), std::move(query), std::move(name),
                          [weak_notifier=std::move(weak_notifier)](std::exception_ptr error) {
         if (auto notifier = weak_notifier.lock())
             notifier->finished_subscribing(error);
@@ -461,7 +461,7 @@ void unsubscribe(Subscription& subscription)
     if (auto result_set_object = subscription.result_set_object()) {
         // The subscription has its result set object, so we can queue up the unsubscription immediately.
         std::weak_ptr<Subscription::Notifier> weak_notifier = subscription.m_notifier;
-        async_unregister_query(*result_set_object, [weak_notifier=std::move(weak_notifier)]() {
+        enqueue_unregistration(*result_set_object, [weak_notifier=std::move(weak_notifier)]() {
             if (auto notifier = weak_notifier.lock())
                 notifier->finished_unsubscribing();
         });
--- a/src/sync/partial_sync.cpp
+++ b/src/sync/partial_sync.cpp
@@ -485,8 +485,13 @@ void unsubscribe(Subscription& subscription)
         }
 
         case SubscriptionState::Error:
+            // We encountered an error when creating the subscription. There's nothing to remove, so just
+            // mark the subscription as removed.
+            subscription.m_notifier->finished_unsubscribing();
+            break;
+
         case SubscriptionState::Invalidated:
-            // Nothing to do. We either failed to create the subscription, or have already removed it.
+            // Nothing to do. We have already removed the subscription.
             break;
 
         case SubscriptionState::Pending:
--- a/src/sync/partial_sync.cpp
+++ b/src/sync/partial_sync.cpp
@@ -369,6 +369,9 @@ util::Optional<Object> Subscription::result_set_object() const
 
 SubscriptionState Subscription::status() const
 {
+    if (!m_notifier->subscription_completed())
+        return SubscriptionState::Creating;
+
     if (m_notifier->error())
         return SubscriptionState::Error;
 
@@ -378,7 +381,9 @@ SubscriptionState Subscription::status() const
         return (SubscriptionState)value;
     }
 
-    return SubscriptionState::Uninitialized;
+    // We may not have an object even if the subscription has completed if the completion callback fired
+    // but the result sets callback is yet to fire.
+    return SubscriptionState::Creating;
 }
 
 std::exception_ptr Subscription::error() const
--- a/src/sync/partial_sync.cpp
+++ b/src/sync/partial_sync.cpp
@@ -23,16 +23,41 @@
 #include "object_schema.hpp"
 #include "results.hpp"
 #include "shared_realm.hpp"
+#include "sync/subscription_state.hpp"
 #include "sync/sync_config.hpp"
+#include "util/event_loop_signal.hpp"
 
 #include <realm/util/scope_exit.hpp>
 
+namespace {
+constexpr const char* result_sets_type_name = "__ResultSets";
+}
+
 namespace realm {
+
+namespace _impl {
+
+void initialize_schema(Group& group)
+{
+    std::string result_sets_table_name = ObjectStore::table_name_for_object_type(result_sets_type_name);
+    if (group.has_table(result_sets_table_name))
+        return;
+
+    TableRef table = sync::create_table(group, result_sets_table_name);
+    size_t indexable_column_idx = table->add_column(type_String, "name"); // Custom property
+    table->add_search_index(indexable_column_idx);
+    table->add_column(type_String, "query");
+    table->add_column(type_String, "matches_property");
+    table->add_column(type_Int, "status");
+    table->add_column(type_String, "error_message");
+    table->add_column(type_Int, "query_parse_counter");
+}
+
+} // namespace _impl
+
 namespace partial_sync {
 
 namespace {
-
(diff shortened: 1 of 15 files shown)


src/sync/subscription_state.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include <cstdint>
+
+namespace realm {
+namespace partial_sync {
+
+// Enum describing the various states a partial sync subscription can have.
+enum class SubscriptionState : int8_t {
+    Error = -1,             // An error occurred with the partial sync subscription..
+    Uninitialized = 0,      // The subscription was created, but has not yet been processed by the sync server.
+    Initialized = 1         // The subscription has been initialized successfully and is syncing data to the device.
```
bdash: I'm thinking that `Incomplete`/ `Complete` or `Pending`/`Complete` may be better terms here, since "uninitialized" doesn't seem like a great term for a subscription that we've created but that the server simply hasn't processed yet. It's accurate in describing the `List` of results associated with the subscription, but not for the subscription itself.

On another note, is `State` or `Status` the tight term here? I used `Subscription::status()` because it felt slightly better to me, but it's obviously bad to be inconsistent within the same API.

src/sync/partial_sync.cpp
```diff
@@ -56,6 +83,153 @@ void update_schema(Group& group, Property matches_property)
...
+    void create_thread()
+    {
+        if (m_thread.joinable())
+            m_thread.join();
+
+        m_thread = std::thread([this] {
+            std::vector<std::function<void()>> queue;
+
+            std::unique_lock<std::mutex> lock(m_mutex);
+            while (!m_stopping &&
```
tgoyne: I think this first m_stopping check is redundant.

src/sync/partial_sync.cpp
```diff
@@ -56,6 +83,153 @@ void update_schema(Group& group, Property matches_property)
...
+
+    std::mutex m_mutex;
+    std::condition_variable m_cv;
+    std::vector<std::function<void()>> m_queue;
+    std::thread m_thread;
+    bool m_stopping = false;
+    bool m_stopped = true;
+};
+
+static auto& work_queue = *new WorkQueue;
```
tgoyne: A single global work queue seems problematic. It means that a write transaction on one Realm can block creating partial sync subscriptions for different Realms. I can't think of a case where this would outright deadlock, but it feels like the sort of thing that'll inevitably cause problems.

tgoyne: The fact that you can't know when the Realm will actually be closed after registering a partial sync query also seems like a potential problem, but I don't know what to do about that.

src/sync/partial_sync.cpp
```diff
@@ -56,6 +83,153 @@ void update_schema(Group& group, Property matches_property)
...
+        realm->begin_transaction();
+        auto cleanup = util::make_scope_exit([&]() noexcept {
+            if (realm->is_in_transaction())
+                realm->cancel_transaction();
+        });
+
+        auto matches_property = std::string(object_type) + "_matches";
+
+        update_schema(realm->read_group(), Property(matches_property, PropertyType::Object|PropertyType::Array, object_type));
+        auto result_sets_schema = std::make_unique<ObjectSchema>(realm->read_group(), result_sets_type_name);
```
tgoyne: This doesn't appear to need to be heap allocated.

src/sync/subscription_state.hpp
tgoyne: New/Pending/Complete?

src/sync/partial_sync.cpp
bdash: > A single global work queue seems problematic. It means that a write transaction on one Realm can block creating partial sync subscriptions for different Realms.

The only problem that came to mind was the potential for unexpected delays when subscribing to multiple Realms, which doesn't see great but also seems unlikely to be problematic. We could avoid that by having the queue be per-`RealmCoordinator` instead if it makes you nervous.

> The fact that you can't know when the Realm will actually be closed after registering a partial sync query also seems like a potential problem, but I don't know what to do about that.

I'm not sure what you mean here. The background Realm will be closed at the point the `Subscription` moves to either the error or complete state.

src/sync/subscription_state.hpp
bdash: What would the distinction be between the new and pending states?

src/sync/partial_sync.cpp
bdash: Quite right. An earlier version heap allocated it to ensure it lived as long as an `Object` that used it, but that's no longer a concern.

src/sync/partial_sync.cpp
nirinchev: Hm... I'm not sure I fully understand this logic. The way I read the code, the background Realm will be closed as soon as the object is written, regardless of what happens to the subscription object. Moving the subscription to the complete/initialized/whatever state happens on the original thread, does it no?

bdash: That's correct. The background Realm is only used for validating and creating the subscription. It is then closed.

src/sync/subscription_state.hpp
bdash: I could see some merit in introducing a state that represents the period while we're waiting on the subscription to be validated and created by the background thread. "New" doesn't seem like a great fit for that though, since there's no guarantee that the background thread will actually create the subscription (it may fail one of the validation checks, or it may be an exact match for an existing subscription). Something like "Creating" may work, though.

src/sync/partial_sync.cpp
bdash: Why's that? Without it the loop will only exit if `wait_for` returns false due to none of the conditions being met before the timeout expires.

tgoyne: I missed that the behavior is different for m_stopping being true in the first check and in the predicate (and that difference is correct).

src/sync/subscription_state.hpp
nirinchev: That was my thinking when I commented on https://github.com/realm/realm-object-store/pull/610#discussion_r164691985, although I may not have articulated that well enough.

I would even argue for adding a callback on the `subscribe` call that will be invoked when the subscription has been created and written to the background Realm. My reasoning (beyond making it symmetric to the way I imagine `unsubscribe` will behave from the public API POV) is that it will differentiate waiting for the server from waiting for the client and once your callback is invoked, you'll have durability guarantees about your subscription. On the other hand, it may be confusing that the callback is invoked when the subscription is created rather than when its data has been synchronized.

I don't have strong feelings about either approach (callback or additional state), but I do feel strongly we need to add a way to notify the user that subscription request is written and waiting to be processed.

bdash: The reason I opted not to use a callback as you describe is that it results in there being two places that errors may be reported for the same subscription: the callback (for an error in creating the subscription), and the subscription state change handler (for an error reported by the server).

> I do feel strongly we need to add a way to notify the user that subscription request is written and waiting to be processed.

To be clear, you're saying that you feel it's important to distinguish between "waiting for the subscription to be created by the background thread" and "waiting for the server to process the new subscription"?

nirinchev: Yes, if that's easy to do, I see some merit in exposing it to developers so there's an early point in time where they can be sure that the subscription is written to the realm - so they can e.g. request additional execution time by the system and ensure that next time the app starts, it'll start synchronizing the subscribed query. I believe that was a pattern available on iOS a while back, not sure if it still applies today, but I imagine there are other similar use cases.

src/sync/partial_sync.cpp
cmelchior: I think this might cause problems for Java. Especially in test setups where we delete Realms after they are closed. We currently use `SyncSessionStopPolicy::Immediately` for the same reason, but this queue seems to outlive the lifecycle of the Realm in nondeterministic ways which look like it could crashes if the underlying Realm is deleted in the meantime.

src/sync/partial_sync.cpp
```diff
@@ -56,6 +83,153 @@ void update_schema(Group& group, Property matches_property)
...
+                                                            {"matches_property", matches_property},
+                                                            {"name", name},
+                                                            {"query", query},
+                                                            {"status", int64_t(0)},
+                                                            {"error_message", std::string()},
+                                                            {"query_parse_counter", int64_t(0)},
+                                                        }, false);
+            }
+        } catch (...) {
+            callback(std::current_exception());
```
cmelchior: This callback seems to happen outside the notifier framework? In order for Java to couple these errors + notifications from the CollectionNotifier, we need to be able to be able to reason about the latest state of both.

cmelchior: After looking at the Subscription `struct the `finished_subscribing` method seems to handle this?

src/sync/partial_sync.cpp
cmelchior: After looking at `partial_sync::reset_for_testing()`, that seems make sure the above doesn't happen, correct?

src/sync/subscription_state.hpp
cmelchior: I'm not sure the distinction matters, but I don't have any strong feelings neither towards State vs. Status nor the individual names used.

bdash: I added a `Creating` state, used until the background thread has finished writing to the Realm, and renamed the other states to `Pending` and `Complete`.

src/sync/partial_sync.cpp
bdash: I moved the work queue to be owned by the `RealmCoordinator`, so it will be torn down as soon as the last reference to a given Realm is dropped.

And yes `partial_sync::reset_for_testing()` existed to prevent that problem with the earlier singleton queue. I removed it since it shouldn't be necessary with the new ownership model.

src/sync/partial_sync.cpp
bdash: The callback that is invoked here is the one within `subscribe` that routes the error, if any, to `Notifier::finished_subscribing`. The `Notifier` handles kicking things over to the originating thread by hooking in to the collection notification subsystem.

bdash:
> There doesn't appear to be any hooks for being notified about when notification start and stop sending, which would be required for Java to adopt this as we talked about. But I guess it can be considered outside the scope of this PR and added later?

Yeah, I haven't tried to tackle this portion of the problem just yet. I figured that since Java is using your object store branch for now, it was a higher priority to get something working for Cocoa before tackling the remaining task that would allow switching Java over to this implementation.

bdash:
I think this is at a good point for it to be reviewed.

kneth:
@bdash I have tried to include it in Realm JS' three build systems, and it compiles fine :smile:

src/sync/partial_sync.cpp
```diff
@@ -56,6 +151,123 @@ void update_schema(Group& group, Property matches_property)
...
+bool validate_existing_subscription(Table& table, ResultSetsColumns const& columns, std::string const& name,
+                                    std::string const& query, std::string const& matches_property)
+{
+    auto existing_row_ndx = table.find_first_string(columns.name, name);
+    if (existing_row_ndx == npos)
+        return false;
+
+    StringData existing_query = table.get_string(columns.query, existing_row_ndx);
+    if (existing_query != query)
+        throw std::runtime_error("An existing subscription exists with the same name, but a different query.");
```
tgoyne: Putting the non-matching queries and result types in these errors could aid debugging.

src/sync/partial_sync.cpp
```diff
@@ -129,5 +342,252 @@ void register_query(std::shared_ptr<Realm> realm, const std::string &object_clas
...
+    , m_coordinator(_impl::RealmCoordinator::get_coordinator(get_realm()->config()).get())
+    {
+    }
+
+    void release_data() noexcept override { }
+    void run() override
+    {
+        std::unique_lock<std::mutex> lock(m_mutex);
+        if (m_has_results_to_deliver)
+            m_changes.modify(0);
```
tgoyne: This could use a comment as it's a bit of a hack.

src/sync/partial_sync.cpp
```diff
@@ -129,5 +342,252 @@ void register_query(std::shared_ptr<Realm> realm, const std::string &object_clas
...
+Subscription subscribe(Results const& results, util::Optional<std::string> user_provided_name)
+{
+    auto realm = results.get_realm();
+
+    auto sync_config = realm->config().sync_config;
+    if (!sync_config || !sync_config->is_partial)
+        throw std::logic_error("A partial sync query can only be registered in a partially synced Realm");
+
+    auto query = results.get_query().get_description(); // Throws if the query cannot be serialized.
+    std::string name = user_provided_name ? std::move(*user_provided_name) : default_name_for_query(query,results.get_object_type());
```
tgoyne: missing a space

src/sync/partial_sync.cpp
```diff
@@ -56,6 +151,123 @@ void update_schema(Group& group, Property matches_property)
...
+                        config=std::move(config)] () {
+        with_open_shared_group(config, [&](SharedGroup& sg) {
+            // Import handed-over object.
+            sg.begin_read(handover->version);
+            Row row = *sg.import_from_handover(std::move(handover->row));
+            sg.unpin_version(handover->version);
+
+            _impl::WriteTransactionNotifyingSync write(config, sg);
+            if (row.is_attached())
+                row.move_last_over();
```
tgoyne: Could do `if (row.is_attached()) { row.move_last_over(); write.commit(); } else { write.cancel(); }` to avoid spurious writes if unregister is called twice.

But meh, maybe not worth caring even that small amount about that.

tests/sync/partial_sync.cpp
```diff
@@ -287,20 +300,146 @@ TEST_CASE("Partial sync", "[sync]") {
...
             REQUIRE(results.size() == 2);
             REQUIRE(results_contains(results, {2, 2, "partial"}));
             REQUIRE(results_contains(results, {3, 8, "sync"}));
         });
     }
+
+    SECTION("unnamed query can be unsubscribed while in creating state") {
+        auto subscription = subscription_with_query("number > 1", partial_config, "object_a", util::none);
+
+        std::atomic<bool> partial_sync_done(false);
```
tgoyne: These shouldn't need to be atomic as both the callback and run_until() are called on the main thread.

bdash:
The sync tests on Linux have been timing out on CI so I ran them locally. It looks like we're hitting a deadlock. `DaemonThread::remove_commit_helper` on the main thread is attempting to acquire `m_mutex` which is held by the notification thread, but the code calling `remove_commit_helper` holds `RealmCoordinator`'s `m_realm_mutex`, which the notification thread is attempting to acquire. This looks like an existing problem that this PR just happens to trigger 😦

bdash:
a8d79de should work around the deadlock for now.

bdash:
Still hanging on CI. Hopefully I'll be able to reproduce it locally if I run in the same docker environment used on CI.

bdash:
I've had absolutely no luck reproducing the hang locally, even when using the same Docker script that CI is running 😫

bdash:
Ok, I'm finally able to reproduce the hang locally by using `cpulimit` to limit the process to only 1% of the available CPU resources. It's the same deadlock as described above, this time hit when the test case itself opens a Realm. I think we'll need to fix the underlying deadlock to get this merged ☹️

bdash:
The deadlock mentioned above was fixed by #633.

kneth:
@bdash Thanks for the hard work on this!

PR #629, opened 2018-01-30
 Fix compilation with sync 3.0

tgoyne:


src/sync/sync_session.cpp
```diff
@@ -89,7 +89,7 @@ struct SyncSession::State {
     // Returns true iff the lock is still locked when the method returns.
     virtual bool access_token_expired(std::unique_lock<std::mutex>&, SyncSession&) const { return true; }
 
-    virtual void nonsync_transact_notify(std::unique_lock<std::mutex>&, SyncSession&, sync::Session::version_type) const { }
+    virtual void nonsync_transact_notify(std::unique_lock<std::mutex>&, SyncSession&, sync::version_type) const { }
```
bdash: Do these types exist directly under the `sync` namespace in older versions of sync as well?

bdash: Perhaps we should just make v2.3.0 the new minimum required version?

tgoyne: I thought it was added in 2.0, but looks like it was in 2.3.

PR #630, opened 2018-02-01
SyncConfig::realm_url() returns a partial sync url including user identity.

morten-krogh:
SyncConfig::realm_url() is changed to return a realm url of the form:

/reference/__partial/user-id/identifier

which is needed for partial sync permissions.

Unit test for this form of the realm_url().

Updated sync dependency to 3.0.0-alpha.2.

REMARK: This PR need not be merged in before Object store updates to the sync 3.x branch.


morten-krogh:
@bdash 
Can I ask you to look at this?

dependencies.list
```diff
@@ -1,4 +1,4 @@
 REALM_CORE_VERSION=5.1.2
-REALM_SYNC_VERSION=2.3.0
+REALM_SYNC_VERSION=3.0.0-alpha.2
```
bdash: It's worth updating this (and the core version) to the most recent version before merging this.

morten-krogh: I will do that.

PR #631, opened 2018-02-06
Add an `apply_ordering` method to `Results`

ironage:
This is a more generic method which will help the js binding easily integrate the parsed sort/distinct descriptors.

### DIFF
--- a/src/results.cpp
+++ b/src/results.cpp
@@ -612,6 +612,24 @@ Results Results::filter(Query&& q) const
     return Results(m_realm, get_query().and_query(std::move(q)), m_descriptor_ordering);
 }
 
+Results Results::apply_ordering(DescriptorOrdering&& ordering)
+{
+    DescriptorOrdering new_order = m_descriptor_ordering;
+    for (size_t i = 0; i < ordering.size(); ++i) {
+        const CommonDescriptor* desc = ordering[i];
+        if (const SortDescriptor* sort = dynamic_cast<const SortDescriptor*>(desc)) {
+            new_order.append_sort(*sort);
+            continue;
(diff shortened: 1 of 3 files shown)


src/results.cpp
```diff
@@ -612,6 +612,24 @@ Results Results::filter(Query&& q) const
     return Results(m_realm, get_query().and_query(std::move(q)), m_descriptor_ordering);
 }
 
+Results Results::apply_ordering(DescriptorOrdering&& ordering)
+{
+    DescriptorOrdering new_order = m_descriptor_ordering;
+    for (size_t i = 0; i < ordering.size(); ++i) {
+        const CommonDescriptor* desc = ordering[i];
+        if (const SortDescriptor* sort = dynamic_cast<const SortDescriptor*>(desc)) {
+            new_order.append_sort(*sort);
```
bdash: `std::move(*sort)` to avoid a copy?

src/results.cpp
```diff
@@ -612,6 +612,24 @@ Results Results::filter(Query&& q) const
...
+            continue;
+        }
+        if (const DistinctDescriptor* distinct = dynamic_cast<const DistinctDescriptor*>(desc)) {
+            new_order.append_distinct(*distinct);
```
bdash: `std::move(*distinct)` to avoid a copy?

ZComwiz:
@ironage and @bdash — please see this crash: https://github.com/realm/realm-js/issues/3562

`m_table_view = m_query.find_all(m_descriptor_ordering);` crashes. 

Any advice on where to look in our source? 
It seems related to this commit possibly.

PR #632, opened 2018-02-06
Add support for Object Level Permissions

cmelchior:
Still work in progress. Just here to facilitate feedback


src/shared_realm.hpp
```diff
@@ -129,6 +134,18 @@ enum class SchemaMode : uint8_t {
...
+    None = 0,
+    Read = (1 << 0),
+    Update = (1 << 1),
+    Delete = (1 << 2),
+    SetPermissions = (1 << 3),
+    Query = (1 << 4),
+    Create = (1 << 5),
+    ModifySchema = (1 << 6),
+    All = 0xFF
+};
```
simonask: Would there be anything wrong with just using the `Privilege` enum from Sync?

src/shared_realm.cpp
```diff
@@ -921,6 +941,53 @@ template Object Realm::resolve_thread_safe_reference(ThreadSafeReference<Object>
...
+{
+    if (!init_permission_cache())
+        return ComputedPrivileges::All;
+    return static_cast<ComputedPrivileges>(m_permissions_cache->get_realm_privileges());
+}
+ComputedPrivileges Realm::get_privileges(StringData object_type)
+{
+    if (!init_permission_cache())
+        return ComputedPrivileges::All;
+    return static_cast<ComputedPrivileges>(m_permissions_cache->get_realm_privileges() & m_permissions_cache->get_class_privileges(object_type));
```
cmelchior: After talking to @simonask we should change this to use `can(Privilegege, obj)` where `obj` is the entry in the `__Class` table for the class in question.

We should use a similar pattern for Realm and Object privileges.

src/shared_realm.cpp
```diff
@@ -921,6 +941,53 @@ template Object Realm::resolve_thread_safe_reference(ThreadSafeReference<Object>
...
+ComputedPrivileges Realm::get_privileges(RowExpr row)
+{
+    if (!init_permission_cache())
+        return ComputedPrivileges::All;
+    auto& table = *row.get_table();
+    auto object_type = ObjectStore::object_type_for_table_name(table.get_name());
+    sync::GlobalID global_id{object_type, sync::object_id_for_row(read_group(), table, row.get_index())};
+    auto privileges = m_permissions_cache->get_realm_privileges()
+                    & m_permissions_cache->get_class_privileges(object_type)
+                    & m_permissions_cache->get_object_privileges(global_id);
```
simonask: This looks like a misunderstanding of the underlying semantics. (Though understandable, because this is how it worked in the very initial implementation a few weeks ago.)

Privileges from different levels can't be and'ed together like this, because they don't mean the same things. `Delete` at the object level only requires `Update` at the class and Realm levels, for example.

`Delete` at the class level means the permission to delete the class itself, not the permission to delete objects of that class.

The method `PermissionsCache::can()` was added in order to encapsulate the multilevel logic — however, it only returns a boolean, not the entire bitfield. I'm confused, though, because we went away from using bitfields in Sync due to the inconvenience, but this is turned booleans back into a bitfield?

cmelchior: It was, and still is, inconvenient if a bitfield is used in the `__Permission` objects because it makes that object un-queryable, but for the purpose of transporting the calculated permissions to the bindings for exposure in the public API, a bitfield is just as easy to use as exposing specific boolean getters.

src/shared_realm.hpp
tgoyne: This abstracts away the difference between sync and non-sync builds so that bindings don't have to worry about that.

src/shared_realm.cpp
tgoyne: Bit fields are only problematic as properties of realm objects, as neither core nor the bindings have proper support for them there.

src/shared_realm.cpp
```diff
@@ -931,6 +950,84 @@ template Object Realm::resolve_thread_safe_reference(ThreadSafeReference<Object>
...
 template Results Realm::resolve_thread_safe_reference(ThreadSafeReference<Results> reference);
 
+#if REALM_ENABLE_SYNC
+static_assert(static_cast<int>(ComputedPrivileges::Read) == static_cast<int>(sync::Privilege::Read), "");
+static_assert(static_cast<int>(ComputedPrivileges::Update) == static_cast<int>(sync::Privilege::Update), "");
+static_assert(static_cast<int>(ComputedPrivileges::Delete) == static_cast<int>(sync::Privilege::Delete), "");
+static_assert(static_cast<int>(ComputedPrivileges::SetPermissions) == static_cast<int>(sync::Privilege::SetPermissions), "");
+static_assert(static_cast<int>(ComputedPrivileges::Query) == static_cast<int>(sync::Privilege::Query), "");
+static_assert(static_cast<int>(ComputedPrivileges::Create) == static_cast<int>(sync::Privilege::Create), "");
+static_assert(static_cast<int>(ComputedPrivileges::ModifySchema) == static_cast<int>(sync::Privilege::ModifySchema), "");
```
simonask: 😒

src/shared_realm.cpp
```diff
@@ -931,6 +950,84 @@ template Object Realm::resolve_thread_safe_reference(ThreadSafeReference<Object>
...
+{
+    if (!init_permission_cache())
+        return ComputedPrivileges::All;
+
+    auto& table = *row.get_table();
+    auto object_type = ObjectStore::object_type_for_table_name(table.get_name());
+    sync::GlobalID global_id{object_type, sync::object_id_for_row(read_group(), table, row.get_index())};
+    auto privileges = inherited_mask(m_permissions_cache->get_realm_privileges())
+                    & inherited_mask(m_permissions_cache->get_class_privileges(object_type))
+                    & m_permissions_cache->get_object_privileges(global_id);
```
simonask: Why not just use `PermissionsCache::can()`?

tests/util/test_file.cpp
```diff
@@ -101,9 +100,25 @@ SyncTestFile::SyncTestFile(SyncServer& server,
...
 
-    sync_config = std::make_shared<SyncConfig>(SyncManager::shared().get_user({ "user", url }, "not_a_real_token"), url);
+    sync_config = std::make_shared<SyncConfig>(SyncManager::shared().get_user({ "test", url }, "not_a_real_token"), url);
+    sync_config->user->set_is_admin(true);
     sync_config->stop_policy = SyncSessionStopPolicy::Immediately;
-    sync_config->bind_session_handler = [=](auto&, auto& config, auto session) { session->refresh_access_token(s_test_token, config.realm_url()); };
+    sync_config->bind_session_handler = [=](auto&, auto& config, auto session) {
+        std::string token, encoded;
+        // Tokens without a path are considered admin tokens by the sync service, so for
+        // non-admin users we need to add a path
```
simonask: Add a FIXME about the fact that this is likely to change within the next few weeks.

tests/util/test_file.hpp
```diff
@@ -85,9 +86,11 @@ struct TestLogger : realm::util::Logger::LevelThreshold, realm::util::Logger {
     static realm::sync::Server::Config server_config();
 };
 
+using StartImmediately = realm::util::TaggedBool<class StartImmediatelyTag>;
```
simonask: Huh, this is a fun idiom! We should be using it all over the place.

src/shared_realm.cpp
cmelchior: I pushed this implementation last week, but looks like it got overridden by a force push: https://github.com/realm/realm-object-store/pull/610/files#diff-55a15b0c9125ec5f7989470e6eb8b217R1018

simonask: Damn. I wish we were a little less laissez faire about rebasing and force pushing. 🙁

PR #633, opened 2018-02-12
Fetch the client's UUID when configuring the SyncManager's file system support

bdash:
This removes the need to open and access a Realm file from within `SyncManager::client_uuid()`, which was problematic as it can be called in the process of opening a Realm file. This is both surprising, and had the potential to cause a deadlock on some platforms.

This fixes the hang I was hitting in #628.

### DIFF
--- a/src/sync/sync_manager.cpp
+++ b/src/sync/sync_manager.cpp
@@ -89,6 +89,8 @@ void SyncManager::configure_file_system(const std::string& base_file_path,
         }
 
         REALM_ASSERT(m_metadata_manager);
+        m_client_uuid = m_metadata_manager->client_uuid();
+
         // Perform any necessary file actions.
         std::vector<SyncFileActionMetadata> completed_actions;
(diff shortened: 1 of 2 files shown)


bdash:
There were multiple locks involved.

On the main thread, `RealmCoordinator::m_realm_mutex` was held by `get_realm`, which ends up using the client UUID as part of constructing the URL of the partially synced Realm. This calls back into `RealmCoordinator::get_realm` as part of opening the sync metadata Realm, and goes on to create an instance of `ExternalCommitHelper`. The `epoll` implementation of `ExternalCommitHelper` then registers itself with the daemon thread, which takes its `m_helpers` mutex.

Simultaneously, `ExternalCommitHelper` is processing an event on the daemon thread. It takes the `m_helpers` mutex, then goes on to call the coordinator's `on_change` function, which attempts to acquire `m_realm_mutex`.

bdash:
I was able to reproduce a crash locally that looks to be the same as what CI saw: https://github.com/realm/realm-sync/issues/1993. It happens on master, so I won't hold off on merging this PR.

PR #634, opened 2018-02-15
Add support for callbacks about start/end of change notification delivery

cmelchior:
This PR extends the `BindingContext` with support for notifying about start/end of notification delivery. This is required if bindings want to group several notifications into one, e.g. in the case of Partial Sync where Subscription and Collection Notifications needs to be combined in Java.

I'm concerned that notifications seem to be delivered in multiple places with hard to discern rules, but I tried to cover all places where `notifiers.deliver()` were being called.


tests/results.cpp
```diff
@@ -2765,3 +2765,159 @@ TEMPLATE_TEST_CASE("results: aggregate", ResultsFromTable, ResultsFromQuery, Res
...
+        };
+        r->m_binding_context.reset(new Context());
+        coordinator->on_change();
+        r->begin_transaction();
+        REQUIRE(binding_context_start_notify_calls == 1);
+        REQUIRE(binding_context_end_notify_calls == 1);
+        r->cancel_transaction();
+    }
+
+    SECTION("notify BindingContext if no callbacks have been registered - remote commit") {
```
tgoyne: I don't see any difference between this test and the one above it.

tests/results.cpp
```diff
@@ -2765,3 +2765,159 @@ TEMPLATE_TEST_CASE("results: aggregate", ResultsFromTable, ResultsFromQuery, Res
         }
     }
 }
+
+
+
+TEST_CASE("BindingContext is notified about delivery of change notifications") {
```
tgoyne: This test case should be in realm.cpp as that's where the the other binding context tests are.

tests/results.cpp
```diff
@@ -2765,3 +2765,159 @@ TEMPLATE_TEST_CASE("results: aggregate", ResultsFromTable, ResultsFromQuery, Res
...
+        r->m_binding_context.reset(new Context());
+        coordinator->on_change();
+        r->begin_transaction();
+        table->add_empty_row();
+        r->commit_transaction();
+        REQUIRE(binding_context_start_notify_calls == 1);
+        REQUIRE(binding_context_end_notify_calls == 1);
+    }
+
+    SECTION("notify BindingContext before and after sending notifications from remote commits") {
```
tgoyne: There's a lot of duplicated code between these two tests which could be eliminated by just pulling it out of the SECTION().

cmelchior:
Tests have been moved and refactored.
I looked into moving the logic to `NotifierPackage::before_advance/after_advance`. But there doesn't seem to be a contract that all 3 are called always. At least in `transact_log_handler.cpp`, `before_advance` might not be called and in `process_available_async()` `deliver` might not be called.

Unless I'm missing something it doesn't seem possible to inline the logic in the `NotifierPackage`?

src/impl/realm_coordinator.cpp
```diff
@@ -860,6 +864,9 @@ void RealmCoordinator::process_available_async(Realm& realm)
     if (notifiers.empty())
         return;
 
+    if (realm.m_binding_context)
+        realm.m_binding_context->will_send_notifications();
```
tgoyne: The error case right below this is missing a call to did_send_notifications()

cmelchior: Fixed

PR #635, opened 2018-02-16
Testing that sort/distinct orderings are applied to partial sync

ironage:
The description of sort/distinct needs to be serialised from the descriptor orderings that are stored in the Results class.

The strange thing is that the added test actually works with the changes commented out. Am I missing something here? How can I test this change?

### DIFF
--- a/src/sync/partial_sync.cpp
+++ b/src/sync/partial_sync.cpp
@@ -482,7 +482,7 @@ Subscription subscribe(Results const& results, util::Optional<std::string> user_
         throw std::logic_error("A partial sync query can only be registered in a partially synced Realm");
 
     auto query = results.get_query().get_description(); // Throws if the query cannot be serialized.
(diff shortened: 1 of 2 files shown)


bdash:
The test works because the `Results` you're looking at is the local instance you created, so the sorting and distincting is simply being performed locally on the results of the query. You could verify whether the distinct operation is being applied by the server by testing that only a subset of the objects are making it into the table. You could verify that the sorting is being applied by the server by looking at a `Results` returned by `SyncSubscription::results` instead of the one you created directly. That `Results` is backed by the `Foo_matches` `LinkList` that the server populates.

ironage:
Thanks for the explanation! I couldn't get `Subscription::results()` working (probably my misunderstanding of the API) but querying the partial realm table directly works for me. Ready for review.

bdash:
It may also just be broken… it's not tested yet, as there's no real reason for bindings to use it yet.

PR #636, opened 2018-02-20
Validate that a given type appears at most once in the schema

bdash:
Fixes the underlying issue identified in https://github.com/realm/realm-cocoa-private/issues/382.


ianpward:
@kneth Did this get included in the latest release of JS?

tgoyne:
No, realm-js has not pulled in a new object store version for a while.

PR #643, opened 2018-03-12
Adding option to allow __partial in URLs.

kneth:
See https://github.com/realm/realm-studio/issues/658


tgoyne:
The `disallow_partial_urls` member variable is never used outside of the constructor and thus doesn't need to exist. The new constructor doesn't particularly need to exist either; you could instead bypass the validation by just setting the url afterwards rather than passing it to the constructor.

kneth:
@tgoyne Thanks for the advice.

PR #648, opened 2018-03-26
Make sure to refetch the schema in case the matches property is missing

nirinchev:
In the .NET implementation, we hit a case where we try to get the subscription results and an `InvalidPropertyException` is thrown because `m_object_schema->property_for_name` returns `null`. I'm not sure how the other SDKs get around this issue, but browsing the code, it seems like the object schema is locked to the point in time the subscription is created and if at that point the matches property isn't already present, `results()` will always throw.

### DIFF
--- a/src/sync/partial_sync.cpp
+++ b/src/sync/partial_sync.cpp
@@ -592,7 +592,7 @@ util::Optional<Object> Subscription::result_set_object() const
     if (m_notifier->state() == Notifier::Complete) {
         if (auto row = m_result_sets.first())
             return Object(m_result_sets.get_realm(), m_object_schema, *row);
-    }
+        }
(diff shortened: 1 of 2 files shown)


tgoyne:
The `matches_property` property is created when the Realm is opened in partial sync mode, so it should always either exist well before a Subscription can be created, or never exist as there's nothing else that creates it (other than creating a subscription with the old partial sync api).

nirinchev:
It's the property with the name stored in `matches_property` that is missing. e.g. `Person_matches` or `Dog_matches`, etc. - sorry for the confusing description.

nirinchev:
We're dropping QBS for now, so that's no longer relevant.

PR #649, opened 2018-03-27
Disable change notifications for internal transient realms

fealebenpae:
A customer using partial sync and thread-safe references with .NET on Android reported a deadlock in a peculiar scenario.

Basically he had a partial realm where he was writing objects that didn't match any of the subscriptions and so the server was deleting them from the partial realm.

On Linux (and Android) this causes a lot of churn in `ExternalCommitHelper` because of the constant opening and closing of realms due to `GetInstance` and `ResolveReference` calls coupled with the sync client integrating frequent changesets. This apparently causes `ExternalCommitHelper::DaemonThread::m_mutex` to deadlock.

The solution is to disable automatic change notifications for realms that don't need them so that `ExternalCommitHelper` instances don't get created needlessly.

The .NET reproduction follows.

```csharp
public class Person : RealmObject
{
    public string Name { get; set; }

    [Backlink(nameof(Dog.Owner))]
    public IQueryable<Dog> Dogs { get; }
}

public class Dog : RealmObject
{
    public string Name { get; set; }

    public Person Owner { get; set; }
}

async Task Repro()
{
    var config = new SyncConfiguration(user, serverUri)
    {
        IsPartial = true,
        ObjectClasses = new[] { typeof(Person), typeof(Dog) }
    };

    var person = new Person { Name = "Пешо" };
    var realm = Realm.GetInstance(config);
    var query = await realm.SubscribeToObjectsAsync<Person>("Name == 'Пешо'");

    realm.Write(() => realm.Add(person));

    var thread = new AsyncContextThread();
    await Enumerable.Range(0, 100).Select(i =>
    {
        var @ref = ThreadSafeReference.Create(person);
        return thread.Factory.Run(() =>
        {
            using (var realm = Realm.GetInstance(config))
            {
                var @object = realm.ResolveReference(@ref);
                realm.Write(() =>
                {
                    realm.Add(new Dog { Name = $"Шаро №{i + 1}", Owner = @object });
                });
            }
        });
    }).WhenAll();
}
```

### DIFF
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -961,6 +961,7 @@ T Realm::resolve_thread_safe_reference(ThreadSafeReference<T> reference)
         if (reference_version < current_version) {
             // Duplicate config for uncached Realm so we don't advance the user's Realm
             Realm::Config config = m_coordinator->get_config();
+            config.automatic_change_notifications = false;
             config.cache = false;
             config.schema = util::none;
             SharedRealm temporary_realm = m_coordinator->get_realm(config);
--- a/src/sync/impl/sync_metadata.cpp
+++ b/src/sync/impl/sync_metadata.cpp
@@ -87,6 +87,7 @@ SyncMetadataManager::SyncMetadataManager(std::string path,
     constexpr uint64_t SCHEMA_VERSION = 2;
 
     Realm::Config config;
+    config.automatic_change_notifications = false;
     config.path = path;
     config.schema = make_schema();
     config.schema_version = SCHEMA_VERSION;
--- a/src/sync/impl/sync_metadata.cpp
+++ b/src/sync/impl/sync_metadata.cpp
@@ -87,7 +87,6 @@ SyncMetadataManager::SyncMetadataManager(std::string path,
     constexpr uint64_t SCHEMA_VERSION = 2;
 
     Realm::Config config;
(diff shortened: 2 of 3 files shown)


tgoyne:
The ThreadSafeReference change is definitely fine, but I'm not totally confident that SyncMetadataManager will work correctly without autorefresh.

fealebenpae:
As far as I can tell all the methods on `SyncMetadataManager` open the metadata realm as a local variable. But `SyncUserMetadata` holds on to a realm instance - I suppose I can re-enable automatic change notifications for the realm it's created with as it's a distinct instance from `SyncMetadataManager`'s.

It was the `client_uuid()` method that caused the issue in this particular test, because it was polled every time the partial realm was opened.

tgoyne:
It's opening the realm with caching enabled, so there's no guarantee that it's actually getting a fresh instance. That said, I'm not very familiar with the workings of the sync metadata realm so I have no idea if it'll ever actually get a cached one. If in practice it's never using a cached one, then disabling both change notifications and caching would be safe. If it is actually sometimes using the cache, then adding an explicit `refresh()` call to each of the read function would suffice.

tgoyne:
For `client_uuid()` specifically it's probably just worth storing the value on an instance variable since it appears to never actually change after it's set, rather than reopening the Realm to read it every time.

PR #654, opened 2018-05-25
Generate CMake files supporting all targets in realm-core

jedelbo:
Allows you to build all realm-core targets without regenerating
CMake files.


jedelbo:
Abandonned

PR #657, opened 2018-06-04
Enabling some more tests

jedelbo:
tests/realm.cpp
tests/thread_safe_reference.cpp

### DIFF
--- a/tests/realm.cpp
+++ b/tests/realm.cpp
@@ -31,15 +31,14 @@
 
 #include "impl/realm_coordinator.hpp"
 
-#include <realm/group.hpp>
+#include <realm/db.hpp>
 
-#if 0
 namespace realm {
 class TestHelper {
 public:
-    static SharedGroup& get_shared_group(SharedRealm const& shared_realm)
+    static DBRef& get_db(SharedRealm const& shared_realm)
     {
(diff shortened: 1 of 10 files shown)


tests/realm.cpp
```diff
@@ -1109,7 +1103,7 @@ TEST_CASE("SharedRealm: SchemaChangedFunction") {
             };
             r1->update_schema(new_schema, 2);
             REQUIRE(schema_changed_called == 1);
-            REQUIRE(changed_fixed_schema.find("object3")->property_for_name("value")->table_column == 0);
+            // REQUIRE(changed_fixed_schema.find("object3")->property_for_name("value")->table_column == 0);
```
tgoyne: These should probably all just check that `column_key` is set to an actual value (without caring what that value is).

tests/realm.cpp
```diff
@@ -493,20 +489,20 @@ TEST_CASE("SharedRealm: schema updating from external changes") {
         }},
     };
 
-    #if 0
     SECTION("newly added columns update table columns but are not added to properties") {
+        // Does this test add any value when column keys are stable?
```
tgoyne: Testing that new columns aren't added to `persisted_properties` in non-dynamic schema modes is still useful. That might be sufficiently covered elsewhere?

PR #660, opened 2018-06-08
Expose a state change listener for SyncSession

cmelchior:
TODO
- [ ] Tests
- [ ] Cleanup


tests/sync/session/state_change_notifications.cpp
```diff
@@ -0,0 +1,91 @@
...
+        std::atomic<bool> listener_called(false);
+        auto token = session->register_state_change_callback([&](SyncSession::PublicState, SyncSession::PublicState) {
+            listener_called = true;
+        });
+
+        // Logging the user out should deactivate the session.
+        EventLoop::main().run_until([&] { return sessions_are_active(*session); });
+        user->log_out();
+        EventLoop::main().run_until([&] { return sessions_are_inactive(*session); });
+        REQUIRE(listener_called == true);
```
kneth: You only test if the callback is called at least once - is that on purpose? Maybe it is better to test for a precise number of calls?

src/sync/sync_session.cpp
```diff
@@ -794,6 +833,24 @@ SyncSession::PublicState SyncSession::state() const
...
+
+SyncSession::PublicState SyncSession::state() const
+{
+    std::unique_lock<std::mutex> lock(m_state_mutex);
+    return get_public_state();
+}
+
+SyncSession::PublicConnectionState SyncSession::connectionState() const
+{
+    switch(m_connection_state) {
```
kneth: Please add space between `h` and `(`.

src/sync/sync_session.cpp
```diff
@@ -935,3 +981,48 @@ std::function<void()> SyncProgressNotifier::NotifierPackage::create_invocation(P
     is_expired = !is_streaming && transferred >= transferrable;
     return [=, notifier=notifier] { notifier(transferred, transferrable); };
 }
+
+uint64_t SyncSession::ConnectionChangeNotifier::add_callback(std::function<ConnectionStateCallback> callback)
+{
+    std::lock_guard<std::mutex> lock(m_callback_mutex);
+    auto token = m_next_token++;
+    m_callbacks.push_back({std::move(callback), token});
+    return token;}
```
tgoyne: Missing a newline

PR #663, opened 2018-06-13
Je/test adaptation

jedelbo:


jedelbo:
@tgoyne There was one list test ("deleting a target row sends a change notification") that I did not manage to get going. Perhaps you could have a look.

PR #664, opened 2018-06-14
Skip some pointless background work for LinkingObjects notifiers

tgoyne:
See https://github.com/realm/realm-cocoa/issues/5816. This fixes the problem in two ways: it skips implicitly creating a notifier at all for LinkingObjects, and skips calculating the change sets for Results which do need a background notifier but don't have any callbacks.

No new tests as this should have no functional impact other than being faster. The obj-c tests do attempt to verify that the implicit notifier is created in the places where we actually want it (and those tests still pass).


bmunkholm:
Can't see anything glaringly wrong :-) outside of adding a changelog note about the performance improvement.

PR #665, opened 2018-06-19
Speed up compact

ironage:
Since we do the optimisation in a read transaction and never commit, it's wasted cpu cycles. LOL.
We were going to put this in a write transaction (https://github.com/realm/realm-object-store/pull/501 ) but we never got the go ahead because we kept finding problems in the enum columns by fuzz testing.  I think we decided to just enable them in core 6 so that they didn't have to be upgraded from 5->6.

### DIFF
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -727,11 +727,11 @@ bool Realm::compact()
         throw InvalidTransactionException("Can't compact a Realm within a write transaction");
     }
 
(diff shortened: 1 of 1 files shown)


tgoyne:
We do still need the call to end_read() if m_group is set.

tgoyne:
The crashing test is also indicating that we need a call to `verify_open()` (which used to happen inside `read_group()`).

PR #667, opened 2018-06-26
Make more List and Results tests work

jedelbo:
The tests not yet passing are commented out. They should - however - be able to compile.

### DIFF
--- a/src/list.cpp
+++ b/src/list.cpp
@@ -127,10 +127,15 @@ static StringData object_name(Table const& table)
 ObjectSchema const& List::get_object_schema() const
 {
     verify_attached();
-    REALM_ASSERT(get_type() == PropertyType::Object);
 
     if (!m_object_schema) {
-        auto object_type = object_name(static_cast<LnkLst&>(*m_list_base).get_target_table());
+        StringData object_type;
+        if (get_type() == PropertyType::Object) {
+            object_type = object_name(static_cast<LnkLst&>(*m_list_base).get_target_table());
+        }
+        else {
+            object_type = object_name(*m_list_base->get_table());
+        }
         auto it = m_realm->schema().find(object_type);
         REALM_ASSERT(it != m_realm->schema().end());
         m_object_schema = &*it;
@@ -364,7 +369,7 @@ Results List::filter(Query q) const
 Results List::as_results() const
 {
     verify_attached();
-    return Results(m_realm, std::dynamic_pointer_cast<LnkLst>(m_list_base));
+    return (m_type == PropertyType::Object) ? Results(m_realm, std::static_pointer_cast<LnkLst>(m_list_base)) : Results(m_realm, m_list_base);
 }
 
 Results List::snapshot() const
--- a/src/list.cpp
+++ b/src/list.cpp
@@ -88,18 +88,34 @@ List& List::operator=(const List&) = default;
 List::List(List&&) = default;
 List& List::operator=(List&&) = default;
 
-List::List(std::shared_ptr<Realm> r, Obj& parent_obj, ColKey col)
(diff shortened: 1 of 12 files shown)


jedelbo:
@tgoyne Perhaps this would be a good time to review this PR. I have made some changes to the implementation that I would like you to evaluate.

src/results.cpp
```diff
@@ -47,47 +47,51 @@ Results::Results(SharedRealm r, Table& table)
...
 
 bool Results::is_valid() const
 {
     if (m_realm) {
         m_realm->verify_thread();
     }
 
-//    if (m_table && !m_table->is_valid())
-//        return false;
+    if (((const Table*)m_table) != nullptr)
```
tgoyne: This is an awfully longwinded way to write `if (m_table)`

src/results.cpp
```diff
@@ -47,47 +47,51 @@ Results::Results(SharedRealm r, Table& table)
...
+        return (bool)m_table;
```
tgoyne: This codebase needs to be `-Wold-style-cast` clean.

src/results.cpp
```diff
@@ -230,11 +242,12 @@ void Results::evaluate_query_if_needed(bool wants_notifications)
         return;
     }
 
-    if (m_notifier)
+    // if (m_notifier) FIXME: Why was this condition added?
```
tgoyne: This looks like it was just a stray paste and should be deleted.

src/results.cpp
```diff
@@ -488,6 +503,9 @@ Query Results::get_query() const
             if (query.get_table()) {
                 return query;
             }
+            if (m_table_view.is_backlink_view()) {
+                return Query();
```
tgoyne: This does need to return a real query as we support filtering and sorting backlink views.

src/results.hpp
```diff
@@ -295,6 +301,28 @@ size_t Results::index_of(Context& ctx, T value)
 {
     return dispatch([&](auto t) { return this->index_of(ctx.template unbox<std::decay_t<decltype(*t)>>(value)); });
 }
+
+template<typename T>
+auto Results::list_get_as(size_t ndx) const
```
tgoyne: There doesn't appear to be any reason for these to be defined in the header.

tests/list.cpp
```diff
@@ -425,6 +425,8 @@ TEST_CASE("list") {
             REQUIRE_MOVES(change, {1, 2});
         }
 
+#if 0
+        // FIXME irrelevant?
```
tgoyne: Yes, everything related to moving rows in tables should just be deleted.

jedelbo: Deleted

src/results.hpp
jedelbo: Moved to .cpp

src/results.cpp
jedelbo: Done

src/results.cpp
jedelbo: Fixed

src/results.cpp
jedelbo: Explanation added

src/results.cpp
jedelbo: I realize that this was just a cover for underlying problems. Made some fixes in core (setting `m_view` when importing) and results_notifier.cpp (setting of `m_results_were_used`)

src/list.cpp
```diff
@@ -111,10 +127,15 @@ static StringData object_name(Table const& table)
 ObjectSchema const& List::get_object_schema() const
 {
     verify_attached();
-    REALM_ASSERT(get_type() == PropertyType::Object);
```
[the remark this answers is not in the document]
jedelbo: Reverted

PR #676, opened 2018-07-18
Je/list test

jedelbo:


tests/util/index_helpers.hpp
```diff
@@ -28,6 +28,18 @@
     } \
 } while (0)
 
+#define REQUIRE_KEYS(index_set, ...) do { \
```
tgoyne: Change notifications are supposed to report indices in the collection that changed, not object keys. If they are currently reporting keys in some cases that's just a bug that needs to be fixed.

jedelbo: Got it!

jedelbo:
@tgoyne commit rebased

PR #679, opened 2018-07-31
Add basic limit() support to Results

tgoyne:
Punts on the potentially harder or more complicated parts (further filtering after limited, notifications) as they aren't needed for the immediate use-case of limiting sync subscriptions.


tgoyne:
Will fail on CI as it depends on https://github.com/realm/realm-core/pull/3079.

src/results.cpp
```diff
@@ -259,8 +259,12 @@ void Results::evaluate_query_if_needed(bool wants_notifications)
...
             m_query.sync_view_if_needed();
-            m_table_view = m_query.find_all();
-            if (!m_descriptor_ordering.is_empty()) {
+            if (m_descriptor_ordering.is_empty()) {
+                auto limit = m_descriptor_ordering.has_limit() ? m_descriptor_ordering.get_limit() : size_t(-1);
+                m_table_view = m_query.find_all(0, -1, limit);
+            }
+            else {
+                m_table_view = m_query.find_all();
                 m_table_view.apply_descriptor_ordering(m_descriptor_ordering);
```
ironage: It looks like having versions of `Query::count` and `Query::find_all` which take a `DescriptorOrdering` would be useful. Having the latter case would also remove the inefficiency of having to run the query twice here, once for `find_all` and having to sync it again in `apply_descriptor_ordering`.

src/results.cpp
```diff
@@ -610,23 +607,31 @@ Results Results::sort(SortDescriptor&& sort) const
 
 Results Results::filter(Query&& q) const
 {
+    if (m_descriptor_ordering.will_apply_limit())
+        throw UnimplementedOperationException("Filtering a Results with a limit is not yet implemented");
```
ironage: We could probably support this by moving the limit into the query itself in core, or emulating ordered behaviour at this level. It is most likely a rare use case though.

src/results.cpp
```diff
@@ -726,6 +731,8 @@ void Results::prepare_async(ForCallback force)
 
 NotificationToken Results::add_notification_callback(CollectionChangeCallback cb) &
 {
+    if (m_descriptor_ordering.will_apply_limit())
+        throw UnimplementedOperationException("Change notifications for Results with a limit are not yet implemented");
```
ironage: I don't know the notification code at all, but I'm a bit surprised to see this will need special handling.

nirinchev:
@tgoyne do you feel this could be merged now or do you prefer to address @ironage's comments? I'd like to update the .NET SDK to use Core 5.8.0, so was wondering if I should go ahead and create a new branch that just fixes the compile errors or wait for this PR to get merged, as it addresses them anyway.

tgoyne:
I was just waiting for a sync release so that bindings can actually use it.

tgoyne:
I'm seeing the test failure that CI is hitting with 3.9.2 locally as well. Looking into that.

tgoyne:
Looks like https://github.com/realm/realm-sync/pull/2338 changed the behavior of access control so that running without a public key no longer works.

tgoyne:
https://github.com/realm/realm-sync/pull/2346 is needed for this.

PR #689, opened 2018-08-21
Rework ThreadSafeReference and restore support for Results

tgoyne:
This adjusts the API to expose just a type-erased wrapper as that's what the binding actually wants, and pulls all of the logic out of shared_realm.cpp as it no longer needs to muck about with the Realm's inner state in the way that it did with core 5. It also drops the restriction again handing over in write transactions, as that now works fine AFAICT.


jedelbo:
Started reviewing this, but will need a bit more time.

src/object.hpp
```diff
@@ -47,6 +47,7 @@ class Object {
     ~Object();
 
     std::shared_ptr<Realm> const& realm() const { return m_realm; }
```
jedelbo: This one seems not to be used

PR #691, opened 2018-08-30
Limit descriptor

jedelbo:


jedelbo:
Not needed. PR targeting master will implement this.

PR #693, opened 2018-09-04
Enable notifications for limited query results

cmelchior:
It seems that notifications for limited query results "just work" so re-enabling them and adding a smoke test.


tests/results.cpp
```diff
@@ -2826,10 +2827,41 @@ TEST_CASE("results: limit") {
...
+            ++notification_calls;
+        });
+        advance_and_notify(*realm);
+        REQUIRE(notification_calls == 1);
+        realm->begin_transaction();
+        table->add_empty_row(1);
+        table->set_int(0, 8, 5);
+        realm->commit_transaction();
+        coordinator->on_change();
+        advance_and_notify(*realm);
```
tgoyne: This needs to check that a notification was actually sent.

PR #699, opened 2018-10-09
Initial implementation of diff-only update of objects

jedelbo:
Lists will be compared if the length is preserved - otherwise rewritten.
Objects with primary key will always just be updated if they already exist in the database.
In the case of single link to object without primary key we compare each value individually.
Lists of objects without primary key wil not yet be compared.

### DIFF
--- a/src/list.hpp
+++ b/src/list.hpp
@@ -26,6 +26,7 @@
 #include <realm/link_view_fwd.hpp>
 #include <realm/row.hpp>
 #include <realm/table_ref.hpp>
+#include <realm/util/any.hpp>
 
 #include <functional>
 #include <memory>
@@ -129,10 +130,12 @@ public:
     void insert(Context&, size_t list_ndx, T&& value, bool update=false);
     template<typename T, typename Context>
     void set(Context&, size_t row_ndx, T&& value, bool update=false);
+    template<typename T, typename Context>
+    void set_if_different(Context&, size_t row_ndx, T&& value, bool update=false);
 
     // Replace the values in this list with the values from an enumerable object
     template<typename T, typename Context>
-    void assign(Context&, T&& value, bool update=false);
+    void assign(Context&, T&& value, bool update=false, bool update_only_diff = false);
 
     // The List object has been invalidated (due to the Realm being invalidated,
     // or the containing object being deleted)
@@ -204,16 +207,57 @@ void List::set(Context& ctx, size_t row_ndx, T&& value, bool update)
     dispatch([&](auto t) { this->set(row_ndx, ctx.template unbox<std::decay_t<decltype(*t)>>(value, true, update)); });
 }
 
+template <class T>
+inline bool help_compare_values(const T& v1, const T& v2)
+{
+    return v1 != v2;
+}
+
+template <>
+inline bool help_compare_values(const RowExpr& v1, const RowExpr& v2)
+{
+    return v1.get_table() != v2.get_table() || v1.get_index() != v2.get_index();
+}
+
+
 template<typename T, typename Context>
-void List::assign(Context& ctx, T&& values, bool update)
+void List::set_if_different(Context& ctx, size_t row_ndx, T&& value, bool update)
+{
+    dispatch([&](auto t) {
+        using U = std::decay_t<decltype(*t)>;
+        auto old_value =  this->get<U>(row_ndx);
+        auto new_value = ctx.template unbox<U>(value, true, update, true);
+        if (help_compare_values(old_value, new_value))
+            this->set(row_ndx, new_value);
+    });
+}
+
+
+template<typename T, typename Context>
+void List::assign(Context& ctx, T&& values, bool update, bool update_only_diff)
 {
     if (ctx.is_same_list(*this, values))
         return;
 
-    remove_all();
-    if (ctx.is_null(values))
+    if (ctx.is_null(values)) {
+        remove_all();
         return;
+    }
 
+    if (update_only_diff) {
+        size_t sz = util::any_cast<std::vector<util::Any>&>(values).size();
+        // If size is the same, we can compare element by element
+        if (size() == sz) {
+            size_t index = 0;
+            ctx.enumerate_list(values, [&](auto&& element) {
+                this->set_if_different(ctx, index, element, update);
+                index++;
+            });
+            return;
+        }
+    }
+
+    remove_all();
     ctx.enumerate_list(values, [&](auto&& element) {
         this->add(ctx, element, update);
     });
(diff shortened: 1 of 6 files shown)


src/object_accessor.hpp
```diff
@@ -64,14 +64,14 @@ ValueType Object::get_property_value(ContextType& ctx, StringData prop_name)
...
-                                     ValueType value, bool try_update, bool is_default)
+                                     ValueType value, bool try_update, bool update_only_diff, bool is_default)
 {
     ctx.will_change(*this, property);
 
     auto& table = *m_row.get_table();
     size_t col = property.table_column;
     size_t row = m_row.get_index();
-    if (is_nullable(property.type) && ctx.is_null(value)) {
+    if (is_nullable(property.type) && ctx.is_null(value) && (!update_only_diff || !table.is_null(col, row))) {
```
tgoyne: The `update_only_diff` check needs to be pushed into a separate `if` inside the body of this or it'll continue on to the rest of the function.

src/object_accessor.hpp
```diff
@@ -98,37 +98,76 @@ void Object::set_property_value_impl(ContextType& ctx, const Property &property,
...
         case PropertyType::Object: {
             ContextType child_ctx(ctx, property);
-            auto link = child_ctx.template unbox<RowExpr>(value, true, try_update);
-            table.set_link(col, row, link.get_index(), is_default);
+            // Check if we are linking to an object already
+            auto curr_link = table.get_link(col,row);
+            auto link = child_ctx.template unbox<RowExpr>(value, true, try_update, update_only_diff);
+            if (!update_only_diff || curr_link != link.get_index()) {
+                table.set_link(col, row, link.get_index(), is_default);
+                ctx.did_change();
```
tgoyne: `ctx.will_change()` and `ctx.did_change()` calls always need to be paired. Either this needs to only call `will_change()` if we're actually setting the value or it needs to call `did_change()` even if we didn't.

src/object_accessor.hpp
```diff
@@ -98,37 +98,76 @@ void Object::set_property_value_impl(ContextType& ctx, const Property &property,
...
+            }
             break;
         }
-        case PropertyType::Bool:
-            table.set(col, row, ctx.template unbox<bool>(value), is_default);
+        case PropertyType::Bool: {
+            auto new_val = ctx.template unbox<bool>(value);
```
tgoyne: This is a lot of duplicated code for each case that looks pretty easy to extract to a function.

jedelbo: I thought about it, but was not really sure if it would be an improvement, but as you mention it, I will extract into a function.

src/object_accessor.hpp
jedelbo: OK. I was not aware of that. I will just call `did_change()` unconditionally.

src/object_accessor.hpp
jedelbo: Indeed. Thanks

src/object_accessor.hpp
bmunkholm: Do we have or should we have tests that would catch this?

tgoyne: This functionality is only tested at the Cocoa level since it's there to support KVO. Part of exposing this in the Cocoa binding will be to extend the existing KVO tests to use the new option.

cmelchior:
Will list support be added to this? Because I suspect that setting list elements will actually be the majority of SET instructions?

I think a simple algorithm that traverses the list and if it finds a change, it goes back to clearing/rewriting the whole list. At worst that should be O(2N) and in the happy path we reduce the transaction log by quite a lot.

Also, from testing in the Java binding it looks like single object references are always overriden as well?

jedelbo:
Added simple diffing on lists. Only done when size is the same.
Currently objects are always overwritten if there is no primary key.

src/list.hpp
```diff
@@ -204,16 +207,57 @@ void List::set(Context& ctx, size_t row_ndx, T&& value, bool update)
...
 
-    remove_all();
-    if (ctx.is_null(values))
+    if (ctx.is_null(values)) {
+        remove_all();
         return;
+    }
+
+    if (update_only_diff) {
+        size_t sz = util::any_cast<std::vector<util::Any>&>(values).size();
```
tgoyne: The concrete type of `values` is not `std::vector<util::Any>` in non-test code. The values passed around have to always be treated as opaque types which are only ever touched by passing them to context functions. The size may also not be known in advance, as we support assigning a non-container enumerator to a list. Fortunately we don't actually need the size in advance, and can do something like the following:

```
size_t sz = size();
size_t index = 0;
ctx.enumerate_list(values, [&](auto&& element) {
    if (index < sz)
        this->set_if_different(ctx, index, element, update);
    else
        this->add(ctx, element, update);
    index++;
});
if (index < sz)
    remove(index, sz);
```

src/object_accessor.hpp
```diff
@@ -90,41 +103,76 @@ void Object::set_property_value_impl(ContextType& ctx, const Property &property,
...
         ctx.did_change();
         return;
     }
 
     switch (property.type & ~PropertyType::Nullable) {
         case PropertyType::Object: {
-            ContextType child_ctx(ctx, property);
-            auto link = child_ctx.template unbox<RowExpr>(value, true, try_update);
-            table.set_link(col, row, link.get_index(), is_default);
+            auto linked_schema = m_realm->schema().find(property.object_type);
```
tgoyne: All of this code is just replicating what's already done by `unbox<RowExpr>()`. There shouldn't be any changes needed here other than forwarding on `update_only_diff` to `unbox()` and only calling `set_link` if it's changed.

src/list.hpp
jedelbo: Good idea. Thanks.

src/object_accessor.hpp
jedelbo: In the case where we link to objects with no primary key, we cannot just call `unbox<RowExpr>` with create flag set to true as this will unconditionally create a new object. I have rearranged the code to make it clearer.

tgoyne: This has to call `unbox()` because there is binding-specific functionality which this is bypassing, with incorrect results. In Cocoa, the linked object may be in the form of an array dictionary of values, an unmanaged object, or a managed objects. For a dictionary of values, this does the correct thing. For an unmanaged object, this will fail to convert the object to a managed object. For a managed object which is not the object which the link currently points to, this will overwrite the values in the linked object rather than changing the link.

I think what this needs to do is if `update_only_diff` is `true`, it should pass the current value of the link to `unbox()`, and the `unbox()` would use that row rather than creating a new object when applicable.

jedelbo: I am getting more and more insight into how the context concept is used :)
I have now created next attempt. A lot of parameters ...

jedelbo:
@tgoyne Can this PR be further improved or can we merge it now?

PR #700, opened 2018-10-10
Send whole property instead of just the property name

cmelchior:
Sending the whole property instead of just the property name enables more flexible logic in custom Context implementations. In my case, I'm looking for the `table_column` as an index into another data structure that stores the actual values.


tgoyne:
This change is fine, but the `property_index` parameter is already there specifically for looking up the equivalent thing in a parallel data structure.

cmelchior:
Yes, but the property_index is the index in the persisted_property array, and from what I could tell there is no guarantee that matches the equivalent table column index?

PR #698, opened 2018-10-09
Don't create __ResultSets

kneth:

PR #701, opened 2018-10-13
Add support for creating a subscription object immediately

cmelchior:
* This adds the capability of creating subscriptions immediately by using `subscribe_blocking()`. This is required by new additions to Java. 
* Removed old methods that were only used by Java, but are no longer relevant.
* Added support for specifying the __ResultSets class in the users Schema. This fixes: https://github.com/realm/realm-object-store/pull/698

TODO
- [x] Tests

### DIFF
--- a/src/object_store.cpp
+++ b/src/object_store.cpp
@@ -57,7 +57,7 @@ const size_t c_zeroRowIndex = 0;
 
 const char c_object_table_prefix[] = "class_";
 
-void create_metadata_tables(Group& group, bool initialize_subscriptions_table) {
+void create_metadata_tables(Group& group) {
     // The tables 'pk' and 'metadata' are treated specially by Sync. The 'pk' table
     // is populated by `sync::create_table` and friends, while the 'metadata' table
     // is simply ignored.
@@ -76,14 +76,6 @@ void create_metadata_tables(Group& group, bool initialize_subscriptions_table) {
         pk_table->insert_column(c_primaryKeyPropertyNameColumnIndex, type_String, c_primaryKeyPropertyNameColumnName);
     }
     pk_table->add_search_index(c_primaryKeyObjectClassColumnIndex);
-
-#if REALM_ENABLE_SYNC
-    // Only add __ResultSets if Realm is a partial Realm
-    if (initialize_subscriptions_table)
-        _impl::initialize_schema(group);
-#else
-    (void)initialize_subscriptions_table;
-#endif
 }
 
 void set_schema_version(Group& group, uint64_t version) {
@@ -258,7 +250,7 @@ void validate_primary_column_uniqueness(Group const& group)
 } // anonymous namespace
 
 void ObjectStore::set_schema_version(Group& group, uint64_t version) {
-    ::create_metadata_tables(group, false);
+    ::create_metadata_tables(group);
     ::set_schema_version(group, version);
 }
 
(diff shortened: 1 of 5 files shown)


src/object_store.cpp
```diff
@@ -786,7 +786,35 @@ void ObjectStore::apply_schema_changes(Group& group, uint64_t schema_version,
                                        util::Optional<std::string> sync_user_id,
                                        std::function<void()> migration_function)
 {
-    create_metadata_tables(group, sync_user_id != util::none);
+    // Temporary fix for https://github.com/realm/realm-object-store/pull/698
```
tgoyne: Rather than doing all this you can just remove the call to ` _impl::initialize_schema()` in `create_metadata_tables()` and add it after applying the changes from the user-specified schema.

src/sync/partial_sync.cpp
```diff
@@ -214,26 +200,61 @@ struct ResultSetsColumns {
     size_t matches_property;
 };
 
-bool validate_existing_subscription(Table& table, ResultSetsColumns const& columns, std::string const& name,
+// Validate the subscription about to be created against existing subscription.
+// If an existing subscription already exists that matches the one  we are about to create, the
+// index of that Subscription is returned. If no current matching subscription exists `npos` is
+// returned.
+static size_t validate_existing_subscription(Table& table, ResultSetsColumns const& columns, std::string const& name,
                                     std::string const& query, std::string const& matches_property)
```
tgoyne: A bunch of spots in this function update the first line of a multi-line function call without adjusting the indent of the following lines to match.

src/sync/partial_sync.cpp
```diff
@@ -214,26 +200,61 @@ struct ResultSetsColumns {
     size_t matches_property;
 };
 
-bool validate_existing_subscription(Table& table, ResultSetsColumns const& columns, std::string const& name,
+// Validate the subscription about to be created against existing subscription.
+// If an existing subscription already exists that matches the one  we are about to create, the
+// index of that Subscription is returned. If no current matching subscription exists `npos` is
+// returned.
+static size_t validate_existing_subscription(Table& table, ResultSetsColumns const& columns, std::string const& name,
```
tgoyne: `static` is redundant as this is within an anonymous namespace.

src/sync/partial_sync.cpp
```diff
@@ -214,26 +200,61 @@ struct ResultSetsColumns {
...
+}
+
+// Performs the logic of actually writing the subscription (if needed) to the Realm and making sure
+// that the `matches_property` field is setup correctly. This method will throw if the query cannot
+// be serialized or the name is already used by another subscription.
+//
+// The row of the resulting subscription is returned. If an old subscription exists that matches
+// the one about to be created, a new subscription is not created, but the old one is returned
+// instead.
+static RowExpr write_subscription(std::string const& object_type, std::string const& name, std::string const& query, Group& group) {
```
tgoyne: `static` is redundant and needs newline before `{`.

src/sync/partial_sync.cpp
```diff
@@ -510,6 +440,25 @@ Subscription subscribe(Results const& results, util::Optional<std::string> user_
     return subscription;
 }
 
+RowExpr subscribe_blocking(Results const& results, util::Optional<std::string> user_provided_name) {
```
tgoyne: Newline before { for functions.

src/object_store.cpp
cmelchior: Ah, much better idea than this 👍

cmelchior:
Updated

tgoyne:
Merging this as I need the schema init changes for JS.

PR #702, opened 2018-10-16
Add a missing virtual destructor to CollectionChangeCallback

tgoyne:
This resulted in the destructor for the callback object passed to `add_notification_callback()` not being called, potentially resulting in memory leaks when removing notification callbacks.


tgoyne:
CI failure was due to that job running on a worker that's out of disk space.

PR #708, opened 2018-11-13
Support setting binding/application user agent

cmelchior:
It turned out that setting both the binding and application user-agent info in `configure` wasn't optimal for how JS uses ObjectStore, so in this refactor I split the two parts. The binding level info is still configured when calling `configure` as it doesn't involve user input. The application level info now has a method similar to how `set_log_level` works.

### DIFF
--- a/src/sync/sync_manager.cpp
+++ b/src/sync/sync_manager.cpp
@@ -515,8 +515,8 @@ std::unique_ptr<SyncClient> SyncManager::create_sync_client() const
         logger = std::move(stderr_logger);
     }
 
-    std::string user_agent(m_user_agent_binding_info + std::string(" ") + m_user_agent_application_info);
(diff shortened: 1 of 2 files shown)


src/sync/sync_manager.cpp
```diff
@@ -508,7 +514,9 @@ std::unique_ptr<SyncClient> SyncManager::create_sync_client() const
         stderr_logger->set_level_threshold(m_log_level);
         logger = std::move(stderr_logger);
     }
-    return std::make_unique<SyncClient>(std::move(logger), m_client_reconnect_mode, m_multiplex_sessions, m_user_agent_application_info);
+
+    std::string user_agent(m_user_agent_binding_info + std::string(" ") + m_user_agent_application_info);
+    return std::make_unique<SyncClient>(std::move(logger), m_client_reconnect_mode, m_multiplex_sessions, user_agent);
```
tgoyne: `util::format("%1 %2", m_user_agent_binding_info, m_user_agent_application_info)`

PR #709, opened 2018-11-13
Add hooks for the audit functionality

tgoyne:
This adds an abstract interface which is implemented by the actual audit implementation for reporting auditable events, along with some tweaks required to get client reset handling working.

### DIFF
--- a/tests/realm.cpp
+++ b/tests/realm.cpp
@@ -1580,7 +1580,7 @@ TEST_CASE("Statistics on Realms") {
     });
 
     SECTION("compute_size") {
-        auto s = r->compute_size();
+        auto s = r->read_group().compute_aggregated_byte_size();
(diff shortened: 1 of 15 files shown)


tests/realm.cpp
```diff
@@ -1580,7 +1580,7 @@ TEST_CASE("Statistics on Realms") {
     });
 
     SECTION("compute_size") {
-        auto s = r->compute_size();
+        auto s = r->read_group().compute_aggregated_byte_size();
```
jedelbo: You probably want to add `SizeAggregateControl::size_of_state`. Otherwise you will always get 0 in return.

tgoyne:
Some basic tests here would certainly be good to have, but for now all of the tests are integration tests of the full functionality.

PR #710, opened 2018-11-13
Fix compilation with sync 3.13.3

tgoyne:
3.13.3 made it so that we have to construct the TableInfoCache ourselves and pass in a reference to the PermissionsCache.

### DIFF
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -41,10 +41,12 @@
 
 #include <realm/sync/history.hpp>
 #include <realm/sync/permissions.hpp>
+#include <realm/sync/version.hpp>
 #else
 namespace realm {
 namespace sync {
     struct PermissionsCache {};
+    struct TableInfoCache {};
 }
 }
 #endif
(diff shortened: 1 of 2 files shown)


src/shared_realm.cpp
```diff
@@ -1020,7 +1024,13 @@ bool Realm::init_permission_cache()
 
     // Admin users bypass permissions checks outside of the logic in PermissionsCache
     if (m_config.sync_config && m_config.sync_config->is_partial && !m_config.sync_config->user->is_admin()) {
+#if REALM_SYNC_VER_MAJOR == 3 && (REALM_SYNC_VER_MINOR < 13 || (REALM_SYNC_VER_MINOR == 13 && REALM_SYNC_VER_PATCH < 3))
```
jedelbo: Why is the dependency on Sync not bumped to 3.13.3 so this can be avoided?

nirinchev: We generally try to make OS backward compatible.

jedelbo: 👍

PR #758, opened 2018-12-03
Allow unsubscribing from a query without a Subscription

tgoyne:
Needed for https://github.com/realm/realm-cocoa/pull/6029.

### DIFF
--- a/src/sync/partial_sync.cpp
+++ b/src/sync/partial_sync.cpp
@@ -33,13 +33,10 @@
 #include <realm/lang_bind_helper.hpp>
 #include <realm/util/scope_exit.hpp>
 
-namespace {
-constexpr const char* result_sets_type_name = "__ResultSets";
-}
-
 namespace realm {
 
 namespace _impl {
+using namespace ::realm::partial_sync;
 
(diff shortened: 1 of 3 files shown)


src/sync/partial_sync.hpp
```diff
@@ -107,6 +108,7 @@ RowExpr subscribe_blocking(Results const&, util::Optional<std::string> name);
 /// The operation is performed asynchronously. Completion will be indicated by the
 /// `Subscription` transitioning to the `Invalidated` state.
 void unsubscribe(Subscription&);
+void unsubscribe(Object&&);
```
cmelchior: Perhaps add a comment here that it needs to be a managed RealmObject of the type `__ResultSets`... at least I was a bit confused about it when just glancing over the code.

src/sync/partial_sync.cpp
```diff
@@ -507,6 +504,18 @@ void unsubscribe(Subscription& subscription)
     }
 }
 
+void unsubscribe(Object&& subscription)
```
cmelchior: Add a test showing this works?

PR #760, opened 2018-12-06
Import the global notifier and adapter from realm-js-private

tgoyne:
The first step for https://github.com/realm/realm-js-private/issues/462 and https://github.com/realm/realm-dotnet-server/issues/4.

This is just the six cpp/hpp files sliced out of realm-js-private and imported here with zero changes as reviewing those changes would be awkward. The next step after this will be to add build system support so that these files are at least compiled and to move the JS/node-specific portions of the GN and adapter to realm-js.


kneth:
We need to move https://github.com/realm/realm-js/blob/master/vendor/json.hpp too.

PR #762, opened 2018-12-21
Fix user operations on a deleted user

ironage:
The `SyncUserMetadata` was operating on a row accessor and using it's own locally cached `m_invalid` state to check for validity. This would only work if no other accessors exist. Instead we should check the row accessor to see if it is valid before operating on it. Behaviour is changed so that invalid operations on a user metadata will now throw an `InvalidUserException`.

Specifically this fixes https://github.com/realm/realm-sync/issues/2586. The included script there both makes a "DELETE" http request on a user and logs the user out locally. This means that the server deletes the user (via ROS handling the DELETE request), and the user is marked for deletion locally the next time the client connects a session. So the next time the local client connects, it sees the locally marked user object which is marked for removal and begins to delete the associated files on disk. While this happens, sync tells the client that the row is now deleted and the accessor becomes invalid. Then when the client finishes cleaning the user's disk state, it removes the (now invalid) row accessor. This causes undefined behaviour later on, including corruption and/or crashes, including the assertion seen in https://github.com/realm/realm-sync/issues/2586.

### DIFF
--- a/src/sync/impl/sync_metadata.cpp
+++ b/src/sync/impl/sync_metadata.cpp
@@ -78,6 +78,10 @@ realm::Schema make_schema()
 
 namespace realm {
 
+InvalidUserException::InvalidUserException(const std::string& msg)
+: std::runtime_error("Operation not permitted on an invalid user; " + msg)
+{}
+
 // MARK: - Sync metadata manager
 
 SyncMetadataManager::SyncMetadataManager(std::string path,
@@ -334,8 +338,15 @@ SyncUserMetadata::SyncUserMetadata(Schema schema, SharedRealm realm, RowExpr row
 , m_row(row)
 { }
 
+void SyncUserMetadata::checkValidOrThrow(const std::string operation) const
+{
+    if (!m_row.is_attached())
+        throw InvalidUserException(operation);
+}
+
 std::string SyncUserMetadata::identity() const
 {
+    checkValidOrThrow("identity");
     REALM_ASSERT(m_realm);
     m_realm->verify_thread();
     return m_row.get_string(m_schema.idx_identity);
@@ -343,6 +354,7 @@ std::string SyncUserMetadata::identity() const
 
 std::string SyncUserMetadata::local_uuid() const
 {
+    checkValidOrThrow("local_uuid");
     REALM_ASSERT(m_realm);
     m_realm->verify_thread();
     return m_row.get_string(m_schema.idx_local_uuid);
@@ -350,6 +362,7 @@ std::string SyncUserMetadata::local_uuid() const
 
 util::Optional<std::string> SyncUserMetadata::user_token() const
 {
+    checkValidOrThrow("user_token");
     REALM_ASSERT(m_realm);
     m_realm->verify_thread();
     StringData result = m_row.get_string(m_schema.idx_user_token);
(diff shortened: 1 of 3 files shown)


bmunkholm:
Nice find @ironage !

src/sync/impl/sync_metadata.cpp
```diff
@@ -334,22 +338,31 @@ SyncUserMetadata::SyncUserMetadata(Schema schema, SharedRealm realm, RowExpr row
 , m_row(row)
 { }
 
+void SyncUserMetadata::checkValidOrThrow(const std::string operation) const
```
tgoyne: This should just take a StringData or const char* to avoid constructing a std::string every time it's called.

tgoyne: The usual name for this function would be just `verify_attached()`.

src/sync/impl/sync_metadata.cpp
```diff
@@ -334,22 +338,31 @@ SyncUserMetadata::SyncUserMetadata(Schema schema, SharedRealm realm, RowExpr row
...
+{
+    if (!m_row.is_attached())
+        throw InvalidUserException(operation);
+}
+
 std::string SyncUserMetadata::identity() const
 {
+    checkValidOrThrow("identity");
```
tgoyne: All of these checks need to happen after `verify_thread()` as that's the only thing that's safe to call from the wrong thread.

src/sync/impl/sync_metadata.cpp
```diff
@@ -358,60 +371,81 @@ util::Optional<std::string> SyncUserMetadata::user_token() const
...
 {
-    m_invalid = true;
+    if (!m_row.is_attached() || !m_realm)
+        return;
+
     m_realm->begin_transaction();
     TableRef table = ObjectStore::table_for_object_type(m_realm->read_group(), c_sync_userMetadata);
-    table->move_last_over(m_row.get_index());
-    m_realm->commit_transaction();
+    if (bool(table) && m_row.is_attached()) { // need to recheck m_row since the write transaction advances versions
```
tgoyne: The check if table is valid is unneeded (we wouldn't get to here if a transaction removed the table).

src/sync/impl/sync_metadata.cpp
```diff
@@ -358,60 +371,81 @@ util::Optional<std::string> SyncUserMetadata::user_token() const
...
+        table->move_last_over(m_row.get_index());
```
tgoyne: m_row.move_last_over()

ironage:
> The SyncMetadata Realm is not a synchronized Realm

Thanks for clarifying that. This invalidates my hypothesis of how the assertion in https://github.com/realm/realm-sync/issues/2586  is triggered. I still think there is a race here though. Maybe there are multiple processes starting at the same time, both try to clean up a logged out user, one will succeed and the other will crash?

> Throwing from the getter functions does not actually produce useful results

How should we behave if a getter is called for an object which has been deleted? I would have thought throwing an exception is an appropriate response, but I'm really familiar with how this code is used.

tgoyne:
Multiple processes starting at once causing problems sounds plausible. I'd bet that a lot of things have awkward race conditions with multiple sync processes, actually; the code was originally written with iOS in mind where that isn't a concern.

As for how to handle it... either it's not something that can actually happen in practice and should probably just be an assert, or a bunch of code will need to be reworked.

PR #765, opened 2019-01-23
Expose fifo fallback path

cmelchior:
In some situations, we cannot create the FIFO files required for Realm to operate (e.g. external storage on Android). In that case, we need to provide a fallback directory where these files can be written.

This PR exposes a `fifo_files_fallback_path` property on the Config object that can be used to configure where fifo files are placed if they cannot be created beside the Realm file. This applies to both Object Store and Core FIFOs.

Required to fix: https://github.com/realm/realm-js/issues/2062

This PR will break current Apple behaviour. They will need to set `config.temp_dir = getenv("TMPDIR")` to get the same behaviour as before this PR.

This PR also split some previous functionality only existing for sync tests, so these helper functions are now also available for non-sync-tests.

TODO
- [x] Handle fallback dirs not ending with `/`
- [x] Figure out if there is a way to cause Object store to take the fallback path.


nirinchev:
I may be misreading the Core API, but I'm not convinced the naming is semantically correct. The new property is called `fallback_path` which implies that it will only be used if the originally intended path is for some reason not usable, yet the shared group property is called `temp_dir`, which implies that it's overwriting whatever the default for that was.

tests/realm.cpp
```diff
@@ -152,6 +152,18 @@ TEST_CASE("SharedRealm: get_shared_realm()") {
...
+    SECTION("should be able to set a FIFO fallback path") {
+        TestFile config;
+        config.fifo_files_fallback_path = "fallback_path";
+        config.schema_version = 1;
+        config.schema = Schema{
+            {"object", {
+                {"value", PropertyType::Int}
+            }},
+        };
+        REQUIRE_NOTHROW(Realm::get_shared_realm(config));
```
tgoyne: This doesn't actually test anything? Should at least verify that the files are in the expected path.

cmelchior: I know, this was mostly a sanity check. I expected that core would only write the files to the backup location in case the primary location failed. If I misunderstood that, then yes, we should check the files

cmelchior:
I didn't know that OS also created that, but yeah, that means this PR needs further work.

cmelchior:
Ready for review

src/util/fifo.hpp
```diff
@@ -0,0 +1,40 @@
...
+#include <string>
+
+// This file contains various helper methods for working with FIFOs.
+
+namespace realm {
+namespace util {
+
+// Creates a fifo at the provided path. If the FIFO could not be created an exception is thrown.
+// This method will also be successful if an existing FIFO already existed at the given location.
+void create_fifo(std::string& path);
```
tgoyne: These functions should take a const reference to the string as they don't mutate the string.

tests/sync/file.cpp
```diff
@@ -20,6 +20,7 @@
 
 #include "shared_realm.hpp"
 #include "sync/sync_manager.hpp"
+#include "util/test_utils.hpp"
```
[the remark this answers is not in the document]
tgoyne: It is used (some functionality was split off from sync_test_utils.hpp).

This include should be grouped with sync_test_utils.hpp and not where it is, though. The current position suggests it's a part of the library being tested rather than a thing that's part of the tests themselves.

PR #766, opened 2019-01-31
Avoid pinning Realm versions when queries are made with automatic change notifications disabled

tgoyne:
The root cause of https://github.com/realm/data-adapters/issues/464

The adapter vends Realm instances to the user code which have automatic change notifications disabled (they wouldn't really work properly with how the adapter integrates into things and disabling them cuts resource usage), but holds on to a long-lived RealmCoordinator for each adapted Realm. If the user code queried one of these Realms we'd implicitly create a ResultsNotifier for the results which pins the current transaction version, which would then never be cleaned up since the worker thread isn't running.

This fixes both of the problems there: if we aren't using the notification worker thread we shouldn't implicitly create notifiers, and if the user explicitly creates a notifier we should clean it up when the Realm is unregistered rather than waiting for the RealmCoordinator to go away.

### DIFF
--- a/src/impl/realm_coordinator.cpp
+++ b/src/impl/realm_coordinator.cpp
@@ -325,18 +325,18 @@ RealmCoordinator::~RealmCoordinator()
 
 void RealmCoordinator::unregister_realm(Realm* realm)
 {
-    {
-        std::lock_guard<std::mutex> lock(m_realm_mutex);
-        auto new_end = remove_if(begin(m_weak_realm_notifiers), end(m_weak_realm_notifiers),
-                                 [=](auto& notifier) { return notifier.expired() || notifier.is_for_realm(realm); });
-        m_weak_realm_notifiers.erase(new_end, end(m_weak_realm_notifiers));
-    }
     // Normally results notifiers are cleaned up by the background worker thread
     // but if that's disabled we need to ensure that any notifiers from this
     // Realm get cleaned up
     if (!m_config.automatic_change_notifications) {
         std::unique_lock<std::mutex> lock(m_notifier_mutex);
         clean_up_dead_notifiers();
+    }
+    {
+        std::lock_guard<std::mutex> lock(m_realm_mutex);
+        auto new_end = remove_if(begin(m_weak_realm_notifiers), end(m_weak_realm_notifiers),
+                                 [=](auto& notifier) { return notifier.expired() || notifier.is_for_realm(realm); });
+        m_weak_realm_notifiers.erase(new_end, end(m_weak_realm_notifiers));
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/impl/realm_coordinator.cpp
```diff
@@ -882,19 +890,25 @@ void RealmCoordinator::process_available_async(Realm& realm)
...
+    if (notifiers.empty())
+        return;
     lock.unlock();
 
     // no before advance because the Realm is already at the given version,
     // because we're either sending initial notifications or the write was
     // done on this Realm instance
 
+    if (realm.m_binding_context) {
+        realm.m_binding_context->will_send_notifications();
```
cmelchior: Isn't it wrong to notify here if we then return immediately after and not send notifications after all? If I remember correctly there is also a `did_send_notifications()` which should always be paired.

src/impl/realm_coordinator.cpp
```diff
@@ -325,10 +325,19 @@ RealmCoordinator::~RealmCoordinator()
...
+    {
+        std::lock_guard<std::mutex> lock(m_realm_mutex);
+        auto new_end = remove_if(begin(m_weak_realm_notifiers), end(m_weak_realm_notifiers),
+                                 [=](auto& notifier) { return notifier.expired() || notifier.is_for_realm(realm); });
+        m_weak_realm_notifiers.erase(new_end, end(m_weak_realm_notifiers));
+    }
+    // Normally results notifiers are cleaned up by the background worker thread
+    // but if that's disabled we need to ensure that any notifiers from this
+    // Realm get cleaned up
+    if (!m_config.automatic_change_notifications) {
```
cmelchior: Is it safe to have this outside the `m_realm_mutex`?

tgoyne: `RealmCoordinate::m_config` is only ever mutated when there's no registered Realms... which I guess means this needs to be *before* the thing right above it.

src/impl/realm_coordinator.cpp
tgoyne: We only return immediately after this if the `will_send_notifications()` callback closes the Realm, and in that case the previous code also didn't call `did_send_notifications()`. That might be a bug that's now just more obvious, but I don't know what it'd actually do in did_send if the Realm is already closed.

tgoyne:
It's hard to write a test which usefully tests this behavior since part of the change is just "don't do this thing which shouldn't have had observable results at all", but I guess it'd be worth at least writing a regression test for the version pinning part.

tgoyne:
I've added a test that fails on master (gives 3 == 2 on the last check) and passes on this branch.

Issue #767, opened 2019-01-30
Allow developers to change the query of a named subscription after it is active


ianpward:
Adding another parameter to subscribe() to let them ask for it to just overwrite the existing one instead wouldn't be hard and would be useful

cc. @roberhofer @astigsen

simonask:
For the record, Sync is fine with all of this. :-)

nirinchev:
@cmelchior why would we need a `setName()`? I believe what Ian wants is something like:

```js
const subscription = realm.objects("Foo").filtered("Bar = 5").subscribe("MainScreenFoos");

const sameSubscription = realm.objects("Foo").filtered("Bar >= 5).subscribe("MainScreenFoos", /* override */ true);
```

simonask:
I don't think it makes a difference from the server's perspective. Assuming the add/remove of a query arrives in the same work unit on the server (which isn't guaranteed, since they happen in separate write transactions, but it is not unlikely), the server will rerun the queries and diff against the current subscription anyway. :-)

tgoyne:
Why do you think it is unlikely that they will arrive separately? A decent amount of stuff happens between when we notify sync about the first write transaction and when we start the second one.

astigsen:
It would be really nice if this could percolate all the way down to the (previous) query itself, so that any UI that observed that query would be updated automatically.

astigsen:
> The API I had in mind for Swift when we talked about this would be something like `realm.objects(Foo.self).filter("query").limit(100).subscribe("Foo query", overwrite: true)`.

An alternative API could be to update it on the subscription itself (which might be a bit more discoverable):

```swift
let newquery = realm.objects(Foo.self).filter("query").limit(100)
subscription.update(newquery)
```

which would also open for the even more concise (since the subscription is already typed):

```swift
subscription.update("query")
```

tgoyne:
I do not think that moving away from an immutable builder API for queries is a good idea.

astigsen:
I agree that we probably don't want to move away from our builder API for queries, but that does not mean that we could not add some convenience methods to make it more discoverable.

It might also be worth considering if we could bind the UI (our listeners) directly to the subscription, rather than the query itself, so that updating the query in the subscription is also reflected in the UI.

cmelchior:
Being able to update the query held by a `Results` class is also something other users are requesting as the use case isn't only restricted to Subscriptions. See e.g. the Java issue here: https://github.com/realm/realm-java/issues/6216

PR #775, opened 2019-03-04
Add support for Subscription timestamps and automatic cleanup

cmelchior:
Closes: https://github.com/realm/realm-object-store/issues/767
Design Issue: https://github.com/realm/engineering/issues/12

This PR adds new capabilities to subscriptions:

* 3 new timestamps to Subscriptions: `created_at`, `updated_at` and `expires_at`
* 1 new int field for defining TTL: `time_to_live` 
* Makes it possible to update the underlying query for a subscription.
* Subscriptions which have parsed their `expires_at` date, will now be removed automatically when the app is started or whenever a subscription is created/modified.

**Design thoughts:**
By hooking the cleanup up to creating/modifying existing subscriptions we do not have truly ephemeral subscriptions. For that, we would need to tie it in with the destruction of the notification token. However, it isn't clear to me how to accomplish that relatively easy. Also, I'm not sure it matters. 

The reasoning being:

1) If we tie the cleanup to the Notification token, it will have to be asynchronous. Meaning an app crash can disrupt the cleanup. Cleaning up on startup catches these cases. Worst case you have subscription living longer on the server if the app isn't re-opened, but this is true either way.

2) In all other cases, the user will navigate to a screen in the app that also contains a subscription (presumably). Which will trigger cleanup?

3) The only concern I have is that running cleanup on all subscription modifications might slow down creating new subscriptions. It is unclear at which point we have so many subscriptions that running the cleanup query actually impacts creating the subscription to a noticeable degree.

**New field names**:
I opted to use `under_score` as it follows the existing naming paradigm for the class. Feel free to object if this is a problem (we already have `error_message`).

TODO:
- [x] Add tests for async API
- [x] Add automatic cleanup when writing subscriptions
- [x] Add tests for timestamps
- [x] Add tests for automatic cleanup
- [x] Determine the name of the underlying fields. E.g. `camelCase` vs. `under_score`
- [x] Add support for both `timeToLive` and `expiresAt`
- [x] Set `expiresAt` to far in the future when upgrading old Realms to mimic old behaviour.
- [x] Create issues in all repos for exposing this.
- [x] Verify the usefulness of this implementation by implementing at least on binding API before merging.


nirinchev:
Wouldn't it make sense to make the `expiresAt` value optional? I realize there's no practical difference between a huge value and no value, but still, would be nice to semantically indicate when you expect a subscription to last forever.

astigsen:
Do we have any logic for updating the `expiresAt`? Sometimes (often?) it would be nice to have it relative to when the data was last used in the app (so you for example could have a subscription stay around for 2 days after last use, just as a cache in case someone did the same query again).

cmelchior:
@astigsen Right now the idea was to set `expiresAt` either when creating or updating the subscription (e.g. replacing the query). Making it relative to `expiresAt` sounds useful, probably even more useful than a fixed `expiresAt`. I see you also made a similar comment in https://github.com/realm/engineering/issues/12, sorry for having missed that.

I agree that this use case is probably more natural (and safer). I'll look into changing `expiresAt` to be an offset from last used.

src/sync/partial_sync.cpp
```diff
@@ -245,28 +295,60 @@ RowExpr write_subscription(std::string const& object_type, std::string const& na
...
-void enqueue_registration(Realm& realm, std::string object_type, std::string query, std::string name,
+void enqueue_registration(Realm& realm, std::string object_type, std::string query, std::string name, Timestamp expires, bool update,
                           std::function<void(std::exception_ptr)> callback)
 {
     auto config = realm.config();
 
     auto& work_queue = _impl::PartialSyncHelper::get_coordinator(realm).partial_sync_work_queue();
     work_queue.enqueue([object_type=std::move(object_type), query=std::move(query), name=std::move(name),
-                        callback=std::move(callback), config=std::move(config)] {
+                        callback=std::move(callback), config=std::move(config), expires=std::move(expires), update=std::move(update)] {
```
tgoyne: Moving a Timestamp is pointless (copying it does the exact same thing).

tgoyne:
Making expiresAt optional would also be good because it makes the migration a no-op. It probably wouldn't ever be an issue, but setting it to max_timestamp after creating the column opens up the possibility of weirdness if the file is then opened with an older version of Realm which creates more subscriptions without setting the timestamp.

Performing cleanup when adding a new subscription is very unlikely to ever be an issue performance-wise. The write transaction overhead is going to dominate everything else, so as long as you can do the cleanup in the same transaction...

cmelchior:
Ready for review

src/sync/partial_sync.cpp
```diff
@@ -33,6 +36,65 @@
...
+
+    /// Turn a system time point value into the 64-bit integer representing ns since the Unix epoch.
+    int64_t ns_since_unix_epoch(const system_clock::time_point& point)
+    {
+        return duration_cast<nanoseconds>(point.time_since_epoch()).count();
+    }
+
+    realm::Timestamp timestamp_now() {
+        int64_t ns_since_epoch = ns_since_unix_epoch(system_clock::now());
+        int64_t s_arg = ns_since_epoch / (int64_t)realm::Timestamp::nanoseconds_per_second;
```
kneth:
```suggestion
        int64_t s_arg = ns_since_epoch / static_cast<int64_t>(realm::Timestamp::nanoseconds_per_second);
```

src/sync/partial_sync.cpp
```diff
@@ -59,12 +122,43 @@ void initialize_schema(Group& group)
...
+    // autogenerated.
     if (table->get_column_index("name") == npos) {
         size_t idx = table->add_column(type_String, "name");
         table->add_search_index(idx);
     }
+
+    // Timestamp for when then the subscription is created. This should only be set the first time the subscription
+    // is created.
+    if (table->get_column_index("created_at") == npos) {
+        table->add_column(type_Timestamp, "created_at");
```
kneth: Can column names be constants and reused other places?

src/sync/partial_sync.cpp
```diff
@@ -195,43 +305,27 @@ struct ResultSetsColumns {
...
 // instead.
-RowExpr write_subscription(std::string const& object_type, std::string const& name, std::string const& query, Group& group)
+//
+// If `update = true` and  if a subscription with `name` already exists, its query and time_to_live
+// will be updated instead of an exception being thrown if the query parsed in was different than
+// the persisted query.
+RowExpr write_subscription(std::string const& object_type, std::string const& name, std::string const& query,
+        util::Optional<int64_t> time_to_live_ms, bool update, Group& group)
 {
+    Timestamp now = timestamp_now(); //
```
kneth:
```suggestion
    Timestamp now = timestamp_now();
```

src/sync/partial_sync.cpp
```diff
@@ -241,32 +335,95 @@ RowExpr write_subscription(std::string const& object_type, std::string const& na
...
+    else {
         // FIXME: Validate that the column type and link target are correct.
     }
 
-    size_t row_ndx = validate_existing_subscription(*table, columns, name, query, matches_property);
-    if (row_ndx == npos) {
+    // Find existing subscription (if any)
+    auto row_ndx = table->find_first_string(columns.name, name);
+    if (row_ndx != npos) {
+        // If an existing subscription exist, we only update the query and TTL if allowed to.
```
kneth:
```suggestion
        // If an subscription exist, we only update the query and TTL if allowed to.
```

src/sync/partial_sync.cpp
```diff
@@ -241,32 +335,95 @@ RowExpr write_subscription(std::string const& object_type, std::string const& na
...
+        // TODO: Consider how Binding API's are going to use this. It might make sense to disallow
+        // updating TTL using this API and instead require updates to TTL to go through a managed Subscription.
+        if (update) {
+            // Check that we don't replace the existing query with a query on a new type.
+            // There is nothing that prevents Sync from handling this, but allowing it
+            // will complicate Binding API's, so for now it is disallowed.
+            auto existing_matching_property = table->get_string(columns.matches_property_name, row_ndx);
+            if (existing_matching_property != matches_property) {
+                throw QueryTypeMismatchException(util::format("Replacing an existing query with a query on "
+                                                              "a different type is not allowed: %1 vså. %2",
```
kneth:
```suggestion
                                                              "a different type is not allowed: %1 vs. %2",
```

src/sync/partial_sync.cpp
```diff
@@ -16,6 +16,9 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
+#define __STDC_LIMIT_MACROS // See https://stackoverflow.com/a/3233069/1389357
+#include <cstdint>
```
tgoyne: These includes should come after the realm ones (the order used in OS files is corresponding header -> local headers -> core/sync headers -> system headers).

src/sync/partial_sync.cpp
```diff
@@ -33,6 +36,65 @@
...
+
+    // FIXME: Replace these methods with: https://github.com/realm/realm-core/pull/3259
+
+    /// Turn a system time point value into the 64-bit integer representing ns since the Unix epoch.
+    int64_t ns_since_unix_epoch(const system_clock::time_point& point)
+    {
+        return duration_cast<nanoseconds>(point.time_since_epoch()).count();
+    }
+
+    realm::Timestamp timestamp_now() {
```
tgoyne: This could be simplified by letting std::chrono handle the conversions rather than doing the math manually:

    auto now = system_clock::now();
    auto sec = time_point_cast<seconds>(now);
    return {sec.time_since_epoch().count(), duration_cast<nanoseconds>(now - sec).count()};

src/sync/partial_sync.cpp
```diff
@@ -241,32 +335,95 @@ RowExpr write_subscription(std::string const& object_type, std::string const& na
...
+    if (row_ndx != npos) {
+        // If an subscription exist, we only update the query and TTL if allowed to.
+        // TODO: Consider how Binding API's are going to use this. It might make sense to disallow
+        // updating TTL using this API and instead require updates to TTL to go through a managed Subscription.
+        if (update) {
+            // Check that we don't replace the existing query with a query on a new type.
+            // There is nothing that prevents Sync from handling this, but allowing it
+            // will complicate Binding API's, so for now it is disallowed.
+            auto existing_matching_property = table->get_string(columns.matches_property_name, row_ndx);
+            if (existing_matching_property != matches_property) {
```
tgoyne: This check exists in both arms of `if (update)`, so it should just be pulled out of the `if`. The error message also should include the name of the subscription.

src/sync/partial_sync.cpp
```diff
@@ -241,32 +335,95 @@ RowExpr write_subscription(std::string const& object_type, std::string const& na
...
+                throw QueryTypeMismatchException(util::format("Replacing an existing query with a query on "
+                                                              "a different type is not allowed: %1 vs. %2",
+                                                              existing_matching_property, matches_property));
+            }
+
+            table->set_string(columns.query, row_ndx, query);
+            if (time_to_live_ms) {
```
tgoyne: Table::set() can just take a `util::Optional<int64_t>` directly: `table->set(columns.time_to_live, row_ndx, time_to_live_ms);`

src/sync/partial_sync.cpp
```diff
@@ -241,32 +335,95 @@ RowExpr write_subscription(std::string const& object_type, std::string const& na
...
+        table->set_null(columns.expires_at, row_ndx);
+    }
+    else {
+        table->set_timestamp(columns.expires_at, row_ndx, calculate_expiry_date(now, table->get_int(columns.time_to_live, row_ndx)));
     }
-    return table->get(row_ndx);
+
+    // Fetch subscription first and return it. Cleanup needs to be performed after as it might delete subscription
+    // causing the row_ndx to change.
+    RowExpr subscription = table->get(row_ndx);
```
tgoyne: This needs to be `Row`. Storing a `RowExpr` is identical to just storing the row index itself.

src/sync/partial_sync.cpp
```diff
@@ -241,32 +335,95 @@ RowExpr write_subscription(std::string const& object_type, std::string const& na
...
-                          std::function<void(std::exception_ptr)> callback)
+void enqueue_registration(Realm& realm, std::string object_type, std::string query, std::string name, util::Optional<int64_t> time_to_live,
+        bool update, std::function<void(std::exception_ptr)> callback)
 {
     auto config = realm.config();
 
     auto& work_queue = _impl::PartialSyncHelper::get_coordinator(realm).partial_sync_work_queue();
     work_queue.enqueue([object_type=std::move(object_type), query=std::move(query), name=std::move(name),
-                        callback=std::move(callback), config=std::move(config)] {
+                        callback=std::move(callback), config=std::move(config), time_to_live=std::move(time_to_live), update=std::move(update)] {
```
tgoyne: Just `time_to_live, update`. Moving primitives is pointless.

src/sync/partial_sync.cpp
```diff
@@ -430,15 +604,15 @@ Subscription subscribe(Results const& results, util::Optional<std::string> user_
 
     Subscription subscription(name, results.get_object_type(), realm);
     std::weak_ptr<Subscription::Notifier> weak_notifier = subscription.m_notifier;
-    enqueue_registration(*realm, results.get_object_type(), std::move(query), std::move(name),
+    enqueue_registration(*realm, results.get_object_type(), std::move(query), std::move(name), std::move(time_to_live_ms), std::move(update),
```
tgoyne: pointless moves

src/sync/partial_sync.cpp
```diff
@@ -33,6 +36,65 @@
...
+
+    realm::Timestamp timestamp_now() {
+        int64_t ns_since_epoch = ns_since_unix_epoch(system_clock::now());
+        int64_t s_arg = ns_since_epoch / static_cast<int64_t>(realm::Timestamp::nanoseconds_per_second);
+        int32_t ns_arg = ns_since_epoch % realm::Timestamp::nanoseconds_per_second;
+        return realm::Timestamp(s_arg, ns_arg);
+    }
+
+    // Calculates the expiry date, claming at the high end if a timestamp overflows
+    realm::Timestamp calculate_expiry_date(realm::Timestamp starting_time, int64_t user_ttl_ms) {
```
tgoyne: Newline before `{` for function definitions (and nothing else). A bunch of these in the file.

src/sync/partial_sync.hpp
```diff
@@ -77,10 +81,16 @@ class Subscription {
...
-    friend Subscription subscribe(Results const&, util::Optional<std::string>);
+    friend Subscription subscribe(Results const&, util::Optional<std::string>, util::Optional<int64_t> time_to_live, bool update);
     friend void unsubscribe(Subscription&);
 };
 
+
+// Delete all subscriptions that are no longer relevant.
+//
+// This method must be called within a write transaction.
+void cleanup_subscriptions(Group& group, Timestamp now);
```
tgoyne: There doesn't appear to be any reason for this to be exposed in the header.

tests/sync/partial_sync.cpp
```diff
@@ -173,16 +209,30 @@ partial_sync::Subscription subscribe_and_wait(Results results, util::Optional<st
...
+    auto results = results_for_query(query, partial_config, object_type);
+    return subscribe_and_wait(std::move(results), std::move(name), std::move(ttl), update, std::move(check));
+}
+
+/// Run a Query-based Sync query, wait for the results, and then perform checks.
+partial_sync::Subscription subscribe_and_wait(std::string const& query, Realm::Config const& partial_config,
                         std::string const& object_type, util::Optional<std::string> name,
                         std::function<void(Results, std::exception_ptr)> check)
 {
     auto results = results_for_query(query, partial_config, object_type);
```
tgoyne: `results` appears to be unused now?

src/sync/partial_sync.cpp
```diff
@@ -241,32 +353,80 @@ RowExpr write_subscription(std::string const& object_type, std::string const& na
...
+        // TODO: Consider how Binding API's are going to use this. It might make sense to disallow
+        // updating TTL using this API and instead require updates to TTL to go through a managed Subscription.
+        if (update) {
+            table->set_string(columns.query, row_ndx, query);
+            table->set(columns.time_to_live, row_ndx, time_to_live_ms);
+        }
+        else {
+            StringData existing_query = table->get_string(columns.query, row_ndx);
+            if (existing_query != query)
+                throw ExistingSubscriptionException(util::format("An existing subscription exists with the same name, "
```
tgoyne: This error should include the name of the subscription.

PR #779, opened 2019-03-14
Reset status flag when updating subscription

cmelchior:
It turned out I forgot to reset the status flag when updating a subscription which meant the server never re-evaluated the query. This PR fixes that.

Also, it turned out there was a bunch of race conditions in the subscription notifications. This PR also fixes the following issues:
- The same subscription being reported twice.
- `Invalidated` is now being reported even if the subscription is deleting manually. Previously it was only reported if you used the explicit unsubscribe API.

TODO
- [x] Support for multiple callbacks on the same notification

### DIFF
--- a/src/sync/partial_sync.cpp
+++ b/src/sync/partial_sync.cpp
@@ -697,6 +697,7 @@ Subscription::Subscription(std::string name, std::string object_type, std::share
 
     auto matches_property = std::string(object_type) + "_matches";
 
+    m_wrapper_created_at = timestamp_now();
     TableRef table = ObjectStore::table_for_object_type(realm->read_group(), result_sets_type_name);
     Query query = table->where();
     query.equal(m_object_schema.property_for_name("name")->table_column, name);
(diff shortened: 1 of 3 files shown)


tgoyne:
The test is failing because the state is already Complete when `subscribe_and_wait()` is called, so there's no guarantee that it'll actually wait.

tgoyne:
I don't think changing the test is the correct fix. The test was doing something that *should* work, and the problem is that the updated subscription claims to be complete when it isn't.

src/sync/partial_sync.cpp
```diff
@@ -729,8 +730,68 @@ util::Optional<Object> Subscription::result_set_object() const
     return util::none;
 }
 
+void Subscription::run_callback(Subscription *sub, std::function<void()> callback) {
```
tgoyne: Why does this function take `this` as a separate parameter?

src/sync/partial_sync.cpp
```diff
@@ -729,8 +730,68 @@ util::Optional<Object> Subscription::result_set_object() const
...
+    //   reaches that state.
+
+    // Errors take precedence over all other notifications
+    if (m_notifier->error())
+        return SubscriptionState::Error;
+
+    // In some cases the subscription already exists. In that case we just report the state of the __ResultSets object.
+    if (auto object = result_set_object()) {
+        CppContext context;
+        auto state = (SubscriptionState) any_cast<int64_t>(object->get_property_value<util::Any>(context, property_status));
```
tgoyne: Use static_cast<SubscriptionState>() rather than a C-style cast.

src/sync/partial_sync.cpp
```diff
@@ -729,8 +730,68 @@ util::Optional<Object> Subscription::result_set_object() const
...
+        auto updated_at = (Timestamp) any_cast<Timestamp>(object->get_property_value<util::Any>(context, property_updated_at));
```
tgoyne: This cast should be pointless?

cmelchior:
I think I managed to close the most obvious errors in the Subscription notification system by now. Ready for another round of review.

src/sync/partial_sync.cpp
```diff
@@ -697,12 +711,16 @@ Subscription& Subscription::operator=(Subscription&&) = default;
 
 SubscriptionNotificationToken Subscription::add_notification_callback(std::function<void ()> callback)
 {
-    auto result_sets_token = m_result_sets.add_notification_callback([callback] (CollectionChangeSet, std::exception_ptr) {
-        callback();
+    auto callback_wrapper = std::make_shared<SubscriptionCallbackWrapper>(SubscriptionCallbackWrapper{callback, none});
+    auto token = std::make_shared<SubscriptionNotificationToken>();
```
tgoyne: This appears to not be used anywhere.

PR #781, opened 2019-03-19
Fix unsubscribing from a subscription which hasn't been created yet

tgoyne:
Unsubscribing from a Subscription which was still in the Creating state only worked correctly if the caller held onto a reference to the Subscription object until it made it to the Invalidated state (as the existing tests happen to). If it was destroyed we would either fail to remove the subscription or even outright crash depending on when the Subscription object was deallocated.

Fixes https://github.com/realm/realm-js-private/issues/520.

### DIFF
--- a/src/sync/partial_sync.cpp
+++ b/src/sync/partial_sync.cpp
@@ -243,24 +243,25 @@ public:
     }
 };
 
-struct RowHandover {
-    RowHandover(Realm& realm, Row row)
-    : source_shared_group(*PartialSyncHelper::get_shared_group(realm))
-    , row(source_shared_group.export_for_handover(std::move(row)))
-    , version(source_shared_group.pin_version())
-    {
-    }
+template<typename... Args>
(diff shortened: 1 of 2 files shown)


src/sync/partial_sync.cpp
```diff
@@ -483,6 +479,42 @@ void enqueue_unregistration(Object result_set, std::function<void()> callback)
...
 
+template<typename Notifier>
+void enqueue_unregistration(Results const& result_set, std::shared_ptr<Notifier> notifier,
+                            std::function<void()> callback)
+{
+    auto realm = result_set.get_realm();
+    auto config = realm->config();
+    auto& work_queue = _impl::PartialSyncHelper::get_coordinator(*realm).partial_sync_work_queue();
+
+    // Export a reference to the __ResultSets row so we can hand it to the worker thread.
```
cmelchior: This comment doesn't seem to be correct anymore? Instead, it seems we are exporting the query and re-running it.

PR #782, opened 2019-03-20
Subscription inclusions

ironage:
This is the basic idea for adding backlink inclusions to a subscription. ~It needs to be rebased on a sync release with https://github.com/realm/realm-sync/pull/2892 and needs some tests.~


ironage:
This is ready for review.
There was one sporadic [failure](https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-object-store/detail/PR-782/7/pipeline) with the new sync which is a bit concerning, but I don't think it is related to this work in particular. @simonask does it appear related to https://github.com/realm/realm-sync/issues/2945?

src/sync/partial_sync.cpp
```diff
@@ -643,12 +642,16 @@ Subscription subscribe(Results const& results, util::Optional<std::string> user_
     auto query = results.get_query().get_description(); // Throws if the query cannot be serialized.
     query += " " + results.get_descriptor_ordering().get_description(results.get_query().get_table());
 
-    std::string name = user_provided_name ? std::move(*user_provided_name)
+    if (options.inclusions.is_valid()) {
+        query += " " + options.inclusions.get_description(results.get_query().get_table());
+    }
+
+    std::string name = options.user_provided_name ? std::move(*options.user_provided_name)
```
tgoyne: Indentation on the following line needs to be updated to match this.

src/sync/partial_sync.hpp
```diff
@@ -71,6 +71,25 @@ struct SubscriptionCallbackWrapper {
     util::Optional<SubscriptionState> last_state;
 };
 
+struct SubscriptionOptions
+{
```
tgoyne: Only function definitions have a newline before brace.

ironage:
I'm fine with leaving this on a branch for a while. It may conflict a bit with the async open work, but that will likely be merged before this so I'll address those conflicts later.

src/sync/partial_sync.cpp
```diff
@@ -643,12 +642,16 @@ Subscription subscribe(Results const& results, util::Optional<std::string> user_
     auto query = results.get_query().get_description(); // Throws if the query cannot be serialized.
     query += " " + results.get_descriptor_ordering().get_description(results.get_query().get_table());
 
-    std::string name = user_provided_name ? std::move(*user_provided_name)
-                                          : default_name_for_query(query, results.get_object_type());
+    if (options.inclusions.is_valid()) {
+        query += " " + options.inclusions.get_description(results.get_query().get_table());
```
cmelchior: Minor, but the unconditional padding of empty whitespace is a bit annoying (especially when writing tests), now, if both descriptor ordering and inclusions are empty you end up with two empty " " at the end of the query string.

simonask:
> There was one sporadic failure with the new sync which is a bit concerning, but I don't think it is related to this work in particular. @simonask does it appear related to realm/realm-sync#2945?

Hmm, I wouldn't say it looks like realm/realm-sync#2945. If it really is the same problem, it's certainly manifesting itself in a very obscure way! :-D

An assertion failure like this typically means that a Client object is being destroyed while Session objects are in use (or similar), i.e. an I/O completion handler is scheduled while the object dies. We've seen this situation a couple of times in tests.

src/sync/partial_sync.cpp
ironage: @cmelchior I've updated things so that there is no unnecessary whitespace at the end of subscriptions without sort/distinct/limit. You'll have to update any binding tests that check subscription syntax, but I figure better to take care of this now than have it as a continual annoyance.

ironage:
@cmelchior I added a [commit](https://github.com/realm/realm-object-store/pull/782/commits/e4b1314d21b521fd604af7f1aacf3ca94272c19a) which should help ease the java binding's integration. Can you let me know if it does what you need?

cmelchior:
Is there anything preventing merging this into Thomas async branch? This seems basically done and then we only have 1 branch with 4.* changes. Right now it is a bit annoying to juggle both when implementing in bindings.

tgoyne:
I split the part of #778 that's finished (updating the sync version and adding async open support) off to a separate `sync-4.x` branch and then retargeted this PR at that branch as a step towards getting all the sync 4 stuff together.

ironage:
Thanks Thomas! I'll be away next week so feel free to merge if the changes here are sufficient.

cmelchior:
Since all comments are addressed and this implementation seems to work for Realm Java I'm merging this to the 4.x branch so bindings can adopt both this and async open at the same time.

If we find any problems we should address them in a new PR.

PR #783, opened 2019-03-25
Add support for field aliases to properties 

cmelchior:
In order to avoid breaking changes in JS (see https://github.com/realm/realm-js/issues/2266), we need to have support for re-mapping fields in the JS schema. This feature is already present in Java, Cocoa and the core query engine, so this PR just adds the tempory data to the schema objects in Object Store similar to computed properties. 

**Usage in bindings**
Bindings are responsible for setting and using the alias field.
Reading/Writing fields are supported by overriding the `Object::value_for_property` method in the Object Accessor. 
Queries support this by creating a specific `KeyPathMapping` with the alias fields.

See https://github.com/realm/realm-js/pull/2306 for an example on how this is used.

**Performance considerations:**
A few comments about performance impact:

We don't currently check if properties are in the schema twice. This PR fixes this. It has currently not been a problem because most (all?) bindings use maps/dictionaries to store properties which effectively makes this impossible. But it is possible to get into this situation using the C++ API's.

With this PR we now validate that no duplicate names, aliases or conflicts between the two exists. Unfortunately, this runs in `(O(n*log(n) + n) + (O(n*log(n) + n) + O(n^2)` in order to avoid duplicated properties being reported multiple times. This is similar to how duplicated types  are reported in `schema.cpp`

This runtime penalty is a bit heavy, so I'm not 100% sure we want to keep this behaviour. Some alternatives could  be:

* Remove the check for duplicated names and document it since all bindings already handle this by using appropriate dictionary types.  This will remove a `(O(n*log(n) + n)` check.

* Remove the check for conflicts between names and aliases. Nothing prevents us from supporting this expect perhaps users being confused by reported names mapping to different things. Removing this would remove an `O(n^2)`

* We could combine all 3 checks into one `O(n^2)` run, the downside would be that duplicated names and aliases are reported multiple times, once for each duplicate.

* The minimal check we would need is checking for duplicated aliases which are not caught by using dictionaries in bindings. Only having this would run in `(O(n*log(n) + n)`.

For now, I have opted for the most user-friendly approach, but thoughts on the above are appreciated.

**Performance update**

The checks now run in pure `O(n^2)` and then we only sort any name/alias conflicts afterwards.

### DIFF
--- a/src/object_schema.cpp
+++ b/src/object_schema.cpp
@@ -108,27 +108,27 @@ Property *ObjectSchema::property_for_name(StringData name)
     return nullptr;
 }
 
-Property *ObjectSchema::property_for_alias(StringData alias)
+Property *ObjectSchema::property_for_public_name(StringData public_name)
 {
-    // If no alias is defined it is the equivalent of name = alias.
+    // If no `public_name` is defined, the internal `name` is also considered the public name.
     for (auto& prop : persisted_properties) {
-        if ((prop.alias.empty() ? StringData(prop.name) :  StringData(prop.alias)) == alias)
+        if ((prop.public_name.empty() ? StringData(prop.name) :  StringData(prop.public_name)) == public_name)
             return &prop;
     }
 
-    // Computed properties are not persisted, so creating an alias for such properties
-    // are a bit pointless since the internal name is already the "alias", but since
+    // Computed properties are not persisted, so creating a public name for such properties
+    // are a bit pointless since the internal name is already the "public name", but since
     // this distinction isn't visible in the Property struct we allow it anyway.
     for (auto& prop : computed_properties) {
-        if ((prop.alias.empty() ? StringData(prop.name) :  StringData(prop.alias)) == alias)
+        if ((prop.public_name.empty() ? StringData(prop.name) :  StringData(prop.public_name)) == public_name)
             return &prop;
     }
     return nullptr;
 }
 
-const Property *ObjectSchema::property_for_alias(StringData alias) const
+const Property *ObjectSchema::property_for_public_name(StringData public_name) const
 {
-    return const_cast<ObjectSchema *>(this)->property_for_alias(alias);
+    return const_cast<ObjectSchema *>(this)->property_for_public_name(public_name);
 }
 
 const Property *ObjectSchema::property_for_name(StringData name) const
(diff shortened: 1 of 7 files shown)


cmelchior:
There seems to be a MacOS CI configuration failure somewhere, but since the rest of the tests parses this PR should be ready for review.

src/property.hpp
```diff
@@ -62,7 +62,22 @@ struct Property {
...
+    // If no alias is defined a property `alias` is equal to its `name`.
+    //
+    // ObjectStore will ensure that no conflicts occur between persisted properties and the alias, so
+    // an alias is just as unique an identifier as the property name in the file.
+    //
+    // In order to respect aliases bindings should use `ObjectSchema::property_for_alias()`
+    // and XXX for read fields defined by an alias.
+    // If alias are allowed in queries, an XXX must be parsed to XXX. It is up to bindings to
+    // keep track of this.
+    std::string alias;
```
kneth: So only one alias?

cmelchior: Yes, I don't think there is any use case for having multiple?

src/object_schema.cpp
```diff
@@ -107,11 +108,31 @@ Property *ObjectSchema::property_for_name(StringData name) {
     return nullptr;
 }
 
-const Property *ObjectSchema::property_for_name(StringData name) const {
+Property *ObjectSchema::property_for_alias(StringData alias)
+{
+    for (auto& prop : persisted_properties) {
+        if (StringData(prop.alias.empty() ? prop.name : prop.alias) == alias)
```
ironage: Shouldn't we check for equality on both the name and the alias? Currently if an alias is set on a property, but the actual name is passed, a match won't be found.

`if (StringData(prop.name) == alias || (!prop.alias.empty() && StringData(prop.alias) == alias)` ?

tgoyne:
What is this actually going to be used for? The linked issues all just say that it's needed for subscription updating and it's not at all obvious to me why it would be, or even why it'd be useful for that considering that in JS Subscription objects aren't managed objects.

src/object_schema.cpp
```diff
@@ -107,11 +108,31 @@ Property *ObjectSchema::property_for_name(StringData name) {
...
+            return &prop;
+    }
+    for (auto& prop : computed_properties) {
+        if (StringData(prop.alias.empty() ? prop.name : prop.alias) == alias)
```
tgoyne: Supporting both an alias and a name for computed properties doesn't make much sense, since there is no associated column name for computed properties.

tests/schema.cpp
```diff
@@ -26,6 +26,7 @@
 #include <realm/descriptor.hpp>
 #include <realm/group.hpp>
 #include <realm/table.hpp>
+#include <util/test_file.hpp>
```
tgoyne: This header is part of the test project, not a system header.

src/property.hpp
tgoyne: Do any of the bindings already call this an alias? It'd be nice to avoid introducing yet another totally different concept for how column names and property names relate to each other.

tgoyne:
WRT the performance of duplicate detection, we can take advantage of that performance only matters in the case where there are no duplicates. This means that you can just do something like dump all the names/aliases into a vector, sort it, check for dupes, and then only do slower things that let you usefully report what the errors are if there are any.

src/object_schema.cpp
cmelchior: True, the alias and name seem to be the same thing for computed properties. However, supporting it doesn't seem to be directly harmful, whereas not supporting it would complicate using the schema API.

src/property.hpp
cmelchior: See https://realmio.slack.com/archives/CA41SMVPV/p1554116647000200

Both Java and .NET support this, but neither call it alias. Instead, they talk about mapping from public names to internal names using slightly different terminology. Java uses `@RealmField("name")` and .NET `[MapTo("name)]`.

Since they both are about mapping public names to internal names, I should probably reword this. However, since the `Property` works on internal names already I probably prefer the word `publicName` as it feels more descriptive. Thoughts?

src/object_schema.cpp
```diff
@@ -107,11 +108,36 @@ Property *ObjectSchema::property_for_name(StringData name) {
...
+    for (auto& prop : persisted_properties) {
+        if ((prop.public_name.empty() ? StringData(prop.name) :  StringData(prop.public_name)) == public_name)
+            return &prop;
+    }
+
+    // Computed properties are not persisted, so creating a public name for such properties
+    // are a bit pointless since the internal name is already the "public name", but since
+    // this distinction isn't visible in the Property struct we allow it anyway.
+    for (auto& prop : computed_properties) {
+        if ((prop.public_name.empty() ? StringData(prop.name) :  StringData(prop.public_name)) == public_name)
```
kneth: For clarity, it might be worth splitting them into two separate statements. I mean, having a `?` in an `if` (branching in the condition for branching) :smile:

cmelchior:
Ready for final review.

See example of usage here: https://github.com/realm/realm-js/pull/2306/files#diff-40fa1bcc04a39e465f1572b1178dfcfbR377

src/property.hpp
tgoyne: `name` (a.k.k `internal_name` but renaming it would be really annoying) +`public_name` seems fine to me. If we stick with the constraint that there needs to still be a field named `name` which means roughly what it currently does to avoid creating a ton of extra work for ourselves I don't think there's a *great* option.

src/object_schema.cpp
```diff
@@ -214,6 +245,78 @@ static void validate_property(Schema const& schema,
 
 void ObjectSchema::validate(Schema const& schema, std::vector<ObjectSchemaValidationException>& exceptions) const
 {
+
```
tgoyne: A rewrite of this which reduces the runtime from O(N^2) to O(N log N):

```c++
    std::vector<StringData> public_property_names;
    std::vector<StringData> internal_property_names;
    internal_property_names.reserve(persisted_properties.size() + computed_properties.size());
    auto gather_names = [&](auto const& properties) {
        for (auto const& prop : properties) {
            internal_property_names.push_back(prop.name);
            if (!prop.public_name.empty())
                public_property_names.push_back(prop.public_name);
        }
    };
    gather_names(persisted_properties);
    gather_names(computed_properties);
    std::sort(public_property_names.begin(), public_property_names.end());
    std::sort(internal_property_names.begin(), internal_property_names.end());

    // Check that property names and aliases are unique
    auto for_each_duplicate = [](auto&& container, auto&& fn) {
        auto end = container.end();
        for (auto it = std::adjacent_find(container.begin(), end); it != end; it = std::adjacent_find(it + 2, end))
            fn(*it);
    };
    for_each_duplicate(public_property_names, [&](auto property_name) {
        exceptions.emplace_back("Alias '%1' appears more than once in the schema for type '%2'.",
                                property_name, name);
    });
    for_each_duplicate(internal_property_names, [&](auto internal_name) {
        exceptions.emplace_back("Property '%1' appears more than once in the schema for type '%2'.",
                                internal_name, name);
    });

    // Check that no aliases conflict with property names
    struct ErrorWriter {
        ObjectSchema const& os;
        std::vector<ObjectSchemaValidationException>& exceptions;

        struct Proxy {
            ObjectSchema const& os;
            std::vector<ObjectSchemaValidationException>& exceptions;
            Proxy& operator=(StringData name) {
                exceptions.emplace_back("Property '%1.%2' has an alias '%3' that conflicts with a property of the same name.",
                                        os.name, os.property_for_public_name(name)->name, name);
                return *this;
            }
        };
        Proxy operator*() { return Proxy{os, exceptions}; }
        ErrorWriter& operator++() { return *this; }
        ErrorWriter& operator++(int) { return *this; }
    } writer{*this, exceptions};
    std::set_intersection(public_property_names.begin(), public_property_names.end(),
                          internal_property_names.begin(), internal_property_names.end(), writer);
```

This makes one of the tests fail due to adding the object type to the error message and changing the order that alias/name collisions are reported in.

src/object_schema.cpp
```diff
@@ -214,6 +240,62 @@ static void validate_property(Schema const& schema,
...
+    // Check that property names and aliases are unique
+    auto for_each_duplicate = [](auto &&container, auto &&fn) {
+        auto end = container.end();
+        for (auto it = std::adjacent_find(container.begin(), end); it != end; it = std::adjacent_find(it + 2, end))
+            fn(*it);
+    };
+    for_each_duplicate(public_property_names, [&](auto public_property_name) {
+        exceptions.emplace_back("Alias '%1' appears more than once in the schema for type '%2'.",
+                                public_property_name, name);
```
nhachicha: where's 'name' defined?

src/object_schema.cpp
```diff
@@ -214,6 +240,62 @@ static void validate_property(Schema const& schema,
...
+    };
+    gather_names(persisted_properties);
+    gather_names(computed_properties);
+    std::sort(public_property_names.begin(), public_property_names.end());
+    std::sort(internal_property_names.begin(), internal_property_names.end());
+
+    // Check that property names and aliases are unique
+    auto for_each_duplicate = [](auto &&container, auto &&fn) {
+        auto end = container.end();
+        for (auto it = std::adjacent_find(container.begin(), end); it != end; it = std::adjacent_find(it + 2, end))
```
nhachicha: you'll get multiple calls to `fn` if there are more than 4 duplicates...
Have you considered using `std::unique`?

src/object_schema.cpp
```diff
@@ -214,6 +240,62 @@ static void validate_property(Schema const& schema,
...
+                exceptions.emplace_back(
+                        "Property '%1.%2' has an alias '%3' that conflicts with a property of the same name.",
+                        os.name, os.property_for_public_name(name)->name, name);
+                return *this;
+            }
+        };
+
+        Proxy operator*() { return Proxy{os, exceptions}; }
+        ErrorWriter &operator=(const ErrorWriter &) { return *this; }
+        ErrorWriter &operator++() { return *this; }
```
nhachicha: why overloading these operators? `++` & `++(int)`

src/property.hpp
```diff
@@ -62,7 +62,24 @@ struct Property {
     using IsPrimary = util::TaggedBool<class IsPrimaryTag>;
     using IsIndexed = util::TaggedBool<class IsIndexedTag>;
 
+    // The internal column name used in the Realm file.
     std::string name;
+
+    // The public name used by the binding to represent the internal column name in the Realm file. Bindings can use
+    // this to expose a different name in the binding API, e.g. to map between different naming conventions.
+    //
+    // Public names are only ever used defined, they are not persisted on disk, so reading the schema from the file
```
nhachicha: user defined?

src/object_schema.cpp
tgoyne: It is a member variable of ObjectSchema.

src/object_schema.cpp
tgoyne: set_intersection takes an OutputIterator, and OutputIterator has a very bad interface. It's a very awkward subset of what you can do with a pointer so that passing a pointer to a buffer that should be written to is a legal OutputIterator, while also constraining what can be done with it so that no-op increment works as well.

src/object_schema.cpp
tgoyne: `std::unique` doesn't tell you what the duplicates were. Maybe something like `for (auto it = std::adjacent_find(container.begin(), end); it != end; it = std::adjacent_find(std::find_if(it + 2, end, [&](auto&& v) { v != *it; }, end))`

cmelchior: I added unit tests for this. The current implementation does the correct thing also, if there are more than 4 duplicates.

PR #789, opened 2019-04-22
Add support for updating the configuration of an existing sync session

tgoyne:
Required for https://github.com/realm/realm-cocoa/pull/6136.

### DIFF
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -811,7 +811,6 @@ void SyncSession::set_multiplex_identifier(std::string multiplex_identity)
     m_multiplex_identity = std::move(multiplex_identity);
 }
 
-
 SyncSession::PublicState SyncSession::get_public_state() const
 {
     if (m_state == nullptr) {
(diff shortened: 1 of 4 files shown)


src/sync/sync_session.cpp
```diff
@@ -838,14 +836,34 @@ SyncSession::PublicState SyncSession::state() const
...
         default:
             REALM_UNREACHABLE();
     }
 }
 
+void SyncSession::update_configuration(SyncConfig new_config)
+{
+    while (true) {
+        std::unique_lock<std::mutex> lock(m_state_mutex);
+        if (m_state != &State::inactive) {
```
cmelchior: Why have this condition + loop? Looking at the implementation for `advance_state` it sets the new state immediately? I'm not 100% sure what happens to the `m_session` variable, but if its destroying is delayed, this isn't caught by this anyway.

tgoyne: Advancing to Inactive releases the lock, so after reacquiring the lock we have to recheck that the state did not get switched away from active while we weren't holding the lock.

tgoyne:
I added some not-very-good tests for this. I'm not really sure how to test that the new config actually makes it to the sync::Session with the tools we have.

I also fixed some unrelated tests that were sometimes failing when I ran the tests locally.

PR #790, opened 2019-04-24
Create columns rather than assert their existence

nirinchev:
When we explicitly specify the schema for the `__ResultSets` table, we may want to specify a subset of the properties in the SDK (e.g. `query_parse_counter` is not something we care about). In this case, rather than assert that the columns were created by someone, we can just go ahead and create those that are missing.

This is based on https://github.com/realm/realm-object-store/pull/782 because I was working on the .NET implementation of subscription inclusions when I ran into this.

### DIFF
--- a/src/sync/partial_sync.cpp
+++ b/src/sync/partial_sync.cpp
@@ -125,19 +125,26 @@ void initialize_schema(Group& group)
     if (!table) {
         // Create the schema required by Sync
         table = sync::create_table(group, result_sets_table_name);
-        table->add_column(type_String, property_query);
-        table->add_column(type_String, property_matches_property_name);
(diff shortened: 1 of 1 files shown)


nirinchev:
This fails consistently on CI with an assertion failure in network.hpp. I can't see how this is caused by a change in this PR - @tgoyne or @cmelchior as the OS pros here, maybe you have a theory?

tgoyne:
It seems to be crashing in a bunch of different places when I run it several times, and I have no idea why as sync-4.x seems to be passing consistently and the change here shouldn't have a functional impact on the tests. It works when I run it locally with the same version of core/sync too.

nirinchev:
No more QBS, so no longer relevant.

PR #796, opened 2019-05-17
Fix using update_only_diff on nullable primitive properties

tgoyne:
`table.get<int64_t>()` and friends don't work on nullable int properties and would hit an assertion failure. Fix this by switching to using `switch_on_type()` which dispatches to `Optional<T>` where applicable.

### DIFF
--- a/src/list.hpp
+++ b/src/list.hpp
@@ -270,8 +270,7 @@ void List::assign(Context& ctx, T&& values, bool update, bool update_only_diff)
             index++;
         });
         while (index < sz) {
-            remove(index);
-            sz--;
+            remove(--sz);
         }
     }
     else {
--- a/src/object_accessor.hpp
+++ b/src/object_accessor.hpp
@@ -69,14 +69,36 @@ ValueType Object::get_property_value(ContextType& ctx, StringData prop_name)
 }
 
 namespace {
-template <class T, typename ValueType, typename ContextType>
-inline void do_update_value(ContextType& ctx, Table& table, ValueType& value, size_t col, size_t row, bool update_only_diff, bool is_default)
-{
-    auto new_val = ctx.template unbox<T>(value);
-    if (!update_only_diff || table.get<T>(col, row) != new_val) {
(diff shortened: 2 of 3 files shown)


src/list.hpp
```diff
@@ -270,8 +270,7 @@ void List::assign(Context& ctx, T&& values, bool update, bool update_only_diff)
             index++;
         });
         while (index < sz) {
-            remove(index);
-            sz--;
+            remove(--sz);
```
ironage: nice optimisation

src/object_accessor.hpp
```diff
@@ -69,14 +69,36 @@ ValueType Object::get_property_value(ContextType& ctx, StringData prop_name)
...
+    bool try_update;
+    bool update_only_diff;
+    bool is_default;
+
+    void operator()(RowExpr*)
+    {
+        ContextType child_ctx(ctx, property);
+        auto curr_link = row.get_link(col);
+        auto link = child_ctx.template unbox<RowExpr>(value, true, try_update, update_only_diff, curr_link);
+        if (!update_only_diff || curr_link != link.get_index()) {
```
ironage: I think null links require special handling? Or is there a test that proves they don't?

tgoyne: They're handled before we get here in both the old and new versions of the code.

src/object_accessor.hpp
```diff
@@ -114,42 +136,9 @@ void Object::set_property_value_impl(ContextType& ctx, const Property &property,
...
-            do_update_value<Timestamp>(ctx, table, value, col, row, update_only_diff, is_default);
-            break;
-        case PropertyType::Any:
-            throw std::logic_error("not supported");
-        default:
-            REALM_COMPILER_HINT_UNREACHABLE();
-    }
+    ValueUpdater<ValueType, ContextType> updater{ctx, property, value,
+        table.get(row),col, try_update, update_only_diff, is_default};
+    switch_on_type(property.type, updater);
```
kneth: Nice simplification!

PR #799, opened 2019-06-06
Better support for asyncOpen

cmelchior:
This PR adds a new class extending the capabilities of asyncOpen. In particular it makes it possible to listen to progress as well as cancel it.

This is a prerequisite for https://github.com/realm/realm-js/pull/2381


src/sync/async_open_task.cpp
```diff
@@ -0,0 +1,74 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+#include "impl/realm_coordinator.hpp"
+#include "sync/async_open_task.hpp"
+#include "sync/sync_manager.hpp"
+#include "sync/sync_session.hpp"
+
+namespace realm {
+
+AsyncOpenTask::AsyncOpenTask(std::string realm_path):
```
tgoyne: RealmCoordinator should just pass these values in rather than looking them up again from the path.

This file doesn't follow the objectstore code formatting (leading colon/comma for initializer lists, newline before { for function definitions but nothing else).

src/sync/async_open_task.cpp
```diff
@@ -0,0 +1,74 @@
...
+ m_coordinator(_impl::RealmCoordinator::get_coordinator(realm_path)),
+ m_session(SyncManager::shared().get_existing_session(realm_path))
+{
+}
+
+void AsyncOpenTask::start(std::function<void(std::shared_ptr<Realm>, std::exception_ptr)> callback) {
+    m_session->wait_for_download_completion([callback, this](std::error_code ec) {
```
tgoyne: This callback needs to capture a weak_ptr to the task and verify that it still exists. Currently calling `cancel()` and then destroying the AsyncOpenTask will result in a use-after-free when the download actually finishes.

src/impl/realm_coordinator.cpp
```diff
@@ -300,6 +300,21 @@ void RealmCoordinator::get_realm(Realm::Config config,
     callback(realm, nullptr);
 }
 
+#if REALM_ENABLE_SYNC
+std::shared_ptr<AsyncOpenTask> RealmCoordinator::get_synchronized_realm(Realm::Config config)
+{
+    if (config.sync_config) {
+        std::unique_lock<std::mutex> lock(m_realm_mutex);
+        set_config(config);
+        create_sync_session(!File::exists(m_config.path));
```
tgoyne: I don't think this check is correct? If the file already exists but no sync_session exists then the returned AsyncOpenTask will just do nothing.

src/sync/async_open_task.cpp
```diff
@@ -0,0 +1,74 @@
...
+        m_canceled = true;
+        m_session->log_out();
+        m_session = nullptr;
+        m_coordinator = nullptr;
+    }
+}
+
+uint64_t AsyncOpenTask::register_download_progress_notifier(std::function<SyncProgressNotifierCallback> callback) {
+    if (m_session) {
+        m_session->register_progress_notifier(callback, realm::SyncSession::NotifierType::download, false);
```
tgoyne: missing `return`

src/sync/async_open_task.hpp
```diff
@@ -0,0 +1,52 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#pragma once
```
tgoyne: Object store uses header guards rather than `#pragma once`.

tests/sync/async_open.cpp
```diff
@@ -0,0 +1,115 @@
...
+        {"link_target", {
+            {"id", PropertyType::Int}
+        },{
+            {"parents", PropertyType::LinkingObjects|PropertyType::Array, "object_a", "link"},
+        }}
+    };
+}
+}
+
+TEST_CASE("Get Realm using Async Open", "[asyncOpen]") {
```
tgoyne: Tests for this should just be in realm.cpp, replacing the existing "SharedRealm: get_shared_realm() with callback" tests.

src/shared_realm.hpp
```diff
@@ -21,6 +21,9 @@
 
 #include "execution_context_id.hpp"
 #include "schema.hpp"
+#if REALM_ENABLE_SYNC
+#include "sync/async_open_task.hpp"
```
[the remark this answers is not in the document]
cmelchior: Trying to forward declare `AsyncOpenTask` results in the following error:

```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/type_traits:3141:38: error: incomplete type 'realm::AsyncOpenTask' used in type trait expression
    : public integral_constant<bool, __is_constructible(_Tp, _Args...)>
                                     ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/memory:4271:20: note: in instantiation of template class 'std::__1::is_constructible<realm::AsyncOpenTask, std::__1::basic_string<char> &>' requested here
    static_assert( is_constructible<_Tp, _Args...>::value, "Can't construct object in make_shared" );
                   ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/memory:4656:29: note: in instantiation of function template specialization 'std::__1::shared_ptr<realm::AsyncOpenTask>::make_shared<std::__1::basic_string<char> &>' requested here
    return shared_ptr<_Tp>::make_shared(_VSTD::forward<_Args>(__args)...);
                            ^
/Users/cm/Realm/realm-object-store-new/src/impl/realm_coordinator.cpp:310:17: note: in instantiation of function template specialization 'std::__1::make_shared<realm::AsyncOpenTask, std::__1::basic_string<char> &>' requested here
    return std::make_shared<AsyncOpenTask>(m_config.path);
                ^
/Users/cm/Realm/realm-object-store-new/src/shared_realm.hpp:35:7: note: forward declaration of 'realm::AsyncOpenTask'
class AsyncOpenTask;
```

I have no idea how to resolve this.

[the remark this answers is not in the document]
tgoyne: realm_coordinator.cpp needs to include async_open_task.hpp

src/impl/realm_coordinator.cpp
cmelchior: Not sure I understand? That method will always create the sync session, the boolean just forces a client reset (and thus faster download)? It is the same implementation as in your implementation of `get_shared_realm(config, callback)`

tgoyne: Yeah that comment's just wrong.

cmelchior:
Replaced by https://github.com/realm/realm-object-store/pull/803

PR #805, opened 2019-06-19
Update protocol error for sync 4.6.0

nhachicha:
needed for https://github.com/realm/realm-java/pull/6540


tgoyne:
This needs to bump the sync version (in dependencies.list) to a version that actually has these error codes.

PR #806, opened 2019-06-19
Fix some async open issues

tgoyne:
This includes changes needed to get the realm-js and realm-cocoa tests passing using the new async open, and fixes a bug which hits Cocoa which wasn't caught by those tests.

AsyncOpenTask now takes more control over its lifetime: the callback holds a strong reference to the task until it's called so that retaining the AsyncOpenTask externally is optional, and cleans itself up after calling the callback passed to `start()` so that the SDK code doesn't have to worry about that. Some thread-safety problems around using `m_session` from multiple threads have been fixed by using `AtomicSharedPtr<>`, and calling `cancel()` while it happens to be processing the callback will no longer crash.

The state dir for async open/resync is now a directory named after the Realm rather than a single shared directory, since it turns out that's what the sync code expects. Sharing a single directory resulted in async opening more than one Realm at a time breaking.

The AsyncOpenTask callback is now passed a `ThreadSafeReference<Realm>` rather than `std::shared_ptr<Realm>`. The TSR can be resolved to an actual Realm by passing it to `Realm::get_shared_realm()`. This avoids creating the EventLoopSignal for the Realm on the background thread (which doesn't work with the libuv implementation because libuv event loops aren't thread-safe), and lets the JS SDK skip closing and re-opening the Realm back on the main thread. This also happened to reveal that the background-opened Realm wasn't really in a valid state, so a few fixes were needed there.

Async open failed to check that the history schema for existing Realm files was compatible before creating the sync session, which resulted in an exception being thrown on the sync client thread that did not get funneled back to the appropraite place.

### DIFF
--- a/src/sync/async_open_task.cpp
+++ b/src/sync/async_open_task.cpp
@@ -38,7 +38,7 @@ void AsyncOpenTask::start(std::function<void(ThreadSafeReference<Realm>, std::ex
         return;
 
     std::shared_ptr<AsyncOpenTask> self(shared_from_this());
-   session->wait_for_download_completion([callback, self, this](std::error_code ec) {
+    session->wait_for_download_completion([callback, self, this](std::error_code ec) {
        auto session = m_session.exchange(nullptr);
         if (!session)
             return; // Swallow all events if the task as been canceled.
--- a/src/sync/async_open_task.cpp
+++ b/src/sync/async_open_task.cpp
@@ -33,17 +33,19 @@ AsyncOpenTask::AsyncOpenTask(std::shared_ptr<_impl::RealmCoordinator> coordinato
 
 void AsyncOpenTask::start(std::function<void(ThreadSafeReference<Realm>, std::exception_ptr)> callback)
 {
+    auto session = m_session.load();
+    if (!session)
+        return;
+
     std::shared_ptr<AsyncOpenTask> self(shared_from_this());
-    m_session->wait_for_download_completion([callback, self](std::error_code ec) {
-        if (self->m_canceled)
+   session->wait_for_download_completion([callback, self, this](std::error_code ec) {
+       auto session = m_session.exchange(nullptr);
+        if (!session)
             return; // Swallow all events if the task as been canceled.
 
-        // Release our references to the session and coordinator after calling
-        // the callback
-        auto session = std::move(self->m_session);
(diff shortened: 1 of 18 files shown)


src/sync/async_open_task.cpp
```diff
@@ -29,45 +31,49 @@ AsyncOpenTask::AsyncOpenTask(std::shared_ptr<_impl::RealmCoordinator> coordinato
...
-    m_session->wait_for_download_completion([callback, weak_self](std::error_code ec) {
-        if (auto self = weak_self.lock()) {
-            if (self->m_canceled)
-                return; // Swallow all events if the task as been canceled.
+    auto session = m_session.load();
+    if (!session)
+        return;
+
+    std::shared_ptr<AsyncOpenTask> self(shared_from_this());
+   session->wait_for_download_completion([callback, self, this](std::error_code ec) {
```
cmelchior:
```suggestion
    session->wait_for_download_completion([callback, self, this](std::error_code ec) {
```

src/sync/partial_sync.cpp
```diff
@@ -176,7 +141,33 @@ void initialize_schema(Group& group)
...
+        if (realm.is_in_transaction())
+            realm.cancel_transaction();
+        if (!was_in_read)
+            realm.invalidate();
+    });
+
+    auto& group = realm.read_group();
+    // Check if the result sets table already has the expected number of columns
+    auto table = ObjectStore::table_for_object_type(group, result_sets_type_name);
+    if (table && table->size() >= 10)
```
cmelchior: Perhaps move the magic constant 10 into a variable with docs, it is pretty opaque what it means.

tests/sync/permission.cpp
```diff
@@ -93,12 +94,17 @@ static void subscribe_to_all(std::shared_ptr<Realm> const& r)
...
         {"matches_property", "object_matches"s},
         {"query", "TRUEPREDICATE"s},
         {"status", int64_t(0)},
         {"error_message", ""s},
         {"query_parse_counter", int64_t(0)},
         {"matches_count", int64_t(0)},
+        {"created_at", Timestamp(0, 0)},
+        {"updated_at", Timestamp(0, 0)},
+        {"expires_at", Timestamp()},
+        {"ttl", Timestamp()},
```
cmelchior: It is currently being called `time_to_live`

Issue #685, opened 2018-08-13
Core6: cache_new_schema refactor throwing a logic_error


nhachicha:
The current implementation of commit in Java calls `shared_realm->commit_transaction();` this will call `m_coordinator->commit_write(*this);` which triggers any registered listeners, say we close the Realm inside the listener, then the OS calls `cache_new_schema();` which in the [new Core6 porting](https://github.com/realm/realm-object-store/blob/core-6/src/shared_realm.cpp#L428) calls `transaction()` -> `read_group()`  then calls `verify_open()` which  fails obviously since the Realm was closed.

The old implementation of  `Realm::cache_new_schema()` simply guarded against this use case by checking at the beginning if the shared_realm is still valid

```C++
if (!m_shared_group)
        return;
```

Note: The point was discussed with @tgoyne and he think this behaviour (relying on transaction needs a refactor)
>Thomas Goyne
`transaction()` is an incorrect hack and nearly everything that uses it is broken in some degree due to it not actually having the same semantics as what it replaced
... I need to change a lot of things related to it eventually

@jedelbo can you look at this while Thomas is off ?

jedelbo:
If Thomas thinks this needs a bigger overhaul, then there is nothing I can do.

nhachicha:
No longer reproducible, could be fixed by https://github.com/realm/realm-object-store/pull/811

PR #811, opened 2019-07-03
Adding check for cache_new_realm

nhachicha:
[cache_new_realm](https://github.com/realm/realm-object-store/blob/master/src/shared_realm.cpp#L574-L577) used to check if the shared_group is still open/valid before proceeding to read the current transaction version id. In Core 6 this checks [is not done](https://github.com/realm/realm-object-store/blob/core-6/src/shared_realm.cpp#L430-L432), which causes the call to `transaction()` to throw. (`transaction()` -> `read_group()` -> `verify_open()` -> Throw `ClosedRealmException`). This PR adds the check to return quickly if the Realm was closed.

### DIFF
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -429,14 +429,16 @@ void Realm::add_schema_change_handler()
 
 void Realm::cache_new_schema()
 {
-    auto new_version = transaction().get_version_of_current_transaction().version;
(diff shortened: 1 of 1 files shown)


nhachicha:
also fixes https://github.com/realm/realm-object-store/issues/685

PR #818, opened 2019-08-07
Use fifo functionality now in Core

cmelchior:
Depends on https://github.com/realm/realm-core/pull/3328 being merged and released.

Remove the code in OS that now live in Core so we only define creating named pipes in one location (as there are a number of bugs we need to account for).


cmelchior:
Replaced by #827

Issue #688, opened 2018-08-15
Core6: transaction.import_copy_of crash when deleting parent of LinkView


nhachicha:
I have some tests involving deleting the parent of LinkView failing:

Example
[accessors_resultsBuiltOnDeletedLinkView_deletionAsALocalCommit](https://github.com/realm/realm-java/blob/3aaf5db578083ffacf858bdb95075ddfd7fa3458/realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java#L286)
and
[changeListener_onResultsBuiltOnDeletedLinkView](https://github.com/realm/realm-java/blob/20a091cfc7da20cf4bb8115a20228d1a47c86376/realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTests.java#L1082)

both cover the same logic, say you have type `Owner` that contains a`RealmList` of `Dog`,  you query for all `dogs` of this owner, then register a listener on the `RealmResults`. Now delete all owners, this will trigger the listeners, but accessing the the list of `dogs` (ex: size) will crash. I traced back this to be triggered by the call to [m_notifier->get_tableview(m_table_view);](https://github.com/realm/realm-object-store/blame/6ceb4a1c0b4ad10163149c083169506895db6553/src/results.cpp#L278) introduced in https://github.com/realm/realm-object-store/commit/eafdd48580816f02cc8a1da743421c4e6e988f99#diff-c0354faf99b53cc5d3c9e6a58ed9ae85 

Here's the crash 
```
08-12 12:17:49.093  5407  5425 F libc    : Fatal signal 11 (SIGSEGV), code 1, fault addr 0x29c in tid 5425 (RunTestInLooper)
08-12 12:17:49.107  5430  5430 I crash_dump32: obtaining output fd from tombstoned
08-12 12:17:49.108   144   144 I /system/bin/tombstoned: received crash request for pid 5407
08-12 12:17:49.108  5430  5430 I crash_dump32: performing dump of process 5407 (target tid = 5425)
08-12 12:17:49.108  5430  5430 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
08-12 12:17:49.108  5430  5430 F DEBUG   : Build fingerprint: 'Android/vbox86p/vbox86p:8.0.0/OPR6.170623.017/32:userdebug/test-keys'
08-12 12:17:49.108  5430  5430 F DEBUG   : Revision: '0'
[...]

PR #821, opened 2019-08-16
Attempt to reproduce #688

cmelchior:


cmelchior:
@jedelbo found the bug and implement a fix here: https://github.com/realm/realm-core-private/issues/328

PR #823, opened 2019-08-20
Port server components to core 6

kneth:


src/server/adapter.cpp
```diff
@@ -60,10 +60,10 @@ static PropertyType from_core_type(DataType type)
         case type_String:    return PropertyType::String;
         case type_Binary:    return PropertyType::Data;
         case type_Timestamp: return PropertyType::Date;
-        case type_Mixed:     return PropertyType::Any;
+        case type_OldMixed:     return PropertyType::Any;
```
tgoyne: Spacing is wrong.

src/server/admin_realm.hpp
```diff
@@ -22,6 +22,7 @@
 #include "results.hpp"
 #include "shared_realm.hpp"
 #include "sync/sync_config.hpp"
+#include "realm/object_id.hpp"
```
tgoyne: Don't need this header; just need to move the forward declaration of ObjectID out of the sync namespace.

src/server/global_notifier.cpp
```diff
@@ -166,8 +164,8 @@ void GlobalNotifier::Impl::register_realm(sync::ObjectID id, StringData path) {
             auto config = info->coordinator->get_config();
             config.force_sync_history = true; // FIXME: needed?
             config.schema = util::none;
-            Realm::open_with_config(config, info->history, info->shared_group, read_only_group, nullptr);
-            info->shared_group->begin_read(old_version);
+            auto realm = info->coordinator->get_realm();
```
tgoyne: We don't want to create a `Realm` here. It adds overhead and makes all of the other code more complicated because it's written to work with the Transaction directly.

src/server/global_notifier.cpp
```diff
@@ -392,29 +394,31 @@ std::unordered_map<std::string, CollectionChangeSet> const& GlobalNotifier::Chan
...
 
     _impl::TransactionChangeInfo info;
     info.track_all = true;
-    _impl::transaction::advance(*sg, info, m_new_version);
+    _impl::transaction::advance(realm->transaction(), info, m_new_version);
 
     m_changes.reserve(info.tables.size());
-    for (size_t i = 0; i < info.tables.size(); ++i) {
+    auto table_keys = g.get_table_keys();
+    int64_t i = 0;
```
tgoyne: This doesn't appear to be used anywhere?

src/server/global_notifier.cpp
```diff
@@ -392,29 +394,31 @@ std::unordered_map<std::string, CollectionChangeSet> const& GlobalNotifier::Chan
...
+    for (auto table_key : table_keys) {
         auto& change = info.tables[i];
         if (!change.empty()) {
-            auto name = ObjectStore::object_type_for_table_name(g.get_table_name(i));
+            auto table = g.get_table(table_key);
```
tgoyne: Doesn't appear to be used?

src/server/global_notifier.hpp
```diff
@@ -21,13 +21,11 @@
 
 #include "impl/collection_notifier.hpp"
 #include "shared_realm.hpp"
+#include "realm/object_id.hpp"
```
tgoyne: This is a core header and not an object store header. It also isn't needed here as it can just be forward-declared.

src/server/admin_realm.cpp
```diff
@@ -75,10 +68,13 @@ void AdminRealmListener::start()
...
+        auto path_col_key = table.get_column_key("path");
 
-        for (size_t i = 0, size = table.size(); i < size; ++i)
-            register_realm(id_for_row(group, table[i]), table.get_string(path_col_ndx, i));
+        for (auto it = table.begin(); it != table.end(); ++it) {
+            auto id = table.get_object_id(it->get_key());
+            auto obj = table.get_object(it->get_key());
+            register_realm(id,  obj.get<StringData>(path_col_key));
+        }
         return;
```
jedelbo:
```
        for (auto& o : table)
            register_realm(table.get_object_id(o.get_key()), o.get<String>(path_col_ndx));
```
is shorter

src/server/admin_realm.cpp
```diff
@@ -75,10 +68,13 @@ void AdminRealmListener::start()
...
-        size_t path_col_ndx = table.get_column_index("path");
+        auto path_col_key = table.get_column_key("path");
 
-        for (size_t i = 0, size = table.size(); i < size; ++i)
-            register_realm(id_for_row(group, table[i]), table.get_string(path_col_ndx, i));
+        for (auto it = table.begin(); it != table.end(); ++it) {
+            auto id = table.get_object_id(it->get_key());
+            auto obj = table.get_object(it->get_key());
+            register_realm(id,  obj.get<StringData>(path_col_key));
+        }
```
jedelbo: Be aware that `it` is a pointer to current Obj, so no need to get the object again.
```
for (auto& o : table)
    register_realm(table.get_object_id(o.get_key()), o.get<String>(path_col_ndx));
```
is shorter.

src/server/admin_realm.cpp
```diff
@@ -123,10 +119,11 @@ void AdminRealmListener::start()
                     return;
 
                 auto& group = self->m_results.get_realm()->read_group();
-                size_t path_col_ndx = self->m_results.get(0).get_column_index("path");
+                auto& table = self->m_results.get_tableview().get_parent();
```
jedelbo: `group` not used

src/server/admin_realm.cpp
```diff
@@ -142,19 +139,20 @@ void AdminRealmListener::start()
...
```
jedelbo: `group` not used

src/server/global_notifier.cpp
```diff
@@ -392,29 +394,31 @@ std::unordered_map<std::string, CollectionChangeSet> const& GlobalNotifier::Chan
...
     info.track_all = true;
-    _impl::transaction::advance(*sg, info, m_new_version);
+    _impl::transaction::advance(realm->transaction(), info, m_new_version);
 
     m_changes.reserve(info.tables.size());
-    for (size_t i = 0; i < info.tables.size(); ++i) {
+    auto table_keys = g.get_table_keys();
+    int64_t i = 0;
+    for (auto table_key : table_keys) {
         auto& change = info.tables[i];
```
jedelbo: I think info.tables should be indexed by table_key.value

kneth:
@jedelbo I believe that I have addressed all comments and suggestion. Please review again.

PR #824, opened 2019-08-20
Allows changing url_prefix of a created sync session

fealebenpae:
On Cloud we're doing away with the Sync Proxy. Previously sync clients would open websockets pointing to `my-awesome-instance.us1.cloud.realm.io/realm-sync/url-encoded-realm-path` and the server-side routing to different sync workers was done by the Sync Proxy sitting on the `/realm-sync` endpoint - it would inspect an incoming request, determine where to proxy it, and open another socket to do so.

We're adding a new mechanism that bypasses proxying and instead exposes every sync worker for an instance on a public path like `/sync/{worker-name}` so it would be possible for the client to make a request to `/sync/alpha/url-encoded-realm-path` and connect directly to the sync worker.

To this end token refresh responses by Cloud will now include the URL path of the sync worker in question which needs to go to the `url_prefix` field of the sync session configuration. However, by the time the first token refresh request is made bindings will have already created a sync session so we need the ability to set the url prefix after the session is created.

### DIFF
--- a/src/sync/sync_session.hpp
+++ b/src/sync/sync_session.hpp
@@ -213,6 +213,13 @@ public:
     // not make the session reconnect.
     void set_multiplex_identifier(std::string multiplex_identity);
 
+    // See SyncConfig::url_prefix
+    //
+    // This method allows to override this value after the session is created but before it is bound
+    // because of Realm Cloud's token refresh service returning the sync worker ingress path with the token response.
+    // Prefer using the SyncConfig field in all other cases.
+    void set_url_prefix(std::string url_prefix);
+
     // Inform the sync session that it should close.
     void close();
 
--- a/src/sync/sync_config.hpp
+++ b/src/sync/sync_config.hpp
@@ -138,6 +138,8 @@ struct SyncConfig {
     util::Optional<std::string> authorization_header_name = none;
     std::map<std::string, std::string> custom_http_headers;
 
+    // Set the URL path prefix sync will use when opening a websocket for this session. Default is `/realm-sync`.
+    // Useful when the sync worker sits behind a firewall or load-balancer that rewrites incoming requests.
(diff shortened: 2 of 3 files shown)


src/sync/sync_session.hpp
```diff
@@ -213,6 +213,13 @@ class SyncSession : public std::enable_shared_from_this<SyncSession> {
     // not make the session reconnect.
     void set_multiplex_identifier(std::string multiplex_identity);
 
+    // See SyncConfig::url_prefix
+    //
+    // This method allows to override this value after the session is created but before it is bound
+    // because of Realm Cloud's token refresh service returning the sync worker ingress path with the token response.
+    // Prefer using the SyncConfig field in all other cases.
+    void set_url_prefix(std::string url_prefix);
```
tgoyne: Since this value comes with the refresh token, shouldn't it just be an optional parameter to refresh_access_token()?

fealebenpae:
> Will the server ever want to change the URL prefix after a session is first connected (e.g. for worker fallover)?

No - the ingress routes on Cloud will automatically reflect the change: if the sync worker behind `/sync/alpha` goes down and fails over to its backup the route will go to the backup instead. There are going to be no cases where `sync_worker.path` in the response ever has a different value for a given realm after it's created on the server.

src/sync/sync_session.hpp
fealebenpae: I modeled this after `set_multiplex_identifier` which is similarly called with a value from the access token refresh response (the sync label, in this case).

I don't think it makes sense for the parameter to be passed on every call to `refresh_access_token()` - it's only needed once, when the session is initially bound after receiving the initial access token from the server. I am not sure, however, what's the best point to insert this in, so I just did what `set_multiplex_identifier` does.

Issue #798, opened 2019-06-06
Bring core-6 branch up-to-date with master


jedelbo:

kneth:
This issue is required in order to complete the Core v6 integrations in the SDKs.

jedelbo:
This is a moving target, but the original requirement was to get OS up to a state where the new async open was supported. This is now achieved.

PR #827, opened 2019-08-23
Use core's new fifo functions

kneth:
Realm Core v5.23.2 adds a couple of new FIFO functions.


nhachicha:
be careful, once merged [realm_coordinator.cpp](https://github.com/realm/realm-object-store/blob/core-6/src/impl/realm_coordinator.cpp#L44) in `core-6` branch needs to be updated as well to use the include from Core 
```Diff
--- a/src/impl/realm_coordinator.cpp
+++ b/src/impl/realm_coordinator.cpp
@@ -41,7 +41,7 @@
 #include <realm/db.hpp>
 #include <realm/history.hpp>
 #include <realm/string_data.hpp>
-#include <util/fifo.hpp>
+#include <realm/util/fifo_helper.hpp>
 
 #include <algorithm>
 #include <unordered_map>

```

kneth:
@nhachicha I hope that #798 will take of that :smile:.

cmelchior:
After restarting the test a few times, they still seem to fail with invalid top refs: `Unable to open a realm at path '/var/folders/01/

  qdkfxkd167338d33x5c6bt540000gl/T/realm-object-server/io.realm.object-server-

  utility/metadata/sync_metadata.realm': Invalid top array (ref: 31432, size:

  11) Path:.
`

workflow/test_coverage.sh
```diff
@@ -13,6 +13,12 @@ fi
 
 set -e
 
+if [ "${TMPDIR}" = "" ]; then
+  TMPDIR="/tmp"
+fi
+
+echo "TMPDIR: ${TMPDIR}"
+rm -rf ${TMP_DIR}/realm*
```
kneth:
```suggestion
rm -rf ${TMPDIR}/realm*
```

PR #828, opened 2019-08-23
Je/merge

jedelbo:


jedelbo:
@tgoyne Although the unittest seem to have passed here, I constantly get `Realm accessed from incorrect thread` in `SharedRealm: get_shared_realm() with callback` unit test. I belive it is because I don't know how the changes you made to `RealmCoordinator::get_realm()` in commit 1998e084ff8709f2aee0c08b7d80c635c2a34edc should be reflected in the Core-6 branch. Perhaps you can help here.

jedelbo:
Just realized that the test fails in exactly the same way on the3e48b69764c0a2aaaa7a3b947d6d0dae215f9a09 commit, so no need to fix this here I guess.

jedelbo:
The problem was fixed in the commits now merged.

PR #845, opened 2019-09-23
Update to catch v2.9.2

ironage:
Update catch from v1.12.2 to v2.9.2 ([release notes](https://github.com/catchorg/Catch2/blob/v2.9.2/docs/release-notes.md)). This update is motivated by getting access to the catch microbenchmarking functionality (introduced in v2.0.1).

As far as I can tell, we were not affected by any of the breaking changes (tests pass locally). However, comparison in `REQUIRE` macros now seem to strictly need arguments to be const which required a few changes, and means that this update is blocked on https://github.com/realm/realm-core/pull/3391.


tests/primitive_list.cpp
```diff
@@ -177,6 +176,20 @@ bool operator==(Results& results, std::vector<T> const& values) {
     }
     return true;
 }
+
+template<typename T>
+bool operator==(Results const& results, std::vector<T> const& values) {
```
tgoyne: This should just replace the non-const one since that'd no longer be used.

tgoyne:
Looks like this fails on android for the usual android reasons.

ironage:
@tgoyne could you take another look now that the tests are passing?
- I updated to catch on master and generate the single include manually for now, this is to obtain the unreleased fix for android builds. Catch will most likely make a release this weekend.
- There were a few failing tests on android CI, which is due to a new catch limitation, I added a workaround and documented it in the comments.

Jenkinsfile
```diff
@@ -148,7 +148,8 @@ stage('prepare') {
     checkout scm
     sh 'git clean -ffdx -e .????????'
     sshagent(['realm-ci-ssh']) {
-      sh 'git submodule update --init --recursive'
+      sh 'git submodule sync && git submodule update --init --recursive'
```
tgoyne: I don't think `submodule sync` should ever do anything on CI since we run from a fresh checkout every time?

ironage: For some reason, this was necessary when I tried changing over to my forked version of catch. Not sure why, maybe due to source caching in CI? However, that is no longer needed, so I'll remove it.

PR #849, opened 2019-09-26
Add work-around for Java schema bug

cmelchior:
This adds a required work-around for https://github.com/realm/realm-java/issues/6619
Unit tests are found in Java PR (https://github.com/realm/realm-java/pull/6621)

### DIFF
--- a/src/impl/realm_coordinator.cpp
+++ b/src/impl/realm_coordinator.cpp
@@ -25,6 +25,7 @@
 #include "binding_context.hpp"
 #include "object_schema.hpp"
 #include "object_store.hpp"
+#include "property.hpp"
 #include "schema.hpp"
 #include "thread_safe_reference.hpp"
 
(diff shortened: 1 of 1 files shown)


src/impl/realm_coordinator.cpp
```diff
@@ -279,6 +280,37 @@ void RealmCoordinator::do_get_realm(Realm::Config config, std::shared_ptr<Realm>
 
     realm_lock.unlock();
     if (schema) {
+#if REALM_ENABLE_SYNC && ANDROID
```
kneth: So it will also be enabled for React Native, right?

cmelchior: I suspect so, I don't think we have defines for each binding? Guess I'll need to add that.

kneth: In Realm JS we already have `REALM_PLATFORM_NODE` so I could add the flags  `REALM_PLATFORM_REACTNATIVE_ANDROID` and `REALM_PLATFORM_REACTNATIVE_IOS`.

cmelchior: Wouldn't it make more sense that I added it to Java so it would become:

```
#if REALM_ENABLE_SYNC && REALM_PLATFORM_JAVA
```

That way we also exclude Xamarin for Android?

PR #854, opened 2019-10-04
Core6 commenting unused code

nhachicha:
Commenting out unused code in OS which causes the compilation to fail in Java (since we use `-Werror -Wall -Wextra ` flags) 

Context https://mongodb.slack.com/archives/CKW2HG3M0/p1569856545059900 cc @ironage


kneth:
@nhachicha Is this still relevant?

ironage:
The code in question has been dealt with at some point in the last few months.

PR #856, opened 2019-10-09
Add support for frozen objects

cmelchior:
This PR is still very much WIP as it is unclear exactly how the bindings are going to use the feature. For now, it is adding things that make it work with Java.

There are some issues around preventing refreshes of frozen Realms that might mean we need to reconsider how Realms are cached. There where similar issues on the Java side.

There are probably a lot of code paths that can be optimized if we can treat a frozen Realm as immutable, but it is unclear how we best can achieve this.

Java PR: https://github.com/realm/realm-java/pull/6590

TODO
- [x] Add `is_frozen()` to Results, List and Object
- [x] Add unit tests for freezing Realms, 
- [x] Unit tests for freezing Results (in its many variants)
- [x] Unit tests for freezing Lists
- [x] Unit tests for freeing Objects
- [x] Add unit for throwing on transactions
- [x] Add unit tests for refresh
- [x] Add unit tests for wait_for_change
- [x] Add unit tests for adding changelisteners to Realm, Results, Lists and Object


src/list.cpp
```diff
@@ -445,6 +447,10 @@ NotificationToken List::add_notification_callback(CollectionChangeCallback cb) &
     return {m_notifier, m_notifier->add_callback(std::move(cb))};
 }
 
+List List::freeze(std::shared_ptr<Realm> frozen_realm) {
```
tgoyne: Newline before { for function definitions (and only function definitions).

src/results.cpp
```diff
@@ -945,6 +947,30 @@ REALM_RESULTS_TYPE(util::Optional<double>)
...
 #undef REALM_RESULTS_TYPE
 
+Results Results::freeze(SharedRealm frozen_realm) {
+    switch (m_mode) {
+        case Mode::Empty:
+            return Results();
+        case Mode::Table:
+            return Results(frozen_realm, *frozen_realm->transaction().import_copy_of(m_table));
+        case Mode::List: {
+            std::shared_ptr<LstBase> shared_list = std::move(frozen_realm->transaction().import_copy_of(*m_list));
```
tgoyne: This std::move() (and the one a few lines below) don't do anything.

src/results.cpp
```diff
@@ -945,6 +947,30 @@ REALM_RESULTS_TYPE(util::Optional<double>)
...
+            return Results(frozen_realm, shared_list, m_descriptor_ordering);
+        }
+        case Mode::LinkList: {
+            std::shared_ptr<LstBase> shared_list = std::move(frozen_realm->transaction().import_copy_of(*m_link_list));
+            return Results(frozen_realm, shared_list); // FIXME: Do we need to parse on query and sort descriptor?
```
tgoyne: This does need to do things for sort and distinct as those are implemented as a layer on top of the LnkLst.

src/shared_realm.cpp
```diff
@@ -484,11 +496,19 @@ static void check_write(const Realm* realm)
...
+        throw InvalidTransactionException(util::format("Number of active versions (%1) in the Realm exceeded the limit of %2",
+                realm->get_number_of_versions(),
+                realm->config().max_number_of_active_versions));
+    }
 }
 
 void Realm::verify_thread() const
 {
-    if (!m_execution_context.contains<std::thread::id>())
+    if (!m_execution_context.contains<std::thread::id>() || m_frozen_version)
```
tgoyne: Checking m_frozen_version first is a small but free perf gain as getting the current thread id is a little bit of work.

src/shared_realm.cpp
```diff
@@ -530,6 +557,8 @@ util::Optional<VersionID> Realm::current_transaction_version() const
     util::Optional<VersionID> ret;
     if (m_group) {
         ret = static_cast<Transaction&>(*m_group).get_version_of_current_transaction();
+    } else if (m_frozen_version) {
```
tgoyne: Newline after }

src/shared_realm.cpp
```diff
@@ -518,6 +538,13 @@ VersionID Realm::read_transaction_version() const
     return static_cast<Transaction&>(*m_group).get_version_of_current_transaction();
 }
 
+uint_fast64_t Realm::get_number_of_versions() const
+{
+    verify_open();
+    check_read_write(this);
```
tgoyne: check_read_write() is now a very misnamed function and it makes all the code using it confusing. check_write() and check_read_write() should be renamed to reflect what they're actually checking (check_write() might be correct but I'm not sure, check_read_write() is actually checking if m_group is a Transaction).

cmelchior:
Ready for review. @finnschiermer I guess the most relevant parts for you are: https://github.com/realm/realm-object-store/pull/856/files#diff-c0354faf99b53cc5d3c9e6a58ed9ae85R957

src/results.cpp
```diff
@@ -945,6 +954,42 @@ REALM_RESULTS_TYPE(util::Optional<double>)
 
 #undef REALM_RESULTS_TYPE
 
+Results Results::freeze(SharedRealm frozen_realm)
+{
+    switch (m_mode) {
+        case Mode::Empty:
+            return Results(true);
```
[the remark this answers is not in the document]
cmelchior: I agree it doesn't seem particularly useful. I was working from the assumption that once you freeze a Realm, anything constructed from it should also be considered frozen. But Java isn't using this functionality, so I wasn't 100% where it was used.

I'll just remove it again.

src/results.cpp
```diff
@@ -945,6 +954,42 @@ REALM_RESULTS_TYPE(util::Optional<double>)
...
+            auto results = Results(frozen_realm, *frozen_realm->transaction().import_copy_of(m_table_view, PayloadPolicy::Copy), m_descriptor_ordering);
+            results.m_table_view.sync_if_needed();
+            return results;
+    }
+    REALM_COMPILER_HINT_UNREACHABLE();
+}
+
+bool Results::is_frozen()
+{
+    return m_frozen;
```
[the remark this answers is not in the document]
cmelchior: The only reason I introduced the m_frozen member was due to the `Results()` constructor which doesn't take a Realm. I don't know whether it is most meaningful that `Results()` is always considered "frozen" or that it is not?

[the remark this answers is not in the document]
tgoyne: I don't think it's particularly important either way, but I'd be inclined to say that a Results with no Realm should count as frozen because it's not thread-confined and the operations you can perform on it are the same as the ones which can be done on a frozen results.

cmelchior:
Ready for 2nd round of review @tgoyne . @finnschiermer can you also take a look to see if you see something being done incorrectly in terms of using the frozen transactions?

PR #867, opened 2019-10-31
Properly propagate update_only_diff through List::assign()

tgoyne:
Calling Object::create(update_only_diff=true) on an object type which had a List of objects would fail to propagate update_only_diff=true to the sub-calls of Object::create(), resulting it those objects not being diffed and incorrect change notifications being generated.

Because passing around 3 booleans all over the place is incredibly awkward and error-prone, I turned create/update/update_only_diff into a single enum that represents all of the valid combinations of those. This should make it impossible to have a similar bug elsewhere, as it'd have to fail to propagate the policy entirely and that's much more likely to be caught by tests.

See https://github.com/realm/realm-cocoa/issues/6321.

### DIFF
--- a/src/object.hpp
+++ b/src/object.hpp
@@ -32,6 +32,18 @@ namespace _impl {
     class ObjectNotifier;
 }
 
+enum class CreatePolicy : int8_t {
+    // Do not create objects if given something that could be converted to a
+    // Realm object but isn't. Used for things like find().
+    Skip,
+    // Throw an exception if an object with the same PK already exists.
+    ForceCreate,
+    // If an object with the same PK already exists, set all fields in the input.
+    UpdateAll,
+    // If an object with the same PK already exists, only set fields which have changed.
+    UpdateModified
+};
+
 class Object {
 public:
     Object();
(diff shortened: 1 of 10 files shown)


tests/object.cpp
```diff
@@ -691,32 +701,34 @@ TEST_CASE("object") {
...
-            {"object array", AnyVec{ AnyDict{{"value", INT64_C(20)}}, AnyDict{{"value", INT64_C(21)}} } },
+            {"object array", AnyVec{AnyDict{{"value", INT64_C(20)}}, AnyDict{{"value", INT64_C(21)}}}},
         };
-        Object obj = create(dict, false);
+        Object obj = create(dict);
 
-        auto token1 = obj.add_notification_callback([&](CollectionChangeSet, std::exception_ptr) {
+        auto obj_table = r->read_group().get_table("class_all types");
+        Results result(r, *obj_table);
+        auto token1 = result.add_notification_callback([&](CollectionChangeSet, std::exception_ptr) {
```
tgoyne: This is the change to the tests which make them actually reproduce the problem. Object notifiers only report direct changes while results notifiers also report changes to any linked objects, so using a results notifier instead makes it fire if any of the objects in the list change. All of the other changes to the tests are just to adapt them to the new API.

src/object.hpp
```diff
@@ -32,6 +32,18 @@ namespace _impl {
     class ObjectNotifier;
 }
 
+enum class CreatePolicy : int8_t {
+    // Do not create objects if given something that could be converted to a
+    // Realm object but isn't. Used for things like find().
+    Skip,
```
cmelchior: This feels a bit weird, but I guess the only other option would be packing the enum class into an `Optional` which would make it rather annoying to use.

tgoyne: Agreed that it's a bit weird but I can't think of anything clearly better.

PR #871, opened 2019-11-07
Automatically use a table iterator for sequential indexing of Results

tgoyne:


src/results.cpp
```diff
@@ -228,6 +228,43 @@ util::Optional<T> Results::try_get(size_t ndx)
...
+    // than indexing into it, but aren't random access and there's nontrivial
+    // overhead to creating one. Experimentally you need to read at least 3
+    // rows for an iterator to be faster, it's very unlikely that trying to use
+    // an iterator for random-access cases will be worthwhile, and it stops
+    // being faster to increment a pre-existing iterator if it's more than
+    // 50-100 rows away.
+    // As such we only create iterators when starting from index 0 and only
+    // advance pre-existing iterators if they're within 50 rows of the
+    // requested index.
+    if (ndx == 0 && (!m_it == m_ndx != 0) && table.size() > 2) {
```
jedelbo: I have a hard time getting my head around `(!m_it == m_ndx != 0)`. What does it mean in practice? Initially `!m_it` is true and `m_ndx != 0` is false, so when can the expression become true?
When running the unit tests, we never get an iterator created.

src/results.cpp
```diff
@@ -228,6 +228,43 @@ util::Optional<T> Results::try_get(size_t ndx)
...
+        m_it = std::make_unique<Table::ConstIterator>(table.begin());
+        m_ndx = 0;
+    }
+    if (!m_it || ndx > m_ndx + 50)
+        return const_cast<Table&>(table).get_object(ndx);
+
+    *m_it += ndx - m_ndx;
```
jedelbo: This is not executed during the unit tests.

src/results.cpp
tgoyne: I appear to have "simplified" this expression at the last moment and actually just turned it into nonsense instead.

There aren't any tests for Results::get() because originally objectstore was tested exclusively via SDK tests, and even after adding testing to OS we only wrote tests for newly added features and not the things ported from existing SDKs. Good time to change that, though.

tgoyne:
Now actually works.

New benchmarks without this:
```
benchmark name                                  samples       iterations    estimated
                                                mean          low mean      high mean
                                                std dev       low std dev   high std dev

Table forwards                                            100             1    198.599 ms 
                                                   1.98441 ms    1.97209 ms    2.00792 ms 
                                                     83.83 us     51.871 us    144.747 us 
                                                                                          
Table reverse                                             100             1    198.429 ms 
                                                   2.01297 ms    2.00222 ms    2.03904 ms 
                                                    80.473 us     41.783 us    166.994 us 
                                                                                          
TableView forwards                                        100             1    60.4143 ms 
                                                   606.617 us    604.965 us    609.189 us 
                                                    10.374 us      7.643 us     16.475 us 
                                                                                          
TableView reverse                                         100             1    62.5538 ms 
                                                   612.382 us    606.369 us    627.753 us 
                                                     45.61 us     21.369 us     93.422 us 
```

With this:
```
benchmark name                                  samples       iterations    estimated
                                                mean          low mean      high mean
                                                std dev       low std dev   high std dev
Table forwards                                            100             1    29.0942 ms 
                                                   278.115 us     275.95 us    281.525 us 
                                                    13.592 us      9.876 us     22.244 us 
                                                                                          
Table reverse                                             100             1    206.686 ms 
                                                   2.07346 ms    2.04761 ms    2.11863 ms 
                                                   170.397 us    112.151 us    261.999 us 
                                                                                          
TableView forwards                                        100             1    59.6611 ms 
                                                   598.426 us    594.041 us    605.489 us 
                                                    27.964 us     19.605 us     40.102 us 
                                                                                          
TableView reverse                                         100             1    62.5616 ms 
                                                   600.063 us    597.084 us    603.808 us 
                                                    17.016 us     14.289 us     20.868 us 
```

bmunkholm:
So this improves "Table Forwards" by a factor of 10 or so. Was that the biggest contributor to the slowdowns seen in iOS? Don't we mostly iterate on the Tableview in the SDKs?

tgoyne:
Table iteration performance is significantly overrepresented in the cocoa performance tests since it's what we do for all the things that are measuring something other than iteration perf, as it used to be by far the cheapest way to get objects. It is something that'll sometimes show up in real code though.

PR #873, opened 2019-11-08
Je/table iterator

jedelbo:
Requires new release of core to pass CI


jedelbo:
Taken over by https://github.com/realm/realm-object-store/pull/878

PR #874, opened 2019-11-12
Add support for creating a sync session manually

cmelchior:
Adding some public functionality for creating sessions before the Realm is opened.

Due to how Java has implemented Async Open and how the Java APIs are, we have a hard time using the OS AsyncOpenTask for download progress listeners. 

Instead, we need to use our SyncSession wrapper which previously was only available after the Realm was fully opened. So by making this change, we are able to create the wrapper before the Realm is opened and allow users to add progress listeners for async open.

### DIFF
--- a/src/impl/realm_coordinator.cpp
+++ b/src/impl/realm_coordinator.cpp
@@ -345,7 +345,7 @@ std::shared_ptr<AsyncOpenTask> RealmCoordinator::get_synchronized_realm(Realm::C
     return std::make_shared<AsyncOpenTask>(shared_from_this(), m_sync_session);
 }
 
-void RealmCoordinator::create_session(Realm::Config config)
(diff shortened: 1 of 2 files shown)


src/impl/realm_coordinator.cpp
```diff
@@ -344,6 +344,15 @@ std::shared_ptr<AsyncOpenTask> RealmCoordinator::get_synchronized_realm(Realm::C
     create_sync_session(!config.sync_config->is_partial && !exists, exists);
     return std::make_shared<AsyncOpenTask>(shared_from_this(), m_sync_session);
 }
+
+void RealmCoordinator::create_session(Realm::Config config)
```
tgoyne: This could take a const reference to the config to avoid copying it.

PR #887, opened 2019-12-06
Populate public -> column names in keypath mappings

tgoyne:
Required for realm-js to be able to use this rather than its own copy of the function. Also renames the function to a more accurate name since it now doesn't only do backlinks.


nirinchev:
@fealebenpae  this may be nice to get in .NET as it'll allows us to alias the MapTo-ed properties.

PR #891, opened 2019-12-13
Update primary key changing to work with latest core

tgoyne:
Depends on a new core6 release and should be merged as part of updating to that.

### DIFF
--- a/src/object_store.cpp
+++ b/src/object_store.cpp
@@ -110,8 +110,6 @@ ColKey add_column(Group& group, Table& table, Property const& property)
         auto key = table.add_column(to_core_type(property.type), property.name, is_nullable(property.type));
         if (property.requires_index())
             table.add_search_index(key);
-        if (property.is_primary)
-            table.set_primary_key_column(key);
         return key;
     }
 }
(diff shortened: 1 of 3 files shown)


src/object_store.cpp
```diff
@@ -110,8 +110,6 @@ ColKey add_column(Group& group, Table& table, Property const& property)
         auto key = table.add_column(to_core_type(property.type), property.name, is_nullable(property.type));
         if (property.requires_index())
             table.add_search_index(key);
-        if (property.is_primary)
```
tgoyne: Setting the property as the primary key here doesn't work because the user hasn't had a chance to populate it yet and all of the values will be the same.

src/object_store.cpp
```diff
@@ -529,7 +527,7 @@ static void apply_pre_migration_changes(Group& group, std::vector<SchemaChange>
         void operator()(ChangePropertyType op) { replace_column(group, table(op.object), *op.old_property, *op.new_property); }
         void operator()(MakePropertyNullable op) { make_property_optional(table(op.object), *op.property); }
         void operator()(MakePropertyRequired op) { make_property_required(group, table(op.object), *op.property); }
-        void operator()(ChangePrimaryKey op) { ObjectStore::set_primary_key_for_object(group, op.object->name.c_str(), op.property ? op.property->name.c_str() : ""); }
+        void operator()(ChangePrimaryKey op) { table(op.object).set_primary_key_column(ColKey{}); }
```
tgoyne: When we're changing the primary key we want to remove the old one at the start of the migration and then add the new one at the end because the user may be populating the new PK during the migration.

src/object_store.cpp
```diff
@@ -572,7 +570,6 @@ static void apply_post_migration_changes(Group& group,
                 auto col = t.get_column_key(op.property->name);
                 REALM_ASSERT(col);
                 t.set_primary_key_column(col);
-                t.validate_primary_column_uniqueness();
```
tgoyne: This is now redundant because set_primary_key_column() does it.

Issue #684, opened 2018-08-13
Lock up when multiple processes write on Windows


kraenhansen:
@kraenhansen commented on [Fri Aug 10 2018](https://github.com/realm/realm-core/issues/3081)

This issue originates from https://github.com/realm/realm-studio/issues/878.

Two processes taking turns on writing to a local Realm file is not an issue, but if
* Process A opens the Realm
* Process B opens the Realm
* Process A writes
* Process A writes again

the second write will lock up process A.

I've made a repository with a simple test using Realm JS, showing this off https://github.com/kraenhansen/realm-js-multi-process-test - it passes on MacOS but times out on Windows. I've seen a single pass of the test on Windows, so its probably something timing related and not completely deterministic.

PR #894, opened 2020-01-08
[Windows] submit external commit notifications on a threadpool

### LINKED ISSUE #872: Multiprocess issue on windows
On Windows, writing from multiple processes while simultaneously listening for notifications causes _something_ to lock up preventing the Realm from refreshing/delivering notifications.

Here's a simple app that reproduces the issue after 3-4 instances of it are run:

```csharp
using System;
using System.Linq;
using System.Threading.Tasks;
using Nito.AsyncEx;
using Realms;

namespace SimpleRealmWriter
{
    internal class Program
    {
        private const string RealmLocation = "C:\\Work\\example.realm";

        private static void Main(string[] args)
        {
            AsyncContext.Run(async () =>
            {
                var realm = Realm.GetInstance(RealmLocation);
                var counter = realm.All<MyCounter>().FirstOrDefault();
                if (counter == null)
                {
                    counter = new MyCounter();
                    realm.Write(() => realm.Add(counter));
                }
                counter.PropertyChanged += (s, e) =>
                {
                    Console.WriteLine($"Counter's {e.PropertyName} changed: {counter.Value}");
                };

                while (true)
                {
                    await Task.Delay(1000);
                    var tsr = ThreadSafeReference.Create(counter);
                    await Task.Run(() =>
                    {
                        using var bgRealm = Realm.GetInstance(RealmLocation);
                        var bgCounter = bgRealm.ResolveReference(tsr);
                        bgRealm.Write(() =>
                        {
                            bgCounter.Value++;
                        });
                    });

                    realm.Refresh(); // Locks up here

                    Console.WriteLine($"Counter: {counter.Value}");
                }
            });
        }
    }

    internal class MyCounter : RealmObject
    {
        public int Value { get; set; }
    }
}
```

fealebenpae:
Fixes #872 and #684 

On Windows, our implementation of the `ExternalCommitHelper` relies on an interprocess condition variable. In some cases this implementation is susceptible to deadlocks, specifically in tight producer-consumer loops with one or more notifiers.

The issue is that `ExternalCommitHelper::notify_others()` calls `InterprocessCondVar::notify_all()` which internally waits on all the condition variable subscribers to process the notification event before continuing, which results in a deadlock if `ExternalCommitHelper::listen()` triggered `RealmCoordinator::on_change()` and it hasn't returned yet because it's processing async notifiers.

The fix is to make `ExternalCommitHelper::notify_others()` signal the condition variable on a thread pool so that it doesn't need to wait for other instances waiting on the condition variable to process the notification. This matches the behavior of our POSIX-based implementations of `ExternalCommitHelper` where `notify_others()` simply signals a named pipe file descriptor without waiting for any remote instances to react to the event.

### DIFF
--- a/src/impl/windows/external_commit_helper.cpp
+++ b/src/impl/windows/external_commit_helper.cpp
@@ -68,7 +68,8 @@ void ExternalCommitHelper::notify_others()
 
 void ExternalCommitHelper::listen()
 {
-    SetThreadDescription(GetCurrentThread(), L"Realm ExternalCommitHelper listener");
+    // uncomment for debugging purposes
+    //SetThreadDescription(GetCurrentThread(), L"Realm ExternalCommitHelper listener");
     std::lock_guard<InterprocessMutex> lock(m_mutex);
     while (m_keep_listening) {
         m_commit_available->wait(m_mutex, nullptr);
--- a/src/impl/windows/external_commit_helper.cpp
+++ b/src/impl/windows/external_commit_helper.cpp
@@ -32,12 +32,13 @@ static std::wstring create_condvar_sharedmemory_name(std::string realm_path) {
 ExternalCommitHelper::ExternalCommitHelper(RealmCoordinator& parent)
 : m_parent(parent)
 , m_condvar_shared(create_condvar_sharedmemory_name(parent.get_path()).c_str())
+, m_commit_available(std::make_shared<InterprocessCondVar>())
 {
     m_mutex.set_shared_part(InterprocessMutex::SharedPart(), parent.get_path(), "ExternalCommitHelper_ControlMutex");
-    m_commit_available.set_shared_part(m_condvar_shared.get(), parent.get_path(),
+    m_commit_available->set_shared_part(m_condvar_shared.get(), parent.get_path(),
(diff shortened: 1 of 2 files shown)


fealebenpae:
Needs additional work.

src/impl/windows/external_commit_helper.cpp
```diff
@@ -32,38 +32,49 @@ static std::wstring create_condvar_sharedmemory_name(std::string realm_path) {
...
+        delete weak_commit_available;
+        CloseThreadpoolWork(work);
+    }, new std::weak_ptr<InterprocessCondVar>(m_commit_available), nullptr);
+    SubmitThreadpoolWork(work);
 }
 
 void ExternalCommitHelper::listen()
 {
+    // uncomment for debugging purposes
+    //SetThreadDescription(GetCurrentThread(), L"Realm ExternalCommitHelper listener");
```
tgoyne: Naming the thread even in release builds would be nice. It helps our users too and afaik there's no reason not to.

fealebenpae: This API is unfortunately only available on a recent enough Windows 10 build, and we still need to support 8.1.

PR #898, opened 2020-01-15
ObjectId and Decimal128

ironage:
Initial support. Requires most recent core/sync on core-6.1 branches.


src/object_accessor.hpp
```diff
@@ -158,6 +158,8 @@ ValueType Object::get_property_value_impl(ContextType& ctx, const Property &prop
         case PropertyType::String: return ctx.box(m_obj.get<StringData>(column));
         case PropertyType::Data:   return ctx.box(m_obj.get<BinaryData>(column));
         case PropertyType::Date:   return ctx.box(m_obj.get<Timestamp>(column));
+        case PropertyType::ObjectId: return ctx.box(m_obj.get<ObjectId>(column));
+        case PropertyType::Decimal:  return ctx.box(m_obj.get<Decimal>(column));
```
kneth: Why not `Decimal128`?

src/property.hpp
```diff
@@ -51,6 +53,9 @@ enum class PropertyType : unsigned char {
     // deprecated and remains only for reading old files
     Any    = 9,
 
+    ObjectId = 10,
+    Decimal = 11,
```
kneth: Why not `Decimal128`?

jedelbo: We don't have Int64 and Int32.

kneth: So why does the PR's title contain `Decimal128`?

ironage: As far as I can tell, according to mongo's type [documentation](https://docs.mongodb.com/manual/reference/bson-types/#document-bson-type-date) the standard is to use `Decimal128` for enum type names, but use "decimal" for the string type description name. I can change the property type enum according to that.

jedelbo:
This PR should be rebased on core-6.1 branch

ironage:
Rebased. This will require a version of core with https://github.com/realm/realm-core/pull/3601 to link properly.

cmelchior:
I guess this branch should be redirected against the v10 branch now?

ironage:
Redirected to v10.
The core version that this requires (containing realm/realm-core#3601) is now available - but it fails to build with sync due to the c++17 changes that come along with it. We are now blocked on getting a c++17 version of sync compatible with core v10.0.0-alpha.1

ironage:
Since a large portion of the original part of this PR was already merged, the remaining work is about adding proper support for Optional<ObjectId> and improving the tests around that.

This now compiles and links correctly with the new sync version provided by the underlying v10 branch. There are however, unrelated segfaults and hangs in the sync tests on v10 so CI won't be clean.

src/results.cpp
```diff
@@ -1037,6 +1037,7 @@ REALM_RESULTS_TYPE(util::Optional<bool>)
 REALM_RESULTS_TYPE(util::Optional<int64_t>)
 REALM_RESULTS_TYPE(util::Optional<float>)
 REALM_RESULTS_TYPE(util::Optional<double>)
+REALM_RESULTS_TYPE(util::Optional<ObjectId>)
```
kneth: I don't see `REALM_RESULTS_TYPE(util::Optional<Decimal128>)`.

src/impl/object_accessor_impl.hpp
```diff
@@ -112,6 +112,7 @@ class CppContext {
     util::Any box(util::Optional<double> v) const { return v; }
     util::Any box(util::Optional<float> v) const { return v; }
     util::Any box(util::Optional<int64_t> v) const { return v; }
+    util::Any box(util::Optional<ObjectId> v) const { return v; }
```
kneth: I don't see `util::Any box(util::Optional<Decimal128> v) const { return v; }`

src/list.cpp
```diff
@@ -482,6 +482,7 @@ REALM_PRIMITIVE_LIST_TYPE(util::Optional<bool>)
 REALM_PRIMITIVE_LIST_TYPE(util::Optional<int64_t>)
 REALM_PRIMITIVE_LIST_TYPE(util::Optional<float>)
 REALM_PRIMITIVE_LIST_TYPE(util::Optional<double>)
+REALM_PRIMITIVE_LIST_TYPE(util::Optional<ObjectId>)
```
kneth: I don't see `REALM_PRIMITIVE_LIST_TYPE(util::Optional<Decimal128>)`

src/results.cpp
tgoyne: `Decimal128` can store `null` internally like StringData etc. while `ObjectId` can't.

PR #899, opened 2020-01-16
To/from core type functions moved to property.cpp

jedelbo:


tgoyne:
`from_core_type()` was originally on `ObjectSchema` because it used a `Table` and there's a correspondence between those two types. With the table parameter removed I don't think there's a reason for it to be there.

I am not a fan of any of the formatting changes made here.

jedelbo:
@tgoyne where should it rather be then? The main motivation is to move the functions to files that are likely to go to core soon.

Of course formatting can always be debated. The changes are formatted by clang-format using the configuration we currently use in core.

tgoyne:
Conceptually they should be functions on PropertyType, but since that isn't possible I think they should just be free functions. Switching PropertyType to a design more like ColKey rather than exposing the bitmask as the public API could be a nice change but is unrelated to this.

jedelbo:
Dropping this for now. Will be fixed when going to monorepo.

PR #901, opened 2020-01-29
Store the table key on ObjectSchema

tgoyne:
Similar to how is done for column keys on Property.

Cocoa has a parallel set of schema data structures which we cache various things on. One of these is the TableRef for each ObjectSchema which is lazily stored the first time it's used. This is pretty important for performance (disabling the caching doubles the runtime of some basic operations), but it doesn't work for frozen Realms. To be thread-safe, we'd need to either add locking or eagerly look up TableRefs for each ObjectSchema.

Fortunately with core 6 we can get away with storing a TableKey rather than a TableRef, and pushing the logic here makes populating the `table_key` property eagerly nearly free, as we're already doing the complicated and expensive parts for column keys. This also lets us eliminate a few places where we look up tables by name in the object store code, which probably gives a performance improvement (but I didn't bother trying to measure that).

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -37,7 +37,6 @@ if(REALM_ENABLE_SYNC)
     )
     list(APPEND SOURCES
         sync/file.cpp
-        sync/global_notifier.cpp
         sync/metadata.cpp
         sync/partial_sync.cpp
         sync/permission.cpp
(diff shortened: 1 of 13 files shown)


tests/CMakeLists.txt
```diff
@@ -51,6 +50,12 @@ if(REALM_ENABLE_SYNC)
     )
 endif()
 
+if(REALM_ENABLE_SERVER)
```
ironage: good catch

PR #905, opened 2020-02-05
Fix a whole bunch of sanitizer failures

tgoyne:
Turns out that while we *built* the tests with asan/tsan on CI, we didn't actually run them in the non-coverage configurations. Unsurprisingly turning it on revealed a pile of issues.

Most of the changes here are fixing lock-order inversions which in practice couldn't actually cause deadlocks due to external constraints around when the locking happened. A few of them could probably cause deadlocks in very specific scenarios, and there's also a few outright data races that could cause issues in niche scenarios.

TSan still sometimes warns about a data race in core. https://github.com/realm/realm-core/pull/3584 attempted to fix that but I think it doesn't actually work.

This is probably best reviewed commit-by-commit due to it being a bunch of unrelated changes.


ironage:
These all seem like good changes to me 👍 
If we don't figure out that core race (or positively identify it as a false positive), we should probably get all these changes in (minus the CI TSAN check until that passes) because that will be a big step in the right direction.

ironage:
I'm totally in favour of all this work, if you can find the time to update it and get it in. If it is difficult to bring this code up to date, maybe consider doing this in the monorepo instead? I'm going to remove myself from being a reviewer, but feel free to add me back if this is ready for reviews.

bmunkholm:
@tgoyne Was this moved to monorepo?

PR #906, opened 2020-02-05
Use NDK21

jedelbo:
The main idea behind the changes made in this PR is to let Sync and Core get into the cmake system by use of `find_package`. This requires that RealmCore_DIR and realm-sync_DIR is set to point to a directory where the respective configuration files are found. This can be either an installed package or a package built locally. The CI system should then just download the Core and Sync binary packages and just build OS. The hurdle is that Sync is not available in all needed flavors, so we anyway have to build that locally. Windows build is not yet working as there is a problem with the Core package configuration file.


tgoyne:
Why is this changing anything other than updating the docker image and the target names? All that code for building core/sync from source is there for a reason.

jedelbo:
@tgoyne The tricky part (for me) is to use the OpenSSL library that is specified in in RealmCoreConfig.cmake. My assumption is that the library needs to be found during the cmake config phase. Any suggestions on how this could be done are welcome.
What exactly is the reason that Core is built from source in the first place?

tgoyne:
Building core and sync from source is extremely valuable for developing things in object store. It's required to be able to step into core/sync when debugging things and is required for testing with tsan/asan/etc. The fact that the build system automatically rebuilds core and sync following changes in those repos makes working on things which touch more than one repository fairly painless, while doing so in the SDKs where you have to manually rebuild a release package every time is a nightmare.

Moving away from this is a very big :-1: from me.

tgoyne:
If things need to happen earlier in the CMake build process for the config file to be available at the applicable time, then they can just happen earlier in the CMake build process. Even separately from the build-from-source issue, moving the downloading from cmake to the Jenkinsfile means that the steps for getting started with working in this repo is no longer "clone, run cmake, open project" and are instead some horribly complicated and undocumented process.

jedelbo:
Ok, I thought the feature to download Core and Sync was just used in CI - I have never used this option as I have always a checkout of Core and Sync that I link to. I found a way to get around the problem with OpenSSL.
(I still find all the functionality in RealmCore.cmake very complicated. The approach I suggested would be much leaner.)

tgoyne:
It's definitely more complicated than it needs to be due to predating core/sync using cmake and being written to work with the old makefile based build system. Ideally it'd just be a pair of `add_subdirectory()` lines, but the sync cmake files need some changes to support that.

PR #908, opened 2020-02-06
Surface client reset

kneth:
In order to fix - at least partially - https://jira.mongodb.org/browse/RJS-347

### DIFF
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -638,7 +638,7 @@ void SyncSession::handle_error(SyncError error)
         case NextStateAfterError::inactive: {
             if (error.is_client_reset_requested()) {
                 std::unique_lock<std::mutex> lock(m_state_mutex);
-                cancel_pending_waits(lock, error.error_code);
+                cancel_pending_waits(lock, make_error_code(realm::sync::ProtocolError::diverging_histories));
             }
 
             {
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -80,7 +80,7 @@ struct SyncSession::State {
     virtual ~State() { }
 
     // Move the given session into this state. All state transitions MUST be carried out through this method.
-    virtual void enter_state(std::unique_lock<std::mutex>&, SyncSession&, bool is_client_reset = false) const { }
+    virtual void enter_state(std::unique_lock<std::mutex>&, SyncSession&) const { }
 
     virtual void refresh_access_token(std::unique_lock<std::mutex>&,
                                       SyncSession&, std::string,
@@ -120,7 +120,7 @@ struct SyncSession::State {
 };
 
 struct sync_session_states::WaitingForAccessToken : public SyncSession::State {
-    void enter_state(std::unique_lock<std::mutex>&, SyncSession& session, bool) const override
+    void enter_state(std::unique_lock<std::mutex>&, SyncSession& session) const override
     {
         session.m_deferred_close = false;
     }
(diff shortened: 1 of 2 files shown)


src/sync/sync_session.cpp
```diff
@@ -637,7 +641,7 @@ void SyncSession::handle_error(SyncError error)
             break;
         case NextStateAfterError::inactive: {
             std::unique_lock<std::mutex> lock(m_state_mutex);
```
ironage: Instead of passing a boolean down the stack, I think it would be cleaner to do something like the following before changing states here:
```
if (error.is_client_reset_requested()) {
    cancel_pending_waits(lock, make_error_code(sync::ProtocolError::bad_client_file));
}
```

kneth: `cancel_pending_waits()` will unlock so I cannot change the state.

kneth: Yes, that's a big simplification.

src/sync/sync_session.cpp
```diff
@@ -636,8 +636,15 @@ void SyncSession::handle_error(SyncError error)
             }
             break;
         case NextStateAfterError::inactive: {
-            std::unique_lock<std::mutex> lock(m_state_mutex);
-            advance_state(lock, State::inactive);
+            if (error.is_client_reset_requested()) {
+                std::unique_lock<std::mutex> lock(m_state_mutex);
+                cancel_pending_waits(lock, error.error_code);
```
ironage: is this the correct error_code that you need at the js level?

kneth: Yeah, I see your point. This could produce a range of different errors, and narrow it down to just one might be easier for users to react to.

src/sync/sync_session.cpp
```diff
@@ -636,8 +636,15 @@ void SyncSession::handle_error(SyncError error)
             }
             break;
         case NextStateAfterError::inactive: {
-            std::unique_lock<std::mutex> lock(m_state_mutex);
-            advance_state(lock, State::inactive);
+            if (error.is_client_reset_requested()) {
+                std::unique_lock<std::mutex> lock(m_state_mutex);
+                cancel_pending_waits(lock, make_error_code(realm::sync::ProtocolError::diverging_histories));
```
cmelchior: Not sure I understand this. Are you faking an error code that result in an Resync? If I remember correctly there are multiple ones that triggers a Resync. Also, I thought the Client Resync would be transparent from the end user perspective? This seems to bubble up to them?

kneth: I don't think the manual case should be fully transparent. As it is now, a manual client resync will lead to an `Operation canceled` while the proposed change will say `Divergent histories`. The SDKs can handle the non-manual cases transparent.

cmelchior: No, the manual case needs to be handled for sure. But I'm still not sure why you are faking the error code because shouldn't you have the "correct" one somewhere?

src/sync/sync_session.cpp
```diff
@@ -636,8 +636,15 @@ void SyncSession::handle_error(SyncError error)
             }
             break;
         case NextStateAfterError::inactive: {
-            std::unique_lock<std::mutex> lock(m_state_mutex);
-            advance_state(lock, State::inactive);
+            if (error.is_client_reset_requested()) {
```
cmelchior, line 639: Why not use this error instead of constructing a new one with `make_error_code`?

ironage: I agree, I think you should be able to use `error.error_code` instead of making a `diverging_histories`

kneth:
@cmelchior @ironage I have resolved your comments by use the error and not a fake one.

PR #910, opened 2020-02-07
Fix thread-safety issues with frozen Results

tgoyne:
Results being lazily evaluated makes reading from it from multiple threads without any locking unsafe even when it's backed by a frozen Realm. To fix this we need to introduce some locking around the members which can be mutated on read. Benchmarking indicated a measurable slowdown from always doing the locking (~10% time increase for iterating over query results from obj-c), so it uses an "optional" mutex which is a no-op for non-frozen realms and only actually locks when the realm is frozen.

The complicated control flow inside Results made figuring out where exactly locks were needed (and needed to not be present to avoid deadlocks), so I introduced the use of clang's [Thread Safety Analysis annotations](https://clang.llvm.org/docs/ThreadSafetyAnalysis.html) which perform static checks at compile time to verify that mutexes are held or not held at the appropriate times. Applying these to RealmCoordinator and CollectionNotifier also revealed a few missing locks in RealmCoordinator.

List and Results also have a lazily populated pointer to the associated object schema, which merely needs to be atomic.

As part of auditing each of the types applicable to frozen Realms for things which might be thread-unsafe I also went ahead and marked a bunch of things as const and/or noexcept to cut down on the number of things which could be unsafe.

### DIFF
--- a/src/impl/collection_notifier.cpp
+++ b/src/impl/collection_notifier.cpp
@@ -441,7 +441,8 @@ NotifierPackage::NotifierPackage(std::exception_ptr error,
 {
 }
 
-void NotifierPackage::package_and_wait(util::Optional<VersionID::version_type> target_version)
+// Clang TSE seems to not like returning a unique_lock from a function
+void NotifierPackage::package_and_wait(util::Optional<VersionID::version_type> target_version) NO_THREAD_SAFETY_ANALYSIS
 {
     if (!m_coordinator || m_error || !*this)
         return;
--- a/src/impl/collection_notifier.cpp
+++ b/src/impl/collection_notifier.cpp
@@ -208,7 +208,7 @@ uint64_t CollectionNotifier::add_callback(CollectionChangeCallback callback)
 {
     m_realm->verify_thread();
 
-    std::lock_guard<std::mutex> lock(m_callback_mutex);
+    util::CheckedLockGuard lock(m_callback_mutex);
     auto token = m_next_token++;
     m_callbacks.push_back({std::move(callback), {}, {}, token, false, false});
     if (m_callback_index == npos) { // Don't need to wake up if we're already sending notifications
@@ -224,7 +224,7 @@ void CollectionNotifier::remove_callback(uint64_t token)
     // it could cause user code to be called
     Callback old;
     {
-        std::lock_guard<std::mutex> lock(m_callback_mutex);
+        util::CheckedLockGuard lock(m_callback_mutex);
         auto it = find_callback(token);
         if (it == end(m_callbacks)) {
             return;
@@ -253,7 +253,7 @@ void CollectionNotifier::suppress_next_notification(uint64_t token)
         m_realm->verify_in_write();
     }
 
-    std::lock_guard<std::mutex> lock(m_callback_mutex);
+    util::CheckedLockGuard lock(m_callback_mutex);
     auto it = find_callback(token);
     if (it != end(m_callbacks)) {
         it->skip_next = true;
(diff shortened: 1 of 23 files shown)


src/schema.cpp
```diff
@@ -28,29 +28,30 @@
...
-Schema& Schema::operator=(Schema&&) = default;
+Schema& Schema::operator=(Schema&&) noexcept = default;
 
 Schema::Schema(std::initializer_list<ObjectSchema> types) : Schema(base(types)) { }
 
-Schema::Schema(base types) : base(std::move(types))
+Schema::Schema(base types) noexcept
+: base(std::move(types))
 {
     std::sort(begin(), end(), [](ObjectSchema const& lft, ObjectSchema const& rgt) {
```
ironage: I'm not convinced about the casual use of `noexcept` everywhere. What if `std::sort` runs out of memory and throws a `bad_alloc`? Do we want to change the behaviour in this scenario from propagating the exception up to the SDKs, to an abort? I doubt the SDK's have any sort of fancy error recovery in such a scenario, but I do want to make sure this change in behaviour is intentional.

ironage:
By the way, I'm getting the following with `Apple LLVM version 10.0.1 (clang-1001.0.46.3)`
```
[22/42] Building CXX object src/CMakeFiles/realm-object-store.dir/impl/realm_coordinator.cpp.o
../src/impl/realm_coordinator.cpp:966:14: warning: acquiring mutex 'lock' that is already held [-Wthread-safety-analysis]
        lock.lock();
             ^
../src/impl/realm_coordinator.cpp:969:14: warning: releasing mutex 'm_notifier_mutex' that was not held [-Wthread-safety-analysis]
        lock.unlock();
             ^
../src/impl/realm_coordinator.cpp:994:10: warning: acquiring mutex 'lock' that is already held [-Wthread-safety-analysis]
    lock.lock();
         ^
../src/impl/realm_coordinator.cpp:1001:5: warning: calling function 'clean_up_dead_notifiers' requires holding mutex 'm_notifier_mutex' exclusively [-Wthread-safety-analysis]
    clean_up_dead_notifiers();
    ^
4 warnings generated.
```
Is this expected, or do I need to upgrade clang?

src/schema.cpp
tgoyne: std::sort is not allowed to throw any exceptions when given valid input and a comparison function which does not throw, which is normally achieved by using a sort algorithm that does not need to allocate any non-stack scratch space.

src/impl/collection_notifier.cpp
```diff
@@ -441,7 +441,8 @@ NotifierPackage::NotifierPackage(std::exception_ptr error,
 {
 }
 
-void NotifierPackage::package_and_wait(util::Optional<VersionID::version_type> target_version)
+// Clang TSE seems to not like returning a unique_lock from a function
```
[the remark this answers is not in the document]
tgoyne: RETURN_CAPABILITY is for returning a reference to the mutex itself. I don't think there's anything that models returning a scoped_capability.

src/schema.cpp
ironage: I couldn't find anything about that in the C++11 standard, doesn't that mean it is implementation dependant? I'm just curious where I can find the information, I don't mean to belabour it too much (not-blocking).

src/impl/collection_notifier.cpp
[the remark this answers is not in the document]
ironage: Ok good to know, thanks for checking.

src/schema.cpp
tgoyne: The standard explicitly lists all exceptions that an algorithm can throw. Since none are listed for std::sort(), that means it's not allowed to throw any itself (but it's not noexcept because the predicate might throw).

tgoyne:
Looks like that's a bug in the version of clang shipped with Xcode 10.x, but working around it turned out to be easy and maybe made the code better anyway.

PR #911, opened 2020-02-07
Realm App + Sync User changes

jsflax:


src/sync/sync_user.cpp
```diff
@@ -21,44 +21,144 @@
...
+    this->token = token;
+    this->expires_at = json["exp"].get<long>();
+    this->issued_at = json["iat"].get<long>();
+    
+    if (json.find("user_data") != json.end()) {
+        this->user_data = json["user_data"].get<std::map<std::string, util::Any>>();
+    }
+}
+
+ObjectSchema SyncUserProfile::schema()
```
cmelchior: If I understand this correctly. The user profile is going to be stored in a Realm on device. While it does have some benefits, it also brings about some problems I'm not entirely comfortable with.

- It exposes a Realm lifecycle and threading constraints into API's that might not be suited for it. E.g. the current `SyncUser` does not have threading restrictions.
- Safety issues. Shouldn't the Realm holding these data be encrypted. Where does the encryption key come from?

src/sync/app.cpp
```diff
@@ -0,0 +1,77 @@
...
+namespace realm {
+
+std::shared_ptr<RealmApp> RealmApp::app(const std::string app_id,
+                                        const realm::util::Optional<RealmApp::Config> config)
+{
+    return std::make_shared<RealmApp>(app_id, config);
+}
+
+void RealmApp::login_with_credentials(const std::shared_ptr<AppCredentials> credentials,
+                                      int timeout,
```
cmelchior: Unclear what unit this is, I assume milliseconds, so `timeoutMs` would be more readable.

src/sync/sync_user.cpp
jsflax: What threading issues does it introduce? SyncUserProfile is a member of SyncUser– they would be stored together.

As far as safety, we did not previously encrypt the user profile in any way. I can ask product and see if this is a requirement.

src/sync/app.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+ - SeeAlso:
+ `RemoteMongoClient`,
+ `RLMPush`,
+ [Functions](https://docs.mongodb.com/stitch/functions/)
+ */
+class RealmApp {
+public:
+    struct Config {
+        realm::util::Optional<std::shared_ptr<realm::GenericNetworkTransport>> transport;
```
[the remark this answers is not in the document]
jsflax: Good call.

src/sync/app.cpp
jsflax: Good catch 👍

src/sync/sync_user.cpp
cmelchior: Maybe I misunderstood the code. I thought this turned it into a RealmObject, where instead it was just a C++ data structure?

Previously the SyncUser didn't contain any sensitive data AFAIK, but now it seems that Stitch sends a lot of data that could potentially be a GDPR violation if leaked.

jsflax: It is turned into a realm object, but so is sync user. Maybe I don’t understand why that introduces threading constraints / exposes the realm lifecycle differently than before.

cmelchior: Previously, the Sync User was maybe stored in a Realm file, but it was not exposed as a live object. In Java, we deserialized the stored JSON to a static Java Object which meant it could be accessed from any thread.

API's that mix live objects and "normal" objects are pretty painful to work with. At least we discovered this when working on the Permission API's for Java.

There might be enough arguments for exposing the SyncUser as a live object, but e.g. in Java I was planning to keep `SyncUser.createConfiguration()`, which becomes pretty difficult.

I guess we could freeze the SyncUser, but still...

I guess I'm asking what use cases are we solving by turning it into a SyncUser? As I can see it having some unforeseen consequences for how the API's currently work.

I guess I'm askin

tgoyne: Using a frozen realm object for SyncUser means that we're keeping the Realm opened (and that version pinned) forever just to avoid copying a small amount of data out, which doesn't seem like a good tradeoff. An unfrozen one means that SyncUser is thread-confined, which probably isn't viable.

tests/sync/app.cpp
```diff
@@ -0,0 +1,130 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
```
kneth: `2016` -> `2020`

src/sync/app.hpp
```diff
@@ -0,0 +1,113 @@
...
+ You can also use it to execute [Functions](https://docs.mongodb.com/stitch/functions/).
+
+ Finally, its `RLMPush` object can register the current user for push notifications.
+
+ - SeeAlso:
+ `RemoteMongoClient`,
+ `RLMPush`,
+ [Functions](https://docs.mongodb.com/stitch/functions/)
+ */
+class RealmApp {
```
kneth: Can't it be `App`? We are in the `realm` namespace.

PR #913, opened 2020-02-11
don't crash on missing file when deleting it

blagoev:
if the file does not exists do not crash when trying to delete it.

### DIFF
--- a/src/server/global_notifier.cpp
+++ b/src/server/global_notifier.cpp
@@ -211,9 +211,7 @@ void GlobalNotifier::Impl::release_version(sync::ObjectID id, VersionID old_vers
         if (info.coordinator) {
             std::string path = info.coordinator->get_config().path;
             m_realms.erase(it);
-            if (File::exists(path)) {
-              File::remove(path);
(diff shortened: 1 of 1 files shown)


src/server/global_notifier.cpp
```diff
@@ -211,7 +211,9 @@ void GlobalNotifier::Impl::release_version(sync::ObjectID id, VersionID old_vers
         if (info.coordinator) {
             std::string path = info.coordinator->get_config().path;
             m_realms.erase(it);
-            File::remove(path);
+            if (File::exists(path)) {
```
tgoyne: This should use `File::try_remove()` rather than checking if the file exists.

Checking if a file exists before performing an operation rather than simply attempting the operation and handling the error if the file doesn't exist is inherently a [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) race condition and is nearly always the wrong thing to do.

blagoev: yes indeed. thanks

PR #915, opened 2020-02-12
Add is_embedded attribute on ObjectSchema

jedelbo:


jedelbo:
@tgoyne @cmelchior I am not sure what kind of support for embedded object is expected here. Is this PR a step in the right direction?

tgoyne:
Embeddedness is a property of ObjectSchema, not Property.

kneth:
Will it also validate the schema? For example, an embedded object cannot define a primary key.

src/object_schema.cpp
```diff
@@ -330,6 +343,9 @@ void ObjectSchema::validate(Schema const& schema, std::vector<ObjectSchemaValida
         validate_property(schema, name, prop, &primary, exceptions);
     }
 
+    if (!primary_key.empty() && is_embedded) {
+        exceptions.emplace_back("Embedded table '%1' cannot have primary key '%2'.", name, primary_key);
```
cmelchior: Just to make it more clear that no primary key is allowed at all.
```suggestion
        exceptions.emplace_back("Embedded table '%1' is not allowed to define a primary key. Found key '%2'.", name, primary_key);
```

tests/schema.cpp
```diff
@@ -450,6 +454,16 @@ TEST_CASE("Schema") {
             REQUIRE_THROWS_CONTAINING(schema.validate(), "Properties 'pk2' and 'pk1' are both marked as the primary key of 'object'.");
         }
 
+        SECTION("rejects primary key on embedded table") {
```
cmelchior: We probably also need a migration test, e.g. if a local Realm adds or remove the embeddedness of a table during a migration.

tests/realm.cpp
```diff
@@ -1284,6 +1284,25 @@ TEST_CASE("SharedRealm: dynamic schema mode doesn't invalidate object schema poi
     REQUIRE(object_schema == &*r2->schema().find("object"));
 }
 
+TEST_CASE("SharedRealm: declaring an object as embedded results in creating an embedded table") {
```
tgoyne: The existing tests for validating that we created the correct things in core from a given schema are in migrations.cpp. Adding sections for is_embedded to each of the existing test cases there will probably reveal some more scenarios that need to be handled.

tests/schema.cpp
jedelbo: It is not possible to change embeddedness of a table after creation

cmelchior: May I ask why not? For local Realms it just seems to be a boolean and some rules for transactions?

jedelbo: Yes, it is just a flag, but the flag indicates that some invariants is upheld. We could probably check if a non embedded table complies to these invariants before making it embedded. This requirement was not captured in the tecnical design.

finnschiermer: Actually - the first implementation allowed this. And then we simplified. Apparently too much.

tests/realm.cpp
jedelbo: Ok, I will look into that.

jedelbo:
@tgoyne Could you have a look at the last commit where I try to create embedded objects based on values in a context? Is it ok to do it that way?

jedelbo:
I will add tests around changing embeddedness in a separate PR

PR #916, opened 2020-02-13
Enable -Werror when building

cmelchior:


src/impl/collection_notifier.cpp
```diff
@@ -170,12 +170,12 @@ bool DeepChangeChecker::check_row(Table const& table, int64_t key, size_t depth)
             return true;
     }
     auto& not_modified = m_not_modified[table_key.value];
-    if (not_modified.contains(key))
+    if (not_modified.contains(static_cast<size_t>(key)))
```
cmelchior: I'm a bit uncertain how safe this cast is

tgoyne: It's not safe for 32-bit and here the warning is pointing at an actual bug. m_not_modified should actually be something like `std::unordered_map<int64_t, std::unordered_set<int64_t>>`.

cmelchior: I replaced the `IndexSet` with `std::unordered_set<ObjectKeyType>`. It should fix the issue. I didn't want to modify `IndexSet` as its internal implementation seemed to suggest some more advanced usecase, but I don't know if there is some unwanted performance impact of this change?

tgoyne: `IndexSet` is optimized for storing sorted sequences of numbers which tend to be sequential. With core 5 it was used for both storing sequences of row numbers and sequences of indexes into collections, but with core 6 it should only be used for the latter as object keys tend not to be sequential. unodered_set is kinda bad but it has a decent chance of being faster here.

PR #920, opened 2020-02-18
Removing Adapter and Global Notifier

kneth:
Removing adapter and global notifier have implications for some SDKs.


tgoyne:
There's some assorted objectstore things which exist purely to support the GN that should be cleaned up too. From memory there's TransactionChangeInfo.track_all and RealmCoordinator::open_with_config(), but I think there may be a few others. shared_realm.hpp also still has a mention of GlobalNotifier.

PR #919, opened 2020-02-17
Draft: Realm App

jsflax:

PR #921, opened 2020-02-18
Realm App

ironage:
This is https://github.com/realm/realm-object-store/pull/919 rebased onto the v10 branch.


src/sync/generic_network_transport.hpp
```diff
@@ -0,0 +1,324 @@
...
+    const std::string message() const { return std::runtime_error::what(); };
+
+    service_error_code code() const { return m_code; };
+
+    const std::string raw_code() const { return m_raw_code; }
+
+private:
+    static service_error_code error_code_for_string(const std::string& code)
+    {
+        if (code == "MissingAuthReq")
```
kneth: Why not use a `std::map` or similar?

src/sync/app.cpp
```diff
@@ -0,0 +1,153 @@
...
+            Method::get,
+            profile_route,
+            timeout_ms,
+            {
+                { "Content-Type", "application/json;charset=utf-8" },
+                { "Accept", "application/json" },
+                { "Authorization", bearer}
+            },
+            std::string()
+        }, [&](const Response profile_response) {
```
cmelchior: One use case in Java is having an API like:

```kotlin
val task: RealmAsyncTask = RealmApp.loginAsync("user", "password", Callback() {
  override fun onSuccess(user: RealmAppUser) { /* User logged in */ }
  override fun onError(error: ObjectServerError) { /* An error happend */ }
});

// Manually cancel. Callbacks should not trigger
task.cancel()
```

This looks kinda doable since the network request is being triggered in Java where we can cancel it. I guess we can wrap that exception with a Java-specific error code that will be ignored when parsed back to the completion block,.

src/sync/app.cpp
```diff
@@ -0,0 +1,153 @@
...
+            // if there is a already an error code, pass the error upstream
+            if (response_code_is_fatal(profile_response.status_code)) {
+                return completion_block(nullptr, handle_error(profile_response));
+            }
+
+            j = nlohmann::json::parse(profile_response.body);
```
cmelchior: If there is a bug or change in the json received, this will crash the caller hard. We should probably wrap most of the logic here in a try-catch so we can return an appropriate error if there is a programmer/server error somewhere we didn't expect.

src/sync/app_credentials.hpp
```diff
@@ -0,0 +1,110 @@
...
+     Construct and return credentials from a username and password.
+     */
+    static std::shared_ptr<AppCredentials> username_password(const std::string username,
+                                                             const std::string password);
+
+    /// The provider of the credential
+    AuthProvider provider() const;
+
+    /// The serialized payload
+    std::string serialize() const;
```
cmelchior: Maybe `serialize_as_json()` to make it explicit what kind of String is returned?

src/sync/app.hpp
```diff
@@ -0,0 +1,119 @@
...
+    /**
+    Log in a user and asynchronously retrieve a user object.
+
+    If the log in completes successfully, the completion block will be called, and a
+    `SyncUser` representing the logged-in user will be passed to it. This user object
+    can be used to open `Realm`s and retrieve `SyncSession`s. Otherwise, the
+    completion block will be called with an error.
+
+    - parameter credentials: A `SyncCredentials` object representing the user to log in.
+    - parameter timeout: How long the network client should wait, in seconds, before timing out.
```
kneth: The parameter is called `timeout_ms` in the signature. It does imply millisec and not sec.

src/sync/app.hpp
```diff
@@ -0,0 +1,119 @@
...
+    - parameter callbackQueue: The dispatch queue upon which the callback should run. Defaults to the main queue.
```
kneth: The method doesn't have this parameter in it signature

src/sync/app_credentials.hpp
```diff
@@ -0,0 +1,110 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or utilied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#ifndef REALM_APP_CREDENTIALS_HPP
+#define REALM_APP_CREDENTIALS_HPP
+
+#include <stdio.h>
```
kneth: What is `stdio.h` needed for?

src/sync/generic_network_transport.hpp
```diff
@@ -0,0 +1,324 @@
...
+     */
+    Method method;
+
+    /**
+     * The URL to which this request will be made.
+     */
+    std::string url;
+
+    /**
+     * The number of seconds that the underlying transport should spend on an HTTP round trip before failing with an
```
kneth: The name implies milliseconds.

src/sync/generic_network_transport.hpp
```diff
@@ -0,0 +1,324 @@
...
+    /**
+     * The body of the request.
+     */
+    std::string body;
+};
+
+/**
+ * The contents of an HTTP response.
+ */
+struct Response {
```
cmelchior: This response also needs to be able to handle other error types than just HTTP error codes, e.g. a socket timeout exception or some other exception thrown from the binding. It is a bit unclear exactly how this should be modeled though, because most likely in Java you want to roundtrip the same exception.

We could probably encode the semantics using special status_codes not used by REST API's and encode the message in body, but it probably does make for a slightly magical API. 

Would it make sense to have a both a `http_status_code` and `binding_status_code` property. Bindings could then define their own semantics to binding_status_codes?

src/sync/generic_network_transport.hpp
```diff
@@ -0,0 +1,324 @@
...
+public:
+    /**
+     * The status code of the HTTP response.
+     */
+    int status_code;
+
+    /**
+     * The headers of the HTTP response.
+     */
+    std::map<std::string, std::string> headers;
```
cmelchior: TIL: Response headers can return multiple values using comma, e.g. `Key: val1, val2, val2`
I assume we are supporting this here in that format? The reason I'm asking is that OkHttp represents header values in their own wrapper class and I need to reconstruct the output for this API.

src/sync/generic_network_transport.hpp
```diff
@@ -0,0 +1,324 @@
...
+    /**
+     * The URL to which this request will be made.
+     */
+    std::string url;
+
+    /**
+     * The number of seconds that the underlying transport should spend on an HTTP round trip before failing with an
+     * error.
+     */
+    int timeout_ms;
```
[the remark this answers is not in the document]
jsflax: The way we had it previously was to have it as a single setting.

src/sync/generic_network_transport.hpp
jsflax: I've modelled this after what we had previously, but we can add binding_status_code as well.

src/sync/generic_network_transport.hpp
jsflax: Hmm... I suppose it could be? I do not believe any headers from Stitch return like that.

src/sync/generic_network_transport.hpp
bmunkholm: We don't use the terminology "binding", but SDK. When it slips, it should at least not go into the code :-)

jsflax: Could just call it as a custom status code then

src/sync/generic_network_transport.hpp
```diff
@@ -0,0 +1,335 @@
...
+        Client,
+        Session,
+    };
+
+    AppError()
+    : std::runtime_error("AppError")
+    , type(Type::Generic)
+    {
+    };
+    AppError(std::string msg, Type classification = Type::Generic)
```
cmelchior: This interface is breaking how errors are currently being reported. Currently, we are reporting this like this:  https://github.com/realm/realm-object-store/blob/master/src/sync/async_open_task.hpp#L45 using an exception or using the error code directly: https://github.com/realm/realm-object-store/blob/master/src/sync/sync_session.hpp#L140

SDK's can then use the error_category and error_code to map to SDK specific errors, e.g like Java is doing here: https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/cpp/io_realm_SyncSession.cpp#L163

I'm not sure exactly what the right API is here, but the App should IMO not to trying to remap errors. Instead, they should be delivered more or less directly as they are to SDK's which will then map them to more proper error types?

src/sync/generic_network_transport.hpp
```diff
@@ -0,0 +1,335 @@
...
+            case ClientError::Code::bad_token:
+                return "Bad Token";
+            case ClientError::Code::bad_response:
+                return "Bad Response";
+        }
+        return util::format("unknown error code: %1", static_cast<int>(code));
+    }
+};
+
+enum class ServiceErrorCode {
```
cmelchior: Would be nice with some docs explaining where these error codes are defined? I assume they come from Stitch somewhere?

src/sync/generic_network_transport.hpp
cmelchior: But I suspect creating an error_category instead of a `Type` and then assign error_codes to all the app specific errors would fit the current pattern better?

src/sync/generic_network_transport.hpp
```diff
@@ -0,0 +1,335 @@
...
+
+    const ServiceErrorCode m_code;
+    const std::string m_raw_code;
+};
+
+} // namespace error
+
+
+enum class Method {
+    get, post, patch
```
cmelchior: Any reason `delete` and `put` is not here?
Also, please add documentation to all public structures.

cmelchior: Also, I would consider renaming this to `HttpMethod` for clarity

src/sync/impl/sync_metadata.cpp
```diff
@@ -339,55 +387,103 @@ std::string SyncUserMetadata::local_uuid() const
...
     m_realm->verify_thread();
     m_realm->begin_transaction();
-    m_obj.set(m_schema.idx_user_is_admin, is_admin);
+    m_obj.set(m_schema.idx_access_token, *user_token);
     m_realm->commit_transaction();
 }
 
+void SyncUserMetadata::set_user_profile(std::shared_ptr<SyncUserProfile>)
+{
+//    m_row.
```
cmelchior:
```suggestion
//    FIXME
```

src/sync/sync_user.cpp
```diff
@@ -21,44 +21,123 @@
...
 namespace realm {
 
+static std::string base64_decode(const std::string &in) {
+    std::string out;
+    out.resize(util::base64_decoded_size(in.size()));
+    util::base64_decode(in, &out[0], out.size());
+    return out;
+}
+
+static std::vector<std::string> split_token(const std::string& jwt) {
```
cmelchior: There don't seem to be any tests for this function? Is it tested implicitly elsewhere because it seems rather complex?

src/sync/sync_user.cpp
```diff
@@ -21,44 +21,123 @@
...
+}
+
+RealmJWT::RealmJWT(const std::string& token)
+{
+    auto parts = split_token(token);
+
+    auto json = nlohmann::json::parse(base64_decode(parts[1]));
+
+    this->token = token;
+    this->expires_at = json["exp"].get<long>();
```
cmelchior: What happens if `exp` isn't in the JSON token?

src/sync/sync_user.cpp
```diff
@@ -21,44 +21,123 @@
...
+    this->issued_at = json["iat"].get<long>();
+
+    if (json.find("user_data") != json.end()) {
+        this->user_data = json["user_data"].get<std::map<std::string, util::Any>>();
+    }
+}
+
+ObjectSchema SyncUserProfile::schema()
+{
+    return { "user_profile",
```
cmelchior: If at all possible I would strongly suggest we use appropriate types for these fields. By the looks of it it could be:
- "birthday" -> Timestamp
- "min_age" -> Int
- "max_age" -> Int

src/sync/app.cpp
```diff
@@ -0,0 +1,163 @@
...
+}
+
+void App::login_with_credentials(const std::shared_ptr<AppCredentials> credentials,
+                                 std::function<void(std::shared_ptr<SyncUser>, std::unique_ptr<error::AppError>)> completion_block) {
+    // construct the route
+    std::string route = util::format("%1/providers/%2/login", m_auth_route, provider_type_from_enum(credentials->m_provider));
+
+    auto handler = [&](const Response response) {
+        // if there is a already an error code, pass the error upstream
+        if (response_code_is_fatal(response.http_status_code)) { // FIXME: handle
```
cmelchior: This doesn't handle sdk specific errors, but only consider the http error code.

src/sync/app.cpp
```diff
@@ -0,0 +1,163 @@
...
+                { "Authorization", bearer}
+            },
+            std::string()
+        }, [&](const Response profile_response) {
+            // if there is a already an error code, pass the error upstream
+            if (response_code_is_fatal(profile_response.http_status_code)) {
+                return completion_block(nullptr, handle_error(profile_response));
+            }
+
+            json = nlohmann::json::parse(profile_response.body);
```
cmelchior: Shouldn't this be in a try-catch like above?

src/sync/app.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+            std::vector<SyncUserIdentity> identities;
+            auto identities_json = json["identities"];
```
cmelchior: Relying on the JSON being well-formed seems a bit dangerous? I would wrap this and the below code in some sort of try-catch?

src/sync/app.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+std::shared_ptr<App> App::app(const std::string app_id,
+                              const realm::util::Optional<App::Config> config)
+{
+    return std::make_shared<App>(app_id, config);
+}
+
+
+
+static std::unique_ptr<error::AppError> handle_error(const Response response) {
```
cmelchior: With the introduction of `custom_error_code` needs to be rewritten I think? As a minimum, we should send the custom error back + message back to bindings. 

Also returning `{}` seems like it would be very hard for bindings to figure out exactly what is going wrong. IMO we should probably alway send on the error code and the message (which can be parsed if relevant).

src/sync/app.cpp
```diff
@@ -0,0 +1,163 @@
...
+            auto body = nlohmann::json::parse(response.body);
+            return std::make_unique<error::AppError>(app::error::ServiceError(body["errorCode"].get<std::string>(),
+                                  body["error"].get<std::string>()));
+        }
+    }
+
+    return {};
+}
+
+inline bool response_code_is_fatal(int status) {
```
cmelchior: Any custom_status_code != 0 should probably also be considered fatal

src/sync/generic_network_transport.hpp
```diff
@@ -0,0 +1,335 @@
...
+struct Response {
+    /**
+     * The status code of the HTTP response.
+     */
+    int http_status_code;
+
+    /**
+     * A custom status code provided by the language binding.
+     */
+    int binding_status_code;
```
cmelchior: We should probably clarify in docs that this `value != 0` indicates an SDK error? Or at least that is the semantics I had in mind?

src/sync/app.cpp
```diff
@@ -0,0 +1,163 @@
...
+                                                                             WRAP_JSON_OPT(profile_data, "gender", std::string),
+                                                                             WRAP_JSON_OPT(profile_data, "birthday", std::string),
+                                                                             WRAP_JSON_OPT(profile_data, "min_age", std::string),
+                                                                             WRAP_JSON_OPT(profile_data, "max_age", std::string)));
+
+            return completion_block(sync_user, {});
+        });
+    };
+
+    std::map<std::string, std::string> headers = {
```
cmelchior: Currently, the SDK's allow you to configure additional headers, but I guess that is something we can add later.

src/sync/app.hpp
```diff
@@ -0,0 +1,121 @@
...
+ Finally, its `RLMPush` object can register the current user for push notifications.
+
+ - SeeAlso:
+ `RemoteMongoClient`,
+ `RLMPush`,
+ [Functions](https://docs.mongodb.com/stitch/functions/)
+ */
+class App {
+public:
+    struct Config {
```
cmelchior: Currently in SDK's we also support setting a custom auth header for proxies and any other custom headers. I suspect defining them here would make sense. But I don't know if that is relevant if we only are going to support MongoDB Realm (compared to selfhosted) as a backend in an initial release?

src/sync/app.hpp
```diff
@@ -0,0 +1,121 @@
...
+    */
+    void login_with_credentials(const std::shared_ptr<AppCredentials> credentials,
+                                std::function<void(std::shared_ptr<SyncUser>, std::unique_ptr<error::AppError>)> completion_block);
+
+
+    App(const std::string app_id,
+        const realm::util::Optional<App::Config> config) :
+    m_app_id(app_id) {
+
+        std::string default_base_url = "https://stitch.mongodb.com";
```
cmelchior: Would perhaps be nice to move constants like outside functions and into class-level properties so they are easier to find?

src/sync/app.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+        if (config) {
+            if (config.value().base_url) {
+                default_base_url = config.value().base_url.value();
+            }
+
+            if (config.value().default_request_timeout_ms) {
+                m_request_timeout_ms = config.value().default_request_timeout_ms.value();
+            } else {
+                m_request_timeout_ms = 60000;
```
cmelchior: Move the default timeout into a class-level constant?

src/sync/app.hpp
```diff
@@ -0,0 +1,121 @@
...
+            }
+
+            if (config.value().transport) {
+                // TODO: Install custom transport
+            }
+        } else {
+            m_request_timeout_ms = 60000;
```
cmelchior: Same?

src/sync/app_credentials.cpp
```diff
@@ -0,0 +1,116 @@
...
+IdentityProvider const IdentityProviderFacebook  = "oauth2-facebook";
+IdentityProvider const IdentityProviderApple     = "oauth2-apple";
+IdentityProvider const IdentityProviderUsernamePassword     = "local-userpass";
+
+std::string AppCredentials::serialize_as_json() const
+{
+    return m_payload_factory();
+}
+
+IdentityProvider provider_type_from_enum(AuthProvider provider)
```
cmelchior: This doesn't seem to cover the full list? https://docs.mongodb.com/stitch/authentication/providers/

src/sync/generic_network_transport.hpp
jsflax: They are defined in the Stitch go codebase 😅

src/sync/generic_network_transport.hpp
jsflax: Seemed pertinent to only add methods as needed.

src/sync/sync_user.cpp
jsflax: I do not think that is a safe assumption since we are getting this data from various third party sources. Until it is sanitised by Stitch, I would prefer not to do this.

src/sync/generic_network_transport.hpp
jsflax: 👍

src/sync/app.hpp
jsflax: We do not need to worry about self hosted right now.

src/sync/app_credentials.cpp
jsflax: I was intending to parallelise the rest of this work so that we can unblock.

src/sync/generic_network_transport.hpp
cmelchior: I just assumed that the first one attempting to add a PUT or DELETE would get rather annoyed if it is missing, but it isn't something I feel strongly about.

src/sync/generic_network_transport.hpp
cmelchior: Can we have a link to them here for easy reference then? Even if it links to a private repo it is rather nice instead of trying to find them manually?

src/sync/app.cpp
jsflax: 👍

src/sync/sync_user.cpp
jsflax: It is tested implicitly by using real JWTs in our tests. However, it should have more robust testing.

jsflax:
@cmelchior PTAL

src/sync/generic_network_transport.hpp
```diff
@@ -248,15 +266,16 @@ struct ServiceError : public AppError
...
 } // namespace error
 
-
-enum class Method {
-    get, post, patch
+/**
+ * An HTTP method type.
+ */
+enum class HttpMethod {
+    get, post, patch, put, del
```
cmelchior:
```suggestion
    get, post, patch, put, delete
```

src/sync/generic_network_transport.hpp
```diff
@@ -302,7 +321,7 @@ struct Response {
     /**
      * A custom status code provided by the language binding.
```
cmelchior:
```suggestion
     * A custom status code provided by the language binding. A value != 0 is interpreted as an error.
```

src/sync/app.cpp
```diff
@@ -0,0 +1,188 @@
...
+#include "../external/json/json.hpp"
+
+// wrap an optional json key into the Optional type
+#define WRAP_JSON_OPT(JSON, KEY, RET_TYPE) \
+JSON.find(KEY) != JSON.end() ? Optional<RET_TYPE>(JSON[KEY].get<RET_TYPE>()) : realm::util::none
+
+namespace realm {
+namespace app {
+
+std::shared_ptr<App> App::app(const std::string app_id,
```
tgoyne: IMO this function just makes the code using it less understandable than calling `make_shared<App>()` directly would be (and there doesn't appear to be any reason the App needs to be wrapped in a shared_ptr at all in the test). We wrap the call to `make_shared()` in other places because of `enabled_shared_from_this()`, which makes it so that an instance of the type which isn't owned by a shared_ptr will crash at runtime, but that isn't used here.

src/sync/app.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+    If the log in completes successfully, the completion block will be called, and a
+    `SyncUser` representing the logged-in user will be passed to it. This user object
+    can be used to open `Realm`s and retrieve `SyncSession`s. Otherwise, the
+    completion block will be called with an error.
+
+    - parameter credentials: A `SyncCredentials` object representing the user to log in.
+    - parameter completion: A callback block to be invoked once the log in completes.
+    */
+    void login_with_credentials(const std::shared_ptr<AppCredentials> credentials,
```
tgoyne:
```suggestion
    void login_with_credentials(std::shared_ptr<AppCredentials> credentials,
```

`const` here does nothing. Whether or not a non-reference parameter is `const` is not actually part of the type of a function, as it merely determines the constness of the local variable visible within the function.

src/sync/app.cpp
```diff
@@ -0,0 +1,188 @@
...
+    return std::make_shared<App>(app_id, config);
+}
+
+const std::string App::default_base_url = "https://stitch.mongodb.com";
+const std::string App::base_path = "/api/client/v2.0";
+const std::string App::app_path = "/app";
+const std::string App::auth_path = "/auth";
+const uint64_t App::default_timeout_ms = 60000;
+
+static std::unique_ptr<error::AppError> handle_error(const Response response) {
```
tgoyne:
```suggestion
static std::unique_ptr<error::AppError> handle_error(const Response& response) {
```

Taking the response by value here copies the entire response.

src/sync/app.cpp
```diff
@@ -0,0 +1,188 @@
...
+    if (response.body.empty()) {
+        return {};
+    }
+
+    if (response.headers.find("Content-Type") != response.headers.end()) {
```
tgoyne:
```suggestion
    if (auto ct = response.headers.find("Content-Type"); ct != response.headers.end() && ct->second == "application/json") {
```

src/sync/generic_network_transport.hpp
```diff
@@ -0,0 +1,354 @@
...
+    {
+        return this->what();
+    }
+
+    int code() const
+    {
+        return m_code;
+    }
+
+    AppError()
```
tgoyne: A default constructor for an exception type doesn't make much sense.

src/sync/generic_network_transport.hpp
```diff
@@ -0,0 +1,354 @@
...
+struct AppError : public std::runtime_error
+{
+    enum class Type {
+        Unknown,
+        JSON,
+        Service,
+        Custom
+    };
+
+    const std::string category() const
```
tgoyne: Returning `const std::string` doesn't do anything useful; it just forces the string to be copied an extra time.

src/sync/generic_network_transport.hpp
```diff
@@ -0,0 +1,354 @@
...
+#include <string>
+#include <memory>
+#include <map>
+#include <vector>
+
+#include <realm/util/to_string.hpp>
+
+namespace realm {
+namespace app {
+namespace error {
```
tgoyne: This namespace just makes things more verbose without disambiguating anything.

src/sync/app.cpp
```diff
@@ -0,0 +1,188 @@
...
+    if (response.custom_status_code != 0) {
+        return std::make_unique<error::AppError>(response.body,
+                                                 response.custom_status_code,
+                                                 error::AppError::Type::Custom);
+    }
+
+    return {};
+}
+
+inline bool response_code_is_fatal(Response response) {
```
tgoyne:
```suggestion
inline bool response_code_is_fatal(Response const& response) {
```

src/sync/generic_network_transport.hpp
```diff
@@ -0,0 +1,354 @@
...
+
+#pragma mark Errors
+
+struct AppError : public std::runtime_error
```
tgoyne: No newline before { for types.

src/sync/generic_network_transport.hpp
```diff
@@ -0,0 +1,354 @@
...
+     * The body of the HTTP response.
+     */
+    std::string body;
+};
+
+#pragma mark GenericNetworkTransport
+
+/// Generic network transport for foreign interfaces.
+struct GenericNetworkTransport {
+    typedef std::unique_ptr<GenericNetworkTransport> (*network_transport_factory)();
```
tgoyne:
```suggestion
    using network_transport_factory = std::unique_ptr<GenericNetworkTransport> (*)();
```

src/sync/generic_network_transport.hpp
```diff
@@ -0,0 +1,354 @@
...
+};
+
+/// Struct allowing for generic error data.
+struct ServiceError : public AppError
+{
+    ServiceError(std::string raw_code, std::string message)
+    : AppError(message, static_cast<int>(error_code_for_string(raw_code)), AppError::Type::Service)
+    {
+    }
+    const std::string message() const { return std::runtime_error::what(); };
```
tgoyne: I don't understand what this is trying to do, but I'm pretty sure it's not doing anything compared to just calling the function defined in the parent class.

src/sync/generic_network_transport.hpp
```diff
@@ -0,0 +1,354 @@
...
+            return ServiceErrorCode::user_already_confirmed;
+        else if (code == "UserNotFound")
+            return ServiceErrorCode::user_not_found;
+        else if (code == "UserDisabled")
+            return ServiceErrorCode::user_disabled;
+        else
+            return ServiceErrorCode::unknown;
+    }
+
+    const std::string m_raw_code;
```
tgoyne: This member does not appear to ever be used.

src/sync/app.cpp
```diff
@@ -0,0 +1,188 @@
...
+    // FIXME: our tests currently only generate codes 0 and 200,
+    // but we need more robust error handling here; eg. should a 300
+    // redirect really be considered fatal or should we automatically redirect?
+    return response.http_status_code >= 300 || response.http_status_code < 200 || response.custom_status_code != 0;
+}
+
+void App::login_with_credentials(const std::shared_ptr<AppCredentials> credentials,
+                                 std::function<void(std::shared_ptr<SyncUser>, std::unique_ptr<error::AppError>)> completion_block) {
```
tgoyne: This completion function could pass `util::Optional<AppError>` and simplify all the code around building errors. Despite having subclasses, `AppError` isn't actually a polymorphic type.

src/sync/app.cpp
```diff
@@ -0,0 +1,188 @@
...
+    // construct the route
+    std::string route = util::format("%1/providers/%2/login", m_auth_route, provider_type_from_enum(credentials->m_provider));
+
+    auto handler = [&](const Response response) {
```
tgoyne:
```suggestion
    auto handler = [&](const Response& response) {
```

src/sync/app.cpp
```diff
@@ -0,0 +1,188 @@
...
+            HttpMethod::get,
+            profile_route,
+            m_request_timeout_ms,
+            {
+                { "Content-Type", "application/json;charset=utf-8" },
+                { "Accept", "application/json" },
+                { "Authorization", bearer}
+            },
+            std::string()
+        }, [&](const Response profile_response) {
```
tgoyne:
```suggestion
        }, [&](const Response& profile_response) {
```

src/sync/app.hpp
```diff
@@ -0,0 +1,125 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#ifndef REALM_APP_HPP
+#define REALM_APP_HPP
+
+#include "sync_user.hpp"
+#include "app_credentials.hpp"
+#include "generic_network_transport.hpp"
```
tgoyne:
```
#include "sync/app_credentials.hpp"
#include "sync/generic_network_transport.hpp"
#include "sync/sync_user.hpp"
```

src/sync/app.hpp
```diff
@@ -0,0 +1,125 @@
...
+    - parameter credentials: A `SyncCredentials` object representing the user to log in.
+    - parameter completion: A callback block to be invoked once the log in completes.
+    */
+    void login_with_credentials(const std::shared_ptr<AppCredentials> credentials,
+                                std::function<void(std::shared_ptr<SyncUser>, std::unique_ptr<error::AppError>)> completion_block);
+
+
+    App(const std::string app_id,
+        const realm::util::Optional<App::Config> config) :
+    m_app_id(app_id) {
```
tgoyne:
```
App(const std::string& app_id,
        const realm::util::Optional<App::Config>& config)
: m_app_id(app_id)
{
```

Leading colon/comma for initializer lists, and `{` on a new line for function definitions (and nothing else). As this a bunch of things this should take the parameters by const reference rather than copying them.

src/sync/app.hpp
```diff
@@ -0,0 +1,125 @@
...
+                // TODO: Install custom transport
+            }
+        } else {
+            m_request_timeout_ms = default_timeout_ms;
+        }
+
+        m_base_route = base_url + base_path;
+        m_app_route = m_base_route + app_path + "/" + app_id;
+        m_auth_route = m_app_route + auth_path;
+    };
```
tgoyne:
```
        std::string base_url;
        if (config) {
            base_url = config->base_url.value_or(default_base_url);
            m_request_timeout_ms = config->default_request_timeout_ms.value_or(default_timeout_ms);

            if (config->transport) {
                // TODO: Install custom transport
            }
        }
        else {
            base_url = default_base_url;
            m_request_timeout_ms = default_timeout_ms;
        }

        m_base_route = base_url + base_path;
        m_app_route = m_base_route + app_path + "/" + app_id;
        m_auth_route = m_app_route + auth_path;
    }
```

src/sync/app_credentials.cpp
```diff
@@ -0,0 +1,116 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or utilied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#include "app_credentials.hpp"
+#include "../external/json/json.hpp"
```
tgoyne:
```suggestion
#include <json.hpp>
```

src/sync/app_credentials.cpp
```diff
@@ -0,0 +1,116 @@
...
+        case AuthProvider::ANONYMOUS:
+            return IdentityProviderAnonymous;
+        case AuthProvider::APPLE:
+            return IdentityProviderApple;
+        case AuthProvider::FACEBOOK:
+            return IdentityProviderFacebook;
+        case AuthProvider::USERNAME_PASSWORD:
+            return IdentityProviderUsernamePassword;
+    }
+    throw std::runtime_error("Error: unknown provider type in provider_type_from_enum");
```
tgoyne: Including "Error:" in the exception message is redundant.

src/sync/app_credentials.hpp
```diff
@@ -0,0 +1,112 @@
...
+    std::string serialize_as_json() const;
+    
+    ~AppCredentials() = default;
+
+    AppCredentials() {};
+
+    AppCredentials(AppCredentials&&) = default;
+    AppCredentials& operator=(AppCredentials&&) = default;
+    AppCredentials(AppCredentials const&) = delete;
+    AppCredentials& operator=(AppCredentials const&) = delete;
```
tgoyne: None of these appear to be necessary.

src/sync/app_credentials.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ Opaque credentials representing a specific Realm Object Server user.
+ */
+struct AppCredentials {
+public:
+    /**
+     Construct and return credentials from a Facebook account token.
+     */
+    static std::shared_ptr<AppCredentials> facebook(const AppCredentialsToken access_token);
```
tgoyne: There doesn't appear to be any reason for these to be returning a `std::shared_ptr<AppCredentials>` rather than just returning a `AppCredentials` (and similarly `login_with_credentials` doesn't have any reason to be taking a `std::shared_ptr<AppCredentials>` rather than a const ref to a `AppCredentials`).

src/sync/sync_user.hpp
```diff
@@ -24,10 +24,14 @@
...
 #include <mutex>
+#include <map>
 
 #include "util/atomic_shared_ptr.hpp"
-
+#include <realm/util/any.hpp>
 #include <realm/util/optional.hpp>
+#include <realm/table.hpp>
+
+#include "../object_schema.hpp"
```
tgoyne:
```suggestion
#include "object_schema.hpp"
```

src/sync/impl/sync_metadata.hpp
```diff
@@ -27,6 +27,7 @@
 
 #include "results.hpp"
 #include "shared_realm.hpp"
+#include "../sync_user.hpp"
```
tgoyne:
```suggestion
#include "sync/sync_user.hpp"
```

src/sync/sync_user.hpp
```diff
@@ -53,16 +57,150 @@ struct SyncUserIdentifier {
...
+
+    RealmJWT(const std::string& token);
+
+    bool operator==(const RealmJWT& other) const
+    {
+        return token == other.token;
+    }
+};
+
+struct SyncUserProfile {
```
tgoyne: There doesn't really appear to be any reason for this to not just be a simple struct that's passed around by value rather than a shared object which then needs getters and private members.

PR #924, opened 2020-02-20
Expose bug with progress notifications

cmelchior:
In some cases the Upload progress notifications report the wrong number of bytes uploaded so `uploadable == uploaded` never happens which can break quite a lot of logic when using the Progress API's.

The output looks like this:
```
Progress 0 : 207
Progress 185 : 207
// Notice Progress 207 : 207 never happens
```

See also https://jira.mongodb.org/browse/RSYNC-101


cmelchior:
Fixed by https://github.com/realm/realm-sync/pull/3261

PR #926, opened 2020-02-24
Restore support for notifications on sorted primitive lists

tgoyne:
As Lists can no longer be accessed via Query/TableView, Lists which have been wrapped in a 
 Results (normally for sort/distinct) need their own special handling for notifications.

The logic for this is very similar to sorted notifications in core 5, with manual updating of row indices to reflect insertions/deletions prior to diffing.

### DIFF
--- a/src/impl/collection_change_builder.cpp
+++ b/src/impl/collection_change_builder.cpp
@@ -282,7 +282,7 @@ void CollectionChangeBuilder::verify()
 
 namespace {
 struct RowInfo {
-    int64_t obj_key;
+    int64_t key;
     size_t prev_tv_index;
     size_t tv_index;
 };
@@ -319,7 +319,7 @@ void calculate_moves_backlinks(std::vector<RowInfo>& new_rows, IndexSet& removed
         // First check if this row even could have moved. If it can't, just
         // treat it as a match and move on, and we'll handle the row we were
         // expecting when we hit it later.
-        if (!move_candidates.contains(row.obj_key)) {
+        if (!move_candidates.contains(row.key)) {
             expected = row.shifted_tv_index + 1;
             continue;
         }
@@ -346,7 +346,7 @@ class LongestCommonSubsequenceCalculator {
 public:
     // A pair of an object key and an index in the table view
     struct Row {
-        int64_t obj_key;
+        int64_t key;
         size_t tv_index;
     };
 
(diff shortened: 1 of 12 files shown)


src/impl/collection_change_builder.cpp
```diff
@@ -282,7 +282,7 @@ void CollectionChangeBuilder::verify()
 
 namespace {
 struct RowInfo {
-    int64_t obj_key;
+    int64_t key;
```
tgoyne: This change from obj_key to key is because it's sometimes an object key, and sometimes an index in a list.

tests/collection_change_indices.cpp
```diff
@@ -363,6 +363,7 @@ TEST_CASE("collection_change: calculate() sorted") {
 
     auto all_modified = [](size_t) { return true; };
     auto none_modified = [](size_t) { return false; };
+    size_t npos = -1;
```
tgoyne: All of the changes in this file are related to disambiguating the size_t/int64_t overloads of calculate() with initializer list arguments.

tests/thread_safe_reference.cpp
```diff
@@ -668,18 +668,15 @@ TEST_CASE("thread safe reference") {
...
             r->begin_transaction();
             obj = create_object(r, "int array", {{"value", AnyVector{INT64_C(1)}}});
-            List list(r, *get_table(*r, "int array"), 0, 0);
+            List list(r, obj.obj(), get_table(*r, "int array")->get_column_key("value"));
             r->commit_transaction();
 
-            auto results = delete_and_resolve(list.sort({{"self", true}}));
-            REQUIRE(results.is_valid());
-            REQUIRE(results.size() == 0);
+            REQUIRE(!delete_and_resolve(list).is_valid());
```
tgoyne: This behavior is inconsistent with lists of objects, but consistent with the non-handover case so I think the list-of-objects one is the one that's wrong.

ironage: I agree, this looks like correct behaviour here to me.
Isn't the difference with the list-of-objects test above because it is testing the `Results` returned by sorting the list and not the list itself? We probably haven't even implemented getting a `Results` on a list of primitives yet (other than a list of objects).

src/impl/results_notifier.hpp
```diff
@@ -26,14 +26,23 @@
...
+};
+
+class ResultsNotifier : public ResultsNotifierBase {
+public:
+    ResultsNotifier(Results& target);
+    bool get_tableview(TableView& out) override;
 
 private:
     std::unique_ptr<Query> m_query;
+    std::shared_ptr<LstBase> m_list;
```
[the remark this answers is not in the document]
tgoyne: Right, it's a remnant of when I was trying to handle lists in the same notifier type.

PR #929, opened 2020-02-25
RCOCOA-537 Address comments from Jasons PR

leemaguire:


src/sync/app.hpp
```diff
@@ -111,7 +108,8 @@ class App {
     std::string m_app_route;
     std::string m_auth_route;
 
-    int m_request_timeout;
+    /// timeout in milliseconds, default is 60 seconds
+    uint64_t m_request_timeout = 60000;
```
ironage: Can we make this `m_request_timeout_ms` for readability?

leemaguire: done

PR #931, opened 2020-02-26
Fix some bugs with creating ThreadSafeReference inside write transactions

tgoyne:
With core 5 this was forbidden, but with core 6 it mostly works fine. The exception is with objects which didn't exist yet at the start of the write transaction, as the object key doesn't actually exist in the stored version. For `ThreadSafeReference<Object>` and `ThreadSafeReference<List>` this has a straightforward fix of refreshing the target Realm if it's at the version which the relevant write was started at before trying to hand over the object/list. For Results derived from a List or LinkingObjects, Query doesn't expose the required members to do this so just disallow it for now.

### DIFF
--- a/src/server/global_notifier.cpp
+++ b/src/server/global_notifier.cpp
@@ -395,11 +395,11 @@ std::unordered_map<std::string, ObjectChangeSet> const& GlobalNotifier::ChangeNo
     auto realm = Realm::get_shared_realm(config);
 
     Realm::Internal::begin_read(*realm, m_old_version);
-    Group const& g = realm->read_group();
+    Group& g = realm->read_group();
 
     _impl::TransactionChangeInfo info;
     info.track_all = true;
-    _impl::transaction::advance(realm->transaction(), info, m_new_version);
+    _impl::transaction::advance(static_cast<Transaction&>(g), info, m_new_version);
 
     m_changes.reserve(info.tables.size());
     auto table_keys = g.get_table_keys();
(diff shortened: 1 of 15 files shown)


src/server/global_notifier.cpp
```diff
@@ -395,11 +395,11 @@ std::unordered_map<std::string, ObjectChangeSet> const& GlobalNotifier::ChangeNo
     auto realm = Realm::get_shared_realm(config);
 
     Realm::Internal::begin_read(*realm, m_old_version);
-    Group const& g = realm->read_group();
+    Group& g = realm->read_group();
 
     _impl::TransactionChangeInfo info;
     info.track_all = true;
-    _impl::transaction::advance(realm->transaction(), info, m_new_version);
+    _impl::transaction::advance(static_cast<Transaction&>(g), info, m_new_version);
```
ironage: This cast will do the correct thing, but this requires knowledge of the implementation. Why not operate on `realm->transaction()` here?

tgoyne: Regardless of how it gets to the Transaction, the global notifier is operating on implementation details that it's not supposed to know about. Exposing `transaction()` for it doesn't really help with letting us change things about the implementation of Realm without breaking this, and having `transaction()` available will inevitably result in it being used in places that should be going through the Realm instance instead.

This code shouldn't actually be using a `Realm` instance at all, but with the GN going away entirely in v10 I don't think it's worth trying to redesign.

ironage: Good points. :shipit:

PR #932, opened 2020-02-27
App updates

ironage:
Addressing review comments on https://github.com/realm/realm-object-store/pull/921

There's still outstanding work, including:

- [x] Remove shared_ptr surrounding AppCredentials
- [x] Redesign the error handling to use an Optional instead of a unique_ptr. Consider merging the different error classes all into one class in the pattern of `SyncError`.

### DIFF
--- a/src/sync/app.cpp
+++ b/src/sync/app.cpp
@@ -48,21 +48,21 @@ App::App(const Config& config)
     REALM_ASSERT(m_config.transport_generator);
 }
 
-static std::unique_ptr<error::AppError> handle_error(const Response& response) {
+static std::unique_ptr<AppError> handle_error(const Response& response) {
     if (response.body.empty()) {
         return {};
     }
 
     if (auto ct = response.headers.find("Content-Type"); ct != response.headers.end() && ct->second == "application/json") {
         auto body = nlohmann::json::parse(response.body);
-        return std::make_unique<error::AppError>(app::error::ServiceError(body["errorCode"].get<std::string>(),
+        return std::make_unique<AppError>(app::ServiceError(body["errorCode"].get<std::string>(),
                               body["error"].get<std::string>()));
     }
 
     if (response.custom_status_code != 0) {
-        return std::make_unique<error::AppError>(response.body,
+        return std::make_unique<AppError>(response.body,
                                                  response.custom_status_code,
-                                                 error::AppError::Type::Custom);
+                                                 AppError::Type::Custom);
     }
 
     return {};
@@ -78,7 +78,7 @@ inline bool response_code_is_fatal(Response const& response) {
 }
 
 void App::login_with_credentials(const AppCredentials& credentials,
-                                 std::function<void(std::shared_ptr<SyncUser>, std::unique_ptr<error::AppError>)> completion_block) {
+                                 std::function<void(std::shared_ptr<SyncUser>, std::unique_ptr<AppError>)> completion_block) {
     // construct the route
     std::string route = util::format("%1/providers/%2/login", m_auth_route, credentials.provider_as_string());
 
@@ -92,7 +92,7 @@ void App::login_with_credentials(const AppCredentials& credentials,
         try {
             json = nlohmann::json::parse(response.body);
         } catch(std::domain_error e) {
-            return completion_block(nullptr, std::make_unique<error::AppError>(app::error::JSONError(app::error::JSONErrorCode::malformed_json,
+            return completion_block(nullptr, std::make_unique<AppError>(app::JSONError(app::JSONErrorCode::malformed_json,
                                                                                                        e.what())));
         }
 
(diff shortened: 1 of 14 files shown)


src/sync/app.cpp
```diff
@@ -30,29 +32,43 @@ JSON.find(KEY) != JSON.end() ? Optional<RET_TYPE>(JSON[KEY].get<RET_TYPE>()) : r
...
+App::App(const std::string& app_id, const realm::util::Optional<App::Config>& config)
+    : m_app_id(app_id)
+    {
+    std::string base_url;
+    if (config) {
+        base_url = config->base_url.value_or(default_base_url);
+        m_request_timeout_ms = config->default_request_timeout_ms.value_or(default_timeout_ms);
+
+        if (config->transport) {
+            // TODO: Install custom transport
```
jsflax: [minor] This should overwrite the static generic transport factory that the SDKs would be overwriting. This configuration field is to give users their own opportunity to use their own generic network transport.

src/sync/impl/sync_metadata.cpp
```diff
@@ -77,6 +78,7 @@ realm::Schema make_schema()
             {c_sync_provider_type, PropertyType::String}
         }},
         {c_sync_profile, {
+            {c_sync_profile_name, PropertyType::String|PropertyType::Nullable},
```
jsflax: [q] What is the purpose of this field?

ironage: I don't know. The property was something you added to `SyncUserProfile` with the comment: "The full name of the user." Since it was the only property not persisted, I assumed that was an oversight and added it in for consistency. Should it instead be removed from `SyncUserProfile`?

src/sync/app.cpp
ironage: I've addressed this by making the transport per App now instead of a global member.

ironage:
I think this addresses all the feedback so far, it also changes the API significantly, so I'm open to additional feedback about that.

src/sync/app.cpp
```diff
@@ -16,67 +16,71 @@
...
+, m_base_route(config.base_url.value_or(default_base_url) + default_base_path)
+, m_app_route(m_base_route + default_app_path + "/" + config.app_id)
+, m_auth_route(m_app_route + default_auth_path)
+, m_request_timeout_ms(config.default_request_timeout_ms.value_or(default_timeout_ms))
+{
+    REALM_ASSERT(m_config.transport_generator);
+}
 
-static std::unique_ptr<error::AppError> handle_error(const Response response) {
+static Optional<AppError> handle_error(const Response& response) {
```
jsflax: I don't believe this should be optional. The idea is to parse the error as best we can, but if for some reason we cannot, we still need to let the binding know that _some_ error has occurred, even if we return "Unknown" or -1.

src/sync/app.cpp
```diff
@@ -16,67 +16,71 @@
...
 
-// wrap an optional json key into the Optional type
-#define WRAP_JSON_OPT(JSON, KEY, RET_TYPE) \
-JSON.find(KEY) != JSON.end() ? Optional<RET_TYPE>(JSON[KEY].get<RET_TYPE>()) : realm::util::none
+#include "sync/app_credentials.hpp"
+#include "sync/generic_network_transport.hpp"
+#include "sync/sync_manager.hpp"
+
+#include <json.hpp>
+#include <sstream>
```
tgoyne: <sstream> doesn't appear to be used anywhere here

src/sync/app.cpp
```diff
@@ -16,67 +16,71 @@
...
-void App::login_with_credentials(const std::shared_ptr<AppCredentials> credentials,
-                                 std::function<void(std::shared_ptr<SyncUser>, std::unique_ptr<error::AppError>)> completion_block) {
+void App::login_with_credentials(const AppCredentials& credentials,
+                                 std::function<void(std::shared_ptr<SyncUser>, Optional<AppError>)> completion_block) {
     // construct the route
-    std::string route = util::format("%1/providers/%2/login", m_auth_route, provider_type_from_enum(credentials->m_provider));
+    std::string route = util::format("%1/providers/%2/login", m_auth_route, credentials.provider_as_string());
 
-    auto handler = [&](const Response response) {
+    auto handler = [&](const Response& response) {
```
tgoyne: This needs to capture `completion_block` by value rather than by reference as `handler` may be called after we return from this function.

src/sync/app.cpp
```diff
@@ -121,7 +120,7 @@ void App::login_with_credentials(const std::shared_ptr<AppCredentials> credentia
                 { "Authorization", bearer}
             },
             std::string()
-        }, [&](const Response profile_response) {
+        }, [&](const Response& profile_response) {
```
tgoyne: Another unsafe capture by reference.

src/sync/app.cpp
ironage: I combined the error check and error generation together so this whole mechanism is more robust.

PR #933, opened 2020-02-27
Fix compiler error on Android

cmelchior:


src/object_store.cpp
```diff
@@ -615,7 +615,7 @@ static void apply_post_migration_changes(Group& group,
 void ObjectStore::apply_schema_changes(Transaction& group, uint64_t schema_version,
                                        Schema& target_schema, uint64_t target_schema_version,
                                        SchemaMode mode, std::vector<SchemaChange> const& changes,
-                                       util::Optional<std::string> sync_user_id,
+                                       util::Optional<std::string> /*sync_user_id*/,
```
jsflax: [q] Why doesn't this compile on Android?

tgoyne:
This parameter should be removed entirely as it was a QBS thing.

cmelchior:
Not entirely sure what is going on. Unit tests can run succesfully without sync. With a sync enabled build this happens:

```
/Users/cm/Realm/realm-object-store/tests/object.cpp:205: FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  SIGSEGV - Segmentation violation signal
```

But the unit test in question there is not at all related to Sync 🤔

PR #934, opened 2020-02-28
Implement missing parts of embedded objects

tgoyne:
Only Object::create() was implemented, and mutating an object or list after creation didn't work. Migrations involving embedded tables didn't work correctly, and some schema validation was missing.

This changes the SDK API for embedded objects a bit. unbox_embedded() and create_embedded_obectj() are gone, and instead the context must implement create_embedded_object(), which uses the object and property passed to the context at construction time. CreatePolicy is now a struct of bools because there's an additional bit of data that needs to be passed along (whether we're in a create operation or add operation) which determines how we handle nested objects which are already managed by the realm (inside create() we copy existing managed embedded objects, but setting a link directly to a managed embedded object throws). Cocoa previously tracked this state externally, but in a way that didn't work for embedded objects.

This depends on a version of core newer than the latest release, so it won't build successfully on CI.

### DIFF
--- a/src/object.hpp
+++ b/src/object.hpp
@@ -31,16 +31,41 @@ namespace _impl {
     class ObjectNotifier;
 }
 
-enum class CreatePolicy : int8_t {
-    // Do not create objects if given something that could be converted to a
-    // Realm object but isn't. Used for things like find().
-    Skip,
-    // Throw an exception if an object with the same PK already exists.
-    ForceCreate,
-    // If an object with the same PK already exists, set all fields in the input.
-    UpdateAll,
-    // If an object with the same PK already exists, only set fields which have changed.
-    UpdateModified
+/// Options for how objects should be unboxed by a context.
+///
+/// unbox<Obj>() is used for several different operations which want an Obj
+/// from a SDK type. CreatePolicy packs together all of the different options
+/// around what unbox() should do.
+struct CreatePolicy {
+    /// If given something that is not a managed Object, should an object be
+    /// created in the Realm? False for pure lookup functions such as find(),
+    /// index_of() and queries, true for everything else.
+    bool create : 1;
+    /// Should the input object be copied into the Realm even if it is already
+    /// an object managed by the current Realm? True for realm.create(), false
+    /// for things like setting a link property.
+    bool copy : 1;
+    /// If the object has a primary key and an object with the same primary key
+    /// already exists, should the existing object be updated rather than
+    /// throwing an exception? Only meaningful if .create is true.
+    bool update : 1;
+    /// When updating an object, should the old and new objects be diffed and
+    /// only the values which are different be set, or should all fields be set?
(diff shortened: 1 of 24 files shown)


src/object_schema.cpp
```diff
@@ -63,27 +63,27 @@ ObjectSchema::ObjectSchema(std::string name, IsEmbedded is_embedded, std::initia
...
+        case col_type_Double:    return PropertyType::Double | flags;
+        case col_type_Bool:      return PropertyType::Bool | flags;
+        case col_type_String:    return PropertyType::String | flags;
+        case col_type_Binary:    return PropertyType::Data | flags;
+        case col_type_Timestamp: return PropertyType::Date | flags;
+        case col_type_OldMixed:  return PropertyType::Any | flags;
+        case col_type_ObjectId:  return PropertyType::ObjectId | flags;
+        case col_type_Decimal:   return PropertyType::Decimal | flags;
+        case col_type_Link:      return PropertyType::Object | PropertyType::Nullable;
+        case col_type_LinkList:  return PropertyType::Object | PropertyType::Array;
```
ironage: Potentially we could now add a case for converting `col_type_BackLink` into a `PropertyType::LinkingObjects` but that's not really part of this PR (not blocking).

src/object_schema.cpp
```diff
@@ -63,27 +63,27 @@ ObjectSchema::ObjectSchema(std::string name, IsEmbedded is_embedded, std::initia
     }
 }
 
-PropertyType ObjectSchema::from_core_type(Table const& table, ColKey col)
+PropertyType ObjectSchema::from_core_type(ColKey col)
```
ironage: not having to pass around the table is a nice simplification all around the codebase 💯

src/object_schema.cpp
ironage: We also have `col_type_OldStringEnum` but that should not be possible to reach since AFAIK if we had any enum columns they should have all been converted to string columns in the migration to core 6.

src/object.hpp
```diff
@@ -31,16 +31,41 @@ namespace _impl {
...
+/// Options for how objects should be unboxed by a context.
+///
+/// unbox<Obj>() is used for several different operations which want an Obj
+/// from a SDK type. CreatePolicy packs together all of the different options
+/// around what unbox() should do.
+struct CreatePolicy {
+    /// If given something that is not a managed Object, should an object be
+    /// created in the Realm? False for pure lookup functions such as find(),
+    /// index_of() and queries, true for everything else.
+    bool create : 1;
```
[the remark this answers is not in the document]
tgoyne: It's not something that the compiler can legally do automatically. Packing them into a bitfield means that you can't take the address of the members, and isn't strictly better performance-wise.

Actually, it's probably not useful to pack them into a bitfield here as we don't have arrays of this type and it fits into a register either way.

PR #935, opened 2020-02-28
Core 10: Cannot create pk values with null and 0

cmelchior:


jedelbo:
This will not compile before the next version of core is used.

tests/object.cpp
```diff
@@ -923,6 +923,18 @@ TEST_CASE("object") {
         REQUIRE(obj.obj().get<String>(col_pk_str) == "value");
     }
 
+    SECTION("create null and 0 primary keys for Int types") {
```
cmelchior: Shouldn't we add unit tests for all the primary key types?

jedelbo:
Now waiting for new sync release

tests/object.cpp
jedelbo: Added a test with ObjectId as primary key. We have one for string.

PR #936, opened 2020-03-01
Add auth provider client stubs

jsflax:
Adds a series of structs and definitions that will allow a user to use functionality based on specific providers that allow it. The two providers that Stitch currently has additional functionality for are:
- `UserAPIKey`
- `UsernamePassword`

See: [CoreUserAPIKeyAuthProviderClient.swift](https://github.com/mongodb/stitch-ios-sdk/blob/master/Core/StitchCoreSDK/Sources/StitchCoreSDK/Auth/Providers/UserAPIKey/CoreUserAPIKeyAuthProviderClient.swift)
See: [CoreUserPasswordAuthProviderClient.swift](https://github.com/mongodb/stitch-ios-sdk/blob/master/Core/StitchCoreSDK/Sources/StitchCoreSDK/Auth/Providers/UserPassword/CoreUserPasswordAuthProviderClient.swift)

Bindings would call this as: 
```cpp
app.provider_client<UsernamePasswordProviderClient>().register_email(
    "bar@foo.com",
    "password", 
    [=](Optional<AppError> error) {
        if (error) {
            ...
        }
    }
);
```
The main purpose of this substructure is to declutter the App namespace from functionality that only certain users will need.

`TODO`: The actual method impls. The reason I have only stubbed these out is so the rest of the work can be parallelised.

### DIFF
--- a/src/sync/app.hpp
+++ b/src/sync/app.hpp
@@ -26,8 +26,6 @@
 namespace realm {
 namespace app {
 
-#pragma mark App
-
 /**
  The `App` has the fundamental set of methods for communicating with a MongoDB Realm application backend.
 
@@ -62,11 +60,11 @@ public:
     */
     struct UserAPIKey {
         // The ID of the key.
-        realm::ObjectId id;
+        ObjectId id;
 
         // The actual key. Will only be included in
         // the response when an API key is first created.
-        std::string key;
+        Optional<std::string> key;
 
         // The name of the key.
         std::string name;
@@ -90,7 +88,7 @@ public:
          *     - completion_block: A callback to be invoked once the call is complete.
         */
         void create_api_key(const std::string& name,
-                            std::function<void(std::shared_ptr<UserAPIKey>, Optional<AppError>)> completion_block);
+                            std::function<void(UserAPIKey, Optional<AppError>)> completion_block);
 
         /**
          * Fetches a user API key associated with the current user.
@@ -99,8 +97,8 @@ public:
          *     - id: The id of the API key to fetch.
          *     - completion_block: A callback to be invoked once the call is complete.
          */
-        void fetch_api_key(const realm::ObjectId& id,
-                           std::function<void(std::shared_ptr<UserAPIKey>, Optional<AppError>)> completion_block);
+        void fetch_api_key(const ObjectId& id,
+                           std::function<void(UserAPIKey, Optional<AppError>)> completion_block);
 
         /**
          * Fetches the user API keys associated with the current user.
@@ -108,7 +106,7 @@ public:
          * - parameters:
          *     - completion_block: A callback to be invoked once the call is complete.
          */
-        void fetch_api_keys(std::function<void(std::vector<UserAPIKey>, Optional<AppError>)> completion_block);
+        void fetch_api_keys(std::function<void(UserAPIKey, Optional<AppError>)> completion_block);
 
         /**
          * Deletes a user API key associated with the current user.
@@ -117,7 +115,7 @@ public:
          *     - id: The id of the API key to delete.
          *     - completion_block: A callback to be invoked once the call is complete.
          */
-        void delete_api_key(const realm::ObjectId& id,
+        void delete_api_key(const ObjectId& id,
                             std::function<void(Optional<AppError>)> completion_block);
 
         /**
@@ -127,7 +125,7 @@ public:
          *     - id: The id of the API key to enable.
          *     - completion_block: A callback to be invoked once the call is complete.
          */
-        void enable_api_key(const realm::ObjectId& id,
+        void enable_api_key(const ObjectId& id,
                             std::function<void(Optional<AppError>)> completion_block);
 
         /**
@@ -137,10 +135,10 @@ public:
          *     - id: The id of the API key to disable.
          *     - completion_block: A callback to be invoked once the call is complete.
          */
-        void disable_api_key(const realm::ObjectId& id,
+        void disable_api_key(const ObjectId& id,
                              std::function<void(Optional<AppError>)> completion_block);
     private:
-        UserAPIKeyProviderClient(std::weak_ptr<App>);
+        UserAPIKeyProviderClient(App*);
     };
 
     /**
@@ -229,7 +227,7 @@ public:
                                           const std::string& args,
                                           std::function<void(Optional<AppError>)> completion_block);
     private:
-        UsernamePasswordProviderClient(std::weak_ptr<App>);
+        UsernamePasswordProviderClient(App*);
     };
 
     App(const Config& config);
--- a/src/sync/app.hpp
+++ b/src/sync/app.hpp
@@ -26,11 +26,10 @@
 namespace realm {
 namespace app {
 
-#pragma mark RealmApp
+#pragma mark App
 
 /**
- The `RealmApp` has the fundamental set of methods for communicating with a MongoDB
- Realm application backend.
+ The `App` has the fundamental set of methods for communicating with a MongoDB Realm application backend.
 
  This class provides access to login and authentication.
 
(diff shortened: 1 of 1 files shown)


src/sync/app.hpp
```diff
@@ -58,6 +57,181 @@ class App {
         realm::util::Optional<uint64_t> default_request_timeout_ms;
     };
 
+    /**
+     * A struct representing a user API key as returned by the App server.
+    */
+    struct UserAPIKey {
```
cmelchior: Maybe rename to UserAPIKeyReponse for clarity?

src/sync/app.hpp
```diff
@@ -58,6 +57,181 @@ class App {
...
+        // The ID of the key.
+        realm::ObjectId id;
+
+        // The actual key. Will only be included in
+        // the response when an API key is first created.
+        std::string key;
```
cmelchior: Should this be an Optional then?

src/sync/app.hpp
```diff
@@ -58,6 +57,181 @@ class App {
...
+        // Whether or not the key is disabled.
+        bool disabled;
+    };
+
+    /**
+     * A client for the user API key authentication provider which
+     * can be used to create and modify user API keys. This
+     * client should only be used by an authenticated user.
+    */
+    class UserAPIKeyProviderClient {
```
cmelchior: How are these methods supposed to be exposed in the SDK's? I haven't seen any mention of them before I think?

cmelchior: A lot of these methods also seem to only work if you are an admin user. Is that correct? Perhaps clarify for each method?

src/sync/app.hpp
```diff
@@ -58,6 +57,181 @@ class App {
...
+         * Disables a user API key associated with the current user.
+         *
+         * - parameters:
+         *     - id: The id of the API key to disable.
+         *     - completion_block: A callback to be invoked once the call is complete.
+         */
+        void disable_api_key(const realm::ObjectId& id,
+                             std::function<void(Optional<AppError>)> completion_block);
+    private:
+        UserAPIKeyProviderClient(std::weak_ptr<App>);
```
cmelchior: Why use a weak_ptr?

src/sync/app.hpp
```diff
@@ -58,6 +57,181 @@ class App {
...
+         * Resets the password of an email identity using the
+         * password reset function set up in the application.
+         *
+         * TODO: Add an overloaded version of this method that takes
+         * TODO: raw, non-serialized args.
+         *
+         * - parameters:
+         *     - email: The email address of the user.
+         *     - password: The desired new password.
+         *     - args: A pre-serialized list of arguments. Must be a JSON array.
```
cmelchior: What extra args are needed here?

jsflax: This method specifically calls a remote function that the user has set up on the Stitch server. The args are passed forward to that function.

src/sync/app.hpp
jsflax: Doesn't have to be a weak_ptr in the final implementation, but the constructor will need to derive route/path information from the outer app.

src/sync/app.hpp
jsflax: Where are you seeing that they only work as an admin user? I do not believe that is the case.

> How are these methods supposed to be exposed in the SDK's? I haven't seen any mention of them before I think?

I left these out mistakenly– to some extent, I was originally thinking a lot of these methods would be on the `User` class (similar to how Realm initially had it), but that created too much bloat.

Ideally, these methods would be exposed the same way in the SDKs as they are here. A generic function or field on the `RealmApp` class that exposes the various provider clients.

src/sync/app.hpp
jsflax: Seems a bit redundant? Or are you trying to differentiate from the actual `string` key within the struct?

src/sync/app.hpp
cmelchior: Ok 👍 Is the name of the function fixed?

src/sync/app.hpp
jsflax: Yes 👍

src/sync/app.hpp
jsflax: Yes– the function ran is chosen and setup on the backend.

src/sync/app.hpp
cmelchior: Maybe I just misunderstood exactly what this is. This is extra login options a user already logged in can configure then?

src/sync/app.hpp
cmelchior: I was mostly thinking about what to make of this name if I just saw it in code somewhere. It is a minor anyway.

src/sync/app.hpp
jsflax: https://docs.mongodb.com/stitch/authentication/api-key/

A Stitch Admin can configure this provider from the backend.

If enabled, a non-anonymous user can effectively generate "login" keys and then login "elsewhere" with them. 

```cpp
/// from main relay
app.provider_client<UseAPIKeyProviderClient>().create_api_key(
    "my_api_key",
    [=](std::shared_ptr<UserAPIKey> api_key, Optional<AppError> error) {
        if (error) {
            ...
        }

        iot_device.send_api_key(api_key.key);
    }
);

// from iot_device
void send_api_key(const std::string& key) {
    app.login_with_credential(AppCredentials::UserAPIKey(key), ...); 
}
```
This would be common amongst IoT users, or anyone doing emulating a primary/replica pattern.

src/sync/app.hpp
jsflax: I see your point, since the actual key is nested within the structure.

src/sync/app.hpp
```diff
@@ -58,6 +57,181 @@ class App {
...
+    public:
+        /**
+         * Creates a user API key that can be used to authenticate as the current user.
+         *
+         * - parameters:
+         *     - name: The name of the API key to be created.
+         *     - completion_block: A callback to be invoked once the call is complete.
+        */
+        void create_api_key(const std::string& name,
+                            std::function<void(std::shared_ptr<UserAPIKey>, Optional<AppError>)> completion_block);
```
tgoyne: Why are these passing shared pointers rather than the struct directly?

src/sync/app.hpp
```diff
@@ -58,6 +55,181 @@ class App {
...
+        void fetch_api_key(const ObjectId& id,
+                           std::function<void(UserAPIKey, Optional<AppError>)> completion_block);
+
+        /**
+         * Fetches the user API keys associated with the current user.
+         *
+         * - parameters:
+         *     - completion_block: A callback to be invoked once the call is complete.
+         */
+        void fetch_api_keys(std::function<void(UserAPIKey, Optional<AppError>)> completion_block);
```
ironage: shouldn't the callback parameter be a list/vector of keys?

src/sync/app.hpp
```diff
@@ -58,6 +55,181 @@ class App {
...
+    public:
+        /**
+         * Creates a user API key that can be used to authenticate as the current user.
+         *
+         * - parameters:
+         *     - name: The name of the API key to be created.
+         *     - completion_block: A callback to be invoked once the call is complete.
+        */
+        void create_api_key(const std::string& name,
+                            std::function<void(UserAPIKey, Optional<AppError>)> completion_block);
```
ironage: consider making the callback parameter give out an optional UserAPIKey, otherwise in the error cases, it will give out the default constructed object which could be confusing (same for `fetch_api_key` below)

src/sync/app.hpp
```diff
@@ -58,6 +55,181 @@ class App {
...
+        void fetch_api_keys(std::function<void(UserAPIKey, Optional<AppError>)> completion_block);
+
+        /**
+         * Deletes a user API key associated with the current user.
+         *
+         * - parameters:
+         *     - id: The id of the API key to delete.
+         *     - completion_block: A callback to be invoked once the call is complete.
+         */
+        void delete_api_key(const ObjectId& id,
```
ironage: It might be nice to pass in a populated `UserAPIKey` object for these functions instead of using an `ObjectId` directly; for one, it hides the implementation a bit, and also means that people can't pass in random ObjectIds.

src/sync/app.hpp
jsflax: Good catch 👍

PR #941, opened 2020-03-03
Stitch tests

ironage:
The first commit (80f90e5) refactors the Jenkinsfile to not rely on shell scripts and adds:
- job timeout wrapper (2 hour max)
- jobs clean up after themselves even if they fail
- don't write to /tmp, instead set the temp directory to be a local folder which means it is also cleaned up
- fix the asan job to actually run
- fix the tsan job to run correctly (but keep it disabled until we fix the races)
- fix an error in the tests, discovered by changing where temp files are written
- remove the custom build shell scripts, because I don't think they are used anymore

The second commit (5550ad9) is about configuring the linux sync docker job to also spin up a stitch instance and configure it with the added app configuration. The `app_id` provided in `tests/mongodb/stitch.json` is not being set because the app is created new and a new id is auto-generated. In order to discover the app_id for the tests to connect to, the tests parse the config file at runtime to fetch the updated id.

The third commit (da24acb) is to massage what existing tests we have to work against a local docker image. The limitations being that it doesn't support any email authentication, so I changed the auth settings to automatically confirm users and installed some custom stitch functions which still allow checking server rejections.


src/sync/app.cpp
```diff
@@ -252,7 +252,6 @@ void App::UsernamePasswordProviderClient::reset_password(const std::string& pass
 
 void App::UsernamePasswordProviderClient::call_reset_password_function(const std::string& email,
                                                                        const std::string& password,
-                                                                       const std::string& args,
```
jsflax: This is mandatory for call_reset_password_function to work.

src/sync/app.hpp
```diff
@@ -227,18 +227,13 @@ class App {
...
-         *
          * - parameters:
          *     - email: The email address of the user.
          *     - password: The desired new password.
-         *     - args: A pre-serialized list of arguments. Must be a JSON array.
          *     - completion_block: A callback to be invoked once the call is complete.
         */
         void call_reset_password_function(const std::string& email,
                                           const std::string& password,
-                                          const std::string& args,
```
jsflax: We need these args.

tests/sync/app.cpp
```diff
@@ -262,89 +267,141 @@ TEST_CASE("app: UsernamePasswordProviderClient integration", "[sync][app]") {
...
-    
-    SECTION("call reset password function") {
+
+    SECTION("reset password function success") {
+        // the imported test app will accept password reset if the password contains "realm_tests_do_reset" via a function
+        std::string accepted_new_password = util::format("realm_tests_do_reset%1", random_string(10));
         app.provider_client<App::UsernamePasswordProviderClient>()
             .call_reset_password_function(email,
-                                          password,
-                                          "[0,1]",
```
jsflax: This previously failed because this is invalid extended json.

src/sync/app.cpp
ironage: It doesn't seem to be mandatory, given that the tests are exercising this successfully 😂 . Passing args in the request body results in a "invalid json" error message from the server. I can't seem to find any documentation on this endpoint, where can I find the source of truth?

jsflax: https://docs.mongodb.com/stitch/authentication/userpass/

Just send an array of strings (from the tests) instead of numbers, it will be simpler. The test passes successfully because leaving out the "arguments" field allows it to be considered an empty array.

tests/sync/app.cpp
jsflax: ^ @ironage if you make the args `"[\"foo\", \"bar\"]`, it will work fine.

ironage: That works, thanks!

src/sync/app.cpp
ironage: Fixed, thanks for the docs link!

Jenkinsfile
```diff
@@ -42,30 +36,126 @@ def publishReport(String label) {
...
-      getSourceArchive()
-      def image = buildDockerEnv("ci/realm-object-store:${flavor}")
-      sshagent(['realm-ci-ssh']) {
-        image.inside("-v /etc/passwd:/etc/passwd:ro -v ${env.HOME}:${env.HOME} -v ${env.SSH_AUTH_SOCK}:${env.SSH_AUTH_SOCK} -e HOME=${env.HOME}") {
-          if(withCoverage) {
-            sh "rm -rf coverage.build ${label}.build && ./workflow/test_coverage.sh ${sync} && mv coverage.build ${label}.build"
-          } else {
-            sh "./workflow/build.sh ${flavor} ${sync} ${sanitizerFlags}"
+    nodeWithSources('docker') {
+      def image = docker.build("ci/realm-object-store:${flavor}")
```
fealebenpae:
```suggestion
      def image = buildDockerEnv('ci/realm-object-store:linux', push: env.BRANCH_NAME == 'master')
```
This will use https://github.com/realm/ci/blob/master/vars/buildDockerEnv.groovy. You can add a `push: true` to the arguments to have the image be cached on our AWS docker registry.

Jenkinsfile
```diff
@@ -42,30 +36,126 @@ def publishReport(String label) {
...
+      def buildSteps = { String dockerArgs = "" ->
+        // The only reason we can't run commands directly is because sync builds need to
+        // use CI's credentials to pull from the private repository. This can be removed
+        // when sync is open sourced.
+        sshagent(['realm-ci-ssh']) {
+          image.inside("-v /etc/passwd:/etc/passwd:ro -v ${env.HOME}:${env.HOME} -v ${env.SSH_AUTH_SOCK}:${env.SSH_AUTH_SOCK} -e HOME=${env.HOME} ${dockerArgs}") {
+            if (enableSync) {
+              sh "curl http://mongodb-realm:9090"
```
fealebenpae: Was this forgotten?

Jenkinsfile
```diff
@@ -74,61 +164,64 @@ def doDockerBuild(String flavor, Boolean withCoverage, Boolean enableSync, Strin
...
-              adb connect emulator
-              timeout 10m adb wait-for-device
-              adb push build/tests/tests /data/local/tmp
-              adb shell '/data/local/tmp/tests || echo __ADB_FAIL__' | tee adb.log
-              ! grep __ADB_FAIL__ adb.log
-            """
+      try {
+        rlmCheckout(scm)
+        wrap([$class: 'AnsiColorBuildWrapper']) {
+          def image = docker.build('realm-object-store:ndk21', '-f android.Dockerfile .')
```
fealebenpae:
```suggestion
          def image = buildDockerEnv('realm-object-store:android', extra_args: '-f android.Dockerfile', push: env.BRANCH_NAME == 'master')
```
See above.

Jenkinsfile
```diff
@@ -141,47 +234,50 @@ def setBuildName(newBuildName) {
...
-stage('prepare') {
-  node('docker') {
-    checkout scm
-    sh 'git clean -ffdx -e .????????'
-    sshagent(['realm-ci-ssh']) {
-      sh 'git submodule update --init --recursive'
-    }
+if (env.BRANCH_NAME == 'master') {
+  env.DOCKER_PUSH = "1"
+}
```
fealebenpae:
```suggestion
```
See suggested changes around `buildDockerEnv` above.

README.md
```diff
@@ -132,6 +132,29 @@ docker run --rm -it -v $(pwd):/tmp -w /tmp objectstore bash
...
 
+## Running [app] tests against a local MongoDB Stitch
+
+Stitch images are published to our private Github CI. Follow the steps here to
+set up authorization from docker to your Github account https://github.com/realm/ci/packages/147854
+Once authorized, run the following docker commands from the top directory to start a local instance:
+
+```
+docker build --rm --tag local-stitch -f workflow/stitch.Dockerfile workflow
+docker run -p 9090:9090 -it -v $(pwd)/tests/mongodb:/app_config:rw local-stitch /bin/bash -c "(import_app.sh&) && run.sh"
```
fealebenpae:
```suggestion
docker build --tag local-stitch -f workflow/stitch.Dockerfile workflow
docker run --rm -p 9090:9090 -v $(pwd)/tests/mongodb:/app_config:rw -it local-stitch /bin/bash -c "(import_app.sh&) && run.sh"
```
`--rm` should be applied on `docker run`, and the order of `-it` is significant in that it usually always needs to immediately precede the image name.

workflow/stitch.Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+        wget
+
+RUN wget -O jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64
+RUN chmod u+x jq
+RUN curl -sL https://rpm.nodesource.com/setup_12.x | bash -
+RUN yum install -y nodejs
+RUN npm install mongodb-stitch-cli
+RUN ln -s ./node_modules/.bin/stitch-cli stitch-cli
+
+COPY import_app.sh /app/
```
fealebenpae: I think this is superseded by https://github.com/realm/ci/tree/master/realm/docker/mongodb-realm - the base image already has importing capabilities.

fealebenpae:
@ironage I have a few small comments on how things can be improved, but I think in general this is in a good enough shape to merge and we can iterate it some more on separate PRs.
Best to get some tests running as soon as possible and then focus on making this better.

Jenkinsfile
ironage: I like the idea that this will fail the job early if there are any errors, rather than compiling everything just to find out that the network connection doesn't work.

fealebenpae: Ah, that's a good idea! Maybe just move it to the `if (enableSync)` block underneath, then?

workflow/stitch.Dockerfile
ironage: I tried it and it's very close, but importing custom apps isn't working. I committed a fix to your run script but I'll have to wait for the nightly job to publish a new docker image to test so I'll fix/simplify all this when that's working.

fealebenpae: https://github.com/realm/ci/runs/512551368?check_suite_focus=true just completed some 20 minutes ago. Thanks for the fix!

ironage: After another fix, the docker image is working (tested locally) but Jenkins needs a way to authenticate to Github to pull the image. Can you give the ci Github user a token or is there a better way to do this?

fealebenpae: https://github.com/realm/ci/pull/283 should do the trick

ironage: that works, thanks! 🎉

PR #942, opened 2020-03-04
Add multiple user semantics

jsflax:
- Add multiple user semantics
    - This defines the behaviour of what happens when multiple users are logged in, logged out, logged back in, or removed.
- Add log_out
- Change user metadata to account for persisted state
- Stop marking users for removal and simply remove them
- Add and re-enable App unit tests (integration tests are still disabled)

NOTE: Tests are passing this time 😄. Smoke tests are still disabled, but have been run manually.
TODO: Add `switch_user`. Add `remove_user`.


src/sync/sync_user.hpp
```diff
@@ -117,24 +106,36 @@ struct SyncUserIdentity {
...
 
 // A `SyncUser` represents a single user account. Each user manages the sessions that
 // are associated with it.
 class SyncUser {
 friend class SyncSession;
 public:
-    enum class State {
+    enum class State : std::size_t {
         LoggedOut,
+        LoggedIn,
```
ironage: It looks like there's some duplication of state here; ie Active implies LoggedIn. Since we're already tracking the active user via `m_metadata_manager->set_current_user_identity` can we simplify this by removing the Active state entirely?

src/sync/sync_manager.cpp
```diff
@@ -349,25 +368,86 @@ std::shared_ptr<SyncUser> SyncManager::get_current_user() const
...
+                return;
+
+            auto user = *it;
+            m_users.erase(it);
+            m_users.push_back(user);
+        }
+    }
+
+    // Set the current active user to the next logged in user, or null if none
+    for (auto& user : m_users) {
```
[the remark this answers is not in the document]
jsflax: Yes 👍

src/sync/sync_manager.cpp
```diff
@@ -349,25 +368,86 @@ std::shared_ptr<SyncUser> SyncManager::get_current_user() const
...
+{
+    // Erase and insert this user as the end of the vector
+    {
+        std::lock_guard<std::mutex> lock(m_user_mutex);
+
+        if (!m_metadata_manager) {
+            return;
+        }
+
+        if (m_users.size() > 1) {
```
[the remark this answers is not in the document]
jsflax: This block of code does not log out the user– it changes the user's place in the `users` vector... something I'm realising we should be persisting.

src/sync/sync_manager.hpp
```diff
@@ -30,6 +30,7 @@
 #include <memory>
 #include <mutex>
 #include <unordered_map>
+#include <map>
```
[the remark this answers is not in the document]
jsflax: Yes 👍

src/sync/sync_user.hpp
jsflax: Good call, that would be a nice bit of simplification.

jsflax:
Tests are passing, @tgoyne @ironage PTAL

src/sync/sync_manager.hpp
```diff
@@ -159,18 +159,28 @@ friend class SyncSession;
 
     // Get a sync user for a given identity, or create one if none exists yet, and set its token.
     // If a logged-out user exists, it will marked as logged back in.
-    std::shared_ptr<SyncUser> get_user(const SyncUserIdentifier& identifier,
+    std::shared_ptr<SyncUser> get_user(const std::string& id,
```
cmelchior: Why are we adding the provider_type to all of these methods? From my understanding, each user has a unique identifier which should be enough? The provider should only be used when logging in, right?

src/sync/sync_manager.hpp
```diff
@@ -207,10 +217,10 @@ friend class SyncSession;
     bool run_file_action(const SyncFileActionMetadata&);
 
     // Protects m_users
-    mutable std::mutex m_user_mutex;
+    mutable std::recursive_mutex m_user_mutex;
```
tgoyne: Rather than changing this to a recursive_mutex (which makes it pretty easy to write subtle deadlocks as most of the things it enables are inherently unsafe), log_out_user() could just call m_metadata_manager->set_current_user_identity() directly.

src/sync/sync_manager.cpp
```diff
@@ -295,48 +298,64 @@ bool SyncManager::perform_metadata_update(std::function<void(const SyncMetadataM
...
+
+        if (user->state() == SyncUser::State::LoggedOut) {
+            user->set_state(SyncUser::State::Active);
+        }
         return user;
     }
 }
 
-std::vector<std::shared_ptr<SyncUser>> SyncManager::all_logged_in_users() const
+std::vector<std::shared_ptr<SyncUser>> SyncManager::all_users()
```
tgoyne: This function can be simplified quite a bit now:

```
    std::lock_guard<std::mutex> lock(m_user_mutex);
    m_users.erase(std::remove_if(m_users.begin(), m_users.end(), [](auto& user) {
        return user->state() == SyncUser::State::Error;
    }), m_users.end());
    return m_users;
```

Most of what it was doing was related to that `m_users` used to be a different type from what this function returns, but with that no longer being the case we can just return `m_users` and let the copy constructor do the copying.

src/sync/sync_manager.cpp
```diff
@@ -349,24 +368,76 @@ std::shared_ptr<SyncUser> SyncManager::get_current_user() const
...
+        auto it = std::find_if(m_users.begin(),
+                               m_users.end(),
+                               [user_id, provider_type](const auto& user) {
+            return user->identity() == user_id && user->provider_type() == provider_type;
+        });
+
+        if (it == m_users.end())
+            return;
+
+        auto user = *it;
```
tgoyne: This is `std::rotate(it, it + 1, m_users.end());`

src/sync/sync_manager.hpp
tgoyne: If user_id+provider_type actually is needed to identify users, then shouldn't set_current_user() also need both? These functions seem to handle the case where two different providers have users with the same user_id (and those are actually different users that need to be distinguished), but the current user persistence is only storing the user_id.

jsflax: This was my mistake that I have since clarified with Stitch and update the PR. We need `provider_type` because only anon-user's are removed on logout (and it can be useful for users), but the `user_id` is actually unique across providers, so we do not need to identity users by user_id+provider_type– just user_id.

jsflax:
@tgoyne @ironage PTAL

src/sync/sync_manager.hpp
```diff
@@ -159,18 +159,27 @@ friend class SyncSession;
...
-    std::shared_ptr<SyncUser> get_user(const SyncUserIdentifier& identifier,
+    std::shared_ptr<SyncUser> get_user(const std::string& id,
+                                       const std::string provider_type,
                                        std::string refresh_token,
                                        std::string access_token);
 
     // Get an existing user for a given identifier, if one exists and is logged in.
-    std::shared_ptr<SyncUser> get_existing_logged_in_user(const SyncUserIdentifier&) const;
+    std::shared_ptr<SyncUser> get_existing_logged_in_user(const std::string& user_id,
+                                                          const std::string& provider_type) const;
```
cmelchior: The provider_type wasn't needed, right?

src/sync/app.cpp
```diff
@@ -183,5 +190,30 @@ void App::login_with_credentials(const AppCredentials& credentials,
...
+        return completion_block(util::none);
+    };
+
+    m_config.transport_generator()->send_request_to_server({
+        HttpMethod::post,
+        route,
+        m_request_timeout_ms,
+        {
+            { "Content-Type", "application/json;charset=utf-8" },
+            { "Accept", "application/json" }
```
ironage: Don't you also need to provide auth tokens to log out? Otherwise that would seem insecure?

src/sync/app.cpp
```diff
@@ -183,5 +190,30 @@ void App::login_with_credentials(const AppCredentials& credentials,
...
 
+void App::log_out(std::function<void (Optional<AppError>)> completion_block) const {
+    std::string route = util::format("%1/session", m_auth_route);
+
+    auto handler = [completion_block, this](const Response& response) {
+        if (auto error = check_for_errors(response)) {
+            return completion_block(error);
+        }
+
+        this->current_user()->log_out();
```
ironage: This may crash if someone calls `log_out` twice concurrently. Best check if the result of `current_user` is valid first.

src/sync/app.cpp
jsflax: Good catch. I'm surprised the smoke test worked.

src/sync/app.cpp
```diff
@@ -575,5 +580,35 @@ void App::login_with_credentials(const AppCredentials& credentials,
     }, handler);
 }
 
+void App::log_out(std::function<void (Optional<AppError>)> completion_block) const {
+    if (!current_user())
```
ironage: There's still a race condition because `current_user()` can return different things (or nullptr) at each call in this method. I'd suggest storing the user it to a local variable and doing all checks against that one instance.

jsflax: In a later PR, I will likely just create a mutex in app that prevents parallelisation of these methods. But I will fix this for now.

src/sync/app.cpp
ironage: We can beef up the integration tests once I get CI to cooperate with the stitch docker image.

jsflax:
@ironage What is preventing this from being merged?

PR #944, opened 2020-03-09
Return custom error message for custom error codes

cmelchior:
Make it possible to also roundtrip a custom error message instead of a hardcoded one.

I couldn't make the unit tests run, so the test is not verified to be correct 😕, but the functionality has been tested through the Java SDK


tests/sync/app.cpp
```diff
@@ -146,6 +164,31 @@ static std::string random_string(std::string::size_type length)
...
+        // TODO: create dummy app using Stitch CLI instead of hardcording
+        auto app = App(App::Config{"translate-utwuv", factory});
+
+        bool processed = false;
+
+        app.login_with_credentials(AppCredentials::anonymous(),
+                                   [&](std::shared_ptr<SyncUser> user, Optional<app::AppError> error) {
+            CHECK(!user);
+            CHECK(error);
+            CHECK(error.value().is_custom_error());
```
tgoyne: `->` is the normal way to write `.value().`

tests/sync/app.cpp
```diff
@@ -146,6 +164,31 @@ static std::string random_string(std::string::size_type length)
   return s;
 }
 
+TEST_CASE("app: custom error handling", "[sync][app][custom_errors]") {
+    SECTION("custom code and message is sent back") {
+        std::unique_ptr<GenericNetworkTransport> (*factory)() = []{
+            return std::unique_ptr<GenericNetworkTransport>(new CustomErrorTransport(1001, "Boom!"));
+        };
+
+        // TODO: create dummy app using Stitch CLI instead of hardcording
```
jsflax: This TODO isn't really necessary since this is a unit test.

tests/sync/app.cpp
```diff
@@ -132,6 +132,24 @@ class IntTestTransport : public GenericNetworkTransport {
     }
 };
 
+class CustomErrorTransport : public GenericNetworkTransport {
+public:
+    CustomErrorTransport(int code, std::string message)
```
jsflax: [nit] `CustomErrorTransport(int code, const std::string& message) : m_code(code), m_message(message) {}`

We can also localise this transport to the test unless you think we will be re-using it.

PR #945, opened 2020-03-11
Add remote mongo stubs

leemaguire:
Adding stubs that allows for interaction between Stitch and Realm


tests/sync/remote_mongo_tests.cpp
```diff
@@ -0,0 +1,34 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#include "catch2/catch.hpp"
+#include "sync/remote_mongo_client.hpp"
+#include "sync/app.hpp"
+
+#ifndef ENABLE_MONGO_CLIENT_TESTS
```
ironage: by convention we use a `REALM_` prefix for project configurations

leemaguire: Noted and addressed

leemaguire:
@ironage I have shared a doc with you privately regarding context. The plan to move forward is to merge the stubs into v10 and then begin the implementation after.

src/sync/app.hpp
```diff
@@ -246,31 +192,32 @@ class App {
...
     T provider_client() {
         return T(this);
     }
-
+    
+    /// Retrieves a general-purpose service client for the Stitch service
+    /// associated with the specified name. Use this for services which do not
+    /// have a well-defined interface in the SDK.
+    /// @param service_name  The name of the desired service in MongoDB Stitch.
+    RemoteMongoClient service_client(std::string service_name);
```
jsflax: [minor] Should be `RemoteMongoClient remote_mongo_client()`

src/sync/remote_mongo_client.hpp
```diff
@@ -0,0 +1,53 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or utilied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#ifndef CORE_REMOTE_MONGO_CLIENT_HPP
```
jsflax: [minor] Change header guard to reflect file name.

src/sync/remote_mongo_client.hpp
```diff
@@ -0,0 +1,53 @@
...
+namespace realm {
+namespace app {
+
+class RemoteMongoDatabase;
+
+/// A client responsible for communication with the Stitch API
+class RemoteMongoClient {
+public:
+
+    RemoteMongoClient(StitchServiceClient service) : m_service(service) { }
```
jsflax: [minor] Remove all references to `Stitch`. Should be `RealmServiceClient`.

src/sync/remote_mongo_collection.hpp
```diff
@@ -0,0 +1,184 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or utilied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#ifndef REALM_CORE_REMOTE_MONGO_COLLECTION_HPP
```
jsflax: Fix all the header guards.

src/sync/remote_mongo_client.hpp
```diff
@@ -0,0 +1,53 @@
...
+    
+    /// Gets a `CoreRemoteMongoDatabase` instance for the given database name.
+    /// @param name the name of the database to retrieve
+    RemoteMongoDatabase operator[](std::string name);
```
jsflax: `const std::string&` for all of these

src/sync/remote_mongo_collection.hpp
```diff
@@ -0,0 +1,184 @@
...
+    /// @param filter  A `Document` that should match the query.
+    /// @param options  Optional `RemoteFindOneAndModifyOptions` to use when executing the command.
+    /// @Returns The resulting `Document` or nil if no such document exists
+    util::Optional<CollectionType> find_one_and_delete(Document filter,
+                                                       util::Optional<RemoteFindOneAndModifyOptions> options);
+
+private:
+    
+    /// Returns a document of database name and collection name
+    Document m_base_operation_args {
```
jsflax: What is the point of this?

src/sync/remote_mongo_collection.hpp
```diff
@@ -0,0 +1,184 @@
...
+namespace realm {
+namespace app {
+
+template<class T>
+class RemoteMongoCollection {
+    
+public:
+    
+    using CollectionType = T;
+    using Document = nlohmann::json;
```
jsflax: [minor] Instead of this, make it a `std::string`, and change the parameter names to be suffixed with `_json`

src/sync/remote_mongo_database.hpp
```diff
@@ -0,0 +1,69 @@
...
+class RemoteMongoDatabase {
+    
+public:
+
+    using Document = nlohmann::json;
+
+    /// The name of this database
+    const std::string name;
+    
+    RemoteMongoDatabase(std::string name, StitchServiceClient service, RemoteMongoClient client) :
```
jsflax: [minor] Service client and RemoteMongoClient should be passed by reference.

src/sync/stitch_service_client.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+namespace realm {
+namespace app {
+
+class AuthRequestClient {
+    /// Performs an authenticated request to the Stitch server, using the current authentication state, and should
+    /// throw when not currently authenticated.
+    /// @param auth_request The request to perform
+    /// @returns Response of the request
+    Response do_authenticated_request(const Request auth_request);
```
jsflax: [minor] This should be virtual.

src/sync/stitch_service_client.hpp
```diff
@@ -0,0 +1,99 @@
...
+};
+
+/// A class representing the authentication API routes on the Stitch server.
+class ServiceRoutes {
```
jsflax: [major] Remove all the routing functionality. It is arguably an unnecessary abstraction.

src/sync/stitch_service_client.hpp
```diff
@@ -0,0 +1,99 @@
...
+    std::string app_metadata_route();
+    /// Returns the route on the server for executing a function.
+    std::string function_call_route();
+private:
+    std::string m_client_app_id;
+};
+
+/// A class providing the core functionality necessary to make authenticated function call requests for a particular
+/// Stitch service.
+class StitchServiceClient {
```
jsflax: [major] Other than modifying the name, make this more idiomatic C++. It does not need to have the same restrictions the Swift one had.

leemaguire: Class renamed and `call_function` has been refactored into an async method

src/sync/stitch_service_client.hpp
leemaguire: Class removed

src/sync/stitch_service_client.hpp
leemaguire: Addressed

src/sync/remote_mongo_database.hpp
leemaguire: Addressed

src/sync/remote_mongo_collection.hpp
leemaguire: Addressed

src/sync/remote_mongo_client.hpp
leemaguire: Addressed

src/sync/remote_mongo_collection.hpp
leemaguire: Addressed

src/sync/remote_mongo_client.hpp
leemaguire: Done

src/sync/remote_mongo_client.hpp
leemaguire: Addressed

src/sync/app.hpp
leemaguire: Addressed

leemaguire:
@jsflax I've addressed you comments PTAL

src/sync/app_service_client.hpp
```diff
@@ -0,0 +1,58 @@
...
+class AppServiceClient {
+public:
+    /// The name of the service
+    const util::Optional<std::string> service_name;
+    
+    /// Calls the MongoDB Stitch function with the provided name and arguments.
+    /// @param name The name of the Stitch function to be called.
+    /// @param args The `BSONArray` of arguments to be provided to the function.
+    /// @param completion_block Returns the result from the intended call, will return an Optional AppError is an error is thrown and a json string if successful
+    void call_function(const std::string& name,
```
jsflax: This should be virtual, and should take the service_name as a parameter.

src/sync/app_service_client.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+namespace realm {
+namespace app {
+
+/// A class providing the core functionality necessary to make authenticated function call requests for a particular
+/// Stitch service.
+class AppServiceClient {
+public:
+    /// The name of the service
+    const util::Optional<std::string> service_name;
```
jsflax: Have this as a parameter of call_function.

src/sync/app_service_client.hpp
leemaguire: Done

src/sync/app_service_client.hpp
leemaguire: Done

src/sync/app_service_client.hpp
```diff
@@ -0,0 +1,50 @@
...
+public:
+    
+    /// Calls the MongoDB Stitch function with the provided name and arguments.
+    /// @param name The name of the Stitch function to be called.
+    /// @param args The `BSONArray` of arguments to be provided to the function.
+    /// @param service_name The name of the service.
+    /// @param completion_block Returns the result from the intended call, will return an Optional AppError is an error is thrown and a json string if successful
+    virtual void call_function(const std::string& name,
+                               const std::string& args_json,
+                               const std::string& service_name,
```
jsflax: service_name should be optional

leemaguire: done

leemaguire:
@tgoyne PTAL

src/sync/app_service_client.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include <string>
+#include <realm/util/optional.hpp>
+#include <sync/generic_network_transport.hpp>
+
+namespace realm {
+namespace app {
+
+/// A class providing the core functionality necessary to make authenticated function call requests for a particular
+/// Stitch service.
+class AppServiceClient {
```
tgoyne: All types with a virtual member function need to also have a virtual destructor, as otherwise deleting objects via a pointer to the base class won't call the destructor for the subclass.

src/sync/app.hpp
```diff
@@ -22,28 +22,25 @@
...
- `RemoteMongoClient`,
- `RLMPush`,
- [Functions](https://docs.mongodb.com/stitch/functions/)
- */
+/// The `App` has the fundamental set of methods for communicating with a MongoDB Realm application backend.
+///
+/// This class provides access to login and authentication.
+///
+/// Using `serviceClient`, you can retrieve services, including the `RemoteMongoClient` for reading
+/// and writing on the database. To create a `RemoteMongoClient`, pass `remoteMongoClientFactory`
```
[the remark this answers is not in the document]
leemaguire: I've updated the docs

src/sync/app.hpp
```diff
@@ -22,28 +22,25 @@
 #include "app_credentials.hpp"
 #include "generic_network_transport.hpp"
 #include "sync_user.hpp"
+#include "remote_mongo_client.hpp"
```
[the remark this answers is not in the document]
leemaguire: done

src/sync/remote_mongo_client.hpp
```diff
@@ -0,0 +1,53 @@
...
+//
+////////////////////////////////////////////////////////////////////////////
+
+#ifndef REMOTE_MONGO_CLIENT_HPP
+#define REMOTE_MONGO_CLIENT_HPP
+
+#include <string>
+#include <map>
+#include <realm/util/optional.hpp>
+#include "app_service_client.hpp"
```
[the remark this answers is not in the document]
leemaguire: done

src/sync/app_service_client.hpp
leemaguire: done

src/sync/remote_mongo_client.hpp
```diff
@@ -0,0 +1,53 @@
...
+    /// Gets a `CoreRemoteMongoDatabase` instance for the given database name.
+    /// @param name the name of the database to retrieve
+    RemoteMongoDatabase operator[](const std::string& name);
+  
+    /// Gets a `CoreRemoteMongoDatabase` instance for the given database name.
+    /// @param name the name of the database to retrieve
+    RemoteMongoDatabase db(const std::string& name);
+    
+private:
+    AppServiceClient m_service;
```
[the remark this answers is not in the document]
leemaguire: done

src/sync/remote_mongo_collection.hpp
```diff
@@ -0,0 +1,183 @@
...
+#include <vector>
+#include <realm/util/optional.hpp>
+#include <json.hpp>
+#include "remote_mongo_read_operation.hpp"
+#include "remote_mongo_result_types.hpp"
+
+namespace realm {
+namespace app {
+
+template<class T>
```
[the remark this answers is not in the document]
leemaguire: T was meant to represent the `Document` class to cast to. After a review it does not make sense do this at the object store level. So instead a json string representing the document will be returned instead.

leemaguire:
@tgoyne @jsflax @ironage  PTAL

src/sync/remote_mongo_result_types.hpp
```diff
@@ -0,0 +1,101 @@
...
+/// The result of an `insert_many` command on a `RemoteMongoCollection`.
+struct RemoteInsertManyResult {
+    /// Map of the index of the inserted document to the id of the inserted document.
+    std::map<long int, BSONValue> inserted_ids;
+};
+
+/// The result of a `delete` command on a `RemoteMongoCollection`.
+struct RemoteDeleteResult {
+    /// The number of documents that were deleted.
+    int deleted_count;
```
ironage: for numeric types, we usually prefer `size_t` as the fastest native unsigned type (with no capacity guarantee). I don't think we ever want a negative value here so best to use an unsigned type as well. Same for several instances of `int` below.

src/sync/remote_mongo_result_types.hpp
```diff
@@ -0,0 +1,101 @@
...
+/// Options to use when executing a `find_one_and_update`, `find_one_and_replace`,
+/// or `find_one_and_delete` command on a `remote_mongo_collection`.
+struct RemoteFindOneAndModifyOptions {
+    /// Limits the fields to return for all matching documents.
+    util::Optional<std::string> projection_json;
+    /// The order in which to return matching documents.
+    util::Optional<std::string> sort_json;
+    /// Whether or not to perform an upsert, default is false
+    /// (only available for find_one_and_replace and find_one_and_update)
+    util::Optional<bool> upsert;
```
ironage: An optional bool gives three states; null, true, false. But I'm not sure we need that, if all you want to do is have a default in place, you can use `bool upsert = false;` Is that sufficient for how this will be used?

src/sync/remote_mongo_collection.hpp
```diff
@@ -0,0 +1,166 @@
...
+    /// The name of the database containing this collection.
+    const std::string database_name;
+
+    RemoteMongoCollection(std::string name, std::string database_name) : name(name), database_name(database_name) { }
+    
+    /// Finds the documents in this collection which match the provided filter.
+    /// @param filter_json A `Document` as a json string that should match the query.
+    /// @param options Optional `RemoteFindOptions` to use when executing the command.
+    /// @returns A json string of the resulting documents.
+    std::string find(const std::string& filter_json, util::Optional<RemoteFindOptions> options);
```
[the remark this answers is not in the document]
leemaguire: I've changed these methods to be `void` reason being that they will perform network calls and therefore a callback is better suited. The callback will have a consistent non optional `std::string`

src/sync/remote_mongo_result_types.hpp
```diff
@@ -0,0 +1,101 @@
...
+
+// TODO: Place using statement here for now until types exists
+using BSONValue = std::string;
+
+// MARK: - Result types
+
+/// Options to use when executing a `find` command on a `RemoteMongoCollection`.
+struct RemoteFindOptions {
+    /// The maximum number of documents to return.
+    util::Optional<long int> limit;
```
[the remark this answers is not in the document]
leemaguire: Changed to `uint64_t` where applicable

src/sync/remote_mongo_result_types.hpp
leemaguire: I've removed this struct so we don't need to specify false. It has been moved to be a param on the method itself instead

src/sync/remote_mongo_collection.hpp
```diff
@@ -0,0 +1,290 @@
...
+    void aggregate(std::vector<std::string> pipline,
+                   std::function<void(std::string, util::Optional<AppError>)> completion_block);
+
+    /// Counts the number of documents in this collection matching the provided filter.
+    /// @param filter_json A `Document` as a json string that should match the query.
+    /// @param options Optional `RemoteCountOptions` to use when executing the command.
+    /// @param completion_block Returns the count of the documents that matched the filter.
+    void count(const std::string& filter_json,
+               util::Optional<RemoteFindOptions> options,
+               std::function<void(u_int64_t, util::Optional<AppError>)> completion_block);
```
ironage, line 126: please use `uint64_t` everywhere as that is part of the standard type system (also for consistency), the `u_` versions are non-standard system defined types

leemaguire: Updated to use uint64_t

PR #948, opened 2020-03-13
Add ability to refresh access token

leemaguire:
…ponse


src/sync/app.cpp
```diff
@@ -611,5 +627,107 @@ void App::log_out(std::function<void (Optional<AppError>)> completion_block) con
...
+        }
+    };
+    
+    // Only handle auth failures
+    if (error.error_code.value() != 401) {
+        completion_block(response);
+        return;
+    }
+        
+    App::refresh_access_token_if_needed(request, access_token_handler);
```
cmelchior: Some things are missing from this logic.

In the past, we automatically started a refresh of the token 10-30 seconds before it expired, in that case, users should hopefully not experience any "hiccups" in the sync connection.

Also the `refresh_access_token_if_needed` doesn't seem to retry itself if it fails. Before all bindings had an exponential backoff algorithm for this, which continuously retried getting the token automatically.

Not sure if it is useful, but the current Java logic is here: https://github.com/realm/realm-java/blob/master/realm/realm-library/src/objectServer/java/io/realm/internal/network/ExponentialBackoffTask.java#L27 and getting the access token logic is here: https://github.com/realm/realm-java/blob/master/realm/realm-library/src/objectServer/java/io/realm/SyncSession.java#L776

jsflax: Agreed that `refresh_access_token_if_needed` should retry itself if it fails.

However, we're trying to explicitly avoid doing any kind of threading in the App Layer. So if we want to do pre-emptive refreshes, it may have to be something the SDKs hook into.

Re: hiccups in the sync connection, I believe @nirinchev said we don't have to refresh the connection anymore? Or did I misunderstand that conversation?

cmelchior: I don't have any opinion on exactly how it is done. If we want to avoid doing threading in OS, then yes, the SDK's need to be involved somehow.

I strongly assume that access tokens still expire? With refresh, I meant getting a new one before the old one expired.

jsflax: Resolved offline. Refreshing a sync session pre-emptively is no longer needed (or at all): "we don’t need to refresh the session. if the client sends the command, the server will accept it and do nothing with it. so it’s safe to remove it but also doesn’t break anything by having it there in the meantime"

leemaguire:
@jsflax @cmelchior PTAL

src/sync/app.hpp
```diff
@@ -280,12 +260,23 @@ class App {
     std::string m_app_route;
     std::string m_auth_route;
     uint64_t m_request_timeout_ms;
+
+    void refresh_access_token(std::function<void(Optional<AppError>)> completion_block) const;
+
+    void handle_auth_failure(const AppError& error,
```
cmelchior, line 266: Some doc for this method would be nice. It seems to do a lot of things that are not immediate obvious from the function name?

src/sync/app.cpp
```diff
@@ -734,6 +730,112 @@ void App::link_user(std::shared_ptr<SyncUser> user,
...
+    {
+        auto handler = [completion_block, request, this](const Response& response) {
+            if (auto error = check_for_errors(response)) {
+                App::handle_auth_failure(error.value(), response, request, completion_block);
+            } else {
+                completion_block(response);
+            }
+        };
+
+        request.headers = get_request_headers(current_user(),
```
cmelchior, line 750: It looks a bit dangerous to use the current user here. I would much prefer it if it was parsed in as a parameter.

Generally, I think there is a very high chance we introduce subtle race conditions by using `current_user()` inside functions, especially when it involves callbacks to bindings which might delegate work to other threads.

IMO, we should parse in the user as a parameter where it is needed. It will not only make it easier to reason about but will also enable more flexible API's in the SDK's.

src/sync/app.hpp
```diff
@@ -280,12 +260,23 @@ class App {
     std::string m_app_route;
     std::string m_auth_route;
     uint64_t m_request_timeout_ms;
+
+    void refresh_access_token(std::function<void(Optional<AppError>)> completion_block) const;
```
cmelchior, line 264: Generally, it would be nice with a high-level doc around how tokens and refreshes work. Perhaps at the top of this file or at `do_authenticated_request`

From my understanding, all REST calls must use `do_authenticated_request` which is responsible for detecting if a token refresh is needed and do that. 

In the past, we attempted to refresh the access token before it expired, but that was mostly due to sync. I guess now that is handled elsewhere, then just reacting to the server response code for REST calls is much simpler and probably doesn't matter much in terms of user experience.

tests/sync/app.cpp
```diff
@@ -733,6 +738,75 @@ TEST_CASE("app: UserAPIKeyProviderClient integration", "[sync][app]") {
...
+
+        app.log_in_with_credentials(AppCredentials::username_password(email, password),
+                               [&](std::shared_ptr<SyncUser> user, Optional<app::AppError> error) {
+            CHECK(user);
+            CHECK(!error);
+        });
+        
+        auto previous_access_token = SyncManager::shared().get_current_user()->access_token();
+        
+        // If we do this call within the same second, the response for the access token will be cached
```
cmelchior, line 772: Who controls that?

tests/sync/app.cpp
```diff
@@ -1812,3 +1910,210 @@ TEST_CASE("app: auth providers", "[sync][app]") {
...
+            }
+            
+            realm::SyncManager::shared().get_user("a_user_id",
+                                                  good_access_token,
+                                                  good_access_token,
+                                                  "anon-user");
+        };
+        
+        /*
+         Expected flow:
```
cmelchior, line 2033: Nice description 👍

src/sync/app.cpp
```diff
@@ -561,19 +563,13 @@ void App::get_profile(std::function<void(std::shared_ptr<SyncUser>, Optional<App
...
-        m_request_timeout_ms,
-        {
-            { "Content-Type", "application/json;charset=utf-8" },
-            { "Accept", "application/json" },
-            { "Authorization", bearer }
-        },
-        std::string(),
-        }, profile_handler);
+    
+    do_authenticated_request({
```
cmelchior, line 567: Shouldn't all REST methods use this method? I assume we want to refresh the token no matter which API called failed because of it?

tests/sync/app.cpp
leemaguire: Its a server side issue, because we are testing the call within the same second. The way around it is to sleep at least a second

src/sync/app.hpp
leemaguire: I've added docs, I don't think all calls need `do_authenticated_request` for example `log_in` and `register` don't require this.

src/sync/app.cpp
leemaguire: Agreed, added the ability to now pass user as a param

src/sync/app.hpp
leemaguire: docs added

src/sync/app.cpp
```diff
@@ -734,6 +732,113 @@ void App::link_user(std::shared_ptr<SyncUser> user,
...
+        
+        // Only handle auth failures
+        if (error.error_code.value() != 401) {
+            completion_block(response);
+            return;
+        }
+        
+        if (request.uses_refresh_token) {
+            if (!sync_user) {
+                sync_user->log_out();
```
ironage, line 790: What is this supposed to do? Dereferencing a null pointer will crash.

src/sync/app.cpp
```diff
@@ -734,6 +732,113 @@ void App::link_user(std::shared_ptr<SyncUser> user,
...
+        
+        App::refresh_access_token(sync_user, access_token_handler);
+    }
+    
+    /// MARK: - refresh access token
+    void App::refresh_access_token(std::shared_ptr<SyncUser> sync_user,
+                                   std::function<void(Optional<AppError>)> completion_block) const
+    {
+        if (!sync_user) {
+            completion_block(AppError(make_custom_error_code(ServiceErrorCode::invalid_session),
```
ironage, line 804: can you use a `ClientErrorCode` here? Either reuse `user_not_found` or create another value that is more appropriate, but don't use the custom error category.

leemaguire: Now using `ClientErrorCode`

src/sync/app.cpp
leemaguire: impl is wrong, changed to `if (sync_user && sync_user->is_logged_in())`

src/sync/app.cpp
```diff
@@ -734,6 +732,113 @@ void App::link_user(std::shared_ptr<SyncUser> user,
...
+                                      "No current user exists"));
+            return;
+        }
+        
+        if (!sync_user->is_logged_in()) {
+            completion_block(AppError(make_custom_error_code(ServiceErrorCode::invalid_session),
```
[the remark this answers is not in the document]
leemaguire: Changed to `(error.is_http_error() && error.error_code.value() != 401)`

src/sync/app.cpp
```diff
@@ -734,6 +732,113 @@ void App::link_user(std::shared_ptr<SyncUser> user,
...
+                request.headers = get_request_headers(sync_user, RequestTokenType::AccessToken);
+                transport_generator->send_request_to_server(request, completion_block);
+            } else {
+                // pass the error back up the chain
+                completion_block(response);
+            }
+        };
+        
+        // Only handle auth failures
+        if (error.error_code.value() != 401) {
```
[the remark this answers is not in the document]
leemaguire: Changed to (error.is_http_error() && error.error_code.value() != 401)

src/sync/app.cpp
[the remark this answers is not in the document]
leemaguire: True, I've changed to ClientErrorCode where applicable

src/sync/app.hpp
```diff
@@ -81,62 +82,38 @@ class App {
...
     class UserAPIKeyProviderClient {
     public:
-        /**
-         * Creates a user API key that can be used to authenticate as the current user.
-         *
-         * - parameters:
-         *     - name: The name of the API key to be created.
-         *     - completion_block: A callback to be invoked once the call is complete.
-        */
+        /// Creates a user API key that can be used to authenticate as the current user.
```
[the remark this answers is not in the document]
leemaguire: 🍻

src/sync/generic_network_transport.hpp
```diff
@@ -109,6 +105,11 @@ enum class ServiceErrorCode {
     none = 0
 };
 
+const std::error_category& custom_error_category() noexcept;
+std::error_code make_custom_error_code(int code) noexcept;
+std::error_code make_custom_error_code(ClientErrorCode) noexcept;
```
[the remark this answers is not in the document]
leemaguire: True, I've removed `make_custom_error_code(ClientErrorCode)` and given it `make_client_error_code` instead

tests/sync/app.cpp
```diff
@@ -733,6 +738,75 @@ TEST_CASE("app: UserAPIKeyProviderClient integration", "[sync][app]") {
...
+            return std::unique_ptr<GenericNetworkTransport>(new IntTestTransport);
+        };
+
+        auto app = App(App::Config{"translate-utwuv", factory});
+
+        static const std::string base_path = realm::tmp_dir();
+        auto tsm = TestSyncManager(base_path);
+        bool processed = false;
+
+        app.refresh_custom_data(SyncManager::shared().get_current_user(), [&](const Optional<AppError>& error) {
```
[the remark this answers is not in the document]
leemaguire: done

tests/sync/app.cpp
```diff
@@ -733,6 +738,75 @@ TEST_CASE("app: UserAPIKeyProviderClient integration", "[sync][app]") {
...
+    SECTION("refresh custom data integration expect success") {
+        
+        auto email = util::format("%1@%2.com", random_string(15), random_string(15));
+        auto password = util::format("%1", random_string(15));
+
+        std::unique_ptr<GenericNetworkTransport> (*factory)() = []{
+            return std::unique_ptr<GenericNetworkTransport>(new IntTestTransport);
+        };
+    
+        auto app = App(App::Config{"translate-utwuv", factory});
```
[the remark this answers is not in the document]
leemaguire: I've changed to hit the local stitch

tests/sync/app.cpp
```diff
@@ -145,7 +155,15 @@ class IntTestTransport : public GenericNetworkTransport {
             response_code = curl_easy_perform(curl);
             int http_code = 0;
             curl_easy_getinfo (curl, CURLINFO_RESPONSE_CODE, &http_code);
-
+            
+            for (auto &header_element : response_headers) {
```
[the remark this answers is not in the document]
leemaguire: Removed for now

src/sync/app.cpp
leemaguire: calls which require authentication now use `do_authenticated_request`

src/sync/generic_network_transport.cpp
```diff
@@ -227,13 +241,16 @@ std::error_code make_custom_error_code(int code) noexcept
 {
     return std::error_code{code, g_custom_error_category};
 }
-    
-std::error_code make_custom_error_code(ClientErrorCode error) noexcept
+
+const std::error_category& client_error_category() noexcept
 {
-    return std::error_code{int(error), g_custom_error_category};
+    return g_http_error_category;
```
ironage, line 247: g_client_error_category

src/sync/generic_network_transport.cpp
```diff
@@ -227,13 +241,16 @@ std::error_code make_custom_error_code(int code) noexcept
...
 }
 
-
+std::error_code make_client_error_code(ClientErrorCode error) noexcept
+{
+    return std::error_code{int(error), g_custom_error_category};
```
ironage, line 252: g_client_error_category

src/sync/generic_network_transport.hpp
```diff
@@ -109,13 +105,19 @@ enum class ServiceErrorCode {
...
 ServiceErrorCode service_error_code_from_string(const std::string& code);
 const std::error_category& service_error_category() noexcept;
 std::error_code make_error_code(ServiceErrorCode) noexcept;
 
 const std::error_category& http_error_category() noexcept;
 std::error_code make_http_error_code(int http_code) noexcept;
 
+const std::error_category& client_error_category() noexcept;
+std::error_code make_client_error_code(ClientErrorCode) noexcept;
+
```
ironage, line 120: Thanks for adding these, can we get a `AppError::is_client_error()` as well (nice to have for SDK's)

src/sync/generic_network_transport.cpp
leemaguire: done 👍🏼

src/sync/generic_network_transport.hpp
leemaguire: done 👍🏼

PR #949, opened 2020-03-16
Fix log out and integration test

leemaguire:


src/sync/app.cpp
```diff
@@ -580,32 +592,37 @@ void App::log_in_with_credentials(const AppCredentials& credentials,
...
-    if (!user)
+    if (!user) {
         return completion_block(util::none);
+    }
 
-    std::string route = util::format("%1/session", m_auth_route);
+    std::string route = util::format("%1/auth/session", m_base_route);
 
     auto handler = [completion_block, user](const Response& response) {
+        if (user) {
```
tgoyne: If we're going to log out the local user even if there's an error, then shouldn't it just happen immediately rather than waiting for the response?

When would user be falsy here?

src/sync/app.cpp
```diff
@@ -73,10 +73,23 @@ App::App(const Config& config)
 
 static Optional<AppError> check_for_errors(const Response& response)
 {
+    auto http_status_code_is_fatal = [response] {
+        if (response.http_status_code >= 300
```
tgoyne: This could just be `bool http_status_code_is_fatal = response.http_status_code >= 300 || || (response.http_status_code < 200 && response.http_status_code != 0)`; there's no real reason for it to be a function.

src/sync/app.cpp
```diff
@@ -73,10 +73,23 @@ App::App(const Config& config)
...
+            || (response.http_status_code < 200 && response.http_status_code != 0)) {
+            return true;
+        }
+        return false;
+    };
+    
     try {
-        if (auto ct = response.headers.find("Content-Type"); ct != response.headers.end() && ct->second == "application/json") {
+        if (auto ct = response.headers.find("Content-Type");
```
tgoyne: Having the wrapped part of the if indented to the same level as the following block makes it hard to tell where the body of the if block starts. When the condition for an `if` statement is too big to reasonable fit on a line it's usually better to extract it to a local variable or helper function.

src/sync/app.cpp
leemaguire: You're right, I've moved `log_out` to be called before we send the request to the server

src/sync/app.cpp
leemaguire: True, I've cleaned this up

src/sync/app.cpp
leemaguire: I agree, I've cleaned it up to be more readable

src/sync/app.cpp
```diff
@@ -73,10 +73,16 @@ App::App(const Config& config)
 
 static Optional<AppError> check_for_errors(const Response& response)
 {
+    bool http_status_code_is_fatal = response.http_status_code >= 300 ||
+        (response.http_status_code < 200 && response.http_status_code != 0);
+    
     try {
-        if (auto ct = response.headers.find("Content-Type"); ct != response.headers.end() && ct->second == "application/json") {
+        auto ct = response.headers.find("Content-Type");
+        if (ct != response.headers.end() && ct->second == "application/json" && http_status_code_is_fatal) {
```
ironage: This change is causing the other integration tests to fail. It is possible for stitch to return an error code with http 200, for example when registering an email account twice the request is processed but denied. I will change it back in https://github.com/realm/realm-object-store/pull/941 but if there's something I'm missing about this let me know

PR #951, opened 2020-03-17
Add support for logging specific user out

cmelchior:
This PR adds support for logging individual users out instead of only the current user.

Note: Does not run the tests until Authentication tests are enabled.

### DIFF
--- a/src/sync/app.cpp
+++ b/src/sync/app.cpp
@@ -586,10 +586,9 @@ void App::log_in_with_credentials(const AppCredentials& credentials,
     }, handler);
 }
 
-void App::log_out(std::function<void (Optional<AppError>)> completion_block) const
+void App::log_out(std::shared_ptr<SyncUser> user, std::function<void (Optional<AppError>)> completion_block) const
 {
-    auto user = current_user();
-    if (!user) {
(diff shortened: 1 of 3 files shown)


src/sync/app.cpp
```diff
@@ -586,10 +586,9 @@ void App::log_in_with_credentials(const AppCredentials& credentials,
     }, handler);
 }
 
-void App::log_out(std::function<void (Optional<AppError>)> completion_block) const
+void App::log_out(std::shared_ptr<SyncUser> user, std::function<void (Optional<AppError>)> completion_block) const
```
jsflax: Should this be user_id instead? For easy compatibility.

cmelchior: I don't have a strong opinion, either works for me. I'll change it.

cmelchior: Or actually, no. 

There are some checks on the user inside `log_out`, so parsing the identity would require looking up the user again? That seems cumbersome and I don't think there are any cases where you don't have a reference to the shared pointer for the user for this?

jsflax: Alright, that's fine.

PR #953, opened 2020-03-18
Add ability to switch and remove user

leemaguire:
App now contains `switch_user(user_id)` & `remove_user(user_id)`

### DIFF
--- a/src/sync/app.cpp
+++ b/src/sync/app.cpp
@@ -661,13 +661,8 @@ void App::remove_user(std::shared_ptr<SyncUser> user,
     }
     
     if (sync_user_idx == -1) {
-        if (!current_user()) {
-            return completion_block(AppError(make_custom_error_code(ClientErrorCode::user_not_found), "No user has been found"));
-        }
-        if (current_user()->is_logged_in()) {
-            current_user()->log_out();
-        }
-        SyncManager::shared().remove_user(current_user()->identity());
+        return completion_block(AppError(make_custom_error_code(ClientErrorCode::user_not_found),
+                                         "No user has been found"));
     } else {
         auto sync_user = SyncManager::shared().all_users()[sync_user_idx];
         if (sync_user->is_logged_in()) {
@@ -682,7 +677,8 @@ void App::remove_user(std::shared_ptr<SyncUser> user,
 void App::remove_user(std::function<void(Optional<AppError>)> completion_block) const
 {
     if (!current_user()) {
-        return completion_block(AppError(make_custom_error_code(ClientErrorCode::user_not_found), "No user has been found"));
+        return completion_block(AppError(make_custom_error_code(ClientErrorCode::user_not_found),
+                                         "No user has been found"));
     }
     
     auto current_user_id = current_user()->identity();
--- a/src/sync/app.cpp
+++ b/src/sync/app.cpp
@@ -631,20 +631,17 @@ void App::switch_user(std::shared_ptr<SyncUser> user,
                                          "User is not longer valid or is logged out"));
     }
     
-    int sync_user_idx = -1;
-    for(size_t i = 0; i < SyncManager::shared().all_users().size(); i++) {
-        if (SyncManager::shared().all_users()[i] == user) {
-            sync_user_idx = static_cast<int>(i);
-            break;
-        }
-    }
+    auto users = SyncManager::shared().all_users();
+    auto it = std::find(users.begin(),
+                        users.end(),
+                        user);
     
-    if (sync_user_idx == -1) {
+    if (it == users.end()) {
         return completion_block(nullptr,
                                 AppError(make_custom_error_code(ClientErrorCode::user_not_found),
                                          "User does not exist"));
     }
-    
+
     SyncManager::shared().set_current_user(user->identity());
     return completion_block(current_user(), {});
 }
@@ -657,15 +654,12 @@ void App::remove_user(std::shared_ptr<SyncUser> user,
                                          "User has already been removed"));
     }
     
-    int sync_user_idx = -1;
-    for(size_t i = 0; i < SyncManager::shared().all_users().size(); i++) {
-        if (SyncManager::shared().all_users()[i] == user) {
-            sync_user_idx = static_cast<int>(i);
-            break;
-        }
-    }
+    auto users = SyncManager::shared().all_users();
+    auto it = std::find(users.begin(),
+                        users.end(),
+                        user);
     
-    if (sync_user_idx == -1) {
+    if (it == users.end()) {
         return completion_block(AppError(make_custom_error_code(ClientErrorCode::user_not_found),
                                          "No user has been found"));
     }
--- a/src/sync/app.cpp
+++ b/src/sync/app.cpp
@@ -631,11 +631,15 @@ void App::switch_user(std::shared_ptr<SyncUser> user,
                                          "User is not longer valid or is logged out"));
     }
     
-    auto it = std::find(SyncManager::shared().all_users().begin(),
-                        SyncManager::shared().all_users().end(),
-                        user);
+    int sync_user_idx = -1;
+    for(size_t i = 0; i < SyncManager::shared().all_users().size(); i++) {
+        if (SyncManager::shared().all_users()[i] == user) {
+            sync_user_idx = static_cast<int>(i);
+            break;
+        }
+    }
     
-    if (it == SyncManager::shared().all_users().end()) {
+    if (sync_user_idx == -1) {
         return completion_block(nullptr,
                                 AppError(make_custom_error_code(ClientErrorCode::user_not_found),
                                          "User does not exist"));
@@ -653,11 +657,15 @@ void App::remove_user(std::shared_ptr<SyncUser> user,
                                          "User has already been removed"));
     }
     
-    auto it = std::find(SyncManager::shared().all_users().begin(),
-                        SyncManager::shared().all_users().end(),
-                        user);
+    int sync_user_idx = -1;
+    for(size_t i = 0; i < SyncManager::shared().all_users().size(); i++) {
+        if (SyncManager::shared().all_users()[i] == user) {
+            sync_user_idx = static_cast<int>(i);
+            break;
+        }
+    }
     
-    if (it == SyncManager::shared().all_users().end()) {
+    if (sync_user_idx == -1) {
         return completion_block(AppError(make_custom_error_code(ClientErrorCode::user_not_found),
                                          "No user has been found"));
     }
--- a/src/sync/app.cpp
+++ b/src/sync/app.cpp
@@ -641,32 +641,35 @@ void App::switch_user(std::shared_ptr<SyncUser> user,
                                          "User does not exist"));
     }
     
-    auto sync_user = *it;
-    SyncManager::shared().set_current_user(sync_user->identity());
-    
+    SyncManager::shared().set_current_user(user->identity());
     return completion_block(current_user(), {});
 }
 
 void App::remove_user(std::shared_ptr<SyncUser> user,
                       std::function<void(Optional<AppError>)> completion_block) const
 {
+    if (user->state() == SyncUser::State::Removed) {
+        return completion_block(AppError(make_custom_error_code(ClientErrorCode::user_not_found),
+                                         "User has already been removed"));
+    }
+    
     auto it = std::find(SyncManager::shared().all_users().begin(),
                         SyncManager::shared().all_users().end(),
                         user);
-    
     
     if (it == SyncManager::shared().all_users().end()) {
         return completion_block(AppError(make_custom_error_code(ClientErrorCode::user_not_found),
                                          "No user has been found"));
     }
     
-    auto sync_user = *it;
-    SyncManager::shared().remove_user(sync_user->identity());
-    
-    if (sync_user->is_logged_in()) {
-        log_out(sync_user, [sync_user, completion_block](const Optional<AppError>& error){
+    if (user->is_logged_in()) {
+        log_out(user, [user, completion_block](const Optional<AppError>& error){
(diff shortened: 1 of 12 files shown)


src/sync/app.hpp
```diff
@@ -269,6 +269,23 @@ class App {
...
+    /// exist in the list of all users who have logged into this application, and
+    /// the user must be currently logged in, otherwise this will return an
+    /// AppError.
+    ///
+    /// @param user_id A String specifying the desired `user_id`
+    /// @param completion_block will return a SyncUser on success, an AppError on failure
+    void switch_user(const std::string& user_id,
+                     std::function<void(std::shared_ptr<SyncUser>, Optional<AppError>)> completion_block) const;
+    
+    /// Logs out and removes the user with the provided id
```
cmelchior: Would probably be helpful to specify that this is only a local operation and doesn't affect the server in any way.

src/sync/app.hpp
```diff
@@ -269,6 +269,23 @@ class App {
...
+    /// if no user ID is provided it will attempt to log out the current user and remove it
```
cmelchior: It would probably be cleaner to have a public `remove_user` method that only accepts a completion block similar to how log_out works. This feels potentially confusing.

src/sync/sync_manager.cpp
```diff
@@ -411,6 +411,32 @@ void SyncManager::set_current_user(const std::string& user_id)
...
+    
+    if (!m_metadata_manager) {
+        return;
+    }
+    
+    if (user->identity() == m_metadata_manager->get_current_user_identity()) {
+        m_metadata_manager->set_current_user_identity("");
+    }
+    
+    m_users.erase(it);
```
cmelchior: This looks dangerous. SDK's might still have a pointer to the User after it has been removed. I suspect doing this will cause all operations to fail after this. See https://mongodb.slack.com/archives/CKW2HG3M0/p1584430859364900 for more context.

IMO it would make sense to rework the current User states, so instead of the current ones we have `LOGGED_IN, LOGGED_OUT and REMOVED. It seems to fit the semantics we offer in a better way.

tests/sync/app.cpp
```diff
@@ -1190,3 +1181,173 @@ TEST_CASE("app: response error handling", "[sync][app]") {
...
+    reset_test_directory(base_path);
+
+    auto tsm = TestSyncManager(base_path);
+    
+    bool processed = false;
+    
+    std::string id_a;
+    std::string id_b;
+    
+    SECTION("remove user expect success") {
```
cmelchior: Missing tests of what happens if you access user after it was removed.

src/sync/app.hpp
```diff
@@ -269,6 +269,23 @@ class App {
...
     void log_out(std::shared_ptr<SyncUser> user, std::function<void(Optional<AppError>)> completion_block) const;
+    
+    /// Switches the active user to the user with the specified id. The user must
+    /// exist in the list of all users who have logged into this application, and
+    /// the user must be currently logged in, otherwise this will return an
+    /// AppError.
+    ///
+    /// @param user_id A String specifying the desired `user_id`
+    /// @param completion_block will return a SyncUser on success, an AppError on failure
+    void switch_user(const std::string& user_id,
```
cmelchior: Why use `user_id` instead of the accepting the `shared_ptr<SyncUser>`? If an SDK wants to Switch a user, the developer would already have a handle to that user which is in the form of a shared_ptr. This kind of API just seems to result in us needing to look things up more times?

src/sync/app.hpp
leemaguire: Makes sense, I've updated the doc to reflect that

src/sync/app.hpp
leemaguire: I've now added another method that just takes the completion block

tests/sync/app.cpp
leemaguire: Added a check to make sure the `user_a` and `user_b` reflect the correct state after they are removed

src/sync/app.cpp
```diff
@@ -622,5 +622,74 @@ void App::log_out(std::function<void (Optional<AppError>)> completion_block) con
     log_out(current_user(), completion_block);
 }
 
+void App::switch_user(std::shared_ptr<SyncUser> user,
+                      std::function<void(std::shared_ptr<SyncUser>, Optional<AppError>)> completion_block) const
+{
+    int sync_user_idx = -1;
+    for(size_t i = 0; i < SyncManager::shared().all_users().size(); i++) {
```
jsflax: [style] Simplify this to `auto it = std::find(SyncManager::shared().all_users().begin(), SyncManager::shared().all_users().end(), user)`. Then replace the below checks using the appropriate iterator comparisons, and read the user by dereferencing it from the iterator.

src/sync/app.cpp
```diff
@@ -622,5 +622,74 @@ void App::log_out(std::function<void (Optional<AppError>)> completion_block) con
...
+    
+    if (sync_user_idx == -1) {
+        return completion_block(nullptr,
+                                AppError(make_custom_error_code(ClientErrorCode::user_not_found), "User does not exist"));
+    }
+    
+    auto sync_user = SyncManager::shared().all_users()[sync_user_idx];
+    if (!sync_user->is_logged_in()) {
+        return completion_block(nullptr,
+                                AppError(make_custom_error_code(ClientErrorCode::user_not_logged_in), "User is not longer valid"));
```
jsflax: This is not entirely accurate, they may be logged out. Please add a clearer error message.

src/sync/app.cpp
```diff
@@ -622,5 +622,74 @@ void App::log_out(std::function<void (Optional<AppError>)> completion_block) con
...
+    SyncManager::shared().set_current_user(user->identity());
+    
+    return completion_block(current_user(), {});
+}
+
+void App::remove_user(std::shared_ptr<SyncUser> user,
+                      std::function<void(Optional<AppError>)> completion_block) const
+{
+    int sync_user_idx = -1;
+    for(size_t i = 0; i < SyncManager::shared().all_users().size(); i++) {
```
jsflax: [style] Same as above.

src/sync/app.cpp
```diff
@@ -622,5 +622,74 @@ void App::log_out(std::function<void (Optional<AppError>)> completion_block) con
...
+        if (SyncManager::shared().all_users()[i] == user) {
+            sync_user_idx = static_cast<int>(i);
+            break;
+        }
+    }
+    
+    if (sync_user_idx == -1) {
```
jsflax: [style] No need to if/else here, just if.

src/sync/app.cpp
```diff
@@ -622,5 +622,74 @@ void App::log_out(std::function<void (Optional<AppError>)> completion_block) con
...
+        return completion_block(AppError(make_custom_error_code(ClientErrorCode::user_not_found),
+                                         "No user has been found"));
+    } else {
+        auto sync_user = SyncManager::shared().all_users()[sync_user_idx];
+        if (sync_user->is_logged_in()) {
```
jsflax: [major] This needs to follow the full logout flow. You are only logging out the user locally.

src/sync/app.cpp
```diff
@@ -622,5 +622,74 @@ void App::log_out(std::function<void (Optional<AppError>)> completion_block) con
...
+            sync_user->log_out();
+        }
+        SyncManager::shared().remove_user(user->identity());
+    }
+
+    return completion_block({});
+}
+
+void App::remove_user(std::function<void(Optional<AppError>)> completion_block) const
```
jsflax: [q] Do we need this function?

src/sync/sync_manager.cpp
```diff
@@ -411,6 +411,33 @@ void SyncManager::set_current_user(const std::string& user_id)
...
+    if (!m_metadata_manager) {
+        return;
+    }
+    
+    if (user->identity() == m_metadata_manager->get_current_user_identity()) {
+        m_metadata_manager->set_current_user_identity("");
+    }
+    
+    user->set_state(SyncUser::State::Removed);
+
```
jsflax: [major] You also need to mark the user for removal. This is not fully removing the user.

leemaguire: done

src/sync/app.cpp
leemaguire: It is potentially confusing as we do not know who we are removing, removed this function as a result of that

src/sync/app.cpp
leemaguire: done

src/sync/app.cpp
leemaguire: done

src/sync/app.cpp
leemaguire: done

src/sync/app.cpp
leemaguire: done

src/sync/app.cpp
leemaguire: done

src/sync/app.hpp
leemaguire: `shared_ptr<SyncUser>` is now the default parameter

src/sync/sync_manager.cpp
leemaguire: Changed, also I have updated the `SyncUser::State` enum to reflect your suggestion

leemaguire:
@cmelchior can you PTAL

src/sync/app.cpp
```diff
@@ -622,5 +622,53 @@ void App::log_out(std::function<void (Optional<AppError>)> completion_block) con
     log_out(current_user(), completion_block);
 }
 
+void App::switch_user(std::shared_ptr<SyncUser> user,
```
cmelchior: Why do we log the user out when switching between them? It should be possible to have multiple users logged in at the same time? This would also remove the need for the callback, which feels really awkward for this type of API.

tests/sync/app.cpp
```diff
@@ -1190,3 +1181,244 @@ TEST_CASE("app: response error handling", "[sync][app]") {
...
+            CHECK(!error);
+            CHECK(SyncManager::shared().get_current_user() == user);
+            user_a = user;
+        });
+        
+        CHECK(user_a->state() == SyncUser::State::LoggedIn);
+        
+        app.log_out(user_a, [&](Optional<app::AppError> error) {
+            CHECK(!error);
+            // a logged out anon user will be marked as Error, not LoggedOut
```
cmelchior:
```suggestion
            // a logged out anon user will be marked as Removed, not LoggedOut
```

tests/sync/app.cpp
```diff
@@ -1190,3 +1181,244 @@ TEST_CASE("app: response error handling", "[sync][app]") {
...
+            CHECK(user_a->state() == SyncUser::State::Removed);
+        });
+        
+        app.remove_user(user_a, [&](Optional<app::AppError> error) {
+            CHECK(error->message == "No user has been found");
```
cmelchior: Is there a reason that removing an already removed user is an error? Couldn't it just be a no-op. I don't have a strong preference though.

But the error message feels a bit weird. Maybe rephrase to "User has already been removed"?

src/sync/app.cpp
```diff
@@ -622,5 +622,53 @@ void App::log_out(std::function<void (Optional<AppError>)> completion_block) con
...
+                        user);
+    
+    
+    if (it == SyncManager::shared().all_users().end()) {
+        return completion_block(AppError(make_custom_error_code(ClientErrorCode::user_not_found),
+                                         "No user has been found"));
+    }
+    
+    auto sync_user = *it;
+    SyncManager::shared().remove_user(sync_user->identity());
```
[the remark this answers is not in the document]
leemaguire: True, I've added it into the remove into the log_out completion block, I've also added an else to handle if the user is logged out

src/sync/app.cpp
leemaguire: we don't `log_out` on `switch_user`, we only perform the log_out on `remove_user`

tests/sync/app.cpp
leemaguire: I agree, a check has been added to see if they have already been removed

src/sync/app.cpp
cmelchior: Hmm, could have sworn there was code doing that, but why is there a callback with an Error in that case. That shouldn't be needed?

leemaguire: If the user is not logged in an error will be returned in the callback

src/sync/app.cpp
leemaguire: using std::find seems to fail tests on linux, using the traditional for loop structure seems to solve the issues with the tests

src/sync/app.cpp
cmelchior: Yes, but why making it a callback based API? That just makes it more annoying for consumers. There is nothing in this method that seems to require it, so I would just throw it as an exception?

src/sync/sync_manager.cpp
```diff
@@ -411,6 +411,34 @@ void SyncManager::set_current_user(const std::string& user_id)
...
+    if (it == m_users.end())
+        return;
+    
+    auto user = *it;
+    
+    if (!m_metadata_manager) {
+        return;
+    }
+    
+    for (size_t i = 0; i < m_metadata_manager->all_unmarked_users().size(); i++) {
```
jsflax: [style] why not use `find_if` again?

src/sync/app.cpp
leemaguire: After talking offline, I've removed the callback and changed to returning a shared pointer to the user, and throwing an exception otherwise

src/sync/sync_manager.cpp
leemaguire: resolved offline

PR #958, opened 2020-03-25
Je/update dependencies

jedelbo:


jedelbo:
@tgoyne We have changed Core in the way that it is not allowed to create `Table::create_object()` on a table with primary key (you should call `Table::create_object_with_primary_key()`). This seems to create some problems in some migration tests. Would it be a solution to always remove the possible primary key before doing the migration?

tgoyne:
Removing all primary keys at the start of a migration and then restoring them at the end is effectively what we do with core 5. I'm not sure that there's actually a good reason to be doing that, though. When adding a primary key to an existing type we can't enforce uniqueness until after the migration has run (as existing objects won't already have unique values if it's a new property), and when we first implemented primary keys the easiest way to achieve that was to just not enforce PK uniqueness during migrations at all.

I think the right thing to do is probably that if a table's primary key is being changed, remove it entirely before the migration and set it to the new pk after the migration, and otherwise don't touch it. There's probably some tests that'll need to be adjusted for this.

jedelbo:
@tgoyne Are these changes ok?

PR #961, opened 2020-03-27
Add link user and more auth clients

leemaguire:


src/sync/app.cpp
```diff
@@ -36,7 +36,7 @@ Optional<T> get_optional(const nlohmann::json& json, const std::string& key)
     return it != json.end() ? Optional<T>(it->get<T>()) : realm::util::none;
 }
 
-static std::map<std::string, std::string> get_request_headers(std::shared_ptr<SyncUser> with_user_authorization = nullptr)
+static std::map<std::string, std::string> get_request_headers(std::shared_ptr<SyncUser> with_user_authorization = nullptr, bool use_access_token = false)
```
cmelchior: These settings are a bit confusing to read at this point, since it in reality is 3 different modes. Perhaps extract it into a single enum `AuthMode` with 3 settings : `NoAuth`, `RefreshToken` `AccessToken`? This was just the first names that popped into my head, so feel free to change.

src/sync/app_credentials.hpp
```diff
@@ -64,10 +66,16 @@ struct AppCredentials {
...
     static AppCredentials anonymous();
 
-     // Construct and return credentials from an Apple account token.
+    // Construct and return credentials from an Apple account token.
     static AppCredentials apple(const AppCredentialsToken id_token);
-
-     // Construct and return credentials from a username and password.
+    
+    // Construct and return credentials from a google account token.
+    static AppCredentials google(const AppCredentialsToken id_token);
```
cmelchior: All of these seem to lack tests? How do we know they actually work?

tests/sync/app.cpp
```diff
@@ -1627,3 +1627,134 @@ TEST_CASE("app: remove user with credentials", "[sync][app]") {
...
+        });
+        
+        CHECK(sync_user->provider_type() == IdentityProviderUsernamePassword);
+
+        app.link_user(sync_user,
+                      custom_credentials,
+                      [&](std::shared_ptr<SyncUser> user, Optional<app::AppError> error) {
+            CHECK(!error);
+            REQUIRE(user);
+            CHECK(user->identity() == sync_user->identity());
```
cmelchior: The doc for this method says that a new user is returned, but this test seems to contradict this as they have a similar identity and provider_type?

src/sync/app_credentials.hpp
leemaguire: I've added unit tests to make sure each auth provider produces the correct json

tests/sync/app.cpp
leemaguire: original impl was wrong, fixed to take a new user id

src/sync/app.cpp
leemaguire: I've added an enum

src/sync/app.cpp
```diff
@@ -36,16 +36,36 @@ Optional<T> get_optional(const nlohmann::json& json, const std::string& key)
...
 
-static std::map<std::string, std::string> get_request_headers(std::shared_ptr<SyncUser> with_user_authorization = nullptr)
+enum class RequestTokenType {
+    NoAuth,
+    AccessToken,
+    RefreshToken
+};
+
+// generate the request headers for a HTTP call, by default it will generate headers with a refresh token if a user is passed
+static std::map<std::string, std::string> get_request_headers(std::shared_ptr<SyncUser> with_user_authorization = nullptr,
```
cmelchior: You can also remove `with_user_authorization`. It seems to be covered by the `NoAuth` case?

leemaguire: The purpose of that is so you can specify the user, as opposed to just relying on the `current_user` method

cmelchior: Ah yeah, sorry. Missed you got the access/refresh token from it.

PR #962, opened 2020-03-27
Set up VS Code container development

fealebenpae:
Set up the container development workflow in Visual Studio Code which allows automatically setting up a development environment against the MDBRealm test server image.


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,27 @@
...
+    "service": "tests",
+    "workspaceFolder": "/realm-object-store",
+
+    // Set *default* container specific settings.json values on container create.
+    "settings": { 
+        "terminal.integrated.shell.linux": null,
+        "cmake.configureSettings": {
+            "REALM_ENABLE_AUTH_TESTS": "ON",
+            "REALM_MONGODB_ENDPOINT": "http://mongodb-realm:9090",
+            "REALM_STITCH_CONFIG": "/realm-object-store/tests/mongodb/stitch.json",
```
ironage: does this absolute path actually work?

.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
+version: "3.4"
```
ironage: can you add a section in README.md about how to get vscode setup? (dependencies etc)
Also, might be worth mentioning that we can now just use `docker-compose up -d -f ./devcontainer/docker-compose.yml` to manually run a local stitch.

dependencies.list
```diff
@@ -2,3 +2,4 @@ REALM_CORE_VERSION=10.0.0-alpha.2
 REALM_SYNC_VERSION=10.0.0-alpha.4
 OPENSSL_VERSION=1.1.1b
 REALM_CORE_PACKAGING=2
+MDBREALM_TEST_SERVER_TAG=2020-03-26
```
ironage: any ideas on how we can easily get this in the Jenkinsfile too? (non-blocking)

.devcontainer/docker-compose.yml
fealebenpae: It should be enough to just open in Visual Studio Code and accept the prompt that suggests to open inside a container (it appears because of the detected devcontainer.json). The config files already list the extensions you need inside the container - CMake, Catch2, C++ debugging, etc., so Visual Studio Code should pull them in automatically.

IMO docker-compose is overkill, because it starts both containers.

.devcontainer/devcontainer.json
fealebenpae: Yes, that's how the workspace is mounted inside the container: https://github.com/realm/realm-object-store/pull/962/files#diff-682a74ffaa0a92339fe8ba23a3fc1febR7

.devcontainer/docker-compose.yml
fealebenpae: Actually you're right, the [Remote - Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) extension isn't installed by default. I'll add it to the list of suggested extensions inside the repo.

Jenkinsfile
```diff
@@ -109,7 +111,7 @@ def doDockerBuild(String flavor, Boolean withCoverage, Boolean enableSync, Strin
           // see https://github.com/realm/ci/tree/master/realm/docker/mongodb-realm
           // we refrain from using "latest" here to optimise docker pull cost due to a new image being built every day
           // if there's really a new feature you need from the latest stitch, upgrade this manually
-        withRealmCloud(version: '2020-03-23', appsToImport: ['auth-integration-tests': "${env.WORKSPACE}/tests/mongodb"]) { networkName ->
+        withRealmCloud(version: dependencies.MDBREALM_TEST_SERVER_TAG, appsToImport: ['auth-integration-tests': "${env.WORKSPACE}/tests/mongodb"]) { networkName ->
```
ironage: 💯  awesome!

PR #963, opened 2020-04-01
Add Bson types and parser

jsflax:
Adds:
* RegularExpression
* MinKey
* MaxKey
* bson_null_t (special null type)
* IndexedMap
    * Map that is ordered by insertion order
* Bson (`std::variant` for Bson compatible types)
* BsonDocument and BsonArray
    * type aliases for `IndexedMap<Bson>` and `std::vector<Bson>`
* parser and `to_json` for serialisation and deserialisation 

TODO:
* Complete bson corpus tests
* More coverage around parser


src/util/bson.hpp
```diff
@@ -0,0 +1,1218 @@
...
+        }
+        return true;
+    }
+
+    /*!
+    @brief an object key was read
+    @param[in] val  object key
+    @return whether parsing should proceed
+    @note It is safe to move the passed string.
+    */
```
ironage, line 1028: I think we prefer single line `//` comments in this repository for documentation, IIRC the reasoning is that it is harder to turn things off experimentally or during development by using multiline comments like this because these doc comments will close things off.

src/util/bson.hpp
```diff
@@ -0,0 +1,1218 @@
...
+
+    /*!
+    @brief an floating-point number was read
+    @param[in] val  floating-point value
+    @param[in] s    raw token value
+    @return whether parsing should proceed
+    */
+    bool number_float(number_float_t, const string_t&) override {
+        throw BsonError(util::format("Invalid SAX instruction for canonical extended json: 'number_float'"));
+        return true;
```
ironage, line 902: I think this will cause a warning for unreachable code because we always throw (there's other place like this too)

src/util/bson.hpp
```diff
@@ -0,0 +1,1218 @@
...
+    }
+
+    /*!
+    @brief an integer number was read
+    @param[in] val  integer value
+    @return whether parsing should proceed
+    */
+    bool number_integer(number_integer_t) override {
+        throw BsonError(util::format("Invalid SAX instruction for canonical extended json: 'number_integer'"));
```
ironage, line 835: So far as I can see, there is a design conflict for error checking, none of these methods ever return false, instead if there's an error they throw an exception. Unless I'm missing something, we should choose either a status checking API or an exception throwing one, but not both.

src/util/bson.hpp
```diff
@@ -0,0 +1,1218 @@
...
+{
+    if (auto value = std::get_if<const char*>(&lhs))
+        return *value == rhs.data();
+    return false;
+}
+
+inline bool operator==(const Bson& lhs, const char rhs[])
+{
+    if (auto value = std::get_if<const char*>(&lhs))
+        return !strcmp(*value, (const char*)rhs);
```
ironage, line 522: this assumes that a null character is the end of the string, so it won't work for strings storing nulls. One fix is to use std::string everywhere, another is to pass in the size of both strings to compare

src/util/bson.hpp
```diff
@@ -0,0 +1,1218 @@
...
+            ss << "{\"$date\":{\"$numberLong\":\"" << bson << "\"}}";
+        },
+        [&ss](MaxKey) {
+            ss << "{\"$maxKey\":1}";
+        },
+        [&ss](MinKey) {
+            ss << "{\"$minKey\":1}";
+        },
+        [&ss](const char* bson) {
+            ss << '"' << bson << '"';
```
ironage, line 469: this won't print anything past a null character

tests/util/bson.cpp
```diff
@@ -0,0 +1,411 @@
...
+    const char* canonical_extjson;
+    corpus_check_t<T> check;
+    bool lossy;
+};
+
+template <typename T>
+using corpus_t = std::vector<corpus_entry_t<T>>;
+
+template <typename T>
+static inline void run_corpus(const char* test_key, const corpus_t<T>& corpus) {
```
tgoyne, line 39: This function mostly just obfuscates the tests and makes them harder to debug. Eliminating the loop and calling the function once per SECTION would be basically the same amount of code, but simpler.

tests/util/bson.cpp
```diff
@@ -0,0 +1,411 @@
...
+    for (auto entry : corpus) {
+        SECTION(entry.description) {
+            std::string canonical_extjson = remove_whitespace(entry.canonical_extjson);
+            auto val = std::get<BsonDocument>(bson::parse(canonical_extjson));
+            auto test_value = val[test_key];
+            CHECK(std::holds_alternative<T>(test_value));
```
tgoyne, line 45: This should be `REQUIRE()` as we don't want to run the later checks if this fails.

src/util/bson.hpp
```diff
@@ -0,0 +1,1218 @@
...
+            {
+                to_json(b, ss);
+                ss << ",";
+            }
+            if (bson.size())
+                ss.seekp(-1, std::ios_base::end);
+            ss << "]";
+        },
+        [&ss](BinaryData bson) {
+            ss << "{\"$binary\":{\"base64\":\"" << (bson.is_null() ? "" : bson.data()) << "\",\"subType\":\"00\"}}";
```
ironage, line 448: This doesn't seem to actually be printing a base64 encoded version of the data?
I think this should work:
```
        util::StringBuffer encode_buffer;
        encode_buffer.resize(util::base64_encoded_size(bson.size()));
        util::base64_encode(bson.data(), bson.size(), encode_buffer.data(), encode_buffer.size());
        ss << encode_buffer.str();
```

tests/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ set(SOURCES
     util/event_loop.cpp
     util/test_file.cpp
     util/test_utils.cpp
+    util/bson.cpp
```
tgoyne, line 29: Keep the list of files sorted. This new test file doesn't go in `util`; that directory is utilities for the tests, not tests for utilities.

src/util/bson.hpp
```diff
@@ -0,0 +1,1218 @@
...
+                    m_instructions.push({state_t::skip});
+                    m_instructions.push({state_t::skip});
+                } else {
+                    // we will ignore the subtype for now
+                    m_marks.top().push_back({instruction.key, BinaryData()});
+                }
+
+                break;
+            case state_t::binary_base64: {
+                char* binary_val = (char *)malloc(sizeof(char) * val.size());
```
tgoyne, line 989: Writing `malloc` is basically always incorrect, and this is no exception. This memory is simply leaked currently.

src/util/bson.hpp
```diff
@@ -0,0 +1,1218 @@
...
+            ss << "{\"$binary\":{\"base64\":\"" << (bson.is_null() ? "" : bson.data()) << "\",\"subType\":\"00\"}}";
+        },
+        [&ss](RegularExpression bson) {
+            std::stringstream options;
+            for (const auto &option : bson.options()) options << option;
+            ss << "{\"$regularExpression\":{\"pattern\":\"" << bson.pattern()
+                << "\",\"options\":\"" << options.str() << "\"}}";
+        },
+        [&ss](Timestamp bson) {
+            ss << "{\"$timestamp\":{\"t\":" << bson.get_seconds() << ",\"i\":" << 1 << "}}";
```
ironage, line 457: What is the "i" parameter?
Are we going to truncate the extra information we have stored in Timestamp::get_nanoseconds()?

src/util/bson.hpp
```diff
@@ -0,0 +1,1218 @@
...
+class Bson : public std::variant<
+    bson_null_t,
+    int32_t,
+    int64_t,
+    bool,
+    float,
+    double,
+    time_t,
+    const char*,
+    BinaryData,
```
tgoyne, line 335: Bson needs to own the data for string and binary values, so these need to be `std::string` and `std::vector<char>`.

src/util/bson.hpp
```diff
@@ -0,0 +1,1218 @@
...
+                ss << bson;
+            }
+            ss << '"' << "}";
+        },
+        [&ss](int64_t bson) {
+            ss << "{" << "\"$numberLong\"" << ":" << '"' << bson << '"' << "}";
+        },
+        [&ss](Decimal128 bson) {
+            ss << "{" << "\"$numberDecimal\"" << ":" << '"';
+            if (bson.is_nan()) {
```
ironage, line 410: Is Decimal128::to_string() insufficient? It should handle everything you are doing here if you don't mind "Inf" and "-Inf"

jsflax: I do not believe Inf and -Inf are valid extjson, hence the conditionals.

src/util/bson.hpp
```diff
@@ -0,0 +1,1218 @@
...
+}
+
+/**
+ ====== BSON =======
+ */
+
+class Bson;
+
+/// A variant of the allowed Bson types.
+class Bson : public std::variant<
```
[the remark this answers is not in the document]
jsflax: Not sure this is possible since the variant contains collections templated to itself.

src/util/bson.hpp
jsflax: I've effectively chosen an exception throwing API that uses status checks. I've no choice but to adhere to nlohman's SAX parser, but you'll notice that I always return true.

src/util/bson.hpp
jsflax: "i" is operations within a given second, which defaults to one.

We do not need the extra information– this field is meant to be internal to the bson implementation: https://docs.mongodb.com/manual/reference/bson-types/#timestamps

jsflax:
@tgoyne @ironage PTAL

src/util/bson/bson.cpp
```diff
@@ -0,0 +1,30 @@
...
+*
+* Copyright 2020 Realm Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unleout required by applicable law or agreed to in writing, software
```
tgoyne, line 11: "Unleout" shows up in a few of these license headers.

src/util/bson/bson.cpp
```diff
@@ -0,0 +1,30 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either expreout or implied.
+* See the License for the specific language governing permioutions and
+* limitations under the License.
+*
+**************************************************************************/
+
+#include "bson.hpp"
```
tgoyne, line 19: All local include paths need to be relative to the `src` directory to work with the large set of build systems that the object store code is built with, so this should be "util/bson.hpp".

(All the other includes need to be adjusted too).

src/util/bson/parser.hpp
```diff
@@ -0,0 +1,165 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either expreout or implied.
+* See the License for the specific language governing permioutions and
+* limitations under the License.
+*
+**************************************************************************/
+
+#ifndef REALM_BSON_PARSER_HPP
+#define REALM_BSON_PARSER_HPP
+
+#include "../../../external/json/json.hpp"
```
tgoyne, line 22: `#include <json.hpp>`

src/util/bson/parser.hpp
```diff
@@ -0,0 +1,165 @@
...
+#include "bson.hpp"
+#include <variant>
+#include <stack>
+
+namespace realm {
+namespace bson {
+namespace detail {
```
tgoyne, line 29: We normally use `_impl` rather than `detail` for private things that have to be exposed in headers.

In this case though, none of this needs to be exposed. Everything in parser.hpp and parser.cpp can just be in bson.cpp and in an anonymous namespace, as it's all just an implementation detail of bson::parse() that doesn't need to be used from anywhere else.

src/util/bson/regular_expression.hpp
```diff
@@ -0,0 +1,95 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*
+**************************************************************************/
+
+#ifndef REALM_BSON_REGULAR_EXPRESSION_HPP
+#define REALM_BSON_REGULAR_EXPRESSION_HPP
+
+#include <iostream>
```
tgoyne, line 22: <iostream>

src/util/bson/bson.hpp
```diff
@@ -0,0 +1,190 @@
...
+#include "regular_expression.hpp"
+#include "min_key.hpp"
+#include "max_key.hpp"
+#include "null.hpp"
+
+#include <realm/binary_data.hpp>
+#include <realm/timestamp.hpp>
+#include <realm/decimal128.hpp>
+#include <realm/object_id.hpp>
+#include <iostream>
```
tgoyne, line 32: This file only needs `<ostream>` and not `<iostream>` (including `<iostream>` can be surprisingly bad due to how the `std::cout` etc. globals have to be implemented).

src/util/bson/bson.hpp
```diff
@@ -0,0 +1,190 @@
...
+    using base::operator=;
+};
+
+using BsonDocument = IndexedMap<Bson>;
+using BsonArray = std::vector<Bson>;
+
+template<class... Ts> struct overloaded : Ts... { using Ts::operator()...; };
+template<class... Ts> overloaded(Ts...) -> overloaded<Ts...>;
+
+inline std::ostream& operator<<(std::ostream& out, const Bson& b)
```
tgoyne, line 91: This should be defined in the cpp file. iostreams produce very large compiled code and so are something you want to avoid putting in headers.

src/util/bson/bson.hpp
```diff
@@ -0,0 +1,190 @@
...
+{
+    std::visit(overloaded {
```
tgoyne, line 93: Using an explicit struct with operator() overloads looks like it'd be about the same amount of boilerplate while being a bit more straightforward and not involving massive template expansions. `std::overloaded` is nice when the lambda captures are saving a whole bunch of code, but here each function is just capturing a single variable.

src/util/bson/bson.hpp
```diff
@@ -0,0 +1,190 @@
...
+                out << "-Infinity";
+            } else {
+                out << bson;
+            }
+            out << '"' << "}";
+        },
+        [&out](ObjectId bson) {
+            out << "{" << "\"$oid\"" << ":" << '"' << bson << '"' << "}";
+        },
+        [&out](BsonDocument bson) {
```
tgoyne, line 135: Copying the argument for BsonDocument and BsonArray seems like it could easily matter sometimes.

src/util/bson/indexed_map.hpp
```diff
@@ -0,0 +1,239 @@
...
+    template <typename V>
+    friend inline bool operator==(const IndexedMap<V>& lhs, const IndexedMap<V>& rhs) noexcept;
+    template <typename V>
+    friend inline bool operator!=(const IndexedMap<V>& lhs, const IndexedMap<V>& rhs) noexcept;
+    std::unordered_map<std::string, T> m_map;
+    std::vector<std::string> m_keys;
+};
+
+template <typename T>
+inline bool operator==(const typename IndexedMap<T>::iterator& lhs,
```
tgoyne, line 96: Template functions are implicitly inline.

src/util/bson/indexed_map.hpp
```diff
@@ -0,0 +1,239 @@
...
+template <typename T>
+typename IndexedMap<T>::iterator IndexedMap<T>::iterator::operator++(int)
+{
+    return IndexedMap<T>::iterator(m_idx++);
+}
+
+template <typename T>
+typename IndexedMap<T>::iterator IndexedMap<T>::iterator::operator--(int)
+{
+    return --(*this);
```
tgoyne, line 147: This needs to do the same thing as the postfix ++ function.

src/util/bson/indexed_map.hpp
```diff
@@ -0,0 +1,239 @@
...
+}
+
+template <typename T>
+using entry = std::pair<std::string, T>;
```
tgoyne, line 151: This is declaring a separate template alias for `entry` outside of the scope of `IndexedMap`, and it shouldn't be.

src/util/bson/indexed_map.hpp
```diff
@@ -0,0 +1,239 @@
...
+
+template <typename T>
+IndexedMap<T>::IndexedMap() {}
+
+template <typename T>
+IndexedMap<T>::IndexedMap(std::initializer_list<entry> entries) {
+    for (auto entry : entries) {
```
tgoyne, line 158: `auto& entries` so that it doesn't copy each element.

src/util/bson/indexed_map.hpp
```diff
@@ -0,0 +1,239 @@
...
+        m_keys.push_back(entry.first);
+        m_map[entry.first] = entry.second;
+    }
+}
+
+/// The size of the map
```
tgoyne, line 164: Doc comments should be in the class declaration.

src/util/bson/indexed_map.hpp
```diff
@@ -0,0 +1,239 @@
...
+template <typename T>
+typename IndexedMap<T>::iterator IndexedMap<T>::end()
+{
+    return iterator(this, m_map.size());
+}
+
+template <typename T>
+typename IndexedMap<T>::iterator IndexedMap<T>::front()
+{
+    return this->begin();
```
tgoyne, line 194: `front()` is `*begin()`, not an alias for `begin()`.

src/util/bson/indexed_map.hpp
```diff
@@ -0,0 +1,239 @@
...
+}
+
+template <typename T>
+typename IndexedMap<T>::iterator IndexedMap<T>::back()
+{
+    return this->end()--;
```
tgoyne, line 200: `*--end()`

src/util/bson/parser.cpp
```diff
@@ -0,0 +1,522 @@
...
+ @param[in] position    the position in the input where the error occurs
+ @param[in] last_token  the last read token
+ @param[in] ex          an exception object describing the error
+ @return whether parsing should proceed (must return false)
+ */
+bool Parser::parse_error(std::size_t,
+                         const std::string&,
+                         const nlohmann::detail::exception& ex) {
+    throw ex;
+    return false;
```
tgoyne, line 502: This is unreachable dead code.

tests/bson.cpp
```diff
@@ -0,0 +1,438 @@
+#include "catch2/catch.hpp"
```
tgoyne, line 1: Test files need license headers too.

src/util/bson/parser.hpp
jsflax: I was exposing it for testing purposes, but fair points.

src/util/bson/bson.hpp
```diff
@@ -0,0 +1,190 @@
...
+            {
+                out << b << ",";
+            }
+            if (bson.size())
+                out.seekp(-1, std::ios_base::end);
+            out << "]";
+        },
+        [&out](std::vector<char> bson) {
+            out << "{\"$binary\":{\"base64\":\"" <<
+                std::string(bson.begin(), bson.end()) << "\",\"subType\":\"00\"}}";
```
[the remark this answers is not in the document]
jsflax: We need the extra information that it's a binary type. The other option is to create a Bson specific "Binary" type that stores the base64 string and subtype.

src/util/bson/bson.hpp
```diff
@@ -0,0 +1,190 @@
...
+        },
+        [&out](Null) {
+            out << "null";
+        },
+        [&out](int32_t bson) {
+            out << "{" << "\"$numberInt\"" << ":" << '"' << bson << '"' << "}";
+        },
+        [&out](double bson) {
+            out << "{" << "\"$numberDouble\"" << ":" << '"';
+            if (isnan(bson)) {
```
ironage, line 105: Be explicit with the namespace when using a std function so that the linux builds are happy:
```
../src/util/bson/bson.hpp: In lambda function:
../src/util/bson/bson.hpp:105:17: error: 'isnan' was not declared in this scope; did you mean 'std::isnan'?
  105 |             if (isnan(bson)) {
      |                 ^~~~~
      |                 std::isnan
```

src/util/bson/bson.hpp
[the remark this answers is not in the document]
ironage: Ok, I would consider this an optimization so not blocking.

src/util/bson/bson.hpp
```diff
@@ -32,41 +32,670 @@
...
+                *document_val = *v.document_val;
+                break;
+            case ARRAY:
+                array_val = new std::vector<Bson>;
+                *array_val = *v.array_val;
+                break;
+        }
+    }
+
+    Bson& operator=(Bson&& v) {
```
tgoyne: The move assignment operator needs to be marked as `noexcept` or it won't actually be used in a lot of places, as the exception safety guarantees provided by things like `std::vector` aren't possible with a move assignment operator that throws exceptions.

src/util/bson/bson.hpp
```diff
@@ -32,41 +32,670 @@
...
+        bool bool_val;
+        double double_val;
+        time_t time_val;
+        Timestamp date_val;
+        ObjectId oid_val;
+        Decimal128 decimal_val;
+        MaxKey max_key_val;
+        MinKey min_key_val;
+        // ref types
+        RegularExpression* regex_val;
```
tgoyne, line 219: These types (except maybe std::vector<Bson>?) don't need to be pointers. Instead they can be inline, and instead of `string_val = new std::string(...); delete string_val` you'd do `new (&string_val) std::string(...)` to construct and `string_val.~std::string()` to destroy.

src/util/bson/bson.hpp
```diff
@@ -32,41 +32,670 @@
...
+    m_type = Bson::MIN_KEY;
+    min_key_val = v;
+}
+
+inline Bson::Bson(MaxKey v) noexcept
+{
+    m_type = Bson::MAX_KEY;
+    max_key_val = v;
+}
+inline Bson::Bson(const RegularExpression& v) noexcept
```
tgoyne, line 262: These constructors which allocate memory aren't `noexcept` as the allocation can throw.

src/util/bson/bson.hpp
```diff
@@ -32,41 +32,670 @@
...
+                break;
+        }
+    }
+
+    Bson(Bson&& v) {
+        std::cout<<"TRYING TO MOVE BSON"<<std::endl;
+        m_type = v.m_type;
+    }
+
+    Bson(const Bson& v) {
```
tgoyne: The copy constructor can be just `m_type = None; *this = v;` (and similar for the move constructor).

src/util/bson/bson.cpp
```diff
@@ -184,45 +184,111 @@ class Parser : public nlohmann::json_sax<json> {
...
 
     /// Convenience class that overloads similar collection functionality
-    class BsonContainer : public std::variant<BsonDocument, BsonArray> {
+    class BsonContainer {
+        typedef enum Type {
+            DOCUMENT, ARRAY
+        } Type;
+        Type m_type;
+        union {
+            BsonDocument* document;
```
tgoyne, line 585: See the note on Bson about how these don't need to be pointers.

src/sync/sync_user.cpp
```diff
@@ -59,14 +59,14 @@ RealmJWT::RealmJWT(const std::string& token)
 {
     auto parts = split_token(token);
 
-    auto json = std::get<bson::BsonDocument>(bson::parse(base64_decode(parts[1])));
+    auto json = (bson::BsonDocument)bson::parse(base64_decode(parts[1]));
```
tgoyne, line 62: Using this cast syntax produces warnings when building with `-Wold-style-casts`, which when building with CocoaPods can end up trickling down to our source files. `static_cast<T>()` avoids that (and doesn't do some dangerous conversions which C casts can do).

src/util/bson/bson.hpp
jsflax: Ah, cool, this will clean things up. What is this syntax though?
`new (&string_val) std::string(...)`
I have not seen it before.

tgoyne: [Placement new](https://en.cppreference.com/w/cpp/language/new#Placement_new)

src/util/bson/bson.cpp
```diff
@@ -0,0 +1,1164 @@
...
+
+Bson::~Bson() noexcept
+{
+    switch (m_type) {
+        case Type::String:
+            string_val.~basic_string();
+            break;
+        case Type::Document:
+            delete document_val;
+            document_val = NULL;
```
tgoyne, line 34: nullptr rather than NULL

src/util/bson/bson.cpp
```diff
@@ -0,0 +1,1164 @@
...
+            max_key_val = v.max_key_val;
+            break;
+        case Type::MinKey:
+            min_key_val = v.min_key_val;
+            break;
+        case Type::RegularExpression:
+            regex_val = std::move(v.regex_val);
+            break;
+        case Type::Binary:
+            if (binary_val) delete binary_val;
```
tgoyne, line 96: This and the similar lines in the copy assignment operator are still wrong when m_type != v.m_type.

src/util/bson/datetime.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+#ifndef REALM_BSON_DATETIME_HPP
+#define REALM_BSON_DATETIME_HPP
+
+#include <ctime>
+
+namespace realm {
+namespace bson {
+
+struct Datetime {
```
tgoyne, line 27: This type doesn't seem to be doing much? It could just be `struct Datetime { std::time_t seconds_since_epoch; };`. Wrapping a call to gmtime() and exposing a very slightly different interface for `tm` doesn't seem useful.

src/sync/sync_user.cpp
```diff
@@ -60,14 +59,15 @@ RealmJWT::RealmJWT(const std::string& token)
...
     auto parts = split_token(token);
 
-    auto json = nlohmann::json::parse(base64_decode(parts[1]));
+    auto json_str = base64_decode(parts[1]);
+    auto json = static_cast<bson::BsonDocument>(bson::parse(json_str));
 
     this->token = token;
-    this->expires_at = app::value_from_json<long>(json, "exp");
-    this->issued_at = app::value_from_json<long>(json, "iat");
+    this->expires_at = (int64_t)json["exp"];
```
tgoyne, line 66: more c-style casts.

jsflax:
@tgoyne ptal

src/util/bson/bson.cpp
```diff
@@ -0,0 +1,1164 @@
...
+            if (m_type == DOCUMENT) {
+                document = new BsonDocument;
+                *document = *v.document;
+            } else {
+                array = new BsonArray;
+                *array = *v.array;
+            }
+        }
+
+        BsonContainer& operator=(const BsonContainer& v)
```
ironage, line 569: this and the below operator= need to check for "this" `if (&v == this) return *this;`
There's also a memory leak as the previous value is not freed.

src/util/bson/bson.cpp
```diff
@@ -0,0 +1,1164 @@
...
+        union {
+            BsonDocument* document;
+            BsonArray* array;
+        };
+    public:
+        ~BsonContainer() noexcept
+        {
+            if (m_type == DOCUMENT) {
+                delete document;
+                document = NULL;
```
ironage, line 529: prefer `nullptr`

src/util/bson/bson.cpp
```diff
@@ -0,0 +1,1164 @@
...
+    }
+}
+
+Bson::Bson(const Bson& v) {
+    m_type = Type::Null;
+    *this = v;
+}
+
+Bson& Bson::operator=(Bson&& v) noexcept {
+    if (m_type != v.m_type) this->~Bson();
```
ironage, line 57: You'll need protection from this here too. `if (this == &v) return *this;`

src/util/bson/bson.cpp
```diff
@@ -0,0 +1,1164 @@
...
+                document = new BsonDocument;
+                *document = *v.document;
+            } else {
+                array = new BsonArray;
+                *array = *v.array;
+            }
+            return *this;
+        }
+
+        BsonContainer& operator=(BsonContainer&& v)
```
ironage, line 582: - `if (&v == this) return *this;`
- this leaks memory if the type changes because you change the type before cleaning up

src/util/bson/bson.cpp
```diff
@@ -0,0 +1,1164 @@
...
+                array = v.array;
+                v.array = NULL;
+            }
+            return *this;
+        }
+
+        BsonContainer(BsonContainer&& v) {
+            m_type = v.m_type;
+            if (m_type == DOCUMENT) {
+                if (document) delete document;
```
ironage, line 600: No need to clean up memory in the constructor because we haven't allocated anything yet. Also don't forget that on linux, pointers aren't zero initialized so this will go and delete random memory somewhere and usually crash.

src/util/bson/bson.cpp
```diff
@@ -0,0 +1,1164 @@
...
+                v.array = NULL;
+            }
+        }
+
+        bool is_array() const { return m_type == ARRAY; }
+        bool is_document() const { return m_type == DOCUMENT; }
+
+        void push_back(std::pair<std::string, Bson> value) {
+            if (m_type == DOCUMENT) {
+                document->operator[](value.first) = value.second;
```
ironage, line 615: personal preference, but I find this to be a bit more readable `(*document)[value.first] ...`

src/util/bson/bson.cpp
```diff
@@ -0,0 +1,1164 @@
...
+            } else {
+                array->emplace_back(value.second);
+            }
+        }
+
+        std::pair<std::string, Bson> back()
```
ironage, line 621: maybe add an assert on the size

src/util/bson/bson.cpp
```diff
@@ -0,0 +1,1164 @@
...
+        void pop_back()
+        {
+            if (m_type == DOCUMENT) {
+                document->pop_back();
+            } else {
+                array->pop_back();
+            }
+        }
+
+        size_t size()
```
ironage, line 640: can this be a const method?

jsflax:
@ironage TPAL

src/util/bson/bson.cpp
```diff
@@ -581,29 +584,30 @@ class Parser : public nlohmann::json_sax<json> {
...
                 if (array) delete array;
                 array = v.array;
-                v.array = NULL;
+                v.array = nullptr;
             }
             return *this;
         }
 
         BsonContainer(BsonContainer&& v) {
+            if (m_type != v.m_type) this->~BsonContainer();
```
ironage, line 604: At this point (in a constructor) none of the member variables have been initialized, so they hold random values (though on mac everything is zero-initialized). Calling the destructor here will go and delete random memory which can cause a crash (if you're lucky). However, you would want to do this type of cleanup in the above `operator=` methods because those are still leaking memory when the type changes.

src/util/bson/bson.cpp
```diff
@@ -0,0 +1,1172 @@
...
+        }
+
+        BsonContainer& operator=(BsonContainer&& v)
+        {
+            if (&v == this) return *this;
+            this->~BsonContainer();
+
+            m_type = v.m_type;
+            if (m_type == DOCUMENT) {
+                if (document) delete document;
```
ironage, line 594: the remaining deletes in this method become redundant because things have already been cleaned up earlier (they are harmless because you did set things to nullptr, but for readability)

src/util/bson/bson.cpp
```diff
@@ -0,0 +1,1170 @@
...
+static constexpr const char * key_date                       = "$date";
+static constexpr const char * key_object_id                  = "$oid";
+static constexpr const char * key_max_key                    = "$maxKey";
+static constexpr const char * key_min_key                    = "$minKey";
+static constexpr const char * key_regular_expression         = "$regularExpression";
+static constexpr const char * key_regular_expression_pattern = "pattern";
+static constexpr const char * key_regular_expression_options = "options";
+static constexpr const char * key_binary                     = "$binary";
+static constexpr const char * key_binary_base64              = "base64";
+static constexpr const char * key_binary_sub_type            = "subType";
```
ironage, line 682: It might be nice to make these public in a header somewhere so that they can be reused, at least I noticed that the "app" code is hardcoding these strings everywhere too. Maybe a future improvement though (non-blocking).

jsflax: Hmm, the "app" code shouldn't be doing that. I'll look into it in a future PR.

jsflax:
@tgoyne Any final comments before merging?

PR #964, opened 2020-04-03
Link user fix impl

leemaguire:


tests/sync/app.cpp
```diff
@@ -1688,14 +1745,10 @@ TEST_CASE("app: link_user", "[sync][app]") {
                       [&](std::shared_ptr<SyncUser> user, Optional<app::AppError> error) {
             CHECK(!error);
             REQUIRE(user);
-            CHECK(user->identity() != sync_user->identity());
-            CHECK(sync_user->provider_type() == IdentityProviderUsernamePassword);
-            CHECK(user->provider_type() == IdentityProviderFacebook);
+            CHECK(user->identity() == sync_user->identity());
```
cmelchior, line 1748: Maybe check that there are now 2 entries in the identities array indicating that linking succeded?

tests/sync/app.cpp
```diff
@@ -1629,6 +1629,63 @@ TEST_CASE("app: remove user with credentials", "[sync][app]") {
...
+        auto app = App(config);
+        std::string base_path = tmp_dir() + "/" + config.app_id;
+        reset_test_directory(base_path);
+        TestSyncManager init_sync_manager(base_path);
+
+        bool processed = false;
+
+        std::shared_ptr<SyncUser> sync_user;
+        
+        auto custom_credentials = realm::app::AppCredentials::facebook("a_token");
```
ironage, line 1656: this isn't used anywhere

src/sync/impl/sync_metadata.cpp
```diff
@@ -572,26 +572,57 @@ void SyncUserMetadata::set_user_profile(const SyncUserProfile& profile)
...
+    Obj obj;
+    if (m_obj.is_null(m_schema.idx_profile)) {
+        obj = m_obj.create_and_set_linked_object(m_schema.idx_profile);
+    } else {
+        obj = m_obj.get_linked_object(m_schema.idx_profile);
+    }
+    
     if (profile.name)
         obj.set(c_sync_profile_name, *profile.name);
+    else
```
tgoyne, line 584: `obj.set(c_sync_profile_name, profile.name)` should work for all of these without manually handling the null/not-null cases.

PR #965, opened 2020-04-03
Add function, user api & server api key auth credentials

leemaguire:


src/sync/app_credentials.cpp
```diff
@@ -48,6 +51,12 @@ IdentityProvider provider_type_from_enum(AuthProvider provider)
             return IdentityProviderCustom;
         case AuthProvider::USERNAME_PASSWORD:
             return IdentityProviderUsernamePassword;
+        case AuthProvider::FUNCTION:
+            return IdentityProviderFunction;
+        case AuthProvider::USER_API_KEY:
+            return IdentityProviderUserAPIKey;
+        case AuthProvider::SERVER_API_KEY:
```
cmelchior, line 58: Don't think I have seen this Auth Provider documented anywhere? Is that something we should expose in the SDK's?

src/sync/app_credentials.hpp
```diff
@@ -78,6 +93,16 @@ struct AppCredentials {
     // Construct and return credentials from a username and password.
     static AppCredentials username_password(std::string username, std::string password);
 
+    // Construct and return credentials with the payload.
+    static AppCredentials function(std::string payload_json);
```
cmelchior, line 97: What is the format of this string? An JSON array? Something else? Maybe add it to the description?

tests/sync/app.cpp
```diff
@@ -1811,4 +1807,25 @@ TEST_CASE("app: auth providers", "[sync][app]") {
         CHECK(credentials.provider_as_string() == IdentityProviderUsernamePassword);
         CHECK(credentials.serialize_as_json() == "{\"password\":\"pass\",\"provider\":\"local-userpass\",\"username\":\"user\"}");
     }
+    
+    SECTION("auth providers function") {
```
cmelchior, line 1811: Missing a test for a function with arguments.

cmelchior: Also, I'm a bit worried that there is no test for this against the server, as it seems a fairly complex piece of functionality.

jsflax: We test this in the current SDKs, I agree with @cmelchior , we should absolutely be testing this.

src/sync/app_credentials.cpp
leemaguire: The Swift Stitch SDK has this, afaik its a valid credential and we should expose it.

src/sync/app_credentials.hpp
leemaguire: its going to be a mongodb document, I'll update the docs to reflect it

src/sync/app_credentials.cpp
cmelchior: Sure, I have no problem with that. I just wasn't aware it was a thing since it isn't documented on the Stitch Website as far as I can tell.

bmunkholm: What is this AuthProvider?
(As a principle, we should not blindly expose everything the existing Stitch API has. Only if it has big value)

jsflax: That principle goes precisely against the original arrangement with the product team. 

Either way, this auth provider allows for logging in with a server generated API key. Which is useful for IoT and debugging.

tests/mongodb/auth_providers/custom-function.json
```diff
@@ -0,0 +1,8 @@
+{
+    "name": "custom-function",
+    "type": "custom-function",
+    "config": {
+        "authFunctionName": "authFunc"
+    },
+    "disabled": false
+}
```
jsflax, line 8: [nit] missing newline

tests/mongodb/functions/authFunc/source.js
```diff
@@ -0,0 +1,17 @@
...
+    The return object must contain a string id, this string id will be used to login with an existing
+    or create a new user. This is NOT the Stitch user id, but it is the id used to identify which user has
+    been created or logged in with. 
+
+    If an error is thrown within the function the login will fail.
+
+    The default function provided below will always result in failure.
+  */
+
+  exports = (loginPayload) => {
```
jsflax, line 15: [nit] You should deconstruct this param:
`exports = ({realmCustomAuthFuncUserId}) => {`

src/sync/app_credentials.cpp
jsflax: @bmunkholm , I've only stated the above to be clear that we should move forward on our current path, and then re-examine the value of features later on. I don't feel that we have time to get caught up in value add discussions right now.

tests/sync/app.cpp
leemaguire: Test has been added

leemaguire:
@cmelchior @ironage PTAL

src/sync/app_credentials.cpp
bmunkholm: @jsflax But if it's a feature that we don't even document I assume it's not being used by users? It must be faster to not add it to all SDKs than adding it though?
It's likely fine to have in OS for now though for the sake of this PR.

tests/sync/app.cpp
```diff
@@ -733,6 +733,41 @@ TEST_CASE("app: UserAPIKeyProviderClient integration", "[sync][app]") {
...
+    bool processed = false;
+
+    SECTION("auth providers function integration") {
+
+        auto credentials = realm::app::AppCredentials::function(nlohmann::json({{"realmCustomAuthFuncUserId", "123456"}}).dump());
+
+        app.log_in_with_credentials(credentials,
+                                    [&](std::shared_ptr<realm::SyncUser> user, Optional<app::AppError> error) {
+            REQUIRE(user);
+            CHECK(user);
```
ironage, line 760: This check is redundant to the require above, (the require is good there though). Instead we could check that the user's id is "123456" and that the user's provider type is "custom-function".

leemaguire: check for provider type added, the id won't necessarily be 123456, stitch will create a new user if one is not found

PR #967, opened 2020-04-12
Add next generation Sync support

cmelchior:


src/sync/sync_session.cpp
```diff
@@ -114,106 +97,17 @@ struct SyncSession::State {
...
-    {
-        session.m_deferred_close = false;
-    }
-
-    void refresh_access_token(std::unique_lock<std::mutex>& lock, SyncSession& session,
-                              std::string access_token,
-                              const util::Optional<std::string>& server_url) const override
     {
         session.create_sync_session();
+        session.m_session->bind();
```
cmelchior, line 109: In Java this crashes with `../src/realm/sync/client.cpp:1569: [realm-core-10.0.0-alpha.4] Assertion failed: !m_initiated`

Looking at the Sync code, I suspect it is because bind is not idempotent, calling it multiple times is a logic error :/

fealebenpae: That's surprising - this is called immediately after a brand new sync session is created. Effectively this means that `bind()` should never be called on the same sync session. Do you have a full stack trace for this?

src/sync/sync_session.cpp
```diff
@@ -530,6 +534,9 @@ void SyncSession::create_sync_session()
             throw sync::BadServerUrl();
         }
         session_config.service_identifier += "/realm-sync";
+        // FIXME: Java needs the fully resolved URL for proxy support, but we also need it before
```
fealebenpae, line 537: Why? Only the hostname and port matter for `ProxySelector`, and those are already known on `App`.

src/sync/sync_session.cpp
cmelchior: I did have a full stack trace, but it didn't show anything interesting except the above failure. When using the debugger it looked like the Active State was entered multiple times, which makes sense given that it would trigger the error. I don't know why that is happening though.

src/sync/sync_session.cpp
cmelchior: Ok. I didn't look closely into exactly what is required for the proxy selector to work. I just noticed that right now we are using the fully qualified URL.

I probably still think exposing the full URL would be nice for debugging, but isn't super critical.

fealebenpae: The URL for sync sessions is always going to be the same - the base URL of their `App` plus `/realm-sync`.

src/sync/sync_session.cpp
fealebenpae: I will look into why that's the case - probably a bug on my part when I removed the `WaitingForAccessToken` state.

src/sync/app.hpp
```diff
@@ -305,26 +315,35 @@ class App {
...
-                     std::function<void(Optional<AppError>)> completion_block) const;
+                     std::function<void(util::Optional<AppError>)> completion_block) const;
 
     // Get a provider client for the given class type.
     template <class T>
     T provider_client() {
         return T(this);
     }
 
+    class Internal {
```
jsflax, line 326: Is this a common pattern? I'm assuming that this is being done to encapsulate this method exclusively for the friend class `realm::SyncSession` (as opposed to making this a private method which would overexpose App to SyncSession). But this smells a bit. Maybe a SyncSession should simply be constructed with the app route?

fealebenpae: We poke holes in the private API in a couple of other places like this in Object Store, such as `Realm`.

PR #971, opened 2020-04-15
Remote Mongo Implementation

leemaguire:


tests/sync/app.cpp
```diff
@@ -824,6 +827,350 @@ TEST_CASE("app: link_user integration", "[sync][app]") {
...
+            if (error) {
+                std::cout << "insert failed, message: " << error->error_code.message() << "+" << error->message << std::endl;
+            } else {
+                try {
+                    auto json = nlohmann::json::parse(document_json);
+                    auto object_id = json.at("insertedId").at("$oid").get<std::string>();
+                    
+                    dog_object_id = object_id;
+                } catch (AppError err) {
+                    CHECK(err.error_code.value() < 0);
```
ironage, line 885: Are you actually testing an error case here, or is this leftover debugging code?

src/sync/app.cpp
```diff
@@ -841,5 +842,53 @@ void App::link_user(std::shared_ptr<SyncUser> user,
     }
 
 
+RemoteMongoClient App::remote_mongo_client(const std::string& service_name)
+{
+    m_service_name = util::Optional<std::string>(service_name);
```
tgoyne, line 847: This should set the service name on the copy of the App made below and the function should be `const`. Even if mutating the member here is harmless, it'll be confusing when trying to debug things.

src/sync/app.cpp
```diff
@@ -841,5 +842,53 @@ void App::link_user(std::shared_ptr<SyncUser> user,
...
+        
+        completion_block(util::none, util::Optional<std::string>(response.body));
+    };
+
+    std::string route = util::format("%1/app/%2/functions/call", m_base_route, m_config.app_id);
+
+    auto args = nlohmann::json::parse(args_json);
+    args.push_back({ "name" , name });
+    if (service_name) {
+        args.push_back({ "service" , service_name.value() });
```
tgoyne, line 880: Getting the value out of an optional is normally done by dereferencing it (i.e. `*service_name`) rather than calling `value()`.

src/sync/app_service_client.hpp
```diff
@@ -32,17 +32,29 @@ namespace app {
...
+    /// @param name The name of the Realm Cloud function to be called.
+    /// @param args The `BSONArray` of arguments to be provided to the function.
+    /// @param completion_block Returns the result from the intended call, will return an Optional AppError is an error is thrown and a json string if successful
+    virtual void call_function(const std::string& name,
+                               const std::string& args_json,
+                               std::function<void (util::Optional<AppError>, util::Optional<std::string>)> completion_block) const = 0;
+    
+private:
+    /// The name of the Realm Cloud service
+    util::Optional<std::string> m_service_name;
```
tgoyne, line 57: This member can't ever be used anywhere since this type doesn't have any non-pure-virtual functions.

src/sync/remote_mongo_collection.cpp
```diff
@@ -18,3 +18,505 @@
...
+
+static void handle_count_response(util::Optional<AppError> error,
+                                  util::Optional<std::string> value,
+                                  std::function<void(uint64_t, util::Optional<AppError>)> completion_block)
+{
+    if (value && !error) {
+        try {
+            auto json = nlohmann::json::parse(value.value());
+            auto count_string = json.at("$numberLong").get<std::string>();
+            return completion_block(std::stoi(count_string), error);
```
tgoyne, line 43: stoi returns an `int`, so this'll truncate values. `stoll` gives a 64-bit int.

tests/mongodb/apps/default-app/functions/confirmFunc/source.js
```diff
@@ -0,0 +1,49 @@
+
```
ironage, line 1: Why add all these things to "default-app" when it's not used by CI and seems to be redundant with the "integration-tests" app? Is it for manually testing?

src/sync/auth_request_client.cpp
```diff
@@ -16,32 +16,4 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-#ifndef REMOTE_MONGO_READ_OPERATION_HPP
```
ironage, line 19: remove the file if not used

src/sync/remote_mongo_collection.cpp
```diff
@@ -18,3 +18,505 @@
...
+         auto args = nlohmann::json( {{"arguments", nlohmann::json::array({base_args} ) }} );
+
+         m_service->call_function("insertMany",
+                                  args.dump(),
+                                  [completion_block](util::Optional<AppError> error, util::Optional<std::string> value) {
+             if (value && !error) {
+                 try {
+                     auto json = nlohmann::json::parse(value.value());
+                     auto inserted_ids = json.at("insertedIds").get<std::vector<nlohmann::json>>();
+                     auto mapped_oids = std::map<uint64_t, std::string>();
```
tgoyne, line 270: It looks like this should just be a vector? The key of the map is just the index, so it doesn't need to be there at all.

src/sync/remote_mongo_collection.cpp
```diff
@@ -18,3 +18,505 @@
...
+}
+
+static void handle_count_response(util::Optional<AppError> error,
+                                  util::Optional<std::string> value,
+                                  std::function<void(uint64_t, util::Optional<AppError>)> completion_block)
+{
+    if (value && !error) {
+        try {
+            auto json = nlohmann::json::parse(value.value());
+            auto count_string = json.at("$numberLong").get<std::string>();
```
ironage, line 42: Can we do the int conversion here with `.get<uint64_t>()` instead of using `stoi`? I'd hope that the json conversion would be more robust than stoi, and would maybe enforce stricter json syntax. Plus if it fails, you'd get a json exception (hopefully more descriptive), rather than a generic "invalid argument".

tests/mongodb/apps/default-app/functions/confirmFunc/source.js
```diff
@@ -0,0 +1,49 @@
...
+        return { status: 'success' };
+      } else {
+        context.functions.execute('sendConfirmationEmail', username, token, tokenId);
+        return { status: 'pending' };
+      }
+  
+      return { status: 'fail' };
+    };
+
+    The uncommented function below is just a placeholder and will result in failure.
```
tgoyne, line 39: This comment appears to be incorrect?

tests/sync/app.cpp
```diff
@@ -824,6 +827,350 @@ TEST_CASE("app: link_user integration", "[sync][app]") {
...
+        realm::app::RemoteMongoCollection::RemoteFindOneAndModifyOptions find_and_modify_options {
+            util::Optional<std::string>(nlohmann::json({{ "name", "fido" }}).dump()), //project
+            util::Optional<std::string>(nlohmann::json({{ "name", 1 }}).dump()), //sort,
+            true, //upsert
+            true // return new doc
+        };
+        
+        collection.find_one_and_update(dog_document, dog_document2, find_and_modify_options, [&](std::string document, Optional<app::AppError> error) {
+            if (error) {
+                std::cout << "find_one_and_update failed, message: " << error->error_code.message() << "+" << error->message << std::endl;
```
tgoyne, line 1045: These need to produce test failures rather than just printing a message.

tests/sync/app.cpp
```diff
@@ -824,6 +827,350 @@ TEST_CASE("app: link_user integration", "[sync][app]") {
...
+    }
+    
+    SECTION("delete") {
+        
+        bool processed = false;
+        
+        auto documents = std::vector<std::string>();
+        documents.push_back(dog_document);
+        documents.push_back(dog_document);
+        documents.push_back(dog_document);
```
tgoyne, line 1140: `auto documents = std::vector(3, dog_document);`

src/sync/remote_mongo_collection.cpp
```diff
@@ -18,3 +18,505 @@
...
+}
+
+void RemoteMongoCollection::insert_one(const std::string& value_json,
+                                       std::function<void(std::string, util::Optional<AppError>)> completion_block)
+{
+    try {
+        auto base_args = m_base_operation_args;
+        auto document_json = nlohmann::json::parse(value_json);
+        base_args.push_back({ "document", document_json });
+        auto args = nlohmann::json( {{"arguments", nlohmann::json::array({base_args} ) }} );
```
tgoyne, line 184: The whitespace whenever braced init is involved is inconsistent both within this PR and with the rest of the codebase. The style we use elsewhere is to treat them like parentheses used for function calls, so no spaces before or after the brace (so this particular line would be `nlohmann::json({{"arguments", nlohmann::json::array({base_args})}});`)

src/sync/remote_mongo_collection.cpp
```diff
@@ -18,3 +18,505 @@
...
+            base_args.push_back({ "project", nlohmann::json::parse(options.projection_json.value()) });
+        }
+        
+        if (options.sort_json) {
+            base_args.push_back({ "sort", nlohmann::json::parse(options.sort_json.value()) });
+        }
+        
+        auto args = nlohmann::json( {{"arguments", nlohmann::json::array({base_args} ) }} );
+                
+        m_service->call_function("find",
```
ironage, line 122: I'm having trouble finding the docs for this. I'm wondering how this works for large collections because I think you'd typically get a cursor back - do we have to handle that or is this actually returning everything?

src/sync/remote_mongo_collection.cpp
```diff
@@ -18,3 +18,505 @@
...
+}
+
+void RemoteMongoCollection::count(const std::string& filter_json,
+                                  std::function<void(uint64_t, util::Optional<AppError>)> completion_block)
+{
+    count(filter_json, 0, completion_block);
+}
+
+void RemoteMongoCollection::insert_many(std::vector<std::string> documents,
+                                        std::function<void(std::map<uint64_t, std::string>, util::Optional<AppError>)> completion_block)
```
ironage, line 250: we're not really using the features of a map since everything is inserted in order, if there isn't a good reason for it, a std::vector will be more efficient

src/sync/remote_mongo_collection.cpp
```diff
@@ -18,3 +18,505 @@
...
+                                                 const std::string& replacement_json,
+                                                 RemoteMongoCollection::RemoteFindOneAndModifyOptions options,
+                                                 std::function<void(std::string, util::Optional<AppError>)> completion_block)
+{
+    try {
+        auto base_args = m_base_operation_args;
+        base_args.push_back({ "query", nlohmann::json::parse(filter_json) });
+        base_args.push_back({ "update", nlohmann::json::parse(replacement_json) });
+        
+        if (options.upsert) {
```
ironage, line 443: consider moving this logic into `RemoteFindOneAndModifyOptions` (something like `std::vector<nlohmann::json> get_json()`) because that would clean this up and allow you to reuse the logic in several places

src/sync/remote_mongo_collection.cpp
leemaguire: we don't have to handle the cursor, the find will return everything, or if a filter is passed, all documents matching the filter

src/sync/remote_mongo_client.cpp
```diff
@@ -17,3 +17,20 @@
 ////////////////////////////////////////////////////////////////////////////
 
 #include "sync/remote_mongo_client.hpp"
+#include "sync/remote_mongo_database.hpp"
```
cmelchior, line 20: Here and for the `remote_mongo_client.hpp`. It would be nice if we could expose the names as public arguments.

src/sync/app.cpp
```diff
@@ -841,5 +842,53 @@ void App::link_user(std::shared_ptr<SyncUser> user,
...
+    RemoteMongoClient remote_client(std::make_unique<App>(*this));
+    return remote_client;
+}
+
+void App::call_function(const std::string& name,
+                        const std::string& args_json,
+                        std::function<void (util::Optional<AppError>, util::Optional<std::string>)> completion_block) const
+{
+    call_function(name, args_json,
+                  util::Optional<std::string>(m_service_name),
```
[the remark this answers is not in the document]
leemaguire: m_service_name will no longer be optional or part of `App` it has been moved to `AppServiceClient` where it is a non optional

src/sync/app.cpp
leemaguire: im no longer setting the service name here, I've also marked this function as const

src/sync/app.cpp
leemaguire: changed to use dereferencing

src/sync/app_service_client.hpp
leemaguire: AppServiceClient is no longer abstract

src/sync/remote_mongo_collection.cpp
leemaguire: changed to use vector

tests/mongodb/apps/default-app/functions/confirmFunc/source.js
leemaguire: These js files are auto generated by stitch

src/sync/remote_mongo_collection.cpp
leemaguire: changed to stoll

tests/sync/app.cpp
leemaguire: removed message

src/sync/remote_mongo_collection.cpp
leemaguire: removed whitespace

tests/sync/app.cpp
leemaguire: leftover debugging code, I've cleaned it up

tests/mongodb/apps/default-app/functions/confirmFunc/source.js
leemaguire: it was for local testing, I was going around in circles with some of the docker stuff locally

src/sync/auth_request_client.cpp
leemaguire: removed file

src/sync/remote_mongo_collection.cpp
leemaguire: issue is the number is returned as a string e.g. `"{\"$numberLong\":\"19\"}"` so a parse from string to int need to be performed

src/sync/remote_mongo_collection.cpp
leemaguire: I've moved to using a vector instead of map, it makes more sense

src/sync/remote_mongo_collection.cpp
```diff
@@ -18,3 +18,472 @@
...
+    if (value && !error) {
+        return completion_block(*value, error);
+    }
+    
+    return completion_block("", error);
+}
+
+static void handle_response_optional(util::Optional<AppError> error,
+                            util::Optional<std::string> value,
+                            std::function<void(Optional<std::string>, util::Optional<AppError>)> completion_block)
```
ironage, line 37: something seems off that we have these two versions, smells like some API inconsistency if we have completion blocks taking an optional string and a non-optional string (but it can be empty) - can we settle on one version for consistency?

src/sync/remote_mongo_collection.cpp
```diff
@@ -18,3 +18,472 @@
...
+                                   util::Optional<std::string> value,
+                                   std::function<void(RemoteMongoCollection::RemoteUpdateResult, util::Optional<AppError>)> completion_block)
+{
+    if (value && !error) {
+        
+        try {
+            auto json = nlohmann::json::parse(*value);
+            auto matched_count_string = json.at("matchedCount").at("$numberInt").get<std::string>();
+            auto matched_count = std::stoull(matched_count_string);
+            auto modified_count_string = json.at("modifiedCount").at("$numberInt").get<std::string>();
```
ironage, line 95: it would be nice to have constants here instead of hard coding these value strings, can we reuse the constants that @jsflax added recently to support extended bson?

tests/sync/app.cpp
```diff
@@ -824,6 +827,368 @@ TEST_CASE("app: link_user integration", "[sync][app]") {
...
+                              [&](std::string, Optional<app::AppError> error) {
+            CHECK(error);
+        });
+        
+        collection.insert_one(dog_document,
+                              [&](std::string document_json, Optional<app::AppError> error) {
+            CHECK(!error);
+            auto json = nlohmann::json::parse(document_json);
+            auto object_id = json.at("insertedId").at("$oid").get<std::string>();
+            dog_object_id = object_id;
```
ironage, line 892: check that the string isn't empty or something?

tests/sync/app.cpp
```diff
@@ -824,6 +827,368 @@ TEST_CASE("app: link_user integration", "[sync][app]") {
...
+    }
+    
+    SECTION("find") {
+        bool processed = false;
+        
+        collection.find(dog_document,
+                        [&](std::string document_json, Optional<app::AppError> error) {
+            CHECK(!error);
+            auto json = nlohmann::json::parse(document_json);
+            CHECK(json.is_array());
```
ironage, line 916: check expected size of array?

tests/sync/app.cpp
```diff
@@ -824,6 +827,368 @@ TEST_CASE("app: link_user integration", "[sync][app]") {
...
+    });
+    
+    SECTION("insert") {
+        
+        bool processed = false;
+        std::string dog_object_id;
+        
+        collection.insert_one("{\"bad\" : \"value\"}",
+                              [&](std::string, Optional<app::AppError> error) {
+            CHECK(error);
```
ironage, line 884: check that the string response is empty

tests/sync/app.cpp
```diff
@@ -824,6 +827,368 @@ TEST_CASE("app: link_user integration", "[sync][app]") {
...
+        
+        collection.find_one(dog_document,
+                            [&](Optional<std::string> document_json, Optional<app::AppError> error) {
+            CHECK(!error);
+            CHECK(!document_json);
+        });
+        
+        collection.insert_one(dog_document,
+                              [&](std::string, Optional<app::AppError> error) {
+            CHECK(!error);
```
ironage, line 927: check that the string response contains a value

tests/sync/app.cpp
```diff
@@ -824,6 +827,368 @@ TEST_CASE("app: link_user integration", "[sync][app]") {
...
+        });
+        
+        collection.find(dog_document,
+                        [&](std::string document_json, Optional<app::AppError> error) {
+            CHECK(!error);
+            auto json = nlohmann::json::parse(document_json);
+            CHECK(json.is_array());
```
ironage, line 934: check the size of the array

tests/sync/app.cpp
```diff
@@ -824,6 +827,368 @@ TEST_CASE("app: link_user integration", "[sync][app]") {
...
+        
+        realm::app::RemoteMongoCollection::RemoteFindOneAndModifyOptions find_and_modify_options {
+            util::Optional<std::string>(nlohmann::json({{"name", "fido"}}).dump()), //project
+            util::Optional<std::string>(nlohmann::json({{"name", 1}}).dump()), //sort,
+            true, //upsert
+            true // return new doc
+        };
+        
+        collection.find_one_and_delete(dog_document, find_and_modify_options, [&](Optional<app::AppError> error) {
+            CHECK(!error);
```
ironage, line 974: check that the result is valid and has been updated

src/sync/remote_mongo_collection.cpp
leemaguire: For the moment we will keep this current implementation, I do agree constants would be good. In a subsequent PR we will move to using BSON completely which removed the need for the hard coded strings.

src/sync/app.cpp
```diff
@@ -298,16 +264,28 @@ void App::UsernamePasswordProviderClient::call_reset_password_function(const std
...
+                                                           auth_path,
+                                                           user_api_key_provider_key_path));
+}
+
+void App::UserAPIKeyProviderClient::create_api_key(const std::string &name, std::shared_ptr<SyncUser> user,
                                                    std::function<void (UserAPIKey, Optional<AppError>)> completion_block)
 {
-    REALM_ASSERT(m_parent);
-    std::string route = util::format("%1/auth/%2", m_parent->m_base_route, user_api_key_provider_key);
+    std::string route = url_for_path("");
```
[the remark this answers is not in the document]
leemaguire: I have added a default param for this method

src/sync/app_service_client.hpp
```diff
@@ -31,18 +32,44 @@ namespace app {
...
-                               const util::Optional<std::string>& service_name,
-                               std::function<void (util::Optional<AppError>, util::Optional<std::string>)> completion_block) = 0;
+    void call_function(const std::string& name,
+                       const std::string& args_json,
+                       const util::Optional<std::string>& service_name,
+                       std::function<void (util::Optional<AppError>, util::Optional<std::string>)> completion_block) const;
+    
+    /// Calls the Realm Cloud function with the provided name and arguments.
+    /// @param name The name of the Realm Cloud function to be called.
+    /// @param args The `BSONArray` of arguments to be provided to the function.
```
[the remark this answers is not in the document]
leemaguire: fixed

src/sync/remote_mongo_client.hpp
```diff
@@ -33,19 +33,19 @@ class RemoteMongoDatabase;
 class RemoteMongoClient {
 public:
 
-    RemoteMongoClient(std::unique_ptr<AppServiceClient> service) :
-    m_service(std::move(service)) { }
+    RemoteMongoClient(const AppServiceClient service) :
```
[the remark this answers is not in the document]
leemaguire: Changed to using this constructor 👍🏼

src/sync/remote_mongo_collection.cpp
leemaguire: I agree, all strings will have an optional type, so I've moved all calls to be consistent with that

tests/sync/app.cpp
leemaguire: now checking for an object id

tests/sync/app.cpp
leemaguire: now checking for an object id

tests/sync/app.cpp
leemaguire: done 👍🏼

tests/sync/app.cpp
leemaguire: I'm now doing a check to see if the size of the collection is 0 after the delete

tests/sync/app.cpp
leemaguire: done

ironage:
I want to point out there's opportunity to improve test coverage of the error handling code paths. Now that we have [coverage reports](https://ci.realm.io/job/realm/job/realm-object-store/job/PR-971/37/cobertura/___src_sync/___src_sync_remote_mongo_collection_cpp/) again, we could use that as a guide but that's not blocking, more for bonus points.

PR #974, opened 2020-04-16
Rework thread checking and event loop signaling

tgoyne:
Combine these two into a "Scheduler" interface which covers both concepts to support Realms which are confined to a scheduler which isn't an event loop running on a single thread. This also subsumes some of the special-casing for frozen Realms by making them just use a no-op scheduler that doesn't perform thread checking or support notifications. By making Scheduler an abstract type rather than a single per-platform type, we can support multiple types of schedulers per platform, along with user-defined schedulers (which is applicable to Swift's Combine and probably other Rx frameworks).

Add a dispatch-queue based scheduler on Apple platforms which posts notifications to a GCD queue rather than a run loop and confines the Realm to that queue (which may not always be on the same thread).

The AbstractExecutionContextId concept is gone entirely because it was only used for the Realm cache, which was removed in the core 6 port. Other than that and the change to the GenericEventLoopSignal API (that's now done by simply subclassing Scheduler) this should have no functional impact on SDKs that done use the new functionality. The changes in tests are just fixing things that previously were broken; the libuv implementation of EventLoopSignal (and now UvMainLoopScheduler) doesn't have the required locking to support creating instances of it on background threads. Previously we didn't have a way to avoid doing this and mostly just got lucky with the sequencing of the tests that happened to do it.


src/shared_realm.hpp
```diff
@@ -259,7 +262,7 @@ class Realm : public std::enable_shared_from_this<Realm> {
 
     // Get a Realm for the given execution context (or current thread if `none`)
     // from the thread safe reference. May return a cached Realm or create a new one.
```
ironage, line 264: some stale comments

src/util/apple/scheduler.hpp
```diff
@@ -0,0 +1,204 @@
...
+//
+////////////////////////////////////////////////////////////////////////////
+
+#include <realm/util/to_string.hpp>
+
+#include <atomic>
+#include <dispatch/dispatch.h>
+#include <objc/runtime.h>
+#include <pthread.h>
+
```
ironage, line 25: Don't we need include guards?

tgoyne: These aren't really intended to be header files. They're just the platform-specific parts of scheduler.cpp split into separate files for organizational reasons, but compiled as part of scheduler.cpp to avoid complicating the build system (both here and in all the SDKs consuming it). Including them anywhere else would result in duplicated symbols.

tgoyne:
@cmelchior @nhachicha Could one of you look at the Android implementation of this? I think I understand what the old code was doing and why, but it'd be good to have a sanity check.

src/util/apple/scheduler.hpp
ironage: ah I didn't catch that thanks, sounds good

nhachicha:
I merged this branch against the current [hash](https://github.com/realm/realm-object-store/tree/66199adbfffbe153e696309a53d4ec03e32c44e3) we're using for the `next-major` branch in Java and the test failed with 
```
object-store/CMakeFiles/realm-core-6.0.1/include/realm/util/optional.hpp:414: [realm-core-6.0.1] Assertion failed: m_engaged
0   tests                               0x0000000106a5b795 _ZN5realm4utilL18terminate_internalERNSt3__118basic_stringstreamIcNS1_11char_traitsIcEENS1_9allocatorIcEEEE + 37
1   tests                               0x0000000106a5c02d _ZN5realm4util9terminateEPKcS2_lOSt16initializer_listINS0_9PrintableEE + 1213
2   tests                               0x0000000105922c0c _ZN5realm4util8OptionalINS_9VersionIDEE5valueEv + 92
3   tests                               0x0000000105917e8d _ZL29____C_A_T_C_H____T_E_S_T____0v + 957
4   tests                               0x0000000105a21933 _ZNK5Catch21TestInvokerAsFunction6invokeEv + 19
5   tests                               0x0000000105a1a2d6 _ZNK5Catch8TestCase6invokeEv + 38
6   tests                               0x0000000105a1a20a _ZN5Catch10RunContext20invokeActiveTestCaseEv + 42
7   tests                               0x0000000105a1831d _ZN5Catch10RunContext14runCurrentTestERNSt3__112basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEES8_ + 1101
8   tests                               0x0000000105a1762d _ZN5Catch10RunContext7runTestERKNS_8TestCaseE + 637
9   tests                               0x0000000105a1d196 _ZN5Catch12_GLOBAL__N_19TestGroup7executeEv + 294
10  tests                               0x0000000105a1ca0b _ZN5Catch7Session11runInternalEv + 491
11  tests                               0x0000000105a1c7bb _ZN5Catch7Session3runEv + 91
12  tests                               0x0000000105a34d1a _ZN5Catch7Session3runIcEEiiPKPKT_ + 90
13  tests                               0x0000000105a34bee main + 206
14  libdyld.dylib                       0x00007fff6eba2cc9 start + 1
```
I pushed the test branch here https://github.com/realm/realm-object-store/tree/nh/test_scheduler

PR #976, opened 2020-04-17
Merge v10

jsflax:


src/sync/app.cpp
```diff
@@ -337,14 +338,14 @@ void App::UsernamePasswordProviderClient::call_reset_password_function(const std
...
     };
-
-    m_parent->m_config.transport_generator()->send_request_to_server({
-        HttpMethod::post,
-        route,
-        m_parent->m_request_timeout_ms,
-        get_request_headers(user),
-        body.dump()
-    }, handler);
+    Request req = {
```
cmelchior, line 341: Doesn't this use the syntax avoided here? https://github.com/realm/realm-object-store/pull/976/files#diff-e169b24827fc964dc2ab40eef9ee1c45R31

Here and multiple other places

jsflax: Yes, it is unclear to me why this was changed

PR #977, opened 2020-04-20
Reuse existing session for anonymous users

jsflax:


src/sync/app.cpp
```diff
@@ -617,6 +617,17 @@ void App::log_in_with_credentials(const AppCredentials& credentials,
         App::get_profile(linking_user ? linking_user : sync_user, completion_block);
     };
 
+    // if we try logging in with an anonymous user while there
+    // is already an anonymous session active, reuse it
+    if (credentials.provider() == AuthProvider::ANONYMOUS) {
```
cmelchior, line 622: I'm not sure this is a good idea? This makes it impossible to create multiple anonymous users If I understand it correctly. At least I'm doing that in a bunch of places in tests, so I suspect that end-users could also have use cases.

jsflax: This is by design, and is the current functionality on the legacy SDKs.

cmelchior: Hmm, ok I guess 👍 . Modifying the tests to use real users isn't super complicated. I guess the semantics just feel a bit weird to me since it isn't obvious this is the result of logging in twice.

Shouldn't the behaviour be the same for other types as well then? So if you log in twice using the same email/password the same user is returned?

fealebenpae: > Shouldn't the behaviour be the same for other types as well then? So if you log in twice using the same email/password the same user is returned?

This should be the case today. The issue with anonymous is that each time you log in, you get a new random user id, whereas any other credential would return the same id.

PR #982, opened 2020-04-27
Use BSON for remote mongo calls

leemaguire:
This PR integrates BSON in the AppServiceClient and RemoteMongoCollection classes. By using BSON we remove a lot of string to json parsing. This also cleans up the object store API.


src/util/bson/bson.cpp
```diff
@@ -102,12 +102,12 @@ Bson& Bson::operator=(Bson&& v) noexcept {
             new (&string_val) std::string(std::move(v.string_val));
             break;
         case Type::Document:
-            if (document_val) delete document_val;
+//            if (document_val) delete document_val;
```
ironage, line 105: Tricky, we have two cases to consider based on above: `if (m_type != v.m_type) this->~Bson();` 
1) Changing from non-document to Document type; then according to above the memory is cleaned up according to the manual destructor call and we should not again delete memory here. `if (document_val) ...` does not protect us because `document_val` coming from the union will be some random memory and not a `nullptr`.
1) Changing from Document type to again Document type; then the destructor is not called above and we leak memory here if not cleaned up.

Using a `std::unique_ptr` would have simplified this. But in lieu of changing the type, I'd suggest we don't set `m_type` until after the switch statement so that we can properly check which type we're coming from in this code and clean up memory only if necessary.

src/sync/generic_network_transport.hpp
```diff
@@ -46,8 +46,9 @@ enum class JSONErrorCode {
     missing_json_key = 3,
 };
 
-const std::error_category& json_error_category() noexcept;
-std::error_code make_error_code(JSONErrorCode) noexcept;
+enum class BSONErrorCode {
```
ironage, line 49: Why not add this value as a new value inside of the existing `JSONErrorCode`? (non-blocking)

src/sync/remote_mongo_collection.cpp
```diff
@@ -137,37 +200,34 @@ void RemoteMongoCollection::find(const std::string& filter_json,
...
     try {
-        auto base_args = m_base_operation_args;
-        base_args.push_back({"query", nlohmann::json::parse(filter_json)});
-        
+        auto base_args = bson::BsonDocument(m_base_operation_args);
+        base_args["query"] = filter_bson;
+
         if (options.limit) {
-            base_args.push_back({"limit", options.limit.value()});
+            base_args["limit"] = static_cast<int32_t>(*options.limit);
```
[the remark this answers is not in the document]
leemaguire: int32_t was incorrect to use here, I've move it to int64_t

src/sync/generic_network_transport.hpp
leemaguire: Your suggestion is a lot cleaner than having a seperate enum. I've added the error state to JSONErrorCode

jsflax:
@ironage PTAL. I fixed up the BSON implementation to use unique pointers, and cleaned up the service client implementation a bit as well. `call_function` has also been exposed.

src/util/bson/bson.cpp
```diff
@@ -520,106 +523,64 @@ class Parser : public nlohmann::json_sax<json> {
...
-        BsonContainer(BsonArray&& v) noexcept : m_type(ARRAY), array(new BsonArray(std::move(v))) {}
+        BsonContainer(BsonArray&& v) noexcept :
+            m_type(ARRAY),
+            array(new BsonArray(std::move(v))) {}
 
-        explicit operator const BsonDocument&&() const {
+        explicit operator const BsonDocument&() noexcept {
             REALM_ASSERT(is_document());
-            return std::move(*document);
+            return *document.release();
```
ironage, line 558: `release()` will cause a leak because it stops managing the memory, you probably want `get()` or simply `*document`
Same for below instances.

src/sync/app.cpp
```diff
@@ -811,13 +811,58 @@ void App::refresh_access_token(std::shared_ptr<SyncUser> sync_user,
...
+    s << bson::Bson(args);
+
+    Request request {
+        .method = HttpMethod::post,
+        .url = route,
+        .body = s.str()
+    };
+
+    do_authenticated_request(request,
+                             SyncManager::shared().get_current_user(),
```
rorbech, line 849: We need to be able to call functions for a specific user and not only a current sync user, so maybe just pass the user as an argument.

src/sync/app.hpp
```diff
@@ -38,23 +44,32 @@ class RemoteMongoClient;
...
-        realm::util::Optional<std::string> local_app_version;
-        realm::util::Optional<uint64_t> default_request_timeout_ms;
+        util::Optional<std::string> base_url;
+        util::Optional<std::string> local_app_name;
+        util::Optional<std::string> local_app_version;
+        util::Optional<uint64_t> default_request_timeout_ms;
     };
 
     App(const Config& config);
+    App() = default;
```
jsflax: Why do we need this constructor?

fealebenpae: You tell me 😝 
https://github.com/realm/realm-object-store/pull/942/files#diff-08c2f8b03a8e4ef38c589102149ed023R59

jsflax: I was but a young padawan back then.

leemaguire: removed

src/sync/app.cpp
leemaguire: I've added the ability to pass a user in call_function

src/util/bson/bson.cpp
leemaguire: moved to just dereferencing document

src/sync/app.cpp
```diff
@@ -258,13 +258,13 @@ void App::UsernamePasswordProviderClient::call_reset_password_function(const std
     auto handler = [completion_block](const Response& response) {
         handle_default_response(response, completion_block);
     };
-
+
```
tgoyne, line 261: Stray trailing whitespace

src/sync/remote_mongo_collection.cpp
```diff
@@ -21,461 +21,411 @@
...
+void RemoteMongoCollection::find(const bson::BsonDocument& filter_bson,
                                  RemoteFindOptions options,
-                                 std::function<void(Optional<std::string>, util::Optional<AppError>)> completion_block)
+                                 std::function<void(util::Optional<bson::BsonArray>, util::Optional<AppError>)> completion_block)
 {
     try {
-        auto base_args = m_base_operation_args;
-        base_args.push_back({"query", nlohmann::json::parse(filter_json)});
-        
+        auto base_args = bson::BsonDocument(m_base_operation_args);
```
tgoyne, line 77: Shouldn't m_base_operation_args just be a BsonDocument rather than converting it to one each time?

src/sync/remote_mongo_collection.cpp
```diff
@@ -21,461 +21,411 @@
...
+        base_args["query"] = filter_bson;
+
         if (options.limit) {
-            base_args.push_back({"limit", options.limit.value()});
+            base_args["limit"] = static_cast<int32_t>(*options.limit);
```
tgoyne, line 81: Is it intentional that options.limit is sometimes int32_t and sometimes int64_t?

src/sync/remote_mongo_collection.cpp
```diff
@@ -21,461 +21,411 @@
...
+                                 [completion_block](util::Optional<AppError> error, util::Optional<bson::Bson> value) {
+            if (error) {
+                return completion_block(util::none, error);
+            }
 
-        m_service.call_function("find",
-                                args.dump(),
-                                [completion_block](util::Optional<AppError> error, util::Optional<std::string> value) {
-            handle_response(error, value, completion_block);
+            return completion_block(util::some<bson::BsonArray>(static_cast<bson::BsonArray>(*value)), util::none);
```
tgoyne, line 100: Are both the `util::some<bson::BsonArray>()` and `static_cast<bson::BsonArray>()` actually needed here? I would expect it to only need one or the other.

src/sync/remote_mongo_collection.cpp
```diff
@@ -21,461 +21,411 @@
...
+                             bson::BsonArray({base_args}),
+                             m_service_name,
+                             [completion_block](util::Optional<AppError> error, util::Optional<bson::Bson> value) {
+        if (error) {
+            return completion_block({}, error);
+        }
+
+        auto bson = static_cast<bson::BsonDocument>(*value);
+        auto inserted_ids = static_cast<bson::BsonArray>(bson["insertedIds"]);
+        auto oid_array = std::vector<ObjectId>();
```
tgoyne, line 248: `std::vector<ObjectId>(inserted_ids.begin(), inserted_ids.end())` is probably sufficient to populate this vector.

src/util/bson/bson.cpp
```diff
@@ -29,34 +29,39 @@ Bson::~Bson() noexcept
...
+        case Type::Array:
+            array_val.reset();
+            break;
         default:
             break;
     }
 }
 
-Bson::Bson(const Bson& v) {
+Bson::Bson(const Bson& v) noexcept {
```
tgoyne, line 49: This and the copy-assignment operator can throw exceptions as they allocate memory.

src/util/bson/bson.cpp
```diff
@@ -520,106 +523,64 @@ class Parser : public nlohmann::json_sax<json> {
...
         }
 
         bool is_array() const { return m_type == ARRAY; }
         bool is_document() const { return m_type == DOCUMENT; }
 
-        void push_back(std::pair<std::string, Bson> value) {
+        void push_back(const std::string& key, Bson&& value) {
             if (m_type == DOCUMENT) {
-                (*document)[value.first] = value.second;
+                document->operator[](key) = std::move(value);
```
tgoyne, line 581: Explicitly calling operator[] is very unusual and the previous version of this code was more idiomatic.

src/util/bson/bson.cpp
```diff
@@ -520,106 +523,64 @@ class Parser : public nlohmann::json_sax<json> {
...
-                document = v.document;
-                v.document = nullptr;
-            } else {
-                array = v.array;
-                v.array = nullptr;
-            }
-            return *this;
+        explicit operator BsonDocument&() noexcept {
+            REALM_ASSERT(is_document());
+            return *document.release();
```
tgoyne, line 568: `release()` is almost certainly not the correct function to call here.

src/util/bson/bson.hpp
```diff
@@ -314,27 +314,31 @@ inline Bson::Bson(ObjectId v) noexcept
 }
 
 inline Bson::Bson(const IndexedMap<Bson>& v) noexcept
+:
```
tgoyne, line 317: This formatting is strange and different.

src/sync/remote_mongo_collection.cpp
jsflax: No, changing.

src/sync/remote_mongo_collection.cpp
jsflax: Yes.

src/sync/app.cpp
```diff
@@ -873,13 +873,88 @@ void App::refresh_access_token(std::shared_ptr<SyncUser> sync_user,
...
+}
 
 RemoteMongoClient App::remote_mongo_client(const std::string& service_name)
 {
-    return RemoteMongoClient(AppServiceClient(service_name,
-                                              m_base_route,
-                                              m_config.app_id,
-                                              *this));
+    // FIXME: Use shared_from_this once enabled
+    return RemoteMongoClient(std::static_pointer_cast<AppServiceClient>(std::make_shared<App>(m_config)),
```
cmelchior, line 956: This looks wrong. It should be `SyncManager::shared().app()` Since the App is still a singleton (found this when debugging with @edualonso )

Not sure if this is done other places

jsflax: That's why the `FIXME` is here. Is it causing a crash on Java?

edualonso: Yes it is.

jsflax: I was hoping to avoid having to make this change in this PR, but I will fix this later today.

src/sync/app.cpp
```diff
@@ -81,8 +81,13 @@ const static uint64_t    default_timeout_ms = 60000;
...
 const static std::string user_api_key_provider_key_path = "api_keys";
 
+SharedApp App::get_shared_app(const Config& config)
+{
+    return std::make_shared<App>(config);
+}
+
 App::App(const Config& config)
-: m_config(config)
+: m_config(std::move(config))
```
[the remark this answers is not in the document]
tgoyne: std::move() here is actually just a no-op. It results in a `const Config&&` and `m_config` will copy it.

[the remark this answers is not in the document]
ironage: Oh you're right! Thanks for clarifying, no harm then.
I guess I'd still like to change it on the grounds that it is confusing to read given it's actually not actually moving anything.

PR #983, opened 2020-04-27
Add app metadata discovery step

jsflax:


src/sync/app.cpp
```diff
@@ -710,23 +710,69 @@ std::string App::url_for_path(const std::string& path="") const
...
+                                   std::function<void (Response)> completion_block)
 {
-    auto handler = [completion_block, request, sync_user, this](const Response& response) {
-        if (auto error = check_for_errors(response)) {
-            App::handle_auth_failure(error.value(), response, request, sync_user, completion_block);
-        } else {
-            completion_block(response);
+    init_app_metadata([completion_block, &request, sync_user, this](const util::Optional<AppError> error){
+        if (error) {
+            return completion_block(Response{.http_status_code = 400});
```
fealebenpae, line 759: Shouldn't we return the real error here, instead of hiding it?

src/sync/app.cpp
```diff
@@ -710,23 +710,69 @@ std::string App::url_for_path(const std::string& path="") const
     return util::format("%1%2", m_base_route, path);
 }
 
+void App::init_app_metadata(std::function<void (util::Optional<AppError>)> completion_block)
```
ironage, line 713: Consider leaving this init to be called by SDK's manually, then they can handle init errors separately and we can leave all the other app methods as const.

jsflax: The problem is, it can break GDPR for some apps if this is not proper initialised.

ironage: I don't disagree that it has to be done, I'm more referring to the design; lazy-loading config from all request sites. You could still force users to call an `App::init` by refusing to make other requests until init has been successful, but this might give more flexibility to users. In the happy path it's nice to hide this request away, but reasoning about the state of things when init fails could be difficult if the first public find or insert request fails with an error about metadata. I guess I liked the property that requests did not modify the App state, but it's not a blocker that we retain that property.

fealebenpae: I think I like @ironage's suggestion - this would mean that all the other methods won't need to be spaghetti-callback'd because they need to call `init_app_metadata` first and then their own asynchronous code. Instead we can just plop an assert to make sure the SDKs have initialized the metadata before calling anything else.

jsflax: Yes but `init_app_metadata` is asynchronous itself, which is going to create a lot of state management SDK side. You cannot asynchronously initialise an App, and there are multiple possible entry points for a user, which creates a much bigger mess in each SDK.

ironage: What if the SDK's don't enforce initialization, instead putting the requirement onto the users? The init method could be publicly exposed so that users must make the call themselves and wait for its results, in this case we could make all other requests return an `AppError::not_initialized` immediately until it has been successfully initialized.

jsflax: We had a release of the legacy SDKs that did that, and it ends up being a bit of a mess. It forced users to asynchronously initialise their Stitch App class, which was not great UX.

ironage: Alright, let's keep the lazy-load init. I can see how moving complexity to the user would create more of a mess.

jsflax: Cool. I agree that this isn't ideal though– let's revisit at a a later date.

PR #986, opened 2020-04-28
Update Core and Sync dependencies

jedelbo:


jedelbo:
@cmelchior Can you tell if `ProtocolError::user_mismatch` is handled in the right place?

src/sync/sync_session.cpp
```diff
@@ -520,6 +520,7 @@ void SyncSession::handle_error(SyncError error)
             case ProtocolError::unsupported_session_feature:
             case ProtocolError::transact_before_upload:
             case ProtocolError::partial_sync_disabled:
+            case ProtocolError::user_mismatch:
```
cmelchior, line 523: I moved this error handling. In the previous location, it would result in the Realm file being deleted. From reading the description in Sync this error seems more like a connection error, so retrying with a correct user would be the correct solution.

PR #987, opened 2020-04-29
Update to Sync alpha.9

cmelchior:


cmelchior:
This is fixed by https://github.com/realm/realm-object-store/pull/986

PR #997, opened 2020-05-07
Fix BSON argument encoding for `call_reset_password_function`

rorbech:


src/sync/app.cpp
```diff
@@ -264,18 +264,21 @@ void App::UsernamePasswordProviderClient::call_reset_password_function(const std
...
+    std::stringstream body;
+    body << bson::Bson(arg);
+
     m_parent->m_config.transport_generator()->send_request_to_server({
         HttpMethod::post,
         route,
         m_parent->m_request_timeout_ms,
         get_request_headers(),
-        body.dump()
+        body.str()
```
cmelchior, line 281: Isn't there a test that would have caught this?

jsflax: It's possible the test was disabled for this before BSON was done, or something in the test was hardcoded where it should not have been.

PR #999, opened 2020-05-07
Update core-6 branch from master

ironage:


tgoyne:
This appears to be missing all of the tests from 5ffe553c44a3a5c24accc3505aec33f60075051d

ironage:
Thanks for catching those missing tests, I deleted them when they didn't compile thinking they weren't relevant anymore, so that's my bad. I've ported them over now.

PR #1001, opened 2020-05-08
Add auth options to log_in_with_credential

leemaguire:
This PR will address sending a set of `options` in the body of the `log_in_with_credential`. The `options` in this PR just contain a `deviceId` parameter, which will be needed for services such as push notifications. 

Consumers of the Object Store will pass the deviceId into `App::Config`. Once a user has successfully logged in the deviceId will be passed to the `SyncUser`.

### DIFF
--- a/src/sync/app.hpp
+++ b/src/sync/app.hpp
@@ -55,6 +55,7 @@ public:
         util::Optional<std::string> local_app_name;
         util::Optional<std::string> local_app_version;
         util::Optional<uint64_t> default_request_timeout_ms;
+        util::Optional<std::string> device_id;
     };
 
     // `enable_shared_from_this` is unsafe with public constructors; use `get_shared_app` instead
@@ -246,7 +247,8 @@ public:
     /// @param completion_block The completion handler to call when the linking is complete.
     ///                         If the operation is  successful, the result will contain the original
     ///                         `SyncUser` object representing the user.
-    void link_user(std::shared_ptr<SyncUser> user, const AppCredentials& credentials,
+    void link_user(std::shared_ptr<SyncUser> user,
+                   const AppCredentials& credentials,
                    std::function<void(std::shared_ptr<SyncUser>, util::Optional<AppError>)> completion_block);
 
     /// Switches the active user with the specified one. The user must
--- a/src/sync/app.hpp
+++ b/src/sync/app.hpp
@@ -426,6 +426,7 @@ private:
                                  const std::shared_ptr<SyncUser> linking_user,
                                  std::function<void(std::shared_ptr<SyncUser>, util::Optional<AppError>)> completion_block);
     
+    /// Provides Realm Cloud with metadata related to the users session
     void attach_auth_options(bson::BsonDocument& body);
 
 };
--- a/src/sync/app.hpp
+++ b/src/sync/app.hpp
@@ -426,7 +426,7 @@ private:
                                  const std::shared_ptr<SyncUser> linking_user,
                                  std::function<void(std::shared_ptr<SyncUser>, util::Optional<AppError>)> completion_block);
     
-    /// Provides Realm Cloud with metadata related to the users session
+    /// Provides MongoDB Realm Cloud with metadata related to the users session
     void attach_auth_options(bson::BsonDocument& body);
 
 };
--- a/src/sync/app.hpp
+++ b/src/sync/app.hpp
@@ -423,6 +423,8 @@ private:
     void log_in_with_credentials(const AppCredentials& credentials,
                                  const std::shared_ptr<SyncUser> linking_user,
                                  std::function<void(std::shared_ptr<SyncUser>, util::Optional<AppError>)> completion_block);
+    
+    void attach_auth_options(bson::BsonDocument& body);
 
 };
 
--- a/src/sync/app.hpp
+++ b/src/sync/app.hpp
@@ -55,7 +55,9 @@ public:
         util::Optional<std::string> local_app_name;
         util::Optional<std::string> local_app_version;
         util::Optional<uint64_t> default_request_timeout_ms;
-        util::Optional<std::string> device_id;
+        util::Optional<std::string> platform;
+        util::Optional<std::string> platform_version;
+        std::string sdk_version;
     };
 
     // `enable_shared_from_this` is unsafe with public constructors; use `get_shared_app` instead
@@ -424,7 +426,7 @@ private:
                                  const std::shared_ptr<SyncUser> linking_user,
                                  std::function<void(std::shared_ptr<SyncUser>, util::Optional<AppError>)> completion_block);
     
-    void attach_auth_options(bson::BsonDocument& body);
+    void attach_auth_options(bson::BsonDocument& body, std::shared_ptr<SyncUser> sync_user);
 
 };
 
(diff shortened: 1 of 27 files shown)


src/sync/app.hpp
```diff
@@ -55,6 +55,7 @@ class App : public std::enable_shared_from_this<App>, public AuthRequestClient,
         util::Optional<std::string> local_app_name;
         util::Optional<std::string> local_app_version;
         util::Optional<uint64_t> default_request_timeout_ms;
+        util::Optional<std::string> device_id;
```
jsflax, line 58: This is not the correct place for this– the other device fields (sdk, platform, etc) need to be on here (the config), but device id is actually associated with the user.

src/sync/sync_manager.cpp
```diff
@@ -56,6 +56,7 @@ void SyncManager::configure(SyncClientConfig config, util::Optional<app::App::Co
         std::string provider_type;
         std::vector<SyncUserIdentity> identities;
         SyncUser::State state;
+        util::Optional<std::string> device_id;
```
jsflax, line 59: This should not be optional.

src/sync/app.hpp
leemaguire: changed

src/sync/sync_manager.cpp
leemaguire: changed

tests/sync/app.cpp
```diff
@@ -1544,7 +1646,10 @@ static const std::string good_access_token2 =  "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpX
 std::string UnitTestTransport::access_token = good_access_token;
 
 static const std::string bad_access_token = "lolwut";
+static const std::string dummy_device_id = "123400000000000000000000";
+static const std::string app_name = "django";
 
+const std::string UnitTestTransport::app_name = "django";
```
ironage, line 1652: why do we need two app_name variables with the same values?

src/sync/impl/sync_metadata.cpp
```diff
@@ -563,6 +574,18 @@ void SyncUserMetadata::set_access_token(util::Optional<std::string> user_token)
     m_realm->commit_transaction();
 }
 
+void SyncUserMetadata::set_device_id(util::Optional<std::string> device_id)
```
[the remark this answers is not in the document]
jsflax: Agreed with @ironage , this was a mistake in the set_access_token impl as well.

src/sync/sync_user.cpp
```diff
@@ -105,12 +105,14 @@ SyncUser::SyncUser(std::string refresh_token,
...
                    std::string access_token,
-                   SyncUser::State state)
+                   SyncUser::State state,
+                   const std::string device_id)
 : m_state(state)
 , m_provider_type(provider_type)
 , m_refresh_token(RealmJWT(std::move(refresh_token)))
 , m_identity(std::move(identity))
 , m_access_token(RealmJWT(std::move(access_token)))
+, m_device_id(std::move(device_id))
```
jsflax, line 115: I'm not sure this `move` actually does anything.

tests/sync/sync_manager.cpp
```diff
@@ -117,19 +118,19 @@ TEST_CASE("sync_manager: user state management", "[sync]") {
...
-        SyncManager::shared().get_user(identity_1, r_token_1, a_token_1, url_2);
-        SyncManager::shared().get_user(identity_1, r_token_1, a_token_1, url_3);
-        SyncManager::shared().get_user(identity_1, r_token_1, a_token_1, url_1); // existing
+        SyncManager::shared().get_user(identity_1, r_token_1, a_token_1, url_1, dummy_device_id);
+        SyncManager::shared().get_user(identity_1, r_token_1, a_token_1, url_2, dummy_device_id);
+        SyncManager::shared().get_user(identity_1, r_token_1, a_token_1, url_3, dummy_device_id);
+        SyncManager::shared().get_user(identity_1, r_token_1, a_token_1, url_1, dummy_device_id); // existing
         auto users = SyncManager::shared().all_users();
         REQUIRE(users.size() == 3);
         CHECK(validate_user_in_vector(users, identity_1, url_1, r_token_1, a_token_1));
```
jsflax, line 136: `validate_user_in_vector` should be changed to accept device_id, which should then be validated. This should be changed everywhere in this file.

tests/sync/user.cpp
```diff
@@ -100,7 +101,7 @@ TEST_CASE("sync_user: SyncManager `get_existing_logged_in_user()` API", "[sync]"
     }
 
     SECTION("properly returns an existing logged-in user") {
-        auto first = SyncManager::shared().get_user(identity, refresh_token, access_token, server_url);
+        auto first = SyncManager::shared().get_user(identity, refresh_token, access_token, server_url, dummy_device_id);
```
jsflax, line 104: We should validate device_id here.

src/sync/impl/sync_metadata.cpp
```diff
@@ -97,7 +98,8 @@ realm::Schema make_schema()
             {c_sync_access_token, PropertyType::String|PropertyType::Nullable},
             {c_sync_identities, PropertyType::Object|PropertyType::Array, c_sync_identityMetadata},
             {c_sync_profile, PropertyType::Object|PropertyType::Nullable, c_sync_profile},
-            {c_sync_state, PropertyType::Int}
+            {c_sync_state, PropertyType::Int},
+            {c_sync_device_id, PropertyType::String}
```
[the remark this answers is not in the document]
leemaguire: Schema updated to version 4, device_id will not be Optional anymore, so I don't think Nullable will be needed

tests/sync/app.cpp
leemaguire: Removed the secondary app_name

src/sync/impl/sync_metadata.cpp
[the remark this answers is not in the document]
leemaguire: Optional has been removed

src/sync/sync_user.cpp
leemaguire: Move, removed

tests/sync/sync_manager.cpp
leemaguire: Updated this method in tests

tests/sync/user.cpp
leemaguire: done

src/sync/sync_user.cpp
```diff
@@ -325,6 +328,18 @@ std::string SyncUser::access_token() const
...
+    std::lock_guard<std::mutex> lock(m_mutex);
+    return m_device_id;
+}
+
+bool SyncUser::has_device_id() const
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
+    return !m_device_id.empty() && m_device_id != "000000000000000000000000";
+}
+
```
ironage, line 342: Should we be providing a way to update the device id as well, or is that missing by design so that people can't update it once a SyncUser has been constructed?

jsflax: This is by design, correct.

src/sync/app.hpp
```diff
@@ -421,6 +425,9 @@ class App : public std::enable_shared_from_this<App>, public AuthRequestClient,
     void log_in_with_credentials(const AppCredentials& credentials,
                                  const std::shared_ptr<SyncUser> linking_user,
                                  std::function<void(std::shared_ptr<SyncUser>, util::Optional<AppError>)> completion_block);
+    
+    /// Provides Realm Cloud with metadata related to the users session
```
kneth, line 429: "Realm Cloud" -> "MongoDB Realm Cloud"

leemaguire: Updated.

PR #1002, opened 2020-05-11
Fix doubly nested arrays in bson parser

jsflax:


jsflax:
@rorbech The parser uses in internal stack of instructions that is pushed to depending on the method called by the internal SAX parser.

We were not pushing an instruction for `start_array` (the beginning of an array)– this was the bug. When the final `end_object` method was called (the instruction from the internal SAX parser denoting the end of a JSON object), and we went to pop our `array` instruction, there was no instruction to pop, and we tried to pop on an empty stack, causing the crash. In fact, there were two missing instructions– one for the start of each array.

The reason it crashed completely for your `bool` case versus your `string` case is that they are handled slightly differently instruction wise– the `string` method recognised not to pop its parent `array` instruction. Because it had one extra instruction, the parser did not crash, but it did pop one extra value (your final int).

PR #1003, opened 2020-05-12
Fix close triggers reopening a Realm

cmelchior:
We have a unit test in Java that caught this https://github.com/realm/realm-java/pull/6838

I tried to create a similar test in Object Store, but so far failed, so I still haven't ruled out this is a bug in the Java SDK. But I thought I would make a PR nonetheless because no matter what, the ObjectStore code looks fishy. `transaction()` will attempt to create a ReadTransaction if one isn't already open, and doing that inside `close()` is definitely wasteful.

### DIFF
--- a/src/shared_realm.cpp
+++ b/src/shared_realm.cpp
@@ -912,7 +912,7 @@ void Realm::close()
         m_coordinator->unregister_realm(this);
     }
     if (!m_config.immutable() && m_group) {
-        dynamic_cast<Transaction&>(*m_group).close();
+        transaction().close();
     }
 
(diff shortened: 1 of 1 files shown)


tgoyne:
The check for m_group makes sense, but it should either be a static_cast or still a call to transaction() after doing the check.

cmelchior:
I don't have any strong opinions, but clang-tidy had a warning that `dynamic_cast` was prefered when Transaction is a subclass of Group. Just wondering what the guideline is between choosing static_cast vs. dynamic_cast?

PR #1005, opened 2020-05-13
Add proper timestamp and datetime support

jsflax:


src/util/bson/bson.cpp
```diff
@@ -829,8 +832,10 @@ bool Parser::string(string_t& val) {
             m_instructions.push({State::Skip});
             break;
         case State::Date: {
-            auto epoch = atol(val.data());
-            m_marks.top().push_back(instruction.key, Datetime(epoch));
+            int64_t millis_since_epoch = atoll(val.data());
+            auto ts = realm::Timestamp(millis_since_epoch/1000,
```
RedBeard0531, line 836: This isn't correct for negative `millis_since_epoch`. Consider when `millis_since_epoch` is -1005. This logic would give `{-1, -5}`, when you really want `{-2, 995}`.

src/util/bson/bson.cpp
```diff
@@ -739,31 +742,31 @@ bool Parser::number_unsigned(number_unsigned_t val) {
...
             } else {
-                m_marks.top().push_back(instruction.key, Timestamp(0, 1));
+                m_marks.top().push_back(instruction.key, MongoTimestamp(0, val));
                 instruction.type = State::Timestamp;
             }
             break;
         case State::TimestampT:
             if (m_marks.top().size() && m_marks.top().back().first == instruction.key) {
-                auto ts = (Timestamp)m_marks.top().back().second;
+                auto ts = static_cast<MongoTimestamp>(m_marks.top().back().second);
```
RedBeard0531, line 760: FYI, you can just say `MongoTimestamp(...)` without the static cast

src/util/bson/bson.hpp
```diff
@@ -71,8 +71,8 @@ class Bson {
     Bson(double) noexcept;
     Bson(MinKey) noexcept;
     Bson(MaxKey) noexcept;
-    Bson(Datetime) noexcept;
-    Bson(Timestamp) noexcept;
+    Bson(MongoTimestamp) noexcept;
+    Bson(realm::Timestamp) noexcept;
```
RedBeard0531, line 75: `/// Produces a mongo Date, not a Timestamp`

src/util/bson/bson.hpp
```diff
@@ -145,16 +145,16 @@ class Bson {
...
+    explicit operator MongoTimestamp() const
     {
-        REALM_ASSERT(m_type == Bson::Type::Datetime);
-        return date_val;
+        REALM_ASSERT(m_type == Bson::Type::Timestamp);
+        return time_val;
     }
 
-    explicit operator Timestamp() const
+    explicit operator realm::Timestamp() const
```
RedBeard0531, line 154: `/// Accessor for mongo Dates, not mongo Timestamps.`

src/util/bson/mongo_timestamp.hpp
```diff
@@ -24,17 +24,31 @@
...
+        return m_seconds;
+    }
+
+    int64_t increment() const
+    {
+        return m_increment;
+    }
+private:
+    friend bool inline operator==(const MongoTimestamp&, const MongoTimestamp&);
+    int64_t m_seconds;
```
RedBeard0531, line 42: These should be `int32_t`

src/util/bson/mongo_timestamp.hpp
```diff
@@ -24,17 +24,31 @@
 namespace realm {
 namespace bson {
 
-struct Datetime {
-    std::time_t seconds_since_epoch;
-    Datetime(time_t epoch);
+struct MongoTimestamp {
+
+    MongoTimestamp(const int64_t seconds, const int64_t increment);
```
RedBeard0531, line 29: There isn't really a reason to put const on non-reference arguments like that. It has no effect when in the declaration. It has some effect on the definition (and they don't have to match in top-level const), but it is rarely useful, so rarely used.

src/util/bson/bson.cpp
RedBeard0531: Looks like this is actually the style of input that `realm::Timestamp` requires.

RedBeard0531:
Seems odd to approve after adding my commits on top but ¯\\_(ツ)_/¯

I think we can address some usability improvements after we get the basic correct functionality in.

jsflax:
I suppose I'll let you take the reins on this one then?

PR #1006, opened 2020-05-13
Nested type tests, remove Optional from access_token & refresh_token in metadata

leemaguire:
This PR focuses on a general cleanup. What is addressed is some documentation in RemoteMongoClient, tests in bson.cpp & app.cpp and the removal of Optional from access_token & refresh_token in sync_metadata.cpp


src/sync/sync_user.cpp
```diff
@@ -214,11 +214,11 @@ void SyncUser::update_access_token(std::string token)
             case State::Removed:
                 return;
             case State::LoggedIn:
-                m_access_token = token;
+                m_access_token = std::move(token);
                 break;
             case State::LoggedOut: {
                 sessions_to_revive.reserve(m_waiting_sessions.size());
-                m_access_token = token;
+                m_access_token = std::move(token);
```
ironage, line 221: same situation here with using a moved from variable below

src/sync/sync_user.cpp
```diff
@@ -175,11 +175,11 @@ void SyncUser::update_refresh_token(std::string token)
             case State::Removed:
                 return;
             case State::LoggedIn:
-                m_refresh_token = token;
+                m_refresh_token = std::move(token);
                 break;
             case State::LoggedOut: {
                 sessions_to_revive.reserve(m_waiting_sessions.size());
-                m_refresh_token = token;
+                m_refresh_token = std::move(token);
```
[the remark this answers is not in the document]
jsflax: I would also change `update_refresh_token(std::string token)` to be `update_refresh_token(std::string& token)` since the only place this is called already receives a copy of the token.

src/sync/sync_user.cpp
jsflax: I would also consider moving the `std::move` to within the `RealmJWT` constructor, and call it explicitly here:
`m_access_token = RealmJWT(token)`

leemaguire: `token` is now only used to construct a `RealmJWT`, anything that needs to access `token` now just used `m_access_token.token` directly

`std::move` has also moved into the constructor of RealmJWT

src/sync/sync_user.cpp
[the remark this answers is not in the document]
leemaguire: `update_refresh_token(std::string token)` has been changed to `update_refresh_token(std::string&& token)` as `SyncManager::get_user` starts the move semantic meta-game and is the first source to call `update_refresh_token `

src/sync/impl/sync_metadata.hpp
```diff
@@ -66,11 +66,11 @@ class SyncUserMetadata {
     std::vector<realm::SyncUserIdentity> identities() const;
     void set_identities(std::vector<SyncUserIdentity>);
 
-    util::Optional<std::string> refresh_token() const;
-    void set_refresh_token(util::Optional<std::string>);
+    std::string refresh_token() const;
+    void set_refresh_token(const std::string&);
```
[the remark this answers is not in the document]
leemaguire: I've added var names, it seems that a few other functions in `sync_metadata.hpp` are missing var names also?

src/sync/sync_user.cpp
```diff
@@ -55,14 +55,15 @@ static std::vector<std::string> split_token(const std::string& jwt) {
     return parts;
 }
 
-RealmJWT::RealmJWT(const std::string& token)
+RealmJWT::RealmJWT(std::string& token)
```
ironage, line 58: This is tricky because callers may pass in their token variable without explicitly moving it and have the contents moved unknowingly. Accepting a variable by reference does technically have the right to modify the variable in any way, but it is usually better to be explicit about having a `f(T&& v)` so that callers can opt in to this efficient no-copy function if they don't intend to use the variable later. It is not uncommon to implement two versions of the constructor though, one that moves and one that copies so users have the flexibility to choose by being explicit about `std::move`. So I'd recommend we have two constructors `RealmJWT(std::string&& token)` and `RealmJWT(const std::string& token)` which will make a copy when assigned to `this->token`. I'm being pedantic here because it can lead to some hard to track down bugs if memory is moved around without explicitly using `std::move` at the call site, because the code reads like it is correct until you look at the function definition and then you'd have to track down all calls to the function and verify they are using it correctly.

leemaguire: I've chosen to only have `RealmJWT(std::string&& token)` I don't think there is any need to have 2 constructors at the moment considering there are very few codepaths initialising a `RealmJWT`

PR #1007, opened 2020-05-18
fix building on windows with vs2019

blagoev:
Fixes building on windows with vs2019

### DIFF
--- a/src/sync/app.cpp
+++ b/src/sync/app.cpp
@@ -332,12 +332,12 @@ void App::UserAPIKeyProviderClient::create_api_key(const std::string &name, std:
     nlohmann::json body = {
         { "name", name }
     };
-    Request req = {
-        .method = HttpMethod::post,
-        .url = route,
-        .body = body.dump(),
-        .uses_refresh_token = true
-    };
+    Request req;
+    req.method = HttpMethod::post;
+    req.url = route;
+    req.body = body.dump();
+    req.uses_refresh_token = true;
+    
     m_auth_request_client.do_authenticated_request(req, user, handler);
 }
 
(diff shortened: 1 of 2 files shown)


src/util/event_loop_dispatcher.hpp
```diff
@@ -54,11 +54,11 @@ class EventLoopDispatcher<void(Args...)> {
     EventLoopDispatcher(std::function<void(Args...)> func)
     : m_state(std::make_shared<State>(std::move(func)))
     {
-        m_scheduler->set_notify_callback([state = m_state] {
+        m_scheduler->set_notify_callback([state = m_state, this] {
```
fealebenpae, line 57:
```suggestion
        m_scheduler->set_notify_callback([state = m_state] {
```

src/util/event_loop_dispatcher.hpp
```diff
@@ -54,11 +54,11 @@ class EventLoopDispatcher<void(Args...)> {
...
             std::unique_lock<std::mutex> lock(state->mutex);
             while (!state->invocations.empty()) {
                 auto& tuple = state->invocations.front();
-                _apply_polyfill::apply(std::move(tuple), state->func);
+                std::apply(m_state->func, std::move(tuple));
```
fealebenpae, line 61:
```suggestion
                std::apply(state->func, std::move(tuple));
```

src/util/event_loop_dispatcher.hpp
blagoev: good catch. It seems this file was not merged correctly.

src/sync/app.cpp
```diff
@@ -910,13 +909,13 @@ void App::call_function(std::shared_ptr<SyncUser> user,
...
 
-    Request request {
-        .method = HttpMethod::post,
-        .url = route,
-        .body = s.str()
-    };
+    Request req;
+    req.method = HttpMethod::post;
+    req.url = route;
+    req.body = s.str();
```
ironage, line 915: It looks like this change caused build failures in the app tests on the v10 branch. The previous way that the Request was initialized would zero initialize the omitted members, namely `uses_refresh_token` should be false here. The linux app tests did fail on this PR in CI. I can see that this error was obscured by a separate windows issue, but please be careful when merging PR's which fail CI.

blagoev: sorry about that

PR #1010, opened 2020-05-19
switch to dom-style json parsing

RedBeard0531:


tests/bson.cpp
```diff
@@ -259,6 +259,56 @@ TEST_CASE("canonical_extjson_corpus", "[bson]") {
...
+        SECTION("Nested Objects") {
+            run_corpus<BsonDocument>("x", {
+                R"({"x": {"value": {"hello": "world", "_id": {"$oid": "5ec38e1e693f9e61e968f701"}}}})",
+                [](auto val) {
+                    CHECK(val.size() == 1);
+                    auto inner = val["value"].operator const BsonDocument&();
+                    CHECK(inner.size() == 2);
+                    CHECK(inner["hello"].operator const std::string&() == "world");
+                    CHECK(inner["_id"].operator ObjectId() == ObjectId("5ec38e1e693f9e61e968f701"));
+                    return true;
```
RedBeard0531, line 271: Was there a reason the other tests were returning a bool rather than using  `CHECK`

src/util/bson/mongo_timestamp.hpp
```diff
@@ -19,39 +19,27 @@
...
 namespace realm {
 namespace bson {
 
 struct MongoTimestamp {
+    MongoTimestamp(uint32_t seconds, uint32_t increment) :seconds(seconds), increment(increment) {}
 
-    MongoTimestamp(const int64_t seconds, const int64_t increment);
-    
-    int64_t seconds_since_epoch() const
+    friend bool inline operator==(const MongoTimestamp& lhs, const MongoTimestamp& rhs)
```
RedBeard0531, line 30: This is the currently recommended way to define these operators. They are called "hidden friends", declared as friends inside the class, and never declared outside of the class, at least not in headers. The main advantage over declaring the outside of the class is that the compiler won't look to see if there is any conversion to MongoTimestamp every time it tries to resolve a usage of `==`. If you've ever seen rediculously long lists of `operator==` functions whenever you have a typo that makes it not resolve, that is because it looks at Every Single One to see if it matches and lists them when none do. This technique keeps is out of that list.

src/util/bson/mongo_timestamp.hpp
```diff
@@ -19,39 +19,27 @@
...
-bool inline operator==(const MongoTimestamp& lhs, const MongoTimestamp& rhs)
-{
-    return lhs.m_seconds == rhs.m_seconds && lhs.m_increment == rhs.m_increment;
-}
 
-bool inline operator!=(const MongoTimestamp& lhs, const MongoTimestamp& rhs)
-{
-    return !(lhs == rhs);
-}
+    uint32_t seconds;
```
RedBeard0531, line 40: Simple structs with public members seems fine for types like this that are just a bag of data.

src/util/bson/bson.cpp
```diff
@@ -443,641 +444,196 @@ std::ostream& operator<<(std::ostream& out, const Bson& b)
...
-                m_instructions.push({State::BinarySubType, top.key});
+    REALM_ASSERT_RELEASE(!"unknown json value type");
+}
+
+// This works around the deleted rvalue constructor in StringData
+inline StringData tosd(const std::string& s) { return s; }
+
+// Keep these sorted by key. This is checked so you can't forget.
+using FancyParser = Bson(*)(const Json& json);
+static constexpr std::pair<std::string_view, FancyParser> bson_fancy_parsers[] = {
```
RedBeard0531, line 533: Using a sorted array rather than a map means that this is part of the binary image rather than being constructed at startup.

src/util/bson/bson.cpp
```diff
@@ -443,641 +444,196 @@ std::ostream& operator<<(std::ostream& out, const Bson& b)
...
- @return whether parsing should proceed
- @note It is safe to move the passed string.
- */
-bool Parser::string(string_t& val) {
-    if (!m_instructions.size()) {
-        m_marks.top().push_back("", std::string(val.begin(), val.end()));
-        return false;
+    iterator erase(iterator) {
+        // This is only used when mutating the DOM which we don't do.
+        REALM_ASSERT_RELEASE(!"LinearMap::erase() should never be called");
```
tgoyne, line 485: `REALM_TERMINATE()` is a more straightforward way to do this.

tests/bson.cpp
jsflax: No, returning a bool was to telegraph to someone calling `run_corpus` that the lambda was meant to return successfully or not.

src/util/bson/mongo_timestamp.hpp
jsflax: We don't care that this is mutable?

RedBeard0531: Nah. In C++ it is more idiomatic to return const references (or const-reference-like objects such as string_view) when you want to disallow mutation rather than than using immutable types.

tests/bson.cpp
RedBeard0531: Is it OK that this is always returning true and using `CHECK()` instead?

jsflax: It's technically OK, but would be cleaner if we rewrote all tests to `CHECK` within their lambdas and eliminate the return value of the lambda.

src/util/bson/mongo_timestamp.hpp
jsflax: 👍

tests/bson.cpp
RedBeard0531: Do you want me to do that in this PR?

jsflax: I can do it in the `jf/add_mongo_timestamp` branch once this is merged.

src/util/bson/bson.cpp
jsflax: 👍

src/util/bson/mongo_timestamp.hpp
jsflax: 👍

src/util/bson/bson.cpp
RedBeard0531: done

tests/bson.cpp
RedBeard0531: I just went ahead and scripted it. Gave me something to do while recompiling 🙂

CMake/RealmCore.cmake
```diff
@@ -215,16 +215,28 @@ function(download_realm_core core_version)
...
+    if(NOT ${flag} STREQUAL "")
+        list(APPEND ${out_var} -D${flag}=${${flag}})
+    endif()
+endmacro()
+
 macro(build_realm_core)
     set(core_prefix_directory "${CMAKE_CURRENT_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/realm-core")
 
+    set(extra_flags "")
+    propagate_flag(extra_flags CMAKE_CXX_COMPILER)
```
RedBeard0531, line 228: I needed to do this in order to be able to test with libc++ on linux. If you'd prefer, I can pull this commit out of this PR and do it separately.

tests/bson.cpp
```diff
@@ -40,7 +38,7 @@ static inline std::string remove_whitespace(const char* c) {
  ======== BSON CORPUS ========
  */
 template <typename T>
-using CorpusCheck = std::function<bool(T)>;
```
RedBeard0531, line 43: `std::function` can be brutal to compile times, especially if you are always constructing it from unique lambdas. Switching to a simple function pointer shaves about a second (10%) off my rebuild times.

src/index_set.hpp
```diff
@@ -189,8 +189,14 @@ class IndexSet : private _impl::ChunkedRangeVector {
     void clear() noexcept;
 
     // An iterator over the individual indices in the set rather than the ranges
-    class IndexIterator : public std::iterator<std::forward_iterator_tag, size_t> {
```
RedBeard0531, line 192: This was unrelated to this PR, but I wanted to get a green build.

CMake/RealmCore.cmake
jsflax: I'm not the best person to ask for this.

src/util/bson/bson.cpp
```diff
@@ -443,641 +444,196 @@ std::ostream& operator<<(std::ostream& out, const Bson& b)
...
+
+// Keep these sorted by key. This is checked so you can't forget.
+using FancyParser = Bson(*)(const Json& json);
+static constexpr std::pair<std::string_view, FancyParser> bson_fancy_parsers[] = {
+    {"$binary", +[](const Json& json) {
+        util::Optional<std::vector<char>> base64;
+        util::Optional<uint8_t> subType;
+        if (json.size() != 2)
+            throw BsonError("invalid extended json $binary");
+        for (auto&& [k, v] : json.items()) {
```
jsflax, line 539: Is this syntax destructuring a pair as an rval ref?

src/util/bson/bson.cpp
```diff
@@ -443,641 +444,196 @@ std::ostream& operator<<(std::ostream& out, const Bson& b)
...
-
-        if (m_instructions.size()) {
-            // if the previous instruction is a key, we don't want it
-            if (m_instructions.top().type == State::JsonKey) {
-                m_instructions.pop();
+        if (!base64 || !subType)
+            throw BsonError("invalid extended json $binary");
+        return Bson(std::move(*base64)); // TODO don't throw away the subType.
+    }},
+    {"$date", +[](const Json& json) {
```
jsflax, line 551: What does the `+` signify here?

src/util/bson/bson.cpp
RedBeard0531: `auto&&` is just generally the best default thing to use for range-for loops because it always does the right thing. Depending on what is yielded from the `json.items()` range it will either be a `&`, a `const&`, or a `&&` (it could also be a `const&&` but best to pretend they don't exist and behave like `const&`). There are two complicating factors to keep in mind though:

1) names are always lvalues, even if they are rvalue-refs. So you still need `std::move(x)` or `std::forward<decltype(x)>(x)` if you want it to be moved-from/forwarded.
2) the cvref qualifications (const, volatile, reference) on a structured binding generally apply to the hidden variable used to capture the whole object, not to the separate bindings. (The actual mechanics of structured bindings are kinda weird but they generally just behave how you expect, as long as you remember that they are just creating new bindings, not actually breaking up the object, since that would cause lifetime and object model issues)

TL;DR: That is the (mostly) full story. Luckily, you basically never need to think about any of that. Just always use `auto&&` in your range-for loops and write whatever code you would write if it was an alias for the elements in the range.

src/util/bson/bson.cpp
RedBeard0531: It decays the lambda object (each lambda expression creates a callable object of a unique type) to an equivalent plain function pointer. This is only possible when the lambda has no captures (ie when it is just fancy syntax for a plain function, not a closure). It probably isn't actually necessary here because the conversion should happen implicitly as part of initializing the array. It is mostly habit to put it there when I want it to happen. It is more important when you are passing a lambda to a context where the type is deduced like `auto f = +[]{};` or `template <typename T> f(T&&); f(+[]{});` or even `std::function(+[]{})`.

CMake/RealmCore.cmake
RedBeard0531: It doesn't matter much since this will be going away with the move to monorepo anyway.

jsflax: 👍

src/util/bson/bson.cpp
jsflax: 👍 Thanks for the tips.

src/util/bson/bson.cpp
jsflax: Ah, that makes sense.

PR #1011, opened 2020-05-19
Add support for push notifications

leemaguire:
This PR will add the functionality to register and deregister a device for use with a push notification service.


src/sync/push_client.cpp
```diff
@@ -0,0 +1,90 @@
...
+    s << bson::Bson(args);
+    
+    Request request {
+        .method = HttpMethod::put,
+        .url = route,
+        .body = s.str()
+    };
+    
+    m_auth_request_client->do_authenticated_request(request,
+                                                   sync_user,
```
jsflax, line 53: Wrong indentation.

src/sync/push_client.cpp
```diff
@@ -0,0 +1,90 @@
...
+    s << bson::Bson(args);
+    
+    Request request {
+        .method = HttpMethod::del,
+        .url = route,
+        .body = s.str()
+    };
+    
+    m_auth_request_client->do_authenticated_request(request,
+                                                   sync_user,
```
jsflax, line 84: Wrong indentation.

src/sync/push_client.cpp
leemaguire: Fixed

src/sync/push_client.cpp
leemaguire: Fixed

src/sync/push_client.cpp
```diff
@@ -0,0 +1,90 @@
...
+    
+    bson::BsonDocument args {
+        {"registrationToken", registration_token}
+    };
+    
+    std::stringstream s;
+    s << bson::Bson(args);
+    
+    Request request {
+        .method = HttpMethod::put,
```
ironage, line 47: Looks like there's a problem using this syntax in vs2019 (https://github.com/realm/realm-object-store/pull/1007). Probably need to initialize members explicitly, or maybe it's time we make a constructor for this struct?

jsflax: Would it be a faux-pas to make a constructor for a struct that is effectively a data bag?

ironage: Maybe it's a bit heavy handed for this type of struct, but it would be clear. I think the problem was that this is a C++ 20 feature that isn't implemented in VS2019 yet, but we should be able to use the unnamed version since it should be supported since C++11 (https://en.cppreference.com/w/cpp/language/aggregate_initialization) so an alternative to making a constructor that should also work in VS could be `Request request { HttpMethod::put, route, m_request_timeout_ms, {}, s.str() }`

leemaguire: changed to use:

`m_auth_request_client->do_authenticated_request({
        HttpMethod::del,
        route,
        m_timeout_ms,
        {},
        s.str(),
        false
    },
    sync_user,
    handler);`

tests/sync/app.cpp
```diff
@@ -1480,6 +1482,77 @@ TEST_CASE("app: remote mongo client", "[sync][app]") {
...
+    });
+    
+    SECTION("register") {
+        bool processed;
+        
+        app->push_notification_client("gcm").register_device("hello",
+                                                                  sync_user,
+                                                                  [&](Optional<app::AppError> error) {
+            CHECK(!error);
+            processed = true;
```
ironage, line 1537: could also check that the user here is the same as the one that we logged in with. Also would be nice to check some error conditions around this (and making sure the error messages are clear), such as:
- what happens when registering a device when there is no current user logged in?
- what happens when registering to a push service that doesn't exist?
- what happens when registering twice for the same user?
- what happens when deregistering for a user that hasn't registered?

PR #1013, opened 2020-05-26
Fix remote mongo findOneAndX to use 'filter'

leemaguire:
This PR addresses a bug fix. The bug is where a findOneAndX remote mongo call was passing up the filtering document with `query` instead of `filter` in the request body. Tests have also been added to make sure the unhappy path is captured

### DIFF
--- a/src/sync/app.cpp
+++ b/src/sync/app.cpp
@@ -927,7 +927,6 @@ void App::call_function(std::shared_ptr<SyncUser> user,
         if (auto error = check_for_errors(response)) {
             return completion_block(error, util::none);
         }
-        
         completion_block(util::none, util::Optional<bson::Bson>(bson::parse(response.body)));
     };
 
--- a/src/sync/app.cpp
+++ b/src/sync/app.cpp
@@ -927,6 +927,10 @@ void App::call_function(std::shared_ptr<SyncUser> user,
         if (auto error = check_for_errors(response)) {
             return completion_block(error, util::none);
         }
+        
+        if (response.body == "null") {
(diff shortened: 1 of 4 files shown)


jsflax:
@leemaguire can we also add a fix here for findOneAndX to use `projection` instead of `project`?

leemaguire:
> @leemaguire can we also add a fix here for findOneAndX to use `projection` instead of `project`?

Sure, all `RemoteFindOneAndModifyOptions` now uses `projection`

src/sync/app.cpp
```diff
@@ -927,6 +927,10 @@ void App::call_function(std::shared_ptr<SyncUser> user,
         if (auto error = check_for_errors(response)) {
             return completion_block(error, util::none);
         }
+        
+        if (response.body == "null") {
```
jsflax, line 931: I'm not sure this correct. "null" is technically valid BSON.

edualonso:
Will using `projection` work before https://jira.mongodb.org/browse/REALMC-5787 is fixed?

src/sync/app.cpp
leemaguire: I've removed this check, I had some weird runtime exceptions where a "null" response body was not parsing correctly but it looks to be all good now

src/sync/app.cpp
```diff
@@ -927,7 +927,7 @@ void App::call_function(std::shared_ptr<SyncUser> user,
         if (auto error = check_for_errors(response)) {
             return completion_block(error, util::none);
         }
-
+
```
jsflax, line 930: bad indent

PR #1015, opened 2020-05-26
Add C++17 deduction guides to EventLoopDispatcher

RedBeard0531:
C++17 For the win!

I was about to start using this type a lot in realm-js, so I thought I'd make it a bit easier to use first.

### DIFF
--- a/src/util/event_loop_dispatcher.hpp
+++ b/src/util/event_loop_dispatcher.hpp
@@ -89,20 +89,36 @@ public:
 namespace _impl::ForEventLoopDispatcher {
 template <typename Sig>
 struct ExtractSignatureImpl {};
-template <typename T, typename... Args, bool isNoexcept>
-struct ExtractSignatureImpl<void(T::*)(Args...) noexcept(isNoexcept)> {
+template <typename T, typename... Args>
+struct ExtractSignatureImpl<void(T::*)(Args...)> {
     using signature = void(Args...);
 };
-template <typename T, typename... Args, bool isNoexcept>
-struct ExtractSignatureImpl<void(T::*)(Args...) const noexcept(isNoexcept)> {
+template <typename T, typename... Args>
+struct ExtractSignatureImpl<void(T::*)(Args...) const> {
     using signature = void(Args...);
 };
-template <typename T, typename... Args, bool isNoexcept>
-struct ExtractSignatureImpl<void(T::*)(Args...) & noexcept(isNoexcept)> {
+template <typename T, typename... Args>
+struct ExtractSignatureImpl<void(T::*)(Args...) &> {
     using signature = void(Args...);
 };
-template <typename T, typename... Args, bool isNoexcept>
-struct ExtractSignatureImpl<void(T::*)(Args...) const & noexcept(isNoexcept)> {
+template <typename T, typename... Args>
+struct ExtractSignatureImpl<void(T::*)(Args...) const &> {
+    using signature = void(Args...);
+};
+template <typename T, typename... Args>
+struct ExtractSignatureImpl<void(T::*)(Args...) noexcept> {
+    using signature = void(Args...);
+};
+template <typename T, typename... Args>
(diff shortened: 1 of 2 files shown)


src/util/event_loop_dispatcher.hpp
```diff
@@ -85,6 +85,25 @@ class EventLoopDispatcher<void(Args...)> {
...
+template <typename Sig>
+struct extract_signature_impl<std::function<Sig>> {
+    using signature = Sig;
+};
+
+template <typename T>
+using extract_signature = typename extract_signature_impl<T>::signature;
+}
+
+// Use std::function deduction guides for EventLoopDispatcher. This works with function pointers, lambdas (without
```
RedBeard0531, line 101: I could write out a version of `extract_signature` that doesn't rely on `std::function` if you'd prefer. It isn't complicated, just long(ish) and repetitive. And since we are already using a `std::function` as the constructor argument, it even seems vaguely correct to just defer to std::function, rather than defining our own.

RedBeard0531: If you are curious, this is the libc++ `std::function` deduction guide impl: https://github.com/llvm/llvm-project/blob/fb38b98338cc87442e3451665e82bf1c8ef9388f/libcxx/include/functional#L2428-L2473

src/util/event_loop_dispatcher.hpp
```diff
@@ -85,6 +85,25 @@ class EventLoopDispatcher<void(Args...)> {
         m_scheduler->notify();
     }
 };
+
+namespace detail {
```
tgoyne, line 89: We use `_impl` rather than `detail` for this.

src/util/event_loop_dispatcher.hpp
RedBeard0531: Blarg. Looks like libc++ added the `std::function` deduction guides too recently for them to be available in the version of xcode we are using: https://github.com/llvm/llvm-project/commit/e1eabcdfad89f67ae575b0c86aa4a72d277378b4. I guess I'll have to not rely on them here anyway.

I'll do that tomorrow. No need to review until then, unless you think you are opposed to the general idea of this commit.

src/util/event_loop_dispatcher.hpp
```diff
@@ -85,6 +85,25 @@ class EventLoopDispatcher<void(Args...)> {
...
+template <typename T>
+struct extract_signature_impl {};
```
tgoyne, line 91: `ExtractSignatureImpl`/`ExtractSignature` is the naming convention we use.

tests/util/event_loop.cpp
```diff
@@ -39,6 +40,29 @@
 
 using namespace realm::util;
 
+template <typename Desired, typename Actual>
+void static_assert_EventLoopDispatcher_guide(const EventLoopDispatcher<Actual>&) {
+    static_assert(std::is_same_v<Actual, Desired>);
+}
+
+[[maybe_ununsed]] void check_EventLoopDispatcher_guides() {
```
tgoyne, line 48: Even though this is just a compile-time check I think it'd be better to put it in one of the files for tests?

RedBeard0531: Oops. I saw it was in `tests/`, but I guess this is utilities *for* tests, not test *of* utilities...

Is there a good existing test file for this stuff or should I make a new one?

tgoyne: There probably isn't a good pre-existing location. There aren't currently any tests for this, and EventLoopDispatcher currently actually isn't even being built on the v10 branch since the things in object store which were using it were deleted.

RedBeard0531: Do you mind if I leave it here for now then, and it can be moved out once the dust settles? As of my latest patch it is now in an anonymous namespace, so it shouldn't even generate any actual code in the object files.

tgoyne: Yeah, it's no big deal.

PR #1016, opened 2020-05-28
Je/update dependencies

jedelbo:


jedelbo:
CI is never really happy - but as happy as can be expected.

PR #1022, opened 2020-06-01
Stitch config sync update

ironage:
This will allow syncing objects.


ironage:
These changes were all added in https://github.com/realm/realm-object-store/pull/1024 so I'm closing this one.

PR #1025, opened 2020-06-03
Update Core and Sync dependencies and adapt

jedelbo:


nirinchev:
This is a client reset type of error so should probably be added here: https://github.com/realm/realm-object-store/blob/master/src/sync/sync_config.hpp#L102-L113

jedelbo:
Updated PR according to suggestion from @nirinchev

jedelbo:
Seems to be failing with the standard error on MacOS

PR #1027, opened 2020-06-04
Add new protocol error to list of errors handled.

cmelchior:
Seems this was forgotten as part of: https://github.com/realm/realm-object-store/pull/1025

Not handling this would otherwise break Realm Java


jsflax:
That was my bad– I removed it in the previous PR. I must have forgotten to update core. There are two places I removed it.

jsflax:
It also needs to be added to the end of `sync_config.hpp`: https://github.com/realm/realm-object-store/commit/29ae92d513cfa2ba3dfd897502ca91d676fdbccd

PR #1028, opened 2020-06-04
Teardown sync session on refresh token expiry

jsflax:


src/sync/sync_session.hpp
```diff
@@ -107,6 +107,8 @@ class SyncProgressNotifier {
     std::unordered_map<uint64_t, NotifierPackage> m_packages;
 };
 
+std::function<void(util::Optional<app::AppError>)> handle_refresh(std::shared_ptr<SyncSession>);
```
tgoyne, line 110: Why is this a friend function? It seems like just a private member function?

src/sync/sync_session.cpp
```diff
@@ -259,23 +259,39 @@ const SyncSession::State& SyncSession::State::active = Active();
...
-SyncSession::SyncSession(SyncClient& client, std::string realm_path, SyncConfig config, bool force_client_resync)
-: m_handle_refresh([this](util::Optional<app::AppError> error) {
-    if (error) {
-        // 10 seconds is arbitrary
-        std::this_thread::sleep_for(std::chrono::milliseconds(10000));
-        if (m_state == &State::active) {
-            user()->refresh_custom_data(m_handle_refresh);
+std::function<void(util::Optional<app::AppError>)> _impl::handle_refresh(std::shared_ptr<SyncSession> session) {
+    return [session](util::Optional<app::AppError> error) {
+        if (!session)
```
tgoyne, line 264: `session` can't be null here since it's capturing a strong reference.

src/sync/sync_session.cpp
```diff
@@ -259,23 +259,39 @@ const SyncSession::State& SyncSession::State::active = Active();
...
+            return;
+
+        auto session_user = session->user();
+        auto is_user_expired = session_user &&
+            session_user->refresh_jwt().expires_at < std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now().time_since_epoch()).count();
```
tgoyne, line 269: Anything using `std::chrono` really wants a `using namespace std::chrono` within the function.

src/sync/sync_session.cpp
```diff
@@ -259,23 +259,39 @@ const SyncSession::State& SyncSession::State::active = Active();
...
+        if (is_user_expired) {
+            std::unique_lock<std::mutex> lock(session->m_state_mutex);
+            session->cancel_pending_waits(lock, error ? error->error_code : std::error_code());
+            if (session->m_config.error_handler) {
+                session->m_config.error_handler(session,
+                                                SyncError(SyncError::ProtocolError::bad_authentication, "expired refresh token", true));
+            }
+        } else if (error || !session_user) {
+            // 10 seconds is arbitrary
+            std::this_thread::sleep_for(std::chrono::milliseconds(10000));
```
tgoyne, line 280: What is this sleep trying to do?

src/sync/sync_session.cpp
```diff
@@ -259,23 +259,39 @@ const SyncSession::State& SyncSession::State::active = Active();
...
-        std::this_thread::sleep_for(std::chrono::milliseconds(10000));
-        if (m_state == &State::active) {
-            user()->refresh_custom_data(m_handle_refresh);
+std::function<void(util::Optional<app::AppError>)> _impl::handle_refresh(std::shared_ptr<SyncSession> session) {
+    return [session](util::Optional<app::AppError> error) {
+        if (!session)
+            return;
+
+        auto session_user = session->user();
+        auto is_user_expired = session_user &&
```
tgoyne, line 268: I think handling the case where session_user is null first will end up simplifying this logic overall since there's a few spots that are awkward due to trying to handle that later.

src/sync/sync_session.hpp
jsflax: Yep, could be. I'll move it.

src/sync/sync_session.cpp
jsflax: I was mimicking what the old cocoa refresh did– giving 10 seconds between retrying an access token refresh.

PR #1029, opened 2020-06-05
Stop holding a strong reference to sessions for refresh

jsflax:


src/sync/sync_session.cpp
```diff
@@ -259,10 +259,14 @@ const SyncSession::State& SyncSession::State::active = Active();
 const SyncSession::State& SyncSession::State::dying = Dying();
 const SyncSession::State& SyncSession::State::inactive = Inactive();
 
-std::function<void(util::Optional<app::AppError>)> SyncSession::handle_refresh(std::shared_ptr<SyncSession> session) {
-    return [session](util::Optional<app::AppError> error) {
+std::function<void(util::Optional<app::AppError>)> SyncSession::handle_refresh(std::weak_ptr<SyncSession> weak_session) {
+    return [weak_session](util::Optional<app::AppError> error) {
         using namespace std::chrono;
+        if (weak_session.expired()) {
```
tgoyne, line 265: This is a TOCTOU race. The correct way to do this is to just call `lock()` and check if the result of that is null.

PR #1030, opened 2020-06-05
Persist app metadata

jsflax:
We need to persist app metadata since init'ing it is not called without logging in.

It appears, however, that something is fundamentally wrong with the pathway for SyncManager, since it is a singleton that isn't really a singleton.


tests/sync/session/session.cpp
```diff
@@ -356,6 +361,7 @@ TEMPLATE_TEST_CASE("sync: stop policy behavior", "[sync]", RegularUser) {
     if (!EventLoop::has_implementation())
         return;
 
+    SyncManager::shared().reset_for_testing();
```
tgoyne, line 364: Nothing should have to ever call reset_for_testing() at the start of the test. The previous test should be cleaning up properly.

tests/sync/session/session.cpp
```diff
@@ -47,6 +47,7 @@ TEST_CASE("SyncSession: management by SyncUser", "[sync]") {
     if (!EventLoop::has_implementation())
         return;
 
+    reset_test_directory(tmp_dir());
```
tgoyne, line 50: This is deleting /tmp or $TMPDIR which can do very bad things to any other running programs.

src/sync/impl/sync_metadata.cpp
```diff
@@ -113,6 +122,12 @@ realm::Schema make_schema()
         }},
         {c_sync_current_user_identity, {
             {c_sync_current_user_identity, PropertyType::String}
+        }},
+        {c_sync_app_metadata, {
+            {c_sync_app_metadata_deployment_model, PropertyType::String},
+            {c_sync_app_metadata_location, PropertyType::String},
+            {c_sync_app_metadata_hostname, PropertyType::String},
+            {c_sync_app_metadata_ws_hostname, PropertyType::String}
```
ironage, line 130: don't forget to bump the schema version, I can't remember where it is but I don't see it modified anywhere here

tgoyne: Adding new types doesn't require bumping the schema version.

ironage: ah ok thanks!

src/sync/impl/sync_metadata.cpp
```diff
@@ -426,6 +449,79 @@ std::shared_ptr<Realm> SyncMetadataManager::get_realm() const
...
+                                           const std::string& hostname,
+                                           const std::string& ws_hostname) const
+{
+    auto realm = get_realm();
+    auto& schema = m_app_metadata_schema;
+    
+    realm->begin_transaction();
+    
+    auto table = ObjectStore::table_for_object_type(realm->read_group(), c_sync_app_metadata);
+    auto obj = table->create_object();
```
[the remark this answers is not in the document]
jsflax: In theory there should only ever been one app metadata object. It is initialized once on first app usage, then never updated.

src/sync/impl/sync_metadata.cpp
```diff
@@ -426,6 +449,79 @@ std::shared_ptr<Realm> SyncMetadataManager::get_realm() const
...
+util::Optional<SyncAppMetadata> SyncMetadataManager::get_app_metadata() const
+{
+    auto realm = get_realm();
+    
+    auto table = ObjectStore::table_for_object_type(realm->read_group(), c_sync_app_metadata);
+    
+    if (!table->size())
+        return util::none;
+    
+    return SyncAppMetadata(m_app_metadata_schema, realm, *table->begin());
```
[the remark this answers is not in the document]
jsflax: See comment above.

src/sync/impl/sync_metadata.cpp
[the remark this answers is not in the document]
cmelchior: I guess my concern was that I saw a comment from Eric that it might update, then we might eventually add that logic and forget to update this. Which will introduce _really_ subtle bugs.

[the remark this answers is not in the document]
cmelchior: Adding a primary key field and assign it a single value, should make this scenario impossible.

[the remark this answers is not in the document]
jsflax: What would the primary key be?

tests/sync/file.cpp
```diff
@@ -147,6 +147,7 @@ TEST_CASE("sync_file: SyncFileManager APIs", "[sync]") {
     const std::string local_identity = "123456789";
     const std::string manager_path = base_path + "syncmanager/";
     prepare_sync_manager_test();
+    auto cleanup = util::make_scope_exit([=]() noexcept { try_remove_dir_recursive(base_path); });
```
fealebenpae, line 150:
```suggestion
    auto cleanup = util::make_scope_exit([=]() noexcept { util::try_remove_dir_recursive(base_path); });
```

src/sync/impl/sync_metadata.cpp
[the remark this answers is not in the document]
cmelchior: Any value would work, but `1` ? That way you can use `table->find_first_int(ColKey(id_col), 1)`

jsflax:
@fealebenpae @cmelchior PTAL

tests/sync/session/wait_for_completion.cpp
```diff
@@ -144,26 +144,28 @@ TEST_CASE("SyncSession: wait_for_upload_completion() API", "[sync]") {
...
-        });
-        REQUIRE(handler_called == false);
-        // Now trigger an error
-        SyncSession::OnlyForTesting::handle_error(*session, {code, "Not a real error message", true});
-        EventLoop::main().run_until([&] {
-            return error_count > 0;
-        });
-        REQUIRE(handler_called == true);
-    }
+    // FIXME: There seems to be a race condition here where the upload completion handler
```
cmelchior, line 147: This is a bit concerning, but I guess outside the context of this PR.

src/sync/app.cpp
```diff
@@ -800,17 +802,18 @@ void App::init_app_metadata(std::function<void (util::Optional<AppError>, util::
...
 
         try {
-            this->m_metadata = AppMetadata(value_from_json<std::string>(json, "deployment_model"),
-                                           value_from_json<std::string>(json, "location"),
-                                           value_from_json<std::string>(json, "hostname"),
-                                           value_from_json<std::string>(json, "ws_hostname"));
-
-            m_base_route = m_metadata->m_hostname + base_path;
+            SyncManager::shared().perform_metadata_update([json](const SyncMetadataManager& manager){
+                manager.set_app_metadata(value_from_json<std::string>(json, "deployment_model"),
```
kneth, line 806: The strings are also defined as constants in `src/sync/impl/sync_metadata.cpp`

src/sync/impl/sync_metadata.cpp
```diff
@@ -426,6 +452,54 @@ std::shared_ptr<Realm> SyncMetadataManager::get_realm() const
...
+    realm->begin_transaction();
+    
+    auto table = ObjectStore::table_for_object_type(realm->read_group(), c_sync_app_metadata);
+    auto obj = table->create_object_with_primary_key(app_metadata_pk);
+    obj.set(schema.idx_deployment_model, deployment_model);
+    obj.set(schema.idx_location, location);
+    obj.set(schema.idx_hostname, hostname);
+    obj.set(schema.idx_ws_hostname, ws_hostname);
+
+    realm->commit_transaction();
```
kneth, line 479: What if the transaction fails?

jsflax: Then the `/location` endpoint will be hit for each network call. I'm not sure this can be fixed properly until we have the monorepo.

PR #1039, opened 2020-06-10
Allow SDKs to override property type names

kneth:
Realm JavaScript is using slightly different names for two types: `object id` and `decimal` (`objectId` and `decimal128`). So it is useful to be able to override the names from the SDK.

I believe only Realm Cocoa and Realm JavaScript are using the changed function.

### DIFF
--- a/src/property.hpp
+++ b/src/property.hpp
@@ -26,6 +26,7 @@
 #include <realm/keys.hpp>
 
 #include <string>
+#include <map>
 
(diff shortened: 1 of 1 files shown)


kneth:
@tgoyne Thanks for take the time to comment.

> If the goal is to have the custom names used for things like objectstore-generated error messages then a different design would be needed.

This is a good point that I haven't taken into account.

PR #1041, opened 2020-06-17
Add fix for refreshing an expired access token on sync start

leemaguire:
This PR addresses an issue where a sync session can not be established with an access token that is invalid. There is some small change in logic w.r.t how sync errors are handled and how a session is created once an error has been resolved.

### DIFF
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -90,7 +90,7 @@ struct SyncSession::State {
     // The session should be closed and moved to `inactive`, in accordance with its stop policy and other state.
     virtual void close(std::unique_lock<std::mutex>&, SyncSession&) const { }
 
-    // Returns true iff the error has been fully handled and the error handler should immediately return.
+    // Returns true if the error has been fully handled and the error handler should immediately return.
     virtual void handle_error(std::unique_lock<std::mutex>&, SyncSession&, const SyncError&) const { }
 
     // Register a handler to wait for sync session uploads, downloads, or synchronization.
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -500,6 +500,13 @@ void SyncSession::handle_error(SyncError error)
                 error.is_unrecognized_by_client = true;
         }
     }
+
+    // Dont't bother invoking m_config.error_handler if the sync is inactive.
+    // It does not make sense to call the handler when the session is closed.
+    if (m_state == &State::inactive) {
+        return;
+    }
+
     switch (next_state) {
         case NextStateAfterError::none:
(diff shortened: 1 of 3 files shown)


src/sync/sync_session.cpp
```diff
@@ -259,16 +253,15 @@ const SyncSession::State& SyncSession::State::active = Active();
 const SyncSession::State& SyncSession::State::dying = Dying();
 const SyncSession::State& SyncSession::State::inactive = Inactive();
 
-std::function<void(util::Optional<app::AppError>)> SyncSession::handle_refresh(std::weak_ptr<SyncSession> weak_session) {
-    return [weak_session](util::Optional<app::AppError> error) {
+std::function<void(util::Optional<app::AppError>)> SyncSession::handle_refresh(std::shared_ptr<SyncSession> session) {
+    return [session](util::Optional<app::AppError> error) {
         using namespace std::chrono;
-        auto session = weak_session.lock();
         if (!session) return;
```
tgoyne, line 259: Session can't be null here any more.

tests/sync/session/session.cpp
```diff
@@ -415,8 +415,8 @@ TEMPLATE_TEST_CASE("sync: stop policy behavior", "[sync]", RegularUser) {
             std::error_code code = std::error_code{static_cast<int>(ProtocolError::bad_syntax), realm::sync::protocol_error_category()};
             SyncSession::OnlyForTesting::handle_error(*session, {code, "Not a real error message", true});
             CHECK(sessions_are_inactive(*session));
-            // The session shouldn't report fatal errors when in the dying state.
-            CHECK(!error_handler_invoked);
+            // The session will move from the dying state to inactive, the error handler should be invoked.
```
tgoyne, line 418: The previous behavior seemed correct here? Reporting errors which happen *after* the session is closed is mostly just confusing.

src/sync/sync_session.cpp
```diff
@@ -90,8 +90,8 @@ struct SyncSession::State {
     // The session should be closed and moved to `inactive`, in accordance with its stop policy and other state.
     virtual void close(std::unique_lock<std::mutex>&, SyncSession&) const { }
 
-    // Returns true iff the error has been fully handled and the error handler should immediately return.
-    virtual bool handle_error(std::unique_lock<std::mutex>&, SyncSession&, const SyncError&) const { return false; }
+    // Returns true if the error has been fully handled and the error handler should immediately return.
```
tgoyne, line 93: This appears to not return anything.

tests/sync/session/session.cpp
leemaguire: It wasn't clear to me what the correct behaviour was. I've changed it to make sure the error handler is not invoked when in an `inactive state`

Issue #1040, opened 2020-06-16
Make AppCredentials::function receive a BsonDocument


RealmBot:
Custom function credentials receive string payloads as shown below:

```java
AppCredentials AppCredentials::function(std::string payload_json)
```

This is outdated and needs to be changed to receive a BsonDocument instead of a string payload.

PR: [https://github.com/realm/realm-object-store/pull/1042](https://github.com/realm/realm-object-store/pull/1042)

PR #1042, opened 2020-06-17
Add support for Bson documents to auth custom functions.

edualonso:
Closes:
https://github.com/realm/realm-object-store/issues/1040

The current implementation receives a string payload, which is outdated. It should be a BsonDocument instead.

### DIFF
--- a/src/sync/app_credentials.cpp
+++ b/src/sync/app_credentials.cpp
@@ -16,10 +16,11 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-#include "util.hpp"
 #include "sync/app_credentials.hpp"
+#include "util/bson/bson.hpp"
 
 #include <json.hpp>
+#include <sstream>
 
 namespace realm {
 namespace app {
@@ -150,7 +151,7 @@ AppCredentials AppCredentials::username_password(std::string username,
                           });
 }
 
-AppCredentials AppCredentials::function(bson::BsonDocument payload)
+AppCredentials AppCredentials::function(bson::BsonDocument& payload)
 {
     return AppCredentials(AuthProvider::FUNCTION,
                           [=] {
--- a/src/sync/app_credentials.cpp
+++ b/src/sync/app_credentials.cpp
@@ -16,9 +16,11 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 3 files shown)


src/sync/app_credentials.cpp
```diff
@@ -16,9 +16,11 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
+#include "util.hpp"
```
jsflax, line 19: Is this includes and `<string>` needed?

edualonso: `<string>` isn't (probably leftover from some testing I did), but `util.hpp` is needed for parsing the credentials.

jsflax: This should be `#include "util/bson/bson.hpp"`.

src/sync/app_credentials.hpp
```diff
@@ -21,6 +21,7 @@
 
 #include <functional>
 #include <string>
+#include <object-store/src/util/bson/bson.hpp>
```
jsflax, line 24: This include should not be here and we should be forward declaring BsonDocument.

```cpp
namespace realm {
namespace bson {
class Bson;
template <typename> class IndexedMap;
using BsonDocument = IndexedMap<Bson>;
}
namespace app {
...
```

And then the method signature should be:
```cpp
static AppCredentials function(bson::BsonDocument& payload_json)
```

src/sync/app_credentials.hpp
```diff
@@ -95,8 +96,8 @@ struct AppCredentials {
 
     // Construct and return credentials with the payload.
     // The payload is a MongoDB document as json
-    static AppCredentials function(std::string payload_json);
-    
+    static AppCredentials function(bson::BsonDocument payload);
```
jsflax, line 99: See above comment.

src/sync/app_credentials.hpp
```diff
@@ -95,8 +100,8 @@ struct AppCredentials {
 
     // Construct and return credentials with the payload.
     // The payload is a MongoDB document as json
-    static AppCredentials function(std::string payload_json);
-    
+    static AppCredentials function(bson::BsonDocument& payload);
```
jsflax, line 103: Should make `payload` `const`

PR #1043, opened 2020-06-17
fix msvc compiler glitch

blagoev:
This change is needed for msvc compiler to produce valid code that initializes the session state classes correctly.


RedBeard0531:
lgtm I guess. Don't love that this is needed, but it seems like an acceptable workaround.

PR #1045, opened 2020-06-18
Add watch() helpers

RedBeard0531:


RedBeard0531:
As discussed, no tests in obj store yet, but putting up for review to get early feedback.

You can see the WIP js integration at https://github.com/realm/realm-js/commit/2d3043df131b36d9cfe32fcac0aa1db9c17003f0#diff-af7f90364fd34f588f43f92c13a92ab6R212-R236

And the js tests at https://github.com/realm/realm-js/commit/2d3043df131b36d9cfe32fcac0aa1db9c17003f0#diff-f24a77a955cf2ca9d84fe9d6d07fe08bR170-R224

jsflax:
It would be nice to see a `watch` method on the RemoteMongoCollection class itself to avoid having every SDK having to put together the arguments. Each SDK also keeps references to the `app` class differently.

src/util/bson/bson.cpp
```diff
@@ -418,23 +418,26 @@ std::ostream& operator<<(std::ostream& out, const Bson& b)
...
+            bool first = true;
             for (auto const& pair : doc) {
-                out << nlohmann::json(pair.first).dump() << ':' << pair.second << ",";
+                if (!first)
+                    out << ',';
+                first = false;
+                out << nlohmann::json(pair.first).dump() << ':' << pair.second;
             }
-            if (doc.size())
-                out.seekp(-1, std::ios_base::end);
```
RedBeard0531, line 425: These seekp() calls cause problems when streaming directly to `cout`.

tests/sync/remote_mongo_tests.cpp
```diff
@@ -16,14 +16,550 @@
...
 ////////////////////////////////////////////////////////////////////////////
 
+#include "sync/remote_mongo_collection.hpp"
+
+#include <sstream>
 #include "catch2/catch.hpp"
 
 #ifndef REALM_ENABLE_MONGO_CLIENT_TESTS
-#define REALM_ENABLE_MONGO_CLIENT_TESTS 0
+#define REALM_ENABLE_MONGO_CLIENT_TESTS 1
```
RedBeard0531, line 25: I wasn't entirely sure what this was for. Should I just remove it?

PR #1046, opened 2020-06-19
Bug - stop termination for ArrayMove Events

rezbuzz:
Fix a bug stop terminating app for receiving ArrayMove Events from realm - instead to print warning message of "Adapter: Ignoring ArrayMove instruction"

### DIFF
--- a/src/server/adapter.cpp
+++ b/src/server/adapter.cpp
@@ -519,7 +519,8 @@ public:
         if (!m_list_property_name.size())
             return; // FIXME
 
-        REALM_TERMINATE("ArrayMove not supported by adapter.");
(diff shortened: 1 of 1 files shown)


tgoyne:
This'll just result in data corruption if an ArrayMove instruction is encountered. Any subsequent operations on the array will modify the wrong item in the array once they are out of sync.

rezbuzz:
the bug crashes the node listener for ArrayMove events from realm. It it causes the data corruption I believe the event should not be sent from Realm 

I put a post on stackoverflow about it https://stackoverflow.com/questions/61949183/error-in-realm-sync-adapter-arraymove-not-supported-by-adapter

PR #1048, opened 2020-06-23
Avoid binding a reference to a null pointer

tgoyne:
We never actually used the reference in the case where it would be invalid so this was probably harmless, but it is UB so don't do it.


bmunkholm:
@tgoyne Do you want to keep this or move to monorepo?

PR #1049, opened 2020-06-26
Better paths for storing synced Realms

cmelchior:
This PR is part of moving a bunch of Java logic to the OS. Mostly around respecting the length restrictions for FAT32 file systems, but it also modifies how synced files are being stored so they are easier to locate and work with. But it ended up touching API's in a bunch of different places.

NOTE: This is a breaking change so when this PR is merged, older Realms on the device will be orphaned.

### Changes
- Synced Realms will now be stored under `<rootDir>/<appId>/<userId>/<partitionKeyValue>.realm`. The PR makes a reasonable effort to create a readable realm file name based on the partition value. If this fails, a hashed value is used. This also makes the file layout easier to reason about when we add support for multiple apps. 

- The folders being used to store Sync data have been renamed to be more in line with new product names, i.e. renaming `realm-object-server` to `mongodb-realm`.

- Realms are now stored with the `.realm` suffix. This makes debugging these Realms a lot easier if pulled directly from a device as you can immediately double-click them to open them in Realm Studio. 

- `SyncManager::path_for_realm()` now has a new parameter that enables "support" for FAT32 systems. Mostly it adds additional checks for the limits on these and tries to work around them if possible.

- It is now possible to provide a user-defined Realm file name (a config option available in Java).

- The public user Id is now being used in the path instead of the local id since the former is known by the user while the later is totally unknown.

- `SyncConfig::partion_value` has changed type from `string` to `Bson` which is more type-safe and checks have been added only allowing valid types.

- `SyncManager::configure()` must now always have the app_config.


src/sync/sync_config.hpp
```diff
@@ -144,10 +145,20 @@ struct SyncConfig {
     util::Optional<std::string> recovery_directory;
     ClientResyncMode client_resync_mode = ClientResyncMode::Recover;
 
-    SyncConfig(std::shared_ptr<SyncUser> user, std::string partition_value)
+    SyncConfig(std::shared_ptr<SyncUser> user, bson::Bson partition_value)
     : user(std::move(user))
     , partition_value(std::move(partition_value))
-    { }
+    {
+        switch(partition_value.type()) {
```
tgoyne, line 152:
```suggestion
        switch (partition_value.type()) {
```

src/sync/impl/sync_file.cpp
```diff
@@ -355,4 +367,19 @@ bool SyncFileManager::remove_metadata_realm() const
     }
 }
 
+std::string SyncFileManager::get_user_directory_path(const std::string& user_identity) const {
```
tgoyne, line 370:
```suggestion
std::string SyncFileManager::get_user_directory_path(const std::string& user_identity) const
{
```

src/sync/sync_manager.cpp
```diff
@@ -168,6 +161,11 @@ void SyncManager::configure(SyncClientConfig config, util::Optional<app::App::Co
             m_users.emplace_back(std::move(user));
         }
     }
+
+    // App must be created last as the constructor depends on the SyncFileManager and
```
tgoyne, line 165: The weird thing being done with make_scope_exit was because we need to create the App even if we return early due to not having metadata storage enabled, so this change breaks all the tests where we have it disabled.

I think it'd be clearer if the metadata storage init was extracted to a separate function that's conditionally called rather than the previous thing of using a scope exit.

src/sync/sync_manager.cpp
```diff
@@ -478,23 +476,73 @@ std::shared_ptr<SyncUser> SyncManager::get_existing_logged_in_user(const std::st
...
+        // Default behavior is to attempt to make a nicer filename which will ease debugging when
+        // locating files in the filesystem. We try the following:
+        // - If the partition key is an integer, convert it to a string directly.
+        // - If the partition key is a string and below 50 chars, use an encoded version which
+        //   convert all illegal filesystem characters. Above 50 chars, hash the value. The 50
+        //   char limit is somewhat arbitrary and set so there is a reasonable chance the whole
+        //   path is below the 256 limit for FAT32 file systems.
+        // - ObjectId partition values are hashed directly.
+        const bson::Bson& value = config.partition_value;
+        switch(value.type()) {
```
tgoyne, line 505:
```suggestion
        switch (value.type()) {
```

src/sync/sync_manager.cpp
```diff
@@ -478,23 +476,73 @@ std::shared_ptr<SyncUser> SyncManager::get_existing_logged_in_user(const std::st
...
+            case bson::Bson::Type::Int32: {
+                std::stringstream ss;
```
tgoyne, line 507: These should use `util::to_string()` rather than stringstream.

src/util/bson/bson.hpp
```diff
@@ -30,6 +30,7 @@
 #include <realm/decimal128.hpp>
 #include <realm/object_id.hpp>
 #include <ostream>
+#include <sstream>
```
tgoyne, line 33: This should be in the cpp file, not the header.

bmunkholm:
> NOTE: This is a breaking change so when this PR is merged, older Realms on the device will be orphaned.

What's the end user consequences of that?

tgoyne:
Any unsynced local changes will be lost and they'll have to redownload everything. Since there's already no migration path to v10 this only effects people using the current beta.

src/sync/impl/sync_file.hpp
```diff
@@ -104,10 +110,11 @@ class SyncFileManager {
 
 private:
     const std::string m_base_path;
+    const std::string m_app_id; // App ID's are guaranteed by MongoDB Realm to be lowercase and only consisting of [a-z0-9\-].
 
-    static constexpr const char c_sync_directory[] = "realm-object-server";
-    static constexpr const char c_utility_directory[] = "io.realm.object-server-utility";
-    static constexpr const char c_recovery_directory[] = "io.realm.object-server-recovered-realms";
+    static constexpr const char c_sync_directory[] = "mongodb-realm";
+    static constexpr const char c_utility_directory[] = "realm.mongodb.server-utility";
```
nirinchev, line 116: These directories already reside in `mongodb-realm` - why do we need to redundantly repeat it at every level?

cmelchior: Good point 👍

ironage:
I changed some of the functionality described in Christian's description, but I think I kept the intent of what was trying to be accomplished here.
- Check if the legacy file path exists, and return that if it is found (no migration is attempted to avoid race conditions when moving files)
- The new preferred file path pattern is `<rootDir>/<appId>/<userId>/<realm_file_name>.realm` or `<rootDir>/<appId>/<userId>/<partition>.realm` but we do a test of creating a file there and if there are errors, we fallback to using `<rootDir>/<hashedPath>.realm`. If there are errors at the hashed path, an exception is thrown with the details.

I believe I have addressed the other review comments as well, so I think this is ready for another round of review @tgoyne @cmelchior

nirinchev:
How is partition encoded in the file path? E.g. if I have a string partition "25" and an int partition 25, it seems like they'll end up with the same path.

ironage:
@nirinchev yes, the current behaviour is that values that are convertible to the same string partition value will produce the same local realm path. I see this as a feature, allowing users to input the partition key into a BSON object in whatever type is convenient, even cross platform. This line of thinking assumes that developers will create partitions based on logical value, not based on type. Do you think it should be otherwise?

nirinchev:
This may be desirable, but it's not the behavior today (and hasn't been requested or planned). This means that "25" and 25 will produce separate Realms on the server and thus cannot be used interchangeably. If a user synchronized with one, then attempted to use the same Realm file when synchronizing with the other, they would get a client reset.

ironage:
Ok, I agree we should not change the existing functionality here. I will change the serialization back to the previous encoding for file paths, and more importantly for server session identifiers.

cmelchior:
There is still value in having the partition values readable IMO. A simple scheme could be something like `s-<stringValue>`, `i-intValue`, `l-longValue`, probably no reason to prefix hashed values.

ianpward:
> This may be desirable, but it's not the behavior today (and hasn't been requested or planned). This means that "25" and 25 will produce separate Realms on the server and thus cannot be used interchangeably. 

What would the use case be for this be from the user perspective? Or is this just an implementation detail? I think user error might cause two different realms but the user's intent would be for this to be the same realm

tgoyne:
I think having separate partitions on 25 and "25" is almost certainly going to be user error, but trying to automatically "fix" that can lead to much more confusing problems as it's very hard to avoid having things which are clearly different values being treated as the same value.

nirinchev:
This was discussed some time ago, but the consensus was that it'd be even more confusing if we started ignoring the types of partition values, considering we don't do it for other fields. E.g. if someone inserted a document like `{ intField: "123" }`, that would be a schema violation and we wouldn't try to convert the string value to an integer. It's the same with partition keys - if users accidentally inserted integers as strings or strings as integers, that's on them and they should fix their data. And while I don't expect this situation to be common at all, I think we should still guard against accidental errors. I like @cmelchior's suggestion of prefixing the partition value with the type - it's a good compromise between safety and readability.

ironage:
Updated according to the suggestions for readability here, local realm paths will be prefixed with a type indication and an underscore which I think is more readable than a hyphen when a negative integer follows.

ironage:
@cmelchior can you take this over again while I'm out on vacation the next two weeks? It needs another round of reviews, but I think it should be close to being mergeable.

cmelchior:
@ironage Will do. Thanks for all the hard work 🥇 🎉

PR #1051, opened 2020-07-07
Disable sync metadata storage for most tests

tgoyne:
This cuts the runtime of the sync-enabled test suite from 4 minutes to 4 *seconds* since nearly all of the time was spent creating, opening, closing, and then deleting the metadata realm without ever actually reading anything from it.

I also ported over the changes to fix code coverage generation from v10 as I wanted to make sure that there wasn't any sync metadata functions only accidentally being covered by other tests.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,7 @@
 #!groovy
+
+@Library('realm-ci') _
+
 def getSourceArchive() {
   deleteDir()
   unstash 'source'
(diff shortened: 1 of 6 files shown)


Jenkinsfile
```diff
@@ -181,7 +180,12 @@ stage('unit-tests') {
...
 stage('publish') {
   node('docker') {
-    publishReport('linux-sync')
-    publishReport('macOS-sync')
+    // we need sources to allow the coverage report to display them
+    rlmCheckout(scm)
+    // coverage reports assume sources are in the parent directory
+    dir("build") {
+      publishCoverageReport('linux-sync')
+      publishCoverageReport('macOS-sync')
```
ironage, line 188: Unfortunately the cobertura plugin we use does not yet support multiple reports (https://github.com/jenkinsci/cobertura-plugin/issues/82) I think the behaviour was that it would just display the last report generated. So even if we run coverage multiple times, we can only display one 😞

PR #1052, opened 2020-07-08
Fix handling of primary keys in migrations

tgoyne:
Creating objects inside migrations which change the primary key property of the object type being created didn't work properly because we called Table::create_object_with_primary_key(), and the Table doesn't actually have a primary key until after the migration ends.

Modifying primary key values for existing objects inside migrations almost entirely worked, but in a very fragile way. In particular, for string primary keys without indexes (which should be all of them, but the core 5 -> core 6 migration unintentionally left existing indexes in place) it only worked because we corrupted the table's state in a way that resulted in the check to see if a primary key was already in use giving false positives, which resulted in using a code path that accidentally worked.

To fix this, trying to mutate a primary key inside a migration will now remove the primary key from the table and restore it at the end of the migration, making things behave the same as when adding a primary key to an existing table.

Fixes https://github.com/realm/realm-core/issues/3765. Fixes https://github.com/realm/realm-cocoa/issues/6613.


src/object.cpp
```diff
@@ -124,6 +124,23 @@ Property const& Object::property_for_name(StringData prop_name) const
...
 
+void Object::validate_property_for_setter(Property const& property) const
+{
+    verify_attached();
+    m_realm->verify_in_write();
+
+    // Modifying primary keys is allowed in migrations to make it possible to
+    // add a new primary key to a type (or change the property type), but it
+    // is otherwise considered the immutable identity of the row
+    if (property.is_primary) {
```
jsflax, line 135: Makes sense 👍

src/object_accessor.hpp
```diff
@@ -185,76 +177,92 @@ Object Object::create(ContextType& ctx, std::shared_ptr<Realm> const& realm,
...
-                throw std::logic_error(util::format("Attempting to create an object of type '%1' with an existing primary key value '%2'.",
-                                                    object_schema.name, ctx.print(*primary_value)));
+        if (!primary_value && !is_nullable(primary_prop->type))
+            throw MissingPropertyValueException(object_schema.name, primary_prop->name);
+
+        // When changing the primary key of a table, we remove the existing pk (if any), call
+        // the migration function, then add the new pk (if any). This means that we can't call
+        // create_object_with_primary_key(), and creating duplicate primary keys is allowed as
+        // long as they're unique by the end of the migration.
+        if (table->get_primary_key_column() == ColKey{}) {
```
jsflax, line 225: Is there a performance hit here for non-migration cases?

src/thread_safe_reference.cpp
```diff
@@ -161,7 +148,7 @@ class ThreadSafeReference::PayloadImpl<Results> : public ThreadSafeReference::Pa
                 // match what happens for other types of handover where the
                 // object doesn't exist.
                 switch_on_type(ObjectSchema::from_core_type(*table, m_col_key), [&](auto* t) -> void {
-                    list = std::make_unique<typename ListType<decltype(*t)>::type>();
+                    list = std::make_unique<Lst<NonObjTypeT<decltype(*t)>>>();
```
jsflax, line 151: Where is `Lst` coming from?

src/object_accessor.hpp
tgoyne: This is just reading an int member variable from the table object and comparing to another int, so should be completely unnoticeable.

src/thread_safe_reference.cpp
tgoyne: Lst is the core List type.

tgoyne:
While working on this I had Object::create() split into a few functions at one point, but then by the time I got it all working they each took way too many parameters, including multiple in/out parameters. Unfortunately here I think the function is hard to follow simply because the actual logic is convoluted and not just because it's factored poorly.

jsflax:
Makes sense.

src/object_accessor.hpp
```diff
@@ -185,76 +177,92 @@ Object Object::create(ContextType& ctx, std::shared_ptr<Realm> const& realm,
...
                       ObjectSchema const& object_schema, ValueType value,
                       CreatePolicy policy, ObjKey current_obj, Obj* out_row)
 {
     realm->verify_in_write();
 
-    // get or create our accessor
+    // When setting each property, we normally want to skip over the primary key
+    // as that's set as part of object creation. However, during migrations the
+    // property marked as the primary key in the schema may not currently be
+    // considered a primary key by core, and so will need to be set.
```
ironage, line 200: all these extra comments are super helpful 👍

PR #1054, opened 2020-07-09
Implement precise, unbatched notification of sync completion events

### LINKED ISSUE #1056: Race condition waiting for upload/download completion events
Currently if there are multiple waiters for upload or download completion, they will all be notified when the first event should fire, even if syncing hasn't reached the point of later-registered waiters.

RedBeard0531:
This avoids a race condition where an earlier upload completion event will
notify a later waiter whose changes haven't been uploaded yet.

Closes #1056.

### DIFF
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -239,9 +239,9 @@ struct sync_session_states::Inactive : public SyncSession::State {
         }
 
         // Inform any queued-up completion handlers that they were cancelled.
-        for (auto& callback : download_waits)
+        for (auto& [id, callback] : download_waits)
             callback(make_error_code(util::error::operation_aborted));
-        for (auto& callback : upload_waits)
+        for (auto& [id, callback] : upload_waits)
             callback(make_error_code(util::error::operation_aborted));
     }
 
@@ -553,9 +553,9 @@ void SyncSession::cancel_pending_waits(std::unique_lock<std::mutex>& lock, std::
     lock.unlock();
 
     // Inform any queued-up completion handlers that they were cancelled.
-    for (auto& callback : download)
+    for (auto& [id, callback] : download)
         callback(error);
-    for (auto& callback : upload)
+    for (auto& [id, callback] : upload)
         callback(error);
 }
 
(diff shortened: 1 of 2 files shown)


src/sync/sync_session.cpp
```diff
@@ -722,11 +722,13 @@ void SyncSession::add_completion_callback(_impl::SyncProgressNotifier::NotifierT
 {
     bool is_download = direction == _impl::SyncProgressNotifier::NotifierType::download;
 
-    int resync_counter = m_client_resync_counter;
-    std::weak_ptr<SyncSession> weak_self = shared_from_this();
     auto waiter = is_download ? &sync::Session::async_wait_for_download_completion
                               : &sync::Session::async_wait_for_upload_completion;
-    (m_session.get()->*waiter)([resync_counter, weak_self, is_download](std::error_code ec) {
+    (m_session.get()->*waiter)([is_download,
+                                resync_counter = m_client_resync_counter,
```
RedBeard0531, line 728: I think the resync_counter logic is now redundant since if there was a resync, it won't find the entry for this id in the map. If you agree, I will remove that logic.

src/sync/sync_session.cpp
```diff
@@ -737,28 +739,33 @@ void SyncSession::add_completion_callback(_impl::SyncProgressNotifier::NotifierT
...
 
 void SyncSession::wait_for_upload_completion(std::function<void(std::error_code)> callback)
 {
     std::unique_lock<std::mutex> lock(m_state_mutex);
-    if (m_upload_completion_callbacks.empty())
-        m_state->wait_for_completion(*this, _impl::SyncProgressNotifier::NotifierType::upload);
-    m_upload_completion_callbacks.push_back(std::move(callback));
+    m_completion_request_counter++; // read in wait_for_completion()
+    m_state->wait_for_completion(*this, _impl::SyncProgressNotifier::NotifierType::upload);
+    m_upload_completion_callbacks.emplace_hint(m_upload_completion_callbacks.end(),
```
RedBeard0531, line 756: Do you think it is worth moving this into `add_completion_callback`? I can't really see any reason for the current distribution of work between the functions, but I didn't want to make too much unnecessary change in a bugfix commit.

src/sync/sync_session.cpp
tgoyne: Yes, this looks like a more general solution to the problem the resync counter was addressing.

src/sync/sync_session.cpp
tgoyne: Do we still need separate upload/download completion callback maps with this change? If not then merging them and pushing this into add_completion_callback() makes sense, but without merging them it'd be another awkward conditional there?

RedBeard0531: No, I don't think so. Merging them cleans up the code nicely.

jbreams:
@tgoyne, do you know what the status of this is? This PR doesn't currently compile - at least not on my machine. I looked at making the fixes/improvements in the comments between you and Mathias, but they cause a bunch of test failures. Before I dig into this anymore, do you know if this is still something we want to do?

tgoyne:
It is a bug that needs to be fixed still, and this is most of the way to a correct fix. I think the only thing left to do is to figure out if we still need to support adding completion handlers while the session is inactive, and then either removing that code (if we don't) or finding a way to make that work correctly.

jbreams:
Closing this in favor of #1118.

PR #1055, opened 2020-07-10
Added Realm::shutdown_sync() which waits for Realm file to be closed

kspangsege:
I am proposing this new feature as a solution to a problem identified by @RedBeard0531, namely that when a unit test needs to delete a Realm file, it has no way of ensuring that the Realm file is closed by everyone before deleting.

One way to solve the problem in Realm JS, is to expose `Realm::shutdown_and_wait()` for the purpose of testing, and have tests call it before deleting Realm files. If done this way, it will be necessary to call `shutdown_sync()` before calling `close()`.

### DIFF
--- a/src/sync/impl/sync_client.hpp
+++ b/src/sync/impl/sync_client.hpp
@@ -115,6 +115,11 @@ struct SyncClient {
         return m_client.decompose_server_url(url, protocol, address, port, path);
     }
 
+    void wait_for_session_terminations()
+    {
+        m_client.wait_for_session_terminations_or_client_stopped();
+    }
+
     ~SyncClient()
     {
         stop();
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -93,6 +93,14 @@ struct SyncSession::State {
     // If the error is fatal advance the state to inactive.
     virtual void handle_error(std::unique_lock<std::mutex>&, SyncSession&, const SyncError&) const { }
 
+    // Transition immediately to `inactive` state. Calling this function must gurantee that any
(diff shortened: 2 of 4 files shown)


src/shared_realm.cpp
```diff
@@ -891,6 +891,12 @@ void Realm::close()
     m_coordinator = nullptr;
 }
 
+void Realm::shutdown_sync()
+{
+    if (m_coordinator)
+        m_coordinator->shutdown_sync();
+}
```
kspangsege, line 898: I'll change this to error if `m_coordinator` is null in order to catch cases where close incorrectly comes first

kspangsege: Changed

kspangsege:
> I don't think the Realm/RealmCoordinator functions are needed here; the SDK has access to the SyncSession and calling it directly on that makes more sense to me.

Ok, interesting. I wonder how. I concluded that they did not, because `RealmCoordinator::m_sync_session` is private, and nothing seems to leak it out of the coordinator.

Anyway, if you are sure, I'll remove the changes to `RealmCoordinator` and `Realm` classes.

Can you confirm your certainty about this?

kspangsege:
> What do you think of a name like shutdown_and_wait()? It isn't obvious to me that something named shutdown_sync() would block until it's done.

Sound good to me. I'll make that change.

tgoyne:
SyncUser::session_for_on_disk_path() gives access to the session without going through the Realm.

kspangsege:
@tgoyne, I made the changes we discussed. Ready for another pass on review.

PR #1061, opened 2020-07-22
Fix facebook credential payload

jsflax:


tgoyne:
Looks like there's a test which needs to be updated.

jsflax:
Should be fixed now.

jsflax:
Test failures seem unrelated to change.

PR #1063, opened 2020-07-30
Fix a version pinning issue with Results

tgoyne:
Core 6 changed how TableViews were passed from ResultsNotifier to Results to be pull-based rather than push based. With Core 5, ResultsNotifier mutated the parent Results to update the TableView as part of change notification delivery, while now Results gets the TableView from the ResultsNotifier when the user reads from the Results and forces a re-evaluation of the query. This change simplified the data flow a lot, but meant that if the Results is never actually read from, the TableView would sit around in the ResultsNotifier forever, and continue to pin that version.

To fix this, add a check to see if there's a stale TV that went unused and if so discard it and release the version pin.

### DIFF
--- a/src/impl/results_notifier.cpp
+++ b/src/impl/results_notifier.cpp
@@ -181,13 +181,21 @@ void ResultsNotifier::do_prepare_handover(Transaction& sg)
 bool ResultsNotifier::prepare_to_deliver()
 {
     auto lock = lock_target();
-    if (!get_realm()) {
+    auto realm = get_realm();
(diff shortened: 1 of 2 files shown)


ironage:
Is this change meant to target the core-6 branch? Or is it targeting master because the change is harmless to core 5, and you plan to merge master into core-6?

tgoyne:
master's been core 6 for a while.

ironage:
ah yes it is, I was confused there 😕

PR #1064, opened 2020-08-01
Merge master into v10

tgoyne:
This currently has a failing test because removing an int primary key now leaves the table in an invalid state. After removing the PK the sequence counter is 0 even though there will already be objects with non-0 keys. This will need to be fixed in core.


jedelbo:
@tgoyne You can update Sync version as well

tgoyne:
There were some spurious conflicts in trailing whitespace in a pile of places when merging, which I resolved by just stripping all trailing whitespace from the files. This makes the diff a bit of a mess, so you'll probably want to filter out the whitespace-only changes by appending `?w=1` to the url.

PR #1073, opened 2020-08-24
ROBJSTORE-109: Don't construct LinkPathPart with invalid tableref

### LINKED ISSUE #1072: generate_include_from_keypaths throws when constructing with forward links

nirinchev:
Fixes #1072 

The `LinkPathPart` constructor throws `NoSuchTable` when `tr` is invalid. Due to that only being used in QBS code that's going away + my limited C++ skills, I haven't added any tests, but @ironage or @fealebenpae feel free to add if you feel it's worth it.


ironage:
Looks good. There is no benefit to having a IncludeDescriptor for forward links though because they are always included automatically in query based sync. Hopefully this code path was short circuited before getting to this method.

nirinchev:
The .NET test in question includes a backlink property over a few forward link hops. E.g. `A.B.Backlinks`. Ultimately, we want to get to `Backlinks`, but the code was throwing when following the forward links.

PR #1075, opened 2020-08-26
Fix bug where a users identities are not being set on init_metadata

leemaguire:
This PR addresses a bug where the `identities` of a user were not being set on creation of an `App` due to the fact that `SyncManager::init_metadata` did not call `update_identities` for each persisted user.

Fixes: https://github.com/realm/realm-cocoa/issues/6647

### DIFF
--- a/src/sync/sync_manager.cpp
+++ b/src/sync/sync_manager.cpp
@@ -157,6 +157,7 @@ void SyncManager::init_metadata(SyncClientConfig config, const std::string& app_
                                                    user_data.access_token,
                                                    user_data.state,
                                                    user_data.device_id);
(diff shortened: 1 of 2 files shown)


bmunkholm:
Remember changelog.md update

jsflax (approved):
Nice job 👍

ironage:
@bmunkholm there's no changelog to update in the object store, SDK's will have to note the changes themselves by looking at the commits as they update to the new object store submodule.

bmunkholm:
@ironage yeah I figured and deleted my comment long before your note :-) (for some reason I thought it was realm-cocoa repo) :-)

PR #1077, opened 2020-08-28
Use global id to compose Realm path

clementetb:
This PR refactor ObjectStore, so it only uses `local_uuid` for legacy Realms. For future Realms, it uses the global server UUID in the path.

### DIFF
--- a/src/sync/impl/sync_file.cpp
+++ b/src/sync/impl/sync_file.cpp
@@ -229,16 +229,16 @@ std::string SyncFileManager::get_base_sync_directory() const
     return sync_path;
 }
 
-std::string SyncFileManager::user_directory(const std::string& local_user_identity) const
+std::string SyncFileManager::user_directory(const std::string& user_identity) const
 {
-    std::string user_path = get_user_directory_path(local_user_identity);
+    std::string user_path = get_user_directory_path(user_identity);
     util::try_make_dir(user_path);
     return user_path;
 }
 
-void SyncFileManager::remove_user_directory(const std::string& local_user_identity) const
+void SyncFileManager::remove_user_directory(const std::string& user_identity) const
 {
-    std::string user_path = get_user_directory_path(local_user_identity);
+    std::string user_path = get_user_directory_path(user_identity);
     util::try_remove_dir_recursive(user_path);
 }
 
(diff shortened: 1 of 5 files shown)


tests/sync/sync_manager.cpp
```diff
@@ -78,6 +78,7 @@ TEST_CASE("sync_manager: `path_for_realm` API", "[sync]") {
     const std::string identity = "foobarbaz";
     auto user = SyncManager::shared().get_user(identity, ENCODE_FAKE_JWT("dummy_token"), ENCODE_FAKE_JWT("not_a_real_token"), auth_server_url, dummy_device_id);
     auto local_identity = user->local_identity();
+    auto server_identity = user->identity();
```
ironage, line 81: I think you'd also want to check `REQUIRE(server_identity == local_identity)` here?

src/sync/impl/sync_file.cpp
```diff
@@ -420,6 +426,15 @@ std::string SyncFileManager::legacy_realm_file_path(const std::string& local_use
...
+std::string SyncFileManager::legacy_local_identity_path(const std::string& local_user_identity, const std::string& realm_file_name) const
+{
+    auto escaped_file_name = validate_and_clean_path(realm_file_name);
+    std::string path_name = util::file_path_by_appending_component(user_directory(local_user_identity), escaped_file_name);
+    std::string path = path_name + c_realm_file_suffix;
+
+    return path;
+}
+
 std::string SyncFileManager::get_user_directory_path(const std::string& local_user_identity) const {
```
ironage, line 438: maybe change the `local_user_identity` parameter name because it's not the local id anymore, same for the `user_directory` parameter?

nirinchev:
Having two types of ids is confusing and bloats the API - e.g. using a global Id, you can easily figure out where the physical file is stored - you grab the user id from the cloud UI and find the file on the filesystem. If we were to keep the concept of a local Id, we would need to expose an API on the user the get the local id, greatly complicating the documentation.

ironage:
We already have an API to get the local user id (at least at the OS level), and I'm already confused 😂  Are we planning to entirely remove the concept of the local user id?
I thought the location of Realms on the filesystem is an implementation detail, there's no guarantee of where we store the Realms because they can be at the shortened fallback hashed path, or the legacy path for compatibility, or the preferred path.

cmelchior:
The plan is to remove the concept of localId at the SDK level, we only keep it for backwards compatibility

ironage:
in that case +1 for reducing confusion 😄

src/sync/impl/sync_file.hpp
```diff
@@ -83,10 +83,10 @@ class SyncFileManager {
     static bool try_file_exists(const std::string& path) noexcept;
 
     /// Return the path for a given Realm, creating the user directory if it does not already exist.
-    std::string realm_file_path(const std::string& local_user_identity, const std::string& realm_file_name) const;
+    std::string realm_file_path(const std::string& user_identity, const std::string& local_user_identity, const std::string& realm_file_name) const;
```
cmelchior, line 86: Would it be cleaner to send in the `User` object instead of having to read two strings from it?

tests/sync/sync_manager.cpp
cmelchior: They shouldn't have the same value? 🤔

tests/sync/file.cpp
```diff
@@ -186,7 +187,7 @@ TEST_CASE("sync_file: SyncFileManager APIs", "[sync]") {
 
     SECTION("Realm path APIs") {
```
cmelchior, line 188: There seems to be a test missing for us hitting the legacy code path?

cmelchior (changes requested):
Looks good. I do think we need a test for hitting the legacy code path though. It does require a bit of fiddling with the filesystem, unfortunately.

src/sync/impl/sync_file.cpp
```diff
@@ -420,6 +426,15 @@ std::string SyncFileManager::legacy_realm_file_path(const std::string& local_use
     return path;
 }
 
+std::string SyncFileManager::legacy_local_identity_path(const std::string& local_user_identity, const std::string& realm_file_name) const
+{
+    auto escaped_file_name = validate_and_clean_path(realm_file_name);
+    std::string path_name = util::file_path_by_appending_component(user_directory(local_user_identity), escaped_file_name);
```
clementetb, line 432: Notice that `user_directory` calls `try_make_dir`, it will create the directory for the local identity path.

tests/sync/sync_manager.cpp
clementetb: Actually there is no need to keep local_identity

src/sync/impl/sync_file.hpp
clementetb: SyncFileManager isn't bound to User, but it is SyncManager handles Users. 

I believe the reason behind this decision is to facilitate testing SyncFileManager as there is no straight way to mock Users.

ironage (approved):
👍

cmelchior (approved):

PR #1079, opened 2020-09-02
Add the version to the frozen scheduler

nirinchev:
Without this change, we would always use the first cached frozen Realm, regardless of whether it's frozen at the required version or not.

Alternatively, we could pass the version in `do_get_cached_realm`, but felt that using the scheduler was a cleaner solution.

### DIFF
--- a/src/util/scheduler.hpp
+++ b/src/util/scheduler.hpp
@@ -20,6 +20,7 @@
 #define REALM_OS_UTIL_SCHEDULER
 
 #include <realm/util/features.h>
+#include <realm/version_id.hpp>
 
 #include <functional>
 #include <memory>
@@ -77,7 +78,7 @@ public:
 
     // Get the scheduler for frozen Realms. This scheduler does not support
     // notifications and does not perform any thread checking.
-    static std::shared_ptr<Scheduler> get_frozen();
+    static std::shared_ptr<Scheduler> get_frozen(VersionID version);
 
     // Create a new instance of the default scheduler for the current platform.
     // This normally will be a thread-confined scheduler using the current
--- a/src/util/scheduler.hpp
+++ b/src/util/scheduler.hpp
@@ -78,7 +78,7 @@ public:
 
     // Get the scheduler for frozen Realms. This scheduler does not support
     // notifications and does not perform any thread checking.
-    static std::shared_ptr<Scheduler> get_frozen(VersionID version);
+    static std::shared_ptr<Scheduler> get_frozen(VersionID version = VersionID());
 
     // Create a new instance of the default scheduler for the current platform.
     // This normally will be a thread-confined scheduler using the current
--- a/src/util/scheduler.hpp
+++ b/src/util/scheduler.hpp
@@ -78,7 +78,7 @@ public:
 
     // Get the scheduler for frozen Realms. This scheduler does not support
     // notifications and does not perform any thread checking.
-    static std::shared_ptr<Scheduler> get_frozen(VersionID version = VersionID());
+    static std::shared_ptr<Scheduler> get_frozen(VersionID version);
 
     // Create a new instance of the default scheduler for the current platform.
     // This normally will be a thread-confined scheduler using the current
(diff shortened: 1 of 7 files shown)


src/util/scheduler.hpp
```diff
@@ -77,7 +78,7 @@ class Scheduler {
 
     // Get the scheduler for frozen Realms. This scheduler does not support
     // notifications and does not perform any thread checking.
-    static std::shared_ptr<Scheduler> get_frozen();
+    static std::shared_ptr<Scheduler> get_frozen(VersionID version = VersionID());
```
ironage, line 81: When the default value is used here repeatedly, we'll return the cached realm again even if it has been frozen on different versions (ie requesting the latest frozen Realm, writing and advancing, then requesting the latest frozen Realm). In that scenario the version stored by the frozen scheduler gets out of sync with the actual version of the Realm, so I think this solution is insufficient?

ironage (commented):
Can you add some tests?

src/util/scheduler.hpp
nirinchev: I only added the default value because I was too lazy to go and modify all the invocations from tests. In reality, all places where `get_frozen` is called from non-test code, a non-default version is passed. In the tests, we don't appear to care about caching Realms, so passing the default version is fine.

nirinchev:
I can try to add some, but I wanted to get feedback on whether this is the preferred approach - I don't have a lot of context into the codebase (especially how the scheduler interacts with other pieces of Object Store), so didn't want to spend a lot of time on it if it turned out to be the wrong solution.

src/util/scheduler.hpp
ironage: I don't know how SDK's are using this, but it is a public API and I'd prefer a design that doesn't give SDK's the option to make to make this mistake. If that means passing the version into `do_get_cached_realm` I'm fine with that even if it seems less elegant.

tgoyne: `Scheduler::get_frozen()` could theoretically be called directly, but in practice none of the SDKs do.

tgoyne:
I sort of dislike this design-wise, but I don't have any better ideas. The version of a frozen realm is not conceptually something that is linked to the scheduler in any way and instead it's just lumping some state into an object that's convenient. However, it is the obvious way to do this with the design of the cache, which points at the actual problem being that `Scheduler:: is_same_as()` is not the correct abstraction for looking up Realms in the cache.

I think it's probably fine to go with this approach for now and maybe rethink how caching is done later (or never if the cache's design is never actually a problem).

src/util/scheduler.hpp
nirinchev: I can also remove the default argument for `get_frozen` and update the tests? This way we'll force consumers to think what version they're passing in.

ironage: Ok sure, not blocking for me as there doesn't seem to be a much better way at this point. Bonus points if you can assert that the scheduler's frozen version matches the Realm's version somewhere.

tests/realm.cpp
```diff
@@ -474,6 +474,26 @@ class SimpleScheduler : public Scheduler {
         Realm::get_shared_realm(config);
         REQUIRE(object_schema == &*realm->schema().find("object"));
     }
+
+    SECTION("should not use cached frozen Realm if versions don't match") {
+        auto realm = Realm::get_shared_realm(config);
+        realm->read_group();
+        auto frozen1 = realm->freeze();
```
ironage, line 481: REQUIRE(frozen1 != realm)

tests/realm.cpp
```diff
@@ -474,6 +474,26 @@ class SimpleScheduler : public Scheduler {
...
+        REQUIRE(realm->read_transaction_version() == frozen1->read_transaction_version());
+
+        auto table = realm->read_group().get_table("class_object");
+        realm->begin_transaction();
+        Obj obj = table->create_object();
+        realm->commit_transaction();
+
+        REQUIRE(realm->read_transaction_version() > frozen1->read_transaction_version());
+
+        auto frozen2 = realm->freeze();
```
ironage, line 492: REQUIRE(frozen2 != frozen1)
REQUIRE(frozen2 != realm)

ironage (approved):

PR #1080, opened 2020-09-03
Custom confirmation logic support

clementetb:
Add support for calling the custom confirmation function from the SDKs

### DIFF
--- a/src/sync/app.cpp
+++ b/src/sync/app.cpp
@@ -228,6 +228,29 @@ void App::UsernamePasswordProviderClient::resend_confirmation_email(const std::s
     }, handler);
 }
 
+void App::UsernamePasswordProviderClient::retry_custom_confirmation(const std::string& email,
+                                                                    std::function<void(Optional<AppError>)> completion_block)
+{
+    REALM_ASSERT(m_parent);
+    std::string route = util::format("%1/providers/%2/confirm/call", m_parent->m_auth_route, username_password_provider_key);
+
+    auto handler = [completion_block](const Response& response) {
+        handle_default_response(response, completion_block);
+    };
+
+    nlohmann::json body {
+        { "email", email }
+    };
+
+    m_parent->do_request(Request {
+        HttpMethod::post,
+        route,
+        m_parent->m_request_timeout_ms,
+        get_request_headers(),
+        body.dump()
+    }, handler);
+}
+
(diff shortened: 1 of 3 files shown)


src/sync/app.cpp
```diff
@@ -228,6 +228,29 @@ void App::UsernamePasswordProviderClient::resend_confirmation_email(const std::s
     }, handler);
 }
 
+void App::UsernamePasswordProviderClient::retry_custom_confirmation(const std::string& email,
```
jsflax, line 231: Is this the correct function name? Maybe I'm misunderstanding, but it seems like `call_custom_confirmation_function` might be a bit clearer.

cmelchior: Sounds like we need to agree on how to expose this to users 😅 https://github.com/realm/realm-java/pull/7079#discussion_r484393051

src/sync/app.hpp
```diff
@@ -188,6 +188,12 @@ class App : public std::enable_shared_from_this<App>, public AuthRequestClient,
         void send_reset_password_email(const std::string& email,
                                        std::function<void(util::Optional<AppError>)> completion_block);
 
+        /// Calls the custom confirmation function on a user for a given email.
+        /// @param email The email address of the user to call the custom confirmation for.
+        /// @param completion_block A callback to be invoked once the call is complete.
+        void call_custom_confirmation_function(const std::string& email,
```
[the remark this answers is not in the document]
ironage: You can add a custom confirmation function to the config of the docker test server in the same way that we test email registration by only allowing emails that contain "realm_tests_do_autoverify" (see [here](https://github.com/realm/realm-object-store/blob/v10/tests/mongodb/functions/confirmFunc/source.js#L44) and [here](https://github.com/realm/realm-object-store/blob/v10/tests/sync/app.cpp#L316))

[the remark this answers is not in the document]
clementetb: Have added test cases for the custom confirmation function, following the same pattern as `call_reset_password_function`.

ironage (approved):

src/sync/app.cpp
cmelchior: By popular vote, we agreed on `retry_custom_confirmation`

cmelchior (approved):

PR #1084, opened 2020-09-08
Expose List API for adding, inserting, and setting embedded objects

nirinchev:
This just proxies Core's API for SDKs that can't use the Accessor APIs.

### DIFF
--- a/src/list.cpp
+++ b/src/list.cpp
@@ -285,6 +285,39 @@ void List::set(size_t row_ndx, Obj o)
     as<Obj>().set(row_ndx, o.get_key());
 }
 
+Obj List::add_embedded()
+{
+    verify_in_transaction();
+
+    if (!m_is_embedded)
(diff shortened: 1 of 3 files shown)


cmelchior (changes requested):
Java could use these methods as well as right now we create a temporary accessor just for this 👏 

The code looks fine from what I can tell, but there seem to be no tests, which would be nice to add?

nirinchev:
Yeah - I can add a test or two. I didn't initially, because it's directly calling the Core API with no extra logic (and I hope those are well tested).

nirinchev:
@cmelchior added some primitive tests, but I'm not super well versed in Object Store semantics, so let me know if it looks too stupid 😅

nirinchev:
Ping @cmelchior 😛

cmelchior (approved):

PR #1085, opened 2020-09-09
Add multiple realm apps support

jsflax:
The main breaking API changes are:

- Remove `SyncManager::shared` singleton. `SyncManagers` are to only be created by the `App`, and uniquely owned by the `App`. SyncManagers therefore also own a unique `SyncClient` and unique `SyncFileManager`
- `App` must now be created with a `SyncClientConfig`, so there is no longer a multistep process to creating an App and then configuring the SyncManager
- Add `App` caching layer

### DIFF
--- a/src/sync/app.cpp
+++ b/src/sync/app.cpp
@@ -91,7 +91,16 @@ const static std::string user_api_key_provider_key_path = "api_keys";
 
 SharedApp App::get_shared_app(const Config& config, const SyncClientConfig& sync_client_config)
 {
-    return std::make_shared<App>(config, sync_client_config);
+    auto app = std::make_shared<App>(config, sync_client_config);
+    m_apps_cache[config.app_id] = app;
+    return app;
+}
+
+std::unordered_map<std::string, std::weak_ptr<App>> App::m_apps_cache = std::unordered_map<std::string, std::weak_ptr<App>>();
+
+std::weak_ptr<App> App::get_cached_app(const std::string& app_id)
+{
+    return m_apps_cache[app_id];
 }
 
 App::App(const Config& config, const SyncClientConfig& sync_client_config)
--- a/src/sync/app.cpp
+++ b/src/sync/app.cpp
@@ -88,10 +88,13 @@ const static std::string sync_path = "/realm-sync";
 const static uint64_t    default_timeout_ms = 60000;
 const static std::string username_password_provider_key = "local-userpass";
 const static std::string user_api_key_provider_key_path = "api_keys";
+static std::unordered_map<std::string, std::shared_ptr<App>> s_apps_cache;
+std::mutex _mutex;
 
 SharedApp App::get_shared_app(const Config& config, const SyncClientConfig& sync_client_config)
 {
-    auto& app = m_apps_cache[config.app_id];
+    std::lock_guard<std::mutex> lock(_mutex);
+    auto& app = s_apps_cache[config.app_id];
     if (!app) {
         app = std::make_shared<App>(config);
         app->configure(sync_client_config);
@@ -99,11 +102,10 @@ SharedApp App::get_shared_app(const Config& config, const SyncClientConfig& sync
     return app;
 }
 
-std::unordered_map<std::string, std::shared_ptr<App>> App::m_apps_cache = std::unordered_map<std::string, std::shared_ptr<App>>();
-
 std::shared_ptr<App> App::get_cached_app(const std::string& app_id)
 {
-    if (auto it = m_apps_cache.find(app_id); it != m_apps_cache.end()) {
+    std::lock_guard<std::mutex> lock(_mutex);
+    if (auto it = s_apps_cache.find(app_id); it != s_apps_cache.end()) {
         return it->second;
     }
 
@@ -112,7 +114,8 @@ std::shared_ptr<App> App::get_cached_app(const std::string& app_id)
 
 void App::clear_cached_apps()
 {
-    m_apps_cache.clear();
+    std::lock_guard<std::mutex> lock(_mutex);
+    s_apps_cache.clear();
 }
 
 App::App(const Config& config)
--- a/src/sync/app.cpp
+++ b/src/sync/app.cpp
@@ -91,19 +91,33 @@ const static std::string user_api_key_provider_key_path = "api_keys";
 
 SharedApp App::get_shared_app(const Config& config, const SyncClientConfig& sync_client_config)
 {
-    auto app = std::make_shared<App>(config, sync_client_config);
+    if (m_apps_cache.find(config.app_id) != m_apps_cache.end()) {
+        return m_apps_cache[config.app_id];
+    }
+
+    auto app = std::make_shared<App>(config);
+    app->configure(sync_client_config);
     m_apps_cache[config.app_id] = app;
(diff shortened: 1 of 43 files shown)


src/sync/impl/sync_file.cpp
```diff
@@ -222,11 +221,11 @@ std::string SyncFileManager::get_special_directory(std::string directory_name) c
 
 std::string SyncFileManager::get_base_sync_directory() const
 {
-    auto sync_path = file_path_by_appending_component(m_base_path,
-                                                      c_sync_directory,
-                                                      util::FilePathType::Directory);
-    util::try_make_dir(sync_path);
-    return sync_path;
+//    auto sync_path = file_path_by_appending_component(m_base_path,
```
kneth, line 224: Should it be removed?

src/sync/sync_user.cpp
```diff
@@ -408,9 +410,10 @@ void SyncUser::set_binding_context_factory(SyncUserContextFactory factory)
 
 void SyncUser::refresh_custom_data(std::function<void(util::Optional<app::AppError>)> completion_block)
 {
-    SyncManager::shared().app()->refresh_custom_data(shared_from_this(), [completion_block](util::Optional<app::AppError> error){
-        completion_block(error);
-    });
+    completion_block(app::AppError(std::error_code(), "FIXME"));
+//    m_sync_manager->app().lock()->refresh_custom_data(shared_from_this(), [completion_block](util::Optional<app::AppError> error){
```
kneth, line 414: Should it be removed?

src/sync/app.cpp
```diff
@@ -87,9 +89,32 @@ const static uint64_t    default_timeout_ms = 60000;
 const static std::string username_password_provider_key = "local-userpass";
 const static std::string user_api_key_provider_key_path = "api_keys";
 
-SharedApp App::get_shared_app(const Config& config)
+SharedApp App::get_shared_app(const Config& config, const SyncClientConfig& sync_client_config)
```
tgoyne, line 92: All of these functions which access the cache need locking.

src/sync/app.cpp
```diff
@@ -87,9 +89,32 @@ const static uint64_t    default_timeout_ms = 60000;
...
 {
-    return std::make_shared<App>(config);
+    if (m_apps_cache.find(config.app_id) != m_apps_cache.end()) {
+        return m_apps_cache[config.app_id];
+    }
+
+    auto app = std::make_shared<App>(config);
+    app->configure(sync_client_config);
+    m_apps_cache[config.app_id] = app;
+    return app;
```
tgoyne, line 101: This is doing a whole bunch of lookups in the map when it only needs to be doing one:

```suggestion
    auto& app = m_apps_cache[config.app_id];
    if (!app) {
        app = std::make_shared<App>(config);
        app->configure(sync_client_config);
    }
    return app;
```

src/sync/sync_user.cpp
```diff
@@ -408,7 +410,7 @@ void SyncUser::set_binding_context_factory(SyncUserContextFactory factory)
 
 void SyncUser::refresh_custom_data(std::function<void(util::Optional<app::AppError>)> completion_block)
 {
-    SyncManager::shared().app()->refresh_custom_data(shared_from_this(), [completion_block](util::Optional<app::AppError> error){
+    m_sync_manager->app().lock()->refresh_custom_data(shared_from_this(), [completion_block](util::Optional<app::AppError> error){
```
tgoyne, line 413: The only reason to use a weak pointer over a raw pointer is to be able to specifically handle the case where lock() returns null. If not checking it here is correct (and it appears to be), then SyncManager::m_app might as well be a raw pointer.

tests/object.cpp
```diff
@@ -379,7 +379,7 @@ TEST_CASE("object") {
             using U = typename decltype(vec)::value_type;
             auto list = row.get_list<U>(col);
             size_t i = 0;
-            for (const auto& value : vec) {
+            for (const auto value : vec) {
```
tgoyne, line 382: auto& should be correct here?

tests/results.cpp
```diff
@@ -1365,8 +1365,9 @@ TEST_CASE("notifications: async error handling") {
 TEST_CASE("notifications: sync") {
     _impl::RealmCoordinator::assert_no_open_realms();
 
-    SyncServer server(false);
-    TestSyncManager init_sync_manager(server);
+    TestSyncManager init_sync_manager({}, { false });
```
tgoyne, line 1368: Normal c++ style isn't to have spaces inside of {} for brace init.

src/sync/impl/sync_client.hpp
```diff
@@ -39,7 +39,7 @@ namespace _impl {
 using ReconnectMode = sync::Client::ReconnectMode;
 
 struct SyncClient {
-    SyncClient(std::unique_ptr<util::Logger> logger, SyncClientConfig const& config)
+    SyncClient(std::unique_ptr<util::Logger> logger, SyncClientConfig const& config, std::shared_ptr<const SyncManager> sync_manager)
```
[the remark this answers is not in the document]
jsflax: It is used if `NETWORK_REACHABILITY_AVAILABLE` is set.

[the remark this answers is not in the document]
cmelchior: Okay, fair enough, but can we add an else statement with `(void) sync_manager` then to avoid triggering the warning when building without that flag? (I think it is an iOS only flag?)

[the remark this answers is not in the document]
jsflax: Done

src/sync/sync_user.cpp
jsflax: Hmm... I suppose we should check. If the app has been deallocated, we should probably let the user know.

src/sync/app.cpp
jsflax: Done

tests/sync/app.cpp
```diff
@@ -1783,8 +1789,11 @@ TEST_CASE("app: sync integration", "[sync][app]") {
 
     // MARK: Expired Session Refresh -
     SECTION("Expired Session Refresh") {
+        TestSyncManager& sync_manager = *new TestSyncManager({
```
tgoyne, line 1792: If this is needed it definitely needs a comment explaining why it's doing this.

jsflax: It is not needed, I'll change it.

src/sync/sync_user.cpp
```diff
@@ -410,9 +410,14 @@ void SyncUser::set_binding_context_factory(SyncUserContextFactory factory)
 
 void SyncUser::refresh_custom_data(std::function<void(util::Optional<app::AppError>)> completion_block)
 {
-    m_sync_manager->app().lock()->refresh_custom_data(shared_from_this(), [completion_block](util::Optional<app::AppError> error){
-        completion_block(error);
-    });
+    if (auto app = m_sync_manager->app().lock()) {
+        app->refresh_custom_data(shared_from_this(), [completion_block](util::Optional<app::AppError> error){
```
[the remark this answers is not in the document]
jsflax: Nope, I'll fix this.

tgoyne (approved):

PR #1086, opened 2020-09-10
Skip validating dispatch queue types on older OSes

tgoyne:
There's no public API to find out if a dispatch queue is serial or concurrent. On more recent OS versions we can check the class to determine it, but this doesn't work on older versions.

It seems fairly safe to skip the checks on old OSes since the only way it'd go uncaught is if a developer tests exclusively on very old devices (and never in the simulator or recent devices) or has separate code paths for old OS versions.

Fixes https://github.com/realm/realm-cocoa/issues/6735.


ironage (approved):
I'm not familiar with this context, but the reasoning seems sound to me 👍

nirinchev (approved):

PR #1087, opened 2020-09-14
Prepend SyncFileManager path with `app_id`

jsflax:


tests/sync/sync_manager.cpp
```diff
@@ -85,34 +85,34 @@ TEST_CASE("sync_manager: `path_for_realm` API", "[sync]") {
 
     SECTION("should work properly without metadata") {
         TestSyncManager init_sync_manager({.base_path = base_path, .metadata_mode = SyncManager::MetadataMode::NoMetadata});
-        const auto expected = base_path + "mongodb-realm/app_id/foobarbaz/realms%3A%2F%2Frealm.example.org%2Fa%2Fb%2F%7E%2F123456%2Fxyz.realm";
+        const auto expected = base_path + "app_id/mongodb-realm/foobarbaz/realms%3A%2F%2Frealm.example.org%2Fa%2Fb%2F%7E%2F123456%2Fxyz.realm";
```
cmelchior, line 88: I'm a bit uncomfortable doing it like this with the app_id as a prefix. At least on Android files are being stored in a shared files folder, which means we will potentially create many folders that might conflict with already existing files or folders. I would much prefer if we only had `mongodb-realm` and then something like `mongodb-realm/<appId>/server-utility` and `mongodb-realm/<appId>/recovered-realms`

cmelchior (changes requested):

jsflax:
@cmelchior PTAL

src/sync/impl/sync_file.hpp
```diff
@@ -111,8 +107,12 @@ class SyncFileManager {
     }
 
 private:
+    // Denotes the root path for any mongodb-realm app for the passed in `base_path`.
+    // Expected to be `base_path` + "mongodb-realm/".
+    const std::string m_root_sync_path;
+    // Denotes the base path for the mongodb-realm app associated with this sync manager.
```
cmelchior, line 113: Maybe it is just me, but the wording of these comments seems to indicate the naming of the variables should be switched?

tests/sync/file.cpp
```diff
@@ -148,15 +148,15 @@ TEST_CASE("sync_file: URL manipulation APIs", "[sync]") {
...
     const std::string identity = "abcdefghi";
     const std::string local_identity = "123456789";
-    const std::string manager_path = base_path + "syncmanager/";
     const std::string app_id = "test_app_id*$#@!%1";
     const std::string expected_clean_app_id = "test_app_id%2A%24%23%40%21%251";
+    const std::string manager_path = base_path + "syncmanager/mongodb-realm/" + expected_clean_app_id + "/";
     prepare_sync_manager_test();
     auto cleanup = util::make_scope_exit([=]() noexcept { util::try_remove_dir_recursive(base_path); });
-    auto manager = SyncFileManager(manager_path, app_id);
+    auto manager = SyncFileManager(base_path + "syncmanager/", app_id);
```
cmelchior, line 156: Minor, but not 100% sure why we need the `syncmanager` subfolder here?

cmelchior (approved):
Nice 💯 . Only had a few cosmetic comments.

src/sync/impl/sync_file.hpp
jsflax: Hmm... I think the `base_path` parameter in the constructor makes in confusing. You are probably correct, but it does mean `m_base_path` becomes moot.

tests/sync/file.cpp
jsflax: I mainly converted this test to make sure it worked as it did previously– my guess is that this is something legacy-ish.

PR #1089, opened 2020-09-15
Add ability for removing a users access/refresh token on log_out

leemaguire:
When we log out and empty a users access/refresh token we do not persist that to the metadata realm. Therefor on the next startup for example, we still have the tokens, breaking some logic that requires both access & refresh tokens to be empty for a logged out user


jsflax (approved):
💯 nice

ironage (commented):
The changes look good to me, but there's some failing app tests now, probably due to not resetting things correctly, 👍  once the tests all pass

PR #1091, opened 2020-09-18
Remove pk col attribute on table when changing nullability

jedelbo:


tgoyne:
I don't see any reason to think that this would ever do anything useful.

jedelbo:
@tgoyne No, not in the perfect world, but we have this issue: https://github.com/realm/realm-cocoa/issues/6724. The error occurs for a table where the primary key string column is changed from optional to required. I have not (yet) been able to find out what goes wrong, but the `m_primary_key_col` seems to be wrong. This change would just make sure it is reset. Well - anyway - I will keep looking for the root cause, so if you think this PR is a bad idea - no problem.

bmunkholm:
@jedelbo Do you want to close or keep this?

jedelbo:
Outdated

PR #1092, opened 2020-09-21
Small optimization in pk handling

jedelbo:


src/object_store.cpp
```diff
@@ -794,7 +794,7 @@ void ObjectStore::set_schema_keys(Group const& group, Schema& schema)
 void ObjectStore::delete_data_for_object(Group& group, StringData object_type)
 {
     if (TableRef table = table_for_object_type(group, object_type)) {
-        ObjectStore::set_primary_key_for_object(group, object_type, "");
+        table->set_primary_key_column({});
```
tgoyne, line 797: Probably don't need this at all now? This code was here to remove the entry from the pk table.

tgoyne (approved):

PR #1094, opened 2020-09-23
Fix order of args for sync_manager.get_user in the header

nirinchev:


src/sync/app.hpp
```diff
@@ -278,10 +278,10 @@ class App : public std::enable_shared_from_this<App>, public AuthRequestClient,
     /// @returns A shared pointer to the new current user
     std::shared_ptr<SyncUser> switch_user(std::shared_ptr<SyncUser> user) const;
 
-    /// Logs out and removes the provided user
-    /// this is a local operation and does not invoke any server side function
```
nirinchev, line 282: This is unrelated, but was too lazy to open a separate PR for it.

jsflax (commented):
This will change soon, so I wouldn't worry too much about it, but the change is fine for now. I'm surprised this doesn't break the sync_manager tests though.

nirinchev:
This change aligns the header and the implementation signature order. I imagine the tests were written with the order used in the implementation and because all arguments are strings, it just worked and continues working. I only noticed it because the VS method hints use the header signature and I had no idea why I was getting mixed up data for the user.

PR #1097, opened 2020-09-24
add a check to detect if an Object is created with an Obj, but without a valid Transaction.

finnschiermer:
This is an attempt to flush out some NoSuchTable exceptions which we do not understand.

It is based on the assumption that one way of getting a NoSuchTable exception, is when accessing an object after the associated Transaction object in Core has been released. The assert checks for any faulty association between the object accessor and the Realm object (representing/holding the transaction).

### DIFF
--- a/src/object.cpp
+++ b/src/object.cpp
@@ -73,6 +73,7 @@ Object::Object(SharedRealm r, Obj const& o)
 , m_object_schema(&*m_realm->schema().find(ObjectStore::object_type_for_table_name(o.get_table()->get_name())))
 , m_obj(o)
 {
+  REALM_ASSERT(!m_obj.get_table() || (m_realm->get_group() == _impl::TableFriend::get_parent_group(*m_obj.get_table())));
 }
 
 Object::Object(SharedRealm r, StringData object_type, ObjKey key)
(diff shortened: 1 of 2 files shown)


src/object.cpp
```diff
@@ -73,6 +73,7 @@ Object::Object(SharedRealm r, Obj const& o)
 , m_object_schema(&*m_realm->schema().find(ObjectStore::object_type_for_table_name(o.get_table()->get_name())))
 , m_obj(o)
 {
+  REALM_ASSERT(!m_obj.get_table() || (m_realm->get_group() == _impl::TableFriend::get_parent_group(*m_obj.get_table())));
```
tgoyne, line 76: This should use realm->read_group() rather than adding another function for the same thing.

finnschiermer: But read_group() appear to have quite some overhead....?

tgoyne: When the realm is already in a read transaction it just does the checks we do on every single property access anyway.

finnschiermer: Ok, I changed it to use read_group(), but can't help notice that read_group() potentially will do a begin_read(). So calling read_group() at the point I'm proposing may change behavior, because the different timing will allow begin_read() to attach to a different version. Are you fine with that @tgoyne ?

tgoyne: If read_group() starts a read transaction here then we're aborting the process anyway because this assertion is going to fail.

PR #1098, opened 2020-09-24
Expose credentials::function with string argument

nirinchev:
.NET doesn't have a way to pass a bson document down to native without essentially serializing it to json in managed code, then parsing it in the native wrappers, only to have OS serialize it back to json. This introduces an overload that eliminates the extra deserialization and serialization step.

Based on #1094, but will rebase on v10 once that goes in.

### DIFF
--- a/src/sync/app_credentials.cpp
+++ b/src/sync/app_credentials.cpp
@@ -161,6 +161,15 @@ AppCredentials AppCredentials::function(const bson::BsonDocument& payload)
                           });
 }
 
+AppCredentials AppCredentials::function(const std::string serialized_document)
+{
+    return AppCredentials(AuthProvider::FUNCTION, 
+        [=] { 
+            return serialized_document; 
+        });
+}
+
+
 AppCredentials AppCredentials::user_api_key(std::string api_key)
 {
     return AppCredentials(AuthProvider::USER_API_KEY,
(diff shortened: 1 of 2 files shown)


src/sync/app_credentials.cpp
```diff
@@ -161,6 +161,15 @@ AppCredentials AppCredentials::function(const bson::BsonDocument& payload)
                           });
 }
 
+AppCredentials AppCredentials::function(const std::string serialized_document)
```
jsflax, line 164: Could have this parameter be a const ref (`const std::string&`) since the lambda captures a copy of it anyways.

src/sync/app_credentials.hpp
```diff
@@ -102,6 +102,10 @@ struct AppCredentials {
     // The payload is a MongoDB document as json
     static AppCredentials function(const bson::BsonDocument& payload);
 
+    // Construct and return credentials with the payload.
+    // The payload is a MongoDB document as json
+    static AppCredentials function(const std::string serialized_document);
```
jsflax, line 107: I would call this `serialized_payload` for consistency.

src/sync/app_credentials.cpp
nirinchev: Hm... I guess it could, but why are the other credentials methods taking `std::string`s directly? Is it an oversight or is it due to some subtle C++-ness that I don't understand?

jsflax (approved):

PR #1101, opened 2020-09-25
Update API calls to use the new Realm endpoints as per Stitch rebranding

edualonso:
Updated parameters as per the Stitch rebrand to Realm Deprecation/EOL Notes:
- `stitch_request` to `baas_request`
- `stitch_at ` to `baas_at`


jsflax (approved):
As long as tests are passing, this LGTM 👍

cmelchior (approved):
We should also modify `default_base_url` in `app.cpp` to point to `realm.mongodb.com` instead of `stitch.mongodb.realm`, but otherwise it looks good 👍

edualonso:
Tests pass after updating the last string. I will merge it.

PR #1103, opened 2020-09-30
Fix a case where notification skipping didn't work

tgoyne:
Notification skipping relies on the fact that the notifier can't be running while the Realm is in a write transaction, which is partially guaranteed by the fact that starting a write transaction blocks until notifications are ready and no other writes can happen while we're in a write transaction. However, other things can also wake up the notification worker thread, so we need to skip running existing notifiers when that happens.


leemaguire (approved):
Nice 👍🏼

PR #1104, opened 2020-10-01
Fix bug around change of nullability

jedelbo:
If a property both had a change in nullability and was set as the new primary key property, we ended up using the wrong ColKey.
When changing the nullability, the col key changes, but the value in the change set was taken before migration started.

Fixes https://github.com/realm/realm-js/issues/3270

### DIFF
--- a/src/object_store.cpp
+++ b/src/object_store.cpp
@@ -169,6 +169,16 @@ void make_property_required(Group& group, Table& table, Property property)
     property.column_key = add_column(group, table, property).value;
 }
 
+void add_search_index(Table& table, Property property)
+{
+    table.add_search_index(table.get_column_key(property.name));
+}
+
+void remove_search_index(Table& table, Property property)
+{
+    table.remove_search_index(table.get_column_key(property.name));
+}
+
 } // anonymous namespace
 
 void ObjectStore::set_schema_version(Group& group, uint64_t version) {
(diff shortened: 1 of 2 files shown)


src/object_store.cpp
```diff
@@ -531,8 +541,8 @@ static void apply_pre_migration_changes(Group& group, std::vector<SchemaChange>
         void operator()(MakePropertyNullable op) { make_property_optional(table(op.object), *op.property); }
         void operator()(MakePropertyRequired op) { make_property_required(group, table(op.object), *op.property); }
         void operator()(ChangePrimaryKey op) { table(op.object).set_primary_key_column(ColKey{}); }
-        void operator()(AddIndex op) { table(op.object).add_search_index(op.property->column_key); }
-        void operator()(RemoveIndex op) { table(op.object).remove_search_index(op.property->column_key); }
+        void operator()(AddIndex op) { add_search_index(table(op.object), *op.property); }
+        void operator()(RemoveIndex op) { remove_search_index(table(op.object), *op.property); }
```
tgoyne, line 545: I think this is the only place that has to look up the properties by name, and in all of the other places modified would only serve to mask bugs. apply_non_migration_changes() and apply_additive_changes() must not change the key of any existing columns. apply_post_migration_changes() is called using the new schema read after the migration block and should have the new keys for columns. replace_column() getting called twice for a single column would also be a bug.

I guess create_initial_tables() should maybe change too since that's just hopefully-dead code there to repair files not created by object store?

jedelbo: Ok - got the point. Removed the changes from the 4 functions you mention.

tgoyne (approved):

Issue #1065, opened 2020-08-11
Prevent cycles in schemas for embedded objects


RealmBot:
Currently embedded objects can be defined with cycles - e.g. an embedded object linking to itself. This is not supported in JSON schema and thus sync. We should disallow such schemas either completely (allowing us to throw a compile-time error) or for synchronized Realms (resulting in a runtime error).

RealmBot:
➤ Brian Munkholm commented:

Can this be done just in ObjStore or is support in each SDK needed as well?

RealmBot:
➤ Christan Melchior commented:

Given that we historically try to avoid having to much difference between Synced and Normal Realms, I would probably prefer we disallow schema cycles between embedded classes completely, which enable us to catch these at compile time. Embedded Objects already have enough caveats, so adding more to the mix would make it even harder for people to understand exactly how they work.

RealmBot:
➤ Christan Melchior commented:

It should also be in ObjectStore, but that can only produce runtime exceptions. Any compile-time checks would need to be done by each SDK.

RealmBot:
➤ Ian Ward commented:

In speaking with [~nikola.irinchev] and [~redbeard0531] support for this appears to be explicitly defined as a Goal in the Scope here:
{quote}An embedded object can also recursively embed objects of the same type.
For example, a “Dog” embedded object class can contain a field named “puppies” which is a list of “Dog”.{quote}
(See bullet directly above this heading: https://docs.google.com/document/d/1VHNKaieX7DL6ZYJd7-swTqQWzAbrxbUWIM6oTEn7G2k/edit#heading=h.li1hf5o5yox6 ( https://docs.google.com/document/d/1VHNKaieX7DL6ZYJd7-swTqQWzAbrxbUWIM6oTEn7G2k/edit#heading=h.li1hf5o5yox6 ))

We need to get an estimate from the Cloud team on what it would take to implement this in the server.

RealmBot:
➤ Brian Munkholm commented:

[~nikola.irinchev] Can you take this with the Cloud team? /cc [~rahul.singhania]

RealmBot:
➤ Adam Chelminski commented:

I just looked into this and very roughly, I think the work involved would be:

[Realm Cloud Core]

  * support recursion in our fork of gojsonschema (see: https://json-schema.org/understanding-json-schema/structuring.html#recursion)
  * add support for recursive embedded schemas in baas's schema generation and validation framework

[Realm Cloud Sync]

[...]

PR #1112, opened 2020-10-06
schema check for embedded object loops

ironage:
Fixes https://github.com/realm/realm-object-store/issues/1065

### DIFF
--- a/src/schema.cpp
+++ b/src/schema.cpp
@@ -80,14 +80,12 @@ Schema::const_iterator Schema::find(ObjectSchema const& object) const noexcept
 
 namespace {
 
-struct CheckObjectPath
-{
+struct CheckObjectPath {
     const ObjectSchema& object;
     std::string path;
 };
 
-struct CheckObjectPathHash
-{
+struct CheckObjectPathHash {
     size_t operator()(CheckObjectPath const& o) const
     {
         return std::hash<std::string>{}(o.object.name);
@@ -108,17 +106,16 @@ std::string do_check(Schema const& schema, const ObjectSchema& start)
 
     while (to_visit.size() > 0) {
         auto current = to_visit.begin();
-        auto insert_pair = visited.insert(current->object.name);
-        if (!insert_pair.second && current->object.is_embedded) {
-            return current->path;
-        }
-        for (auto prop : current->object.persisted_properties) {
+        visited.insert(current->object.name);
+        for (auto& prop : current->object.persisted_properties) {
             if (prop.type == PropertyType::Object) {
                 auto it = schema.find(prop.object_type);
-                REALM_ASSERT(it != schema.end()); // this will be there if the schema is otherwise valid
+                REALM_ASSERT(it != schema.end()); // this succeeds if the schema is otherwise valid
                 auto next_path = current->path + "." + prop.name;
-                auto insert_pair = to_visit.insert(CheckObjectPath{*it, next_path});
-                if (!insert_pair.second && it->is_embedded) {
+                if (visited.find(prop.object_type) == visited.end()) {
+                    to_visit.insert(CheckObjectPath{*it, next_path});
+                }
+                else if (it->is_embedded) {
                     return next_path;
                 }
             }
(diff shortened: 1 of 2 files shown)


src/schema.cpp
```diff
@@ -77,6 +78,70 @@ Schema::const_iterator Schema::find(ObjectSchema const& object) const noexcept
     return const_cast<Schema *>(this)->find(object);
 }
 
+namespace {
+
+struct CheckObjectPath
+{
```
tgoyne, line 84:
```suggestion
struct CheckObjectPath {
```

Only functions have a newline before {

src/schema.cpp
```diff
@@ -77,6 +78,70 @@ Schema::const_iterator Schema::find(ObjectSchema const& object) const noexcept
...
+    const ObjectSchema& object;
+    std::string path;
+};
+
+struct CheckObjectPathHash
+{
```
tgoyne, line 90:
```suggestion
struct CheckObjectPathHash {
```

src/schema.cpp
```diff
@@ -77,6 +78,70 @@ Schema::const_iterator Schema::find(ObjectSchema const& object) const noexcept
...
+    std::unordered_set<std::string> visited;
+    to_visit.insert(CheckObjectPath{start, start.name});
+
+    while (to_visit.size() > 0) {
+        auto current = to_visit.begin();
+        auto insert_pair = visited.insert(current->object.name);
+        if (!insert_pair.second && current->object.is_embedded) {
+            return current->path;
+        }
+        for (auto prop : current->object.persisted_properties) {
```
tgoyne, line 115:
```suggestion
        for (auto& prop : current->object.persisted_properties) {
```

src/schema.cpp
```diff
@@ -77,6 +78,70 @@ Schema::const_iterator Schema::find(ObjectSchema const& object) const noexcept
...
+}
+
+void check_for_embedded_objects_loop(Schema const& schema, std::vector<ObjectSchemaValidationException>& exceptions)
+{
+    for (auto const& object : schema) {
+        if (object.is_embedded) {
+            std::string loop = do_check(schema, object);
+            if (!loop.empty()) {
+                exceptions.push_back(util::format("Cycles containing embedded objects are not currently supported: '%1'", loop));
+                return;
```
tgoyne, line 138: This shouldn't break early as we want to report all schema errors and not just the first one.

src/schema.cpp
```diff
@@ -77,6 +78,70 @@ Schema::const_iterator Schema::find(ObjectSchema const& object) const noexcept
...
+        }
+        for (auto prop : current->object.persisted_properties) {
+            if (prop.type == PropertyType::Object) {
+                auto it = schema.find(prop.object_type);
+                REALM_ASSERT(it != schema.end()); // this will be there if the schema is otherwise valid
+                auto next_path = current->path + "." + prop.name;
+                auto insert_pair = to_visit.insert(CheckObjectPath{*it, next_path});
+                if (!insert_pair.second && it->is_embedded) {
+                    return next_path;
+                }
```
tgoyne, line 123: This whole check seems to be redundant? Couldn't to_visit just be a queue with the duplicate check only being the insert into visited?

src/schema.cpp
ironage: Ah ok, I wasn't sure about that. I changed it so that there will be one report for each embedded object involved.

src/schema.cpp
ironage: Agreed. I cleaned this up so there's only one check now.

src/schema.cpp
```diff
@@ -77,6 +78,66 @@ Schema::const_iterator Schema::find(ObjectSchema const& object) const noexcept
...
+
+inline bool operator==(const CheckObjectPath& lhs, const CheckObjectPath& rhs)
+{
+    return lhs.object.name == rhs.object.name;
+}
+
+// a non-recursive search that returns a path to any embedded object that has multiple paths from the start
+std::string do_check(Schema const& schema, const ObjectSchema& start)
+{
+    std::unordered_set<CheckObjectPath, CheckObjectPathHash> to_visit;
```
tgoyne, line 103: This could now be just `std::queue<CheckObjectPath>` since there's no lookups in it.

tgoyne (approved):

PR #1113, opened 2020-10-06
Refactor remote_mongo_collection to expose raw bson callbacks

nirinchev:
This PR exposes `RemoteMongoCollection` API that return the raw bson rather than attempt to parse it into a concrete C++ type. This has the following benefits (mainly for .NET):

1. In .NET we don't have a way to interop with the c++ API, which means that we need to wrap every return type in its own struct with a native and managed counterpart. By returning the raw bson, we are just going to transmit the string which dramatically reduces the complexity of the feature.
2. For a lot of the return types, we're going to have to serialize them anyway since they're unstructured or of unknown type (e.g. the inserted id).
3. The post processing OS does on the raw response is quite minimal and can easily be replicated in .NET.

The PR may seem fairly massive, but in reality all of it is moving code from the `collection::XXX` methods to `collection::XXX_bson` methods and no new functionality has been introduced.

Additionally, I changed the `handle_xxx_response` methods to return a func as that cleaned up the API quite nicely, but don't have strong feelings about this change.

### DIFF
--- a/src/sync/remote_mongo_collection.cpp
+++ b/src/sync/remote_mongo_collection.cpp
@@ -22,109 +22,89 @@
 namespace realm {
 namespace app {
 
-static void handle_delete_count_response(util::Optional<AppError> error,
-                                         util::Optional<bson::Bson> value,
+inline std::function<void(util::Optional<AppError>, util::Optional<bson::Bson>)> handle_delete_count_response(
                                          std::function<void(uint64_t, util::Optional<AppError>)> completion_block)
 {
-    if (value && !error) {
+    return [completion_block](util::Optional<AppError> error, util::Optional<bson::Bson> value) {
+        if (value && !error) {
+            try {
+                auto document = static_cast<bson::BsonDocument>(*value);
+                auto count = static_cast<int32_t>(document["deletedCount"]);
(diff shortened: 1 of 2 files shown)


src/sync/remote_mongo_collection.cpp
```diff
@@ -22,109 +22,89 @@
 namespace realm {
 namespace app {
 
-static void handle_delete_count_response(util::Optional<AppError> error,
-                                         util::Optional<bson::Bson> value,
+inline std::function<void(util::Optional<AppError>, util::Optional<bson::Bson>)> handle_delete_count_response(
```
jsflax, line 25: Cool choice 👍

jsflax (commented):
Looks good 👍 I know all the old methods call into the new ones, but I'd consider writing some unit tests that call into the new methods as well.

nirinchev:
I'll redo that after https://github.com/realm/realm-object-store/pull/1114 gets merged and will add a few tests for the new methods, but yeah, since the old methods call into the new ones, I won't go overboard with the tests 😄

nirinchev:
Closing in favor of https://github.com/realm/realm-object-store/pull/1117.

PR #1115, opened 2020-10-06
ROBJSTORE-124 insert functions should return non-ObjectID inserted IDs

jbreams:
The insertedId/insertedIds returned by the mongo server for calls to insert_one/insert_many can be any valid value. This changes our implementation of those functions to reflect that.


leemaguire (approved):
Nice 👍🏼

PR #1116, opened 2020-10-07
ROBJSTORE-122 Set precision to max when serializing doubles to extended json

jbreams:


jsflax (approved):
Nice! Wish I knew about this when we were building it 😅

PR #1117, opened 2020-10-08
Re-expose _bson alternatives for MongoCollection methods

nirinchev:


src/sync/remote_mongo_collection.cpp
```diff
@@ -22,114 +22,93 @@
 namespace realm {
 namespace app {
 
-static void handle_delete_count_response(util::Optional<AppError> error,
-                                         util::Optional<bson::Bson> value,
-                                         std::function<void(uint64_t, util::Optional<AppError>)> completion_block)
+inline std::function<void(util::Optional<AppError>, util::Optional<bson::Bson>)> handle_delete_count_response(
```
jbreams, line 25: A few things, I don't think this needs to be inlined. Also, can we give this return type a shorter name? Maybe:

using ResponseHandler = std::function<void(util::Optional<AppError>, util::Optional<bson::Bson>)>;
ResponseHandler handle_delete_count_response(...)
{
    ...
}

jbreams: Also, can we put all these bare functions in an anonymous namespace?

src/sync/remote_mongo_collection.cpp
```diff
@@ -22,114 +22,93 @@
...
-    return completion_block(0, error);
+        return completion_block(0, error);
+    };
 }
 
-static void handle_update_response(util::Optional<AppError> error,
-                                   util::Optional<bson::Bson> value,
-                                   std::function<void(MongoCollection::UpdateResult,
-                                                      util::Optional<AppError>)> completion_block)
+inline std::function<void(util::Optional<AppError>, util::Optional<bson::Bson>)> handle_update_response(
```
jbreams, line 44: Same comments as above.

src/sync/remote_mongo_collection.cpp
```diff
@@ -22,114 +22,93 @@
...
+    std::function<void(MongoCollection::UpdateResult, util::Optional<AppError>)> completion_block)
 {
+    return [completion_block](util::Optional<AppError> error, util::Optional<bson::Bson> value) {
+        if (error) {
+            return completion_block({}, error);
```
jbreams, line 49: completion_block(util::none, error);

src/sync/remote_mongo_collection.cpp
```diff
@@ -22,114 +22,93 @@
...
+        catch (const std::exception& e) {
+            return completion_block({}, AppError(make_error_code(JSONErrorCode::bad_bson_parse), e.what()));
+        }
+    };
 }
 
-static void handle_document_response(util::Optional<AppError> error,
-                                     util::Optional<bson::Bson> value,
-                                     std::function<void(util::Optional<bson::BsonDocument>, util::Optional<AppError>)> completion_block)
+inline std::function<void(util::Optional<AppError>, util::Optional<bson::Bson>)> handle_document_response(
```
jbreams, line 75: Same questions about inlining and using shorter names.

src/sync/remote_mongo_collection.cpp
```diff
@@ -407,24 +261,230 @@ void MongoCollection::find_one_and_replace(const bson::BsonDocument& filter_bson
...
+                                          std::function<void(util::Optional<bson::BsonDocument>, util::Optional<AppError>)> completion_block)
+{
+    find_one_and_delete(filter_bson, {}, completion_block);
+}
+
+void MongoCollection::find_bson(const bson::BsonDocument& filter_bson,
+                                FindOptions options,
+                                std::function<void(util::Optional<AppError>, util::Optional<bson::Bson>)> completion_block)
+{
+    try {
```
jbreams, line 278: If you want to just wrap the entire function in a try catch block, you can make this a little cleaner by doing:
```
void FunctionDeclaration(foo bar, bizz buzz) try
{
   {function body}
} catch(const std::exception& e) {
   { catch block 1 }
} catch(const OtherException& e) {
   { catch block 2 }
}
```

See https://en.cppreference.com/w/cpp/language/function-try-block for more details.

src/sync/remote_mongo_collection.cpp
```diff
@@ -407,24 +261,230 @@ void MongoCollection::find_one_and_replace(const bson::BsonDocument& filter_bson
...
+
+        if (options.projection_bson) {
+            base_args["project"] = *options.projection_bson;
+        }
+
+        if (options.sort_bson) {
+            base_args["sort"] = *options.sort_bson;
+        }
+
+        m_service->call_function("find",
```
jbreams, line 294: Hmm, is this rebased on my patch that binds the MongoCollection to a SyncUser? Check out how this looks in the current v10 branch https://github.com/realm/realm-object-store/blob/6b44209e6fcac0137e193c96444f93c50d184d06/src/sync/remote_mongo_collection.cpp#L115

jbreams (changes requested):

src/sync/remote_mongo_collection.cpp
nirinchev: Good catch! Must have botched something when copy-pasting stuff.

src/sync/remote_mongo_collection.cpp
nirinchev: `MongoCollection::UpdateResult` is not optional (and I didn't want to make changes to the current API).

src/sync/remote_mongo_collection.cpp
nirinchev: That's some overly fancy syntax here 😄

nirinchev:
Ping @jsflax @jbreams.

src/sync/remote_mongo_collection.cpp
jbreams: Gotcha, I missed that.

jbreams (changes requested):
Just one more ask, otherwise this lgtm.

src/sync/remote_mongo_collection.cpp
```diff
@@ -19,117 +19,101 @@
 #include "sync/remote_mongo_collection.hpp"
 #include "realm/util/uri.hpp"
 
-namespace realm {
-namespace app {
+using namespace realm;
```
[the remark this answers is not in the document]
nirinchev: Ugh, my bad, sorry 🤦‍♂️

jbreams:
🚢 🇮🇹

PR #1118, opened 2020-10-09
ROBJSTORE-98 Implement precise and unbatched notification of sync completion events

jbreams:
Opening this in lieu of #1054.

### DIFF
--- a/src/sync/sync_session.cpp
+++ b/src/sync/sync_session.cpp
@@ -124,7 +124,7 @@ struct sync_session_states::Active : public SyncSession::State {
         SyncSession::CompletionCallbacks callbacks_to_register;
         std::swap(session.m_completion_callbacks, callbacks_to_register);
 
-        for (auto& [ id, callback_tuple ] : callbacks_to_register) {
+        for (auto& [id, callback_tuple] : callbacks_to_register) {
             session.add_completion_callback(lock, std::move(callback_tuple.second), callback_tuple.first);
         }
     }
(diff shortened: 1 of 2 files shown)


jbreams:
@tgoyne let me know if you have time to review this.

src/sync/sync_session.cpp
```diff
@@ -125,10 +121,12 @@ struct sync_session_states::Active : public SyncSession::State {
...
         // potentially add a redundant callback if we're coming from the Dying
         // state, but that's okay (we won't call the user callbacks twice).
-        if (!session.m_upload_completion_callbacks.empty())
-            session.add_completion_callback(_impl::SyncProgressNotifier::NotifierType::upload);
-        if (!session.m_download_completion_callbacks.empty())
-            session.add_completion_callback(_impl::SyncProgressNotifier::NotifierType::download);
+        SyncSession::CompletionCallbacks callbacks_to_register;
+        std::swap(session.m_completion_callbacks, callbacks_to_register);
+
+        for (auto& [ id, callback_tuple ] : callbacks_to_register) {
```
tgoyne, line 127: Spacing on structured binding is inconsistent here; some have spaces in [] and some don't. I think no spaces is the normal thing?

src/sync/sync_session.cpp
```diff
@@ -379,23 +360,17 @@ void SyncSession::handle_error(SyncError error)
...
+                // when the session becomes active again.
                 {
                     std::unique_lock<std::mutex> lock(m_state_mutex);
                     m_force_client_resync = true;
 
-                    ++m_client_resync_counter;
-                    auto download_handlers = std::move(m_download_completion_callbacks);
-                    auto upload_handlers = std::move(m_upload_completion_callbacks);
-
+                    auto completion_callbacks = std::move(m_completion_callbacks);
```
tgoyne, line 370: In the other spots we do this you changed it to swapping with a local which is probably better, so this one should as well.

tgoyne (approved):
The logic looks correct to me and just a few trivial nits.

src/sync/sync_session.cpp
jbreams: Will do. I also added a FIXME comment since I realized advance_state is not exception safe so we should make sure that we always move the callbacks back in place by using a guard type, but I don't think we need to fix that in this PR.

Issue #923, opened 2020-02-20
Enforce PK restrictions for synchronized Realms


RealmBot:
We'll be enforcing some restrictions on the PK type and name for synchronized Realms only. Since there's no obvious compile-time way to do that, the sanest approach seems to be to do it is to validate the schema before opening a synchronized Realm.

Internal doc link: https://docs.google.com/document/d/1eGwS7kid-zjBI3Q79ogyZ9mJk8JLLEmlVCbcOczcjuU/edit#

RealmBot:
➤ Brian Munkholm commented:

Mathias will align with Eric to not do this and allow all existing PK's.

RealmBot:
➤ Brian Munkholm commented:

The restriction will be kept until private beta, but lifted before public beta.
So this task will not be done.

RealmBot:
➤ Nikola Irinchev commented:

I'm reopening this because even if we don't enforce a particular type limitation, we need to enforce that all top-level objects have PK property defined and that it is called _id. This will be enforced on the server, but the error message is not exactly obvious - it tells the client that there's a type mismatch (server expects ObjectId but gets GlobalKey which is the identity type for objects without PK).

RealmBot:
➤ Brian Munkholm commented:

[~nikola.irinchev] Is this still relevant? Are there changes needed in the mini scope given new PK types allowed?

RealmBot:
➤ Nikola Irinchev commented:

As mentioned in my last comment - object store should throw an error if a top level object doesn't have the _id primary key. 

I don't believe this has been done, so the ticket is still relevant.

ironage:
I have requested better server generated error messages for schema errors here: https://jira.mongodb.org/browse/REALMC-7052 currently I only see "Bad changeset (DOWNLOAD)" for these types of errors.

nirinchev:
Why do we need better server messages though (I mean, we do, but this is kind of orthogonal to the OS ticket)?

ironage:
From my perspective, PK restrictions are entirely server side, and therefore the error should be server generated. If we implement the restrictions client side, the server could potentially lift these restrictions and then the clients would be left artificially enforcing an obsolete rule.

nirinchev:
[...]

PR #1120, opened 2020-10-10
non _id pks generate a schema error for sync realms

ironage:
Requested by https://github.com/realm/realm-object-store/issues/923

Hopefully this doesn't break too many SDK level tests.

### DIFF
--- a/tests/sync/app.cpp
+++ b/tests/sync/app.cpp
@@ -1745,7 +1745,7 @@ TEST_CASE("app: sync integration", "[sync][app]") {
         });
         return app;
     };
-    auto setup_and_get_config = [&base_path](std::shared_ptr<App> app) -> realm::Realm::Config {
+    auto setup_and_get_config = [&base_path](std::shared_ptr<App> app, util::Optional<Schema> schema = util::none) -> realm::Realm::Config {
         realm::Realm::Config config;
         config.sync_config = std::make_shared<realm::SyncConfig>(app->current_user(), bson::Bson("foo"));
         config.sync_config->client_resync_mode = ClientResyncMode::Manual;
@@ -1768,7 +1768,7 @@ TEST_CASE("app: sync integration", "[sync][app]") {
             realm::Property("lastName", PropertyType::String),
             realm::Property("realm_id", PropertyType::String | PropertyType::Nullable)
         });
-        config.schema = realm::Schema({dog_schema, person_schema});
+        config.schema = schema ? *schema : realm::Schema({dog_schema, person_schema});
         return config;
     };
     auto get_dogs = [&](realm::SharedRealm r, std::shared_ptr<SyncSession> session) -> Results {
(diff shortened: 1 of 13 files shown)


tests/sync/app.cpp
```diff
@@ -1908,6 +1908,67 @@ TEST_CASE("app: sync integration", "[sync][app]") {
...
+        Schema schema = {dog_schema, person_schema};
+        auto config = setup_and_get_config(app, schema);
+        std::string exception_message;
+        try {
+            realm::Realm::get_shared_realm(config);
+        }
+        catch (const std::exception& e) {
+            exception_message = e.what();
+        }
+        CHECK(exception_message.find("The primary key property on a synchronized Realm must be named '_id' but found 'my_primary_key' for type 'Dog'") != std::string::npos);
```
tgoyne, line 1940: This is `REQUIRE_THROWS_WITH`

tests/sync/app.cpp
```diff
@@ -1908,6 +1908,67 @@ TEST_CASE("app: sync integration", "[sync][app]") {
...
+        });
+        const auto person_schema = realm::ObjectSchema("Person", {
+            realm::Property("_id", PropertyType::ObjectId | PropertyType::Nullable, true),
+            realm::Property("age", PropertyType::Int),
+            realm::Property("dogs", PropertyType::Object | PropertyType::Array, "Dog"),
+            realm::Property("firstName", PropertyType::String),
+            realm::Property("lastName", PropertyType::String),
+            realm::Property("realm_id", PropertyType::String | PropertyType::Nullable)
+        });
+        Schema schema = {dog_schema, person_schema};
```
tgoyne, line 1931: This should define the schema in the same way as all of the other tests rather than doing something different.

tests/sync/app.cpp
```diff
@@ -1908,6 +1908,67 @@ TEST_CASE("app: sync integration", "[sync][app]") {
         util::EventLoop::main().run_until([&]{ return error_did_occur.load(); });
         REQUIRE(error_did_occur.load());
     }
+
+    SECTION("invalid pk schema error handling")
+    {
```
tgoyne, line 1913: Only functions have a newline before brace.

tests/sync/session/session.cpp
```diff
@@ -856,27 +865,28 @@ TEST_CASE("sync: client resync") {
...
     }
 
     config.sync_config->client_resync_mode = ClientResyncMode::Recover;
 
     SECTION("add table without pk in recovered transaction") {
-        auto realm = trigger_client_reset([](auto& realm) {
+        auto realm = trigger_client_reset([&](auto& realm) {
             realm.update_schema({
                 {"object2", {
+                    {"_id", PropertyType::Int, Property::IsPrimary{true}},
```
tgoyne, line 885: This is no longer testing what the test says it's testing. This section should probably just be deleted since it's no longer applicable.

tests/sync/app.cpp
ironage: There's a prefix to the exception message that I didn't want to check for, but I found `REQUIRE_THROWS_CONTAINING` elsewhere in the code so I moved to using that.

tests/sync/app.cpp
ironage: Updated to use the existing schema and modify the properties manually which should make it more clear what is being tested instead of having to read the schema diff.

tests/sync/session/session.cpp
ironage: 👍

ironage:
Friendly reminder for reviews @tgoyne @nirinchev

nirinchev (approved):

tgoyne (approved):

Issue #1119, opened 2020-10-09
Access token is not refreshed when user uses the http API


nirinchev:
I encountered this in a .NET Core app - I logged in an anonymous user, then a few days later, I tried running the app again. When trying to use the remote mongo API, I get `invalid session`. When I inspected the user object, the access token had expired more than 5 days ago. So I see two orthogonal problems exposed by this:
1. When calling `app.LoginAsync` with anonymous credentials, we should probably attempt to refresh the access token before returning a cached user.
2. When calling any of the http methods, we should refresh the access token if it has expired.

jsflax:
We have logic that ensures that tokens are auto refreshed, so you've likely uncovered a bug here.

nirinchev:
I briefly reviewed the code and may be missing something, but the way I read it, in `App::handle_auth_failure`, we only handle http errors:

https://github.com/realm/realm-object-store/blob/6b44209e6fcac0137e193c96444f93c50d184d06/src/sync/app.cpp#L981

But in `check_for_errors`, we will return `service_error` if the content type is application/json:

https://github.com/realm/realm-object-store/blob/6b44209e6fcac0137e193c96444f93c50d184d06/src/sync/app_utils.hpp#L49-L64

This means that if the server responds with 401, but `application/json` content, we're constructing a service error which then circumvents the 401 check and prevents the token refresh.

nirinchev:
@leemaguire I vaguely remember you working on the token refresh code - can you confirm if my understanding is correct or if I'm missing some piece of the puzzle?

leemaguire:
@nirinchev Your understanding is correct. This does look like a bug. `static util::Optional<AppError> check_for_errors(const Response& response)` likely needs a rework because if something does have `application/json` as a header, it will try parse an error message from the body and http status code.

nirinchev:
Fixed by https://github.com/realm/realm-object-store/pull/1124

PR #1124, opened 2020-10-12
Add http_status_code as a param to AppError.

leemaguire:
This PR addresses an uncaught bug. The bug is where a http response which contains `application/json` would not write the http status to an AppError object. Therefor if the user got a 401 http response because of a bad token we would never be able to catch it in `App::handle_auth_failure` as there is no http status code to look up.

In this PR I've added `http_status_code` as an optional attribute of `AppError`. I have also wrapped the static `check_for_errors` method in a class.

Addresses: https://github.com/realm/realm-object-store/issues/1119

### DIFF
--- a/src/sync/generic_network_transport.hpp
+++ b/src/sync/generic_network_transport.hpp
@@ -21,6 +21,7 @@
 #define REALM_GENERIC_NETWORK_TRANSPORT_HPP
 
 #include <realm/util/to_string.hpp>
+#include <realm/util/optional.hpp>
 
 #include <functional>
 #include <iosfwd>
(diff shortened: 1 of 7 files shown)


src/sync/generic_network_transport.hpp
```diff
@@ -123,11 +124,17 @@ std::error_code make_client_error_code(ClientErrorCode) noexcept;
 struct AppError {
 
     std::error_code error_code;
+    util::Optional<std::error_code> http_status_code;
```
nirinchev, line 127: Does it make sense for this to be an error code as opposed to just int? The http error code is just an int value + the http category and the only place where we use the new field, we only read the int value.

nirinchev (approved):
Looks good, I only have a question/preference for using a simple int instead of an error code for the new field.

src/sync/generic_network_transport.hpp
leemaguire: Yeah, its a bit overkill, I'll change it to an int.

PR #1127, opened 2020-11-06
Add C++ SDK

jsflax:
Remove the `sync_metadata` abstraction in favour of a (somewhat spartan) C++ SDK. This is used internally by `sync`/`App` functionality. Tests displaying usage are also located in `tests/sdk.cpp`.

Ideally with some polish, it will be spun into its own fully fledged SDK. 

## Property
Add `sdk::Property` template specialisation class. Properties act as a facade for an underlying Realm type, and inherit
`PropertyBase` from their `Property` specialization struct.

Each Property contains an accessor (`AccessorBase`) that acts as a reader, writer, and translator between Core and Value
types. Accessor's have specialisations for each `Property` type. E.g., an `Accessor<std::string>` would understand how to
set and get core `realm::StringData` and `std::string`s.

Properties use the accessor to read/write actual values from/to the realm (once attached), and then emulate the functionality
of their specialized type, e.g.,:
```cpp
struct Property<std::string> {
   Property& operator +=(const std::string& other_value)  {
       if (!is_managed()) {
           m_unmanaged_value += other_value;
           return *this;
       }
       auto specialized_accessor = m_accessor.as<std::string>();
       auto core_value = specialized_accessor.get();
       auto actual_value = specialized_accessor.to_value_type(core_value);
       specialized_accessor.set(other_value + actual_value);
       return *this;
   }
};

struct Foo : Object<Foo> {
    REALM(std::string) my_str;

    REALM_EXPORT(my_str);
};

auto foo = realm.get_objects<Foo>().first();
realm.write([&foo] { foo.my_str += "hello world"; });
assert(foo.my_str == "hello world");
```

Properties generally hold a reference to an unmanaged version of the underlying type. If the Property is unmanaged, all requested operations should be conducted as normal on the underlying type.

## Object
Add `sdk::Object` class. `Object` is a base class for model objects representing data stored in Realms. User's define model classes by subclassing `Object` and adding properties to be managed. Any managed property should have its type wrapped in the `REALM` macro. Properties should then be `REALM_EXPORT`'d within the class body.
 ```cpp
 struct Dog : Object<Dog> {
    REALM(std::string) name;
    REALM(bool) adopted;

    REALM_EXPORT(name, adopted);
 };
 ```
 ### Supported property types
 - `std::string`
 -  `int`, `float`, and `double`
 - `bool`
 - `std::chrono::time_point`
 - `enum class`
 - `realm::ObjectId`
 - `X*` where  `X` is an `Object` subclass, creating a link
 - `realm::util::Optional<X>` where `X` is a supported property type
 - `EmbeddedObject`
 - `std::vector<X*>`, where `X` is an `Object` subclass, to model many-to-many relationships.
 - `std::vector<X>`, where `X` is a supported property type, including `EmbeddedObject`

## EmbeddedObject

`EmbeddedObject` is a base class used to define embedded Realm model objects.

The property types supported on `EmbeddedObject` are the same as for `Object`, except for that embedded objects cannot link to top-level objects, so `Object` and `std::vector<Object*>` properties are not supported (`EmbeddedObject` and
`std::vector<EmbeddedObject>` *are*).

Embedded objects cannot have primary keys or indexed properties.

 ```cpp
struct Dog: EmbeddedObject<Dog> {
   REALM(std::string) name;
   REALM(bool) adopted;

   REALM_EXPORT(name, adopted);
}

struct Owner: Object<Owner> {
   REALM(std::string) name;
   REALM(std::vector<Dog>) dogs;
   REALM_EXPORT(name, dogs);
}
```

## Link
Add `Property` specialisation for user inherited `Object` types. A link is an auto-updating link to another Realm Object. Links are declared as a class member as pointers to the actual Object.
```cpp
class Foo : Object<Foo> {
    REALM(std::string) my_str;

    REALM_EXPORT(my_str);
};

class Bar : Object<Bar> {
    REALM(Foo*) foo;
    REALM_EXPORT(foo);
};

Foo foo;
Bar bar;
bar.foo = &foo;
realm.write([&bar]{ realm.add_object(bar); });
realm.write([&bar]{ bar.foo->my_str = "hello world"; });
assert(bar.foo->my_str == "hello world");
assert(foo.my_str == "hello world");
```

This notation felt natural considering the actual meaning of what a Realm linked object is.

## List
Add `Property` specialisation for std::vector<X> where `X` is an `Object` subclass pointer (a Link List), or another supported property type including `EmbeddedObject`.

Lists are meant to be used as if they were plain old `std::vector`s.

## Realm
Add `Realm` wrapped around `realm::object_store::Realm`. This acts as a standard Realm type, with methods template for `Object` inheritors.

## Results
Add `Results` wrapped around `realm::object_store::Results`. This is a simple wrapper with methods templated for `Object` inheritors.

Notable Missing features:
- this is not rebased on monorepo
- `Object`s should no longer inherit from `AccessorBase` to avoid having to deal with virtual destructors. This will allow user `Objects` to use designated initialisers.
- Binary `Property` type (trivial to add)
- Collection observation (also trivial to add)
- Add as many method bodies to .cpp as possible (lots of logic is currently in the headers)
- Certain logic between how a sync user manages their sessions needs to be re-implemented
- More `std` library / general methods to add to `Property` facades
- More static assertions to provide users with valuable information about "out of bounds" actions


src/sync/app.cpp
```diff
@@ -501,149 +532,147 @@ void App::UserAPIKeyProviderClient::fetch_api_keys(std::shared_ptr<SyncUser> use
...
-                                                           get_optional<std::string>(profile_data, "last_name"),
-                                                           get_optional<std::string>(profile_data, "gender"),
-                                                           get_optional<std::string>(profile_data, "birthday"),
-                                                           get_optional<std::string>(profile_data, "min_age"),
-                                                           get_optional<std::string>(profile_data, "max_age")));
-
-            sync_user->set_state(SyncUser::State::LoggedIn);
-            m_sync_manager->set_current_user(sync_user->identity());
+            realm().write([this, &user, &profile_json, &identities_json] {
+                auto profile_data = value_from_json<nlohmann::json>(profile_json, "data");
```
leemaguire, line 645: nice!

src/sync/sync_manager.hpp
```diff
@@ -95,6 +93,36 @@ struct SyncClientConfig {
     SyncClientTimeouts timeouts;
 };
 
+// A facade for a metadata Realm object representing a pending action to be carried out upon a specific file(s).
+struct SyncFileAction : sdk::Object<SyncFileAction> {
+    enum class Action {
```
leemaguire, line 98: Cleans things up nicely 👍🏼

src/sdk/util.hpp
```diff
@@ -0,0 +1,155 @@
...
+} // namespace type_traits
+
+template <typename T>
+using EmbeddedObject = Property<T, util::type_traits::embedded_object>;
+
+static inline std::string demangle(const char* name) {
+    int status = -4; // some arbitrary value to eliminate the compiler warning
+
+    std::unique_ptr<char, void(*)(void*)> res {
+        abi::__cxa_demangle(name, NULL, NULL, &status),
```
leemaguire, line 44: Does this sort of call guarantee consistency between different platforms and vendors?

src/sdk/object.hpp
```diff
@@ -0,0 +1,616 @@
...
+        *this = std::move(o);
+    }
+
+    Property& operator=(Property&& o)
+    {
+        Property::~Property();
+        if (o.m_value) {
+            new (&m_value) std::unique_ptr<removed_pointer_t>(std::move(o.m_value));
+        }
+        m_unmanaged_value = o.m_unmanaged_value;
```
simonask, line 305: This is super undefined behavior. Accessing the memory of an object that has been destroyed is not legal, which can happen because the placement-new above is conditional.

leemaguire (commented):
Great job on this! I think it opens up a lot of doors w.r.t different applications and use cases (cars, small linux devices etc.), as well as cleaning up some of the sync code.

src/sdk/object.hpp
```diff
@@ -0,0 +1,616 @@
...
+
+    auto operator->() {
+        return value();
+    }
+
+    removed_pointer_t& operator*() const {
+        return *value();
+    }
+
+    operator bool() const {
```
simonask, line 349: Use `explicit operator bool`.

src/sdk/object.hpp
```diff
@@ -0,0 +1,616 @@
...
+// MARK: EmbeddedObjectAccessor
+template <typename T>
+struct Accessor<T, typename std::enable_if_t<util::is_embedded_v<T>>> : public AccessorBase
+{
+    using type_info = TypeInfo<T>;
+    using base = AccessorBase;
+    using base::base;
+    using decayed_t = std::decay_t<T>;
+
+    Accessor(const AccessorBase& o) : base(o)
```
simonask, line 549: Do you intend this class to be implicitly convertible from any other `Accessor<T>`?

src/sdk/optional.hpp
```diff
@@ -0,0 +1,127 @@
...
+    }
+};
+
+template <typename T>
+struct OptionalBase : public PrimitiveBase<Property<optional<T>>, optional<T>>
+{
+    using base = PrimitiveBase<Property<optional<T>>, optional<T>>;
+    using base::base;
+    using base::operator=;
+    operator T() const {
```
simonask, line 68: Why would you want this?

src/sdk/primitive.hpp
```diff
@@ -0,0 +1,152 @@
...
+ Base `Property` specialization for any primitive type. "Primitive"
+ is defined here as any non collection or polymorphic type, such as
+ int, long, ObjectId, time_point, std::string, etc..
+ */
+template <typename Impl, typename T>
+struct PrimitiveBase : PropertyBase
+{
+    using value_type = T;
+    using impl = Impl;
+    using PropertyBase::PropertyBase;
```
simonask, line 62: Do you really want any `PrimitiveBase` to be constructible from any `const PropertyBase&`?

src/sdk/util.hpp
```diff
@@ -0,0 +1,155 @@
...
+// trim from start (in place)
+static inline void ltrim(std::string &s) {
+    s.erase(s.begin(), std::find_if(s.begin(), s.end(), [](unsigned char ch) {
+        return !std::isspace(ch);
+    }));
+}
+
+// trim from end (in place)
+static inline void rtrim(std::string &s) {
+    s.erase(std::find_if(s.rbegin(), s.rend(), [](unsigned char ch) {
```
simonask, line 75: Does this actually work? `std::string::reverse_iterator` should not be convertible to `std::string::iterator`.

src/sdk.hpp
```diff
@@ -0,0 +1,187 @@
...
+#include "sdk/arithmetic.hpp"
+#include "sdk/list.hpp"
+#include "sdk/realm.hpp"
+#include "sdk/time_point.hpp"
+#include "sdk/object_id.hpp"
+#include "sdk/property.hpp"
+
+// MARK: Macro Utilities
+
+#define SECOND(a, b, ...) b
```
simonask, line 35: All these macros clobber the global macro namespace for any user. They have some pretty generic names, too.

src/sdk.hpp
```diff
@@ -0,0 +1,187 @@
...
+    realm::sdk::Property<arg, realm::sdk::util::type_traits::primary_key>
+
+/**
+ MARK: REALM
+ Wrap a cpp type in a Realm Property specialization.
+ For EmbeddedObjects, this wrapping will yield the original type, so that
+ users can access its fields directly.
+ */
+
+#define REALM(arg) \
```
simonask, line 158: Why this macro? It seems better to just typedef/rename `realm::sdk::util::conditional_property_type<arg>::type` to something more digestible.

src/sdk.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+/**
+ MARK: REALM_EXPORT
+ Add metadata to an Object about the exported properties. This will add:
+ - a method that provides access to a tuple of the exported properties
+ - an inlined Schema generated based on the exported properties
+ - an enum listing the Properties so that user's can better understand what has changed (during observation)
+ - an observe method to allow Object observation
+ */
+#define REALM_EXPORT(...) \
```
simonask, line 169: Why this macro? It seems that all of the stuff it does can be achieved with a `static constexpr` function defined by the user.

src/sdk/accessor.hpp
```diff
@@ -0,0 +1,106 @@
...
+    template <typename T>
+    const Accessor<T>& as() const
+    {
+        return static_cast<const Accessor<T>&>(*this);
+    }
+
+    template <typename T>
+    Accessor<T>& as()
+    {
+        return static_cast<Accessor<T>&>(*this);
```
simonask, line 91: Deeply unsafe. Must be either private or guarded in some way (`dynamic_cast` is one way, but since it comes with a performance overhead, it probably indicates a deeper design issue).

src/sdk/arithmetic.hpp
```diff
@@ -0,0 +1,94 @@
...
+    using typename base::value_type;
+    using base::operator=;
+    using base::is_managed;
+    using base::value;
+
+    friend base;
+    template <typename Impl>
+    friend struct ObjectBase;
+
+    ArithmeticBase& operator+=(ArithmeticT rhs) {
```
simonask, line 22: You probably want a const-reference here. This will currently increment and decrement the `SharedRealm` refcount unnecessarily.

simonask (commented):
Ok, so a lot to unpack here. I posted some immediate feedback from glancing over the code, but I don't have time for a really in-depth review.

I have some issues with the fundamental design - would be good discuss at some point. :-)

simonask:
Design-wise, I'll take the opportunity to weigh in with what I think would be a more attractive approach from a user's perspective that avoids excessive use of macros.

Consider this syntax:

```c++
class Bar;

class Foo : public realm::object {
public:
    realm::property<int64_t> m_int;
    realm::property<std::string> m_string;
    realm::property<Bar> m_bar;

    static constexpr auto realm_model() {
        return realm::model {
            "Foo",
            {
                {"int", &Foo::m_int, realm::property::is_primary_key},
                {"string", &Foo::m_string},
                {"bar", &Foo::m_bar},
            },
        };
    }
};

class Bar : public realm::embedded_object {
public:
    realm::property<float> m_float;
    realm::linking_objects m_owner;

    static constexpr auto realm_model() {
        return realm::model {
            "Bar",
            {
                {"float", &Bar::m_float},
                {&Bar::m_owner},
            },
        };
    }
};
```

Major points:

- Constexpr metadata allows compile-time type introspection, which allows for some interesting magic around attaching/detaching objects.
- The design can avoid virtual methods and inheritance-related cans-of-worms altogether.
- Inheritance between model classes can be supported.
- There is slightly more boilerplate, but this is usually necessary anyway to accomodate different naming conventions and property options (primary key flag, indexed flag, default values, etc.).
- We can avoid duplicating the `SharedRealm` instance in every property.
- We can support custom access methods for properties.
- Importantly (to me), we would favor template magic over macro magic.

src/sdk/optional.hpp
jsflax: To implicitly convert between one of our primitive Property specializations and its actual types.

ianpward:
@jsflax is this true?

`The property types supported on EmbeddedObject are the same as for Object, except for that embedded objects cannot link to top-level objects, so Object and std::vector<Object*> properties are not supported (EmbeddedObject and
std::vector<EmbeddedObject> are).`

My understanding is that embedded objects do allow links to other top-level objects, you just cannot link to an embedded object that is embedded in another object.

src/sdk/list.hpp
```diff
@@ -0,0 +1,404 @@
...
+        base::m_accessor.as<std::vector<T>>().add(value);
+    }
+
+    T& at(size_t idx)
+    {
+        if (!base::is_managed()) {
+            return m_value.at(idx);
+        }
+
+        m_value[idx] = base::m_accessor.as<std::vector<T>>().get(idx);
```
tgoyne, line 50: Everything involving `m_value` for managed lists seems invalid. There's nothing here that guarantees that `m_value.size() > idx` here. Returning a reference to an entry in `m_value` just allows nonsense (e.g. `list.at(0) += 1` will compile and mutate a value that can never be read again). The reference returned is only valid until the next index is read, so there's also no need to store anything more than a single value.

PR #1128, opened 2020-11-08
hold the mutex while calling notify_all

blagoev:
according to this https://github.com/realm/realm-core/blob/master/src/realm/util/interprocess_condvar.cpp#L459-L461
the caller should be holding the mutex before calling `notify_all`

In what other branch/repo this changes should be cherry-picked ?

### DIFF
--- a/src/impl/windows/external_commit_helper.cpp
+++ b/src/impl/windows/external_commit_helper.cpp
@@ -54,6 +54,7 @@ ExternalCommitHelper::~ExternalCommitHelper()
 
 void ExternalCommitHelper::notify_others()
 {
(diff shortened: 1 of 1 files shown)


fealebenpae (approved):
This needs to be cherry-picked in the master branch here so that v10 SDKs can pick it up, and in realm-core/monorepo so that vNext SDKs can pick it up.

blagoev:
@fealebenpae  Could the win32 CI hang be related to this

fealebenpae:
@blagoev I think it's very likely. I think additional debugging is warranted.

blagoev:
this seems not the correct fix. Closing this PR

PR #1129, opened 2020-11-09
fix SharedMemory on Win32 platforms

blagoev:
don't close the mapped file handle cause the mapped file can not be opened from another process. This fixes the sharing of memory needed to synchronize realm file access between windows processes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@ compile_commands.json
 
 # vim
 *.swp
+/openssl/
+/out/build/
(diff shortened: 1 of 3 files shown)


finnschiermer (commented):
I'm not super familiar with Windows API, but don't we need to still call CloseHandle() somewhere to avoid leaking handles?

blagoev:
good point will fix

finnschiermer (approved):
LGTM

PR #1133, opened 2020-11-12
Stop the sync client before tearing down the sync file manager

tgoyne:
The sync client may still be using that directory so it needs to be stopped first. This is probably the source of occasional make_dir() errors in the Cocoa sync tests.


jsflax (approved):
Nice catch 👍

PR #1135, opened 2020-11-13
Add subscription feature to App/User

jsflax:


src/sync/subscribable.hpp
```diff
@@ -0,0 +1,104 @@
...
+        : m_subscribable(subscribable)
+        , m_token(std::make_unique<uint64_t>(token))
+        {
+        }
+        Token(Token&& other)
+        : m_subscribable(std::move(other.m_subscribable))
+        , m_token(std::move(other.m_token))
+        {
+        }
+        Token& operator=(Token&& other) {
```
tgoyne, line 45: This looks like it could just be `= default`. If it can't be, it needs to be noexcept and actually return `*this`.

src/sync/subscribable.hpp
```diff
@@ -0,0 +1,104 @@
...
+            }
+        }
+
+        uint64_t value() const
+        {
+            return *m_token;
+        }
+    private:
+        std::reference_wrapper<Subscribable> m_subscribable;
+        std::unique_ptr<uint64_t> m_token;
```
tgoyne, line 65: This could just be a `uint64_t` and `m_subscribable` could be a pointer that's set to null when it's moved from.

src/sync/subscribable.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+        template <class U> friend struct Subscribable;
+    };
+
+    using observer_t = std::function<void(const T&)>;
```
tgoyne, line 70: `Observer`

src/sync/subscribable.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+    /// Subscribe to notifications for class type T. Any mutation to the T class
+    /// will trigger the observer. Notifying subscribers must be done manually
+    /// by the Subscribable.
+    /// @param observer callback to be called on mutation
+    /// @returns a token identifying the observer
+    [[maybe_unused]] Token subscribe(observer_t&& observer)
```
tgoyne, line 77: This is a very strange function to mark as maybe_unused as it seems like something where we'd specifically want a warning if the return value is unused?

src/sync/app.hpp
```diff
@@ -359,6 +360,7 @@ class App : public std::enable_shared_from_this<App>, public AuthRequestClient,
     std::string m_auth_route;
     uint64_t m_request_timeout_ms;
     std::shared_ptr<SyncManager> m_sync_manager;
+//    std::unordered_map<size_t, observer_t> m_subscribers;
```
tgoyne, line 363: commented code

src/sync/subscribable.hpp
jsflax: I was under the impression that we tried to avoid raw pointers, but yes, I agree, this seems like applicable an applicable case.

src/sync/subscribable.hpp
jsflax: I'll go with `[[nodiscard]]` then. I suppose I did `[[maybe_unused]]` for testing, but will get rid of it.

src/sync/subscribable.hpp
tgoyne: We avoid *owning* raw pointers. Pointers for non-owning nullable references are totally fine.

src/sync/app.cpp
```diff
@@ -628,6 +628,9 @@ void App::get_profile(std::shared_ptr<SyncUser> sync_user,
 
             sync_user->set_state(SyncUser::State::LoggedIn);
             m_sync_manager->set_current_user(sync_user->identity());
+            for (auto& [_, subscriber] : m_subscribers) {
```
leemaguire, line 631: Is there any way to add a facade over all of these `for`'s so it would become `emit_change_to_subscribers(*this)`?

leemaguire (approved):
Nice work! I've left some minor comments / suggestions

src/sync/app.cpp
```diff
@@ -763,11 +766,14 @@ void App::log_out(std::shared_ptr<SyncUser> user, std::function<void (Optional<A
         util::format("Bearer %1", refresh_token)
     });
 
-    do_request(req, [completion_block, req](Response response) {
+    do_request(req, [&, completion_block = std::move(completion_block)](Response response) {
```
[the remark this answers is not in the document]
jsflax: I don't think it matters that much.

src/sync/app.cpp
jsflax: Yeah, would clean things up a bit.

src/sync/subscribable.hpp
```diff
@@ -0,0 +1,110 @@
...
+    /// Token that identifies an observer.
+    /// Unsubscribes when deconstructed to
+    /// avoid dangling observers.
+    struct Token {
+        Token(Subscribable* subscribable, uint64_t token)
+        : m_subscribable(subscribable)
+        , m_token(token)
+        {
+        }
+        Token(Token&& other)
```
tgoyne, line 40: Needs to be noexcept.

src/sync/subscribable.hpp
```diff
@@ -0,0 +1,110 @@
...
+        : m_subscribable(std::move(other.m_subscribable))
+        , m_token(std::move(other.m_token))
+        {
+            other.m_subscribable = nullptr;
+        }
+        Token& operator=(Token&& other) = default;
```
tgoyne, line 46: Shifting to storing a raw pointer actually make it so that this can no longer be defaulted as the default move won't set the source token's pointer to null.

src/sync/subscribable.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+    /// Subscribe to notifications for class type T. Any mutation to the T class
+    /// will trigger the observer. Notifying subscribers must be done manually
+    /// by the Subscribable.
+    /// @param observer callback to be called on mutation
+    /// @returns a token identifying the observer
+    [[nodiscard]] Token subscribe(Observer&& observer)
+    {
+        static uint64_t m_token;
+        m_subscribers.insert({m_token, std::move(observer)});
```
tgoyne, line 78: This type probably needs a mutex which guards access to m_subscribers? App and User are thread-safe so the subscribe functions need to be usable from any thread as well.

src/sync/app.cpp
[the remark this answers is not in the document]
jbreams: Def don't capture `&` for callbacks that will complete after the current scope returns. If you're trying to capture just `this` you can do [this, completion_block = std::move(completion_block)] to effectively make the lambda a member function.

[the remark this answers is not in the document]
jbreams: Better yet, you can do `[anchor = shared_from_this(), std::move(completion_block)]` so that you explicitly extend the lifetime of app to the lifetime of the lambda.

src/sync/subscribable.hpp
```diff
@@ -0,0 +1,134 @@
...
+    }
+    /// Subscribe to notifications for class type T. Any mutation to the T class
+    /// will trigger the observer. Notifying subscribers must be done manually
+    /// by the Subscribable.
+    /// @param observer callback to be called on mutation
+    /// @returns a token identifying the observer
+    [[nodiscard]] Token subscribe(Observer&& observer)
+    {
+        std::lock_guard<std::mutex> lock(m_mutex);
+        static uint64_t m_token;
```
jbreams, line 98: I think this needs to be initialized to zero, or at least value initialized.

src/sync/subscribable.hpp
```diff
@@ -0,0 +1,134 @@
...
+    size_t subscribers_count() const
+    {
+        return m_subscribers.size();
+    }
+protected:
+    /// Emit a change event to all subscribers.
+    void emit_change_to_subscribers(const T& subject) const
+    {
+        std::lock_guard<std::mutex> lock(m_mutex);
+        for (auto& [_, subscriber] : m_subscribers) {
```
jbreams, line 123: `const auto&` ?

jbreams (changes requested):

src/sync/subscribable.hpp
jsflax: Yep, missed that 👍

jsflax:
Moving to monorepo.

PR #1136, opened 2020-11-13
Fix multi process sync

blagoev:
This fixes multi-process sync when one process writes to the realm and notifies the current process that changes occured.

### DIFF
--- a/src/impl/realm_coordinator.cpp
+++ b/src/impl/realm_coordinator.cpp
@@ -796,6 +796,14 @@ void RealmCoordinator::on_change()
     for (auto& realm : m_weak_realm_notifiers) {
         realm.notify();
     }
+
+#if REALM_ENABLE_SYNC
+    // Invoke realm sync if another process has notified for a change
+    if (m_sync_session) {
+        auto versionId = m_db->start_read()->get_version_of_current_transaction();
+        SyncSession::Internal::nonsync_transact_notify(*m_sync_session, versionId.version);
+    }
+#endif
 }
 
 namespace {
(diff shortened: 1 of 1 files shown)


src/impl/realm_coordinator.cpp
```diff
@@ -796,6 +796,14 @@ void RealmCoordinator::on_change()
     for (auto& realm : m_weak_realm_notifiers) {
         realm.notify();
     }
+
+#if REALM_ENABLE_SYNC
+    // Invoke realm sync if another process has notified for a change
+    if (m_sync_session) {
+        auto versionId = m_db->start_read()->get_version_of_current_transaction();
```
tgoyne, line 803: `m_db->get_version_of_latest_snapshot()` skips creating and destroying a read transaction

blagoev: thanks will fix

blagoev: @tgoyne This comment was bugging me. Probably I could fix this point as well ?
https://github.com/realm/realm-object-store/blob/d8ff6335be0872270489917edceeec5628e7a6fa/src/impl/realm_coordinator.cpp#L967

tgoyne: That spot we need a VersionID rather than just the version number. I guess it could use get_version_of_latest_snapshot() for the check to see if it's a spurious wakeup and then only get the VersionID if it isn't.

blagoev: It seems that I could do it (the index seems not used anywhere after that point) but will need to construct a new VersionID only from the version_type which might introduce bugs later if the index is used somewhere from that semi VersionID

blagoev: yep that's good optimization

tgoyne: `version` is eventually passed to `advance_read()`, which needs the index.

blagoev: I won't change that since I am worried changing that to non atomic read of the version.

blagoev: but I may be overthinking it

tgoyne (approved):

PR #1141, opened 2020-11-18
update embedded object cycle detection

ironage:
It was [reported](https://jira.mongodb.org/browse/RJAVA-710) that a cycle is incorrectly detected in the case of a diamond pattern of paths; eg. given the relationship graph A->B->C, and A->D->C. 

Also fixed detecting cycles through top level objects, as this should be allowed by the server. Eg. Given a top level object with an embedded object which also has a link to the same top level object, this is not considered a loop we should restrict.

### DIFF
--- a/src/schema.cpp
+++ b/src/schema.cpp
@@ -82,29 +82,35 @@ Schema::const_iterator Schema::find(ObjectSchema const& object) const noexcept
 namespace {
 
 struct CheckObjectPath {
-    const ObjectSchema& object;
-    std::string path;
+    const ObjectSchema& object; // the schema to check
+    std::string path; // a printable path for error messaging
+    std::unordered_set<std::string> visited; // the list of object types encountered along this path so far
 };
 
-// a non-recursive search that returns a path to any embedded object that has multiple paths from the start
+// a non-recursive search that returns a property path to the first embedded object cycle detected
 std::string do_check(Schema const& schema, const ObjectSchema& start)
 {
     std::queue<CheckObjectPath> to_visit;
-    std::unordered_set<std::string> visited;
-    to_visit.push(CheckObjectPath{start, start.name});
+    to_visit.push(CheckObjectPath{start, start.name, {start.name}});
 
     while (to_visit.size() > 0) {
         auto current = to_visit.front();
-        visited.insert(current.object.name);
         for (auto& prop : current.object.persisted_properties) {
             if (prop.type == PropertyType::Object) {
                 auto it = schema.find(prop.object_type);
                 REALM_ASSERT(it != schema.end()); // this succeeds if the schema is otherwise valid
-                auto next_path = current.path + "." + prop.name;
-                if (visited.find(prop.object_type) == visited.end()) {
-                    to_visit.push(CheckObjectPath{*it, next_path});
+                if (!it->is_embedded) {
+                    // the server does support links to top level objects (serialized as a PK)
+                    // so if we encounter this type of link, no need to check further along this path
+                    continue;
                 }
-                else if (it->is_embedded) {
+                auto next_path = current.path + "." + prop.name;
+                if (current.visited.find(prop.object_type) == current.visited.end()) {
+                    auto visited_copy = current.visited;
+                    visited_copy.insert(current.object.name);
+                    to_visit.push(CheckObjectPath{*it, next_path, std::move(visited_copy)});
+                }
+                else {
                     return next_path;
                 }
             }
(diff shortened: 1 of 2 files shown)


ironage:
Related to https://github.com/realm/realm-object-store/issues/1065 and https://github.com/realm/realm-object-store/pull/1112.
@adamchel could you also double check the test cases here and make sure they agree with the cases the server will restrict?

tests/schema.cpp
```diff
@@ -338,7 +338,19 @@ TEST_CASE("Schema") {
...
-            REQUIRE_THROWS_CONTAINING(schema.validate(), "Cycles containing embedded objects are not currently supported: 'EmbeddedObject.link_to_top_level_object.link_to_embedded_object'");
+            REQUIRE_NOTHROW(schema.validate());
+        }
+
+        SECTION("does not reject a top level loop via embedded object link") {
+            Schema schema = {
+                {"TopLevelObject", {
+                    {"link_to_self", PropertyType::Object|PropertyType::Nullable, "TopLevelObject"}
+                }},
+                {"EmbeddedObject", ObjectSchema::IsEmbedded{true}, {
```
nirinchev, line 349: Nothing is linking to this embedded object, so that schema is probably invalid (we have a ticket to detect freestanding embedded objects, but that hasn't been completed yet).

adamchel: so the server doesn't actually outright reject freestanding embedded object schemas, it will be stored as an "unbound embedded schema" in memory as long as the connection is open, and will be persisted once something links to it and makes it a "bound" embedded schema

adamchel (approved):
The test cases LGTM!

tests/schema.cpp
nirinchev: Right, my point was that it would likely be an error on the developer's end to include a freestanding embedded object in their schema, as there would be nothing meaningful they can do with it. Obviously, it doesn't matter much in a test case, but it would probably be a good idea to use realistic schemas in tests. For example, for this one, I would probably add a link to the embedded object in `TopLevelObject` to verify that both cycles `parent->parent` and `parent->embedded->parent` are allowed.

ironage: Thanks for the clarifications both! I've updated this particular test schema to be more realistic.

ironage:
@nirinchev @jbreams could I get one of you to review the updated loop checking code before merging?

src/schema.cpp
```diff
@@ -82,29 +82,35 @@ Schema::const_iterator Schema::find(ObjectSchema const& object) const noexcept
...
                 REALM_ASSERT(it != schema.end()); // this succeeds if the schema is otherwise valid
+                if (!it->is_embedded) {
+                    // the server does support links to top level objects (serialized as a PK)
+                    // so if we encounter this type of link, no need to check further along this path
+                    continue;
+                }
                 auto next_path = current.path + "." + prop.name;
-                if (visited.find(prop.object_type) == visited.end()) {
-                    to_visit.push(CheckObjectPath{*it, next_path});
+                if (current.visited.find(prop.object_type) == current.visited.end()) {
```
jbreams, line 108: Nit: I think it's a little clearer to see the error path be the thing the conditional is testing for, so could you change this so it's more like this:
```
if (current.visited.find(prop.object_type) != current.visited.end()) {
    return next_path;
}

auto visited_copy = current.visted;
visited_copy.insert(current.object_name);
to_visit.push(...);
```

src/schema.cpp
```diff
@@ -82,29 +82,35 @@ Schema::const_iterator Schema::find(ObjectSchema const& object) const noexcept
 namespace {
 
 struct CheckObjectPath {
-    const ObjectSchema& object;
-    std::string path;
+    const ObjectSchema& object; // the schema to check
+    std::string path; // a printable path for error messaging
+    std::unordered_set<std::string> visited; // the list of object types encountered along this path so far
```
jbreams, line 87: I know you're just copying this from elsewhere in this function, but how many items do we expect visited to contain? std::unordered_set might be overkill here.

jbreams (approved):
Overall, it looks good! Just a few questions.

src/schema.cpp
ironage: I agree that does read better; updated.

src/schema.cpp
ironage: There will only be a handful of items yes. I think the data structure does does convey the intent though; ie a bucket of objects that we've already seen on the path in no particular order. I don't think this is worth measuring performance for different structures because schema checking is not a hot path.

jbreams: That's fine if this is a cold path. I agree the data structure matches the intent. I at least wanted to ask if we need to construct a hash table to store ~10-100 short strings when a vector and `std::find()` would work just as well.

PR #1142, opened 2020-11-19
hold the mutex while calling notify_all

blagoev:
This is a precondition when calling notify_all and fixes a dead lock in Electron on Windows

### DIFF
--- a/src/impl/windows/external_commit_helper.cpp
+++ b/src/impl/windows/external_commit_helper.cpp
@@ -86,7 +86,9 @@ void ExternalCommitHelper::listen()
     while (m_keep_listening) {
         m_commit_available.wait(m_mutex, nullptr);
         if (m_keep_listening) {
+            m_mutex.unlock();
 			m_parent.on_change();
+            m_mutex.lock();
         }
     }
 }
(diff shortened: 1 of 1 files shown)


blagoev:
I have verified this PR fixes multi-process write in Electron. There was a failing test in Realm JS now when on_change is called with the mutex released this RJS tests are green on Windows. 
Also https://github.com/realm/realm-object-store/issues/872 is fixed as well. 

There is still another deadlock on windows on process exit.

fealebenpae:
I think I might know why there's a deadlock on process exit. The `ExternalCommitHelper` destructor attempts to acquire `m_mutex` to signal to the worker thread that it should stop listening. However it's the same mutex that we now use to coordinate calls to `notify_all`, and that's because the worker thread is always waiting on the condition variable so the only way to wake it up and let it know it needs to wrap things up is to "notify" it.

Ordinarily I would have used a second mutex just for that, and waited with `WaitForMultipleObjectsEx`, but the API of `InterprocessConditionVariable` doesn't allow for that.

Under what conditions does the deadlock on process exit occur? Can we afford to merge this PR without breaking anything else?

nirinchev:
The deadlock on process exit existed before this PR (at least when running the repro in #872).

fealebenpae (approved):
Alright, if the deadlock on process exit is an existing issue then I don't see why we can't merge this as it already fixes some other known deadlocks.

blagoev:
yes indeed it's a separate deadlock. 
While its true that the exit deadlock is related to the destructor in ExternalCommitHelper I think it's not the cause. I suspect our design of syncing  multiple process is not robust enough. 
My investigations in Electron showed that when the window is closed the renderer process exits succesfully (exitcode = 0) and the main is still waiting for a specific number of listeners to notify it in `ExternalCommitHelper::listen->m_commit_available.wait`

RedBeard0531:
https://github.com/realm/realm-core/issues/4174 is probably the cause of the deadlock on shutdown

PR #1144, opened 2020-12-01
Allow revival of removed users

ironage:
Fixes https://github.com/realm/realm-core/issues/4142

I think this is all that is required, but I'm not super familiar with the Removed user state and how it differentiates from the logged out state. I also considered doing the user removal at this point, and creating a new object instance but I didn't want to deviate the behaviour from what the logged out state does. So I'll ask reviewers to please confirm that this seems correct.

### DIFF
--- a/src/sync/sync_manager.cpp
+++ b/src/sync/sync_manager.cpp
@@ -348,11 +348,7 @@ std::shared_ptr<SyncUser> SyncManager::get_user(const std::string& user_id,
         // if the order of these were flipped).
         user->update_access_token(std::move(access_token));
         user->update_refresh_token(std::move(refresh_token));
-
         user->set_state(SyncUser::State::LoggedIn);
-        if (!m_metadata_manager)
-            m_current_user = user;
-
         return user;
     }
 }
--- a/src/sync/sync_manager.cpp
+++ b/src/sync/sync_manager.cpp
@@ -322,7 +322,7 @@ std::shared_ptr<SyncUser> SyncManager::get_user(const std::string& user_id,
     auto it = std::find_if(m_users.begin(),
                            m_users.end(),
                            [user_id, provider_type](const auto& user) {
-        return user->identity() == user_id && user->provider_type() == provider_type;
+        return user->identity() == user_id && user->provider_type() == provider_type && user->state() != SyncUser::State::Removed;
     });
     if (it == m_users.end()) {
         // No existing user.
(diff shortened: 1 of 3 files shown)


src/sync/sync_manager.cpp
```diff
@@ -337,22 +337,20 @@ std::shared_ptr<SyncUser> SyncManager::get_user(const std::string& user_id,
         if (!m_metadata_manager)
             m_current_user = new_user;
         return new_user;
-    } else {
+    }
+    else { // LoggedOut || Removed => LoggedIn
         auto user = *it;
```
leemaguire, line 342: I think the issue here is that the `find_if` is returning a user and then it is hitting this `else` case, where what should really happen is that no user gets returned at all. Maybe the solution lies within the search conditions:

```
auto it = std::find_if(m_users.begin(),
                           m_users.end(),
                           [user_id, provider_type](const auto& user) {
        return user->identity() == user_id && user->provider_type() == provider_type && user->state() != SyncUser::State::Removed;
    });
```

ironage: Thanks for the suggestion! I looked into this and there are a few subtle differences. 
In your suggestion, a new user object instance will be created after the second log in, although the underlying sync metadata does revive it's representation in either case (removed -> logged_in). In the changes I proposed, the original user object remains valid after the second log in and changes it's state from Removed to LoggedIn.
So I guess the question is, what is appropriate behaviour here from a user or SDK perspective?

leemaguire: From a lifecycle perspective, reviving a removed user seems incorrect. This user is marked to be discarded. A new user being created after login is fine from my point of view, as it just invokes the user lifecycle from the beginning.

ironage: Thanks for confirming, I agree this makes the most sense. I've updated the code to reflect this.

src/sync/sync_manager.cpp
```diff
@@ -350,9 +349,10 @@ std::shared_ptr<SyncUser> SyncManager::get_user(const std::string& user_id,
         user->update_access_token(std::move(access_token));
         user->update_refresh_token(std::move(refresh_token));
 
-        if (user->state() == SyncUser::State::LoggedOut) {
-            user->set_state(SyncUser::State::LoggedIn);
-        }
+        user->set_state(SyncUser::State::LoggedIn);
+        if (!m_metadata_manager)
+            m_current_user = user;
+
```
leemaguire, line 355: Why is there a need for this?

ironage: I reasoned that we were forgetting to set the current user, but I just tested this and it is being done in a subsequent call to `App::get_profile()` elsewhere so indeed this is not necessary. I will remove it thanks!

ironage:
Bump for reviews :)

jsflax (approved):

Issue #925, opened 2020-02-24
Validate that schemas don't contain orphaned embedded tables


RealmBot:
When opening a Realm (or at compile time for statically typed SDKs), we should validate that we don't have any unlinked to embedded tables. It's inevitably an error in production as those can't be queried or created directly. In addition, it simplifies the way schema changes are handled by the sync server because we won't have to persist unbounded embedded schemas.

RealmBot:
➤ Thomas Goyne commented:

Making it a runtime error to merely define a class and not reference it anywhere sounds very unpleasant during development. I think it's reasonable to make it an error to explicitly include an embedded object type in the schema, but automatic class discovery should simply ignore unused embedded object types.

RealmBot:
➤ Nikola Irinchev commented:

That sounds reasonable to me.

bmunkholm:
Moved to https://github.com/realm/realm-core/issues/4185

PR #1145, opened 2020-12-02
Add a schema check for unreachable embedded objects

ironage:
Fixes https://github.com/realm/realm-object-store/issues/925

### DIFF
--- a/src/schema.cpp
+++ b/src/schema.cpp
@@ -132,7 +132,57 @@ void check_for_embedded_objects_loop(Schema const& schema, std::vector<ObjectSch
         }
     }
 }
+
+void check_for_embedded_object_orphans(Schema const& schema, std::vector<ObjectSchemaValidationException>& exceptions)
+{
+    std::unordered_set<std::string> orphans;
+    std::unordered_set<std::string> visitable;
+    for (auto const& object : schema) {
+        if (!object.is_embedded) {
+            for (auto const& prop : object.persisted_properties) {
+                if (prop.type == PropertyType::Object) {
+                    auto it = schema.find(prop.object_type);
+                    REALM_ASSERT(it != schema.end());
+                    if (it->is_embedded) {
+                        visitable.insert(it->name);
+                    }
+                }
+            }
+        }
+        else {
+            orphans.insert(object.name);
+        }
+    }
+    for (auto it = orphans.begin(); it != orphans.end();) {
+        if (visitable.find(*it) != visitable.end()) {
+            it = orphans.erase(it);
+        }
+        else {
+            ++it;
+        }
+    }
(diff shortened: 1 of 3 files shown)


tgoyne:
This rejects orphans unconditionally, while we only want to reject orphans which were explicitly added to the schema and not just auto-discovered. Object store doesn't currently have a way to distinguish those, so either it needs a way to or this needs to be something explicitly called on the schema by the SDK when it is applicable.

ironage:
Thanks, @tgoyne. Since I'm not familiar with how SDK's do their automatic schema discovery the simplest thing is to provide a method on Schema to do the check that SDK's can opt-in to calling at the appropriate time. What would be the best output of such a method, a list of the names of orphans, or would a list of `ObjectSchemaValidationException` still be appropriate?

tgoyne:
I think the behavior we want is:

1. If the schema was discovered, simply do not create tables for any orphaned embedded types. Having types in the schema which do not have backing tables is already something we support for the sake of read-only Realms. This is so that merely defining a class and not using it isn't an error, as that'd be pretty annoying during development.
2. If the embedded object type is explicitly included in the schema by the user, throw an exception. 
3. If the table for the orphaned embedded object type already exists, validate that it's compatible with the schema even if the schema was auto-discovered.

Satisfying all three of these probably requires pushing the knowledge about if the schema was auto-discovered down pretty far into the logic. A pre-emptive validator-or-filter function probably couldn't handle both skipping creating tables but still validating that the schema matches the table if it already exists. The most natural thing would probably be another schema mode? Having separate AdditiveWithEmbeddedOrphansAllowed and AdditiveWithEmbeddedOrphansForbidden modes (preferably with better names) modes would fit with the existing design and leave it to the SDK to pick the appropriate one without having to pass around addition bits of state. Hopefully the actual logic to make that work isn't awful...

ironage:
For sure, I can add another schema mode. Perhaps AdditiveExplicit and AdditiveDiscovered so that we can maybe reuse the code paths for some other future schema checks?
Would it be useful to separate out the explicit/discovered state from the schema mode so that it can be combined with all other schema modes and not just additive? I just want to double check the design matches the conceptual model because I don't have any experience with the discoverable schema code. But I have no problem with adding it if that's what we need.

tgoyne:
Orphaned embedded tables are only a problem for the server, so the check is only needed for synchronized Realms and synchronized Realms always use the additive-only schema mode. I think having this as a flag that can be added to any schema mode would just complicate things, but if I'm wrong and it would be simpler then there's nothing wrong with that approach. I mostly wouldn't want a separate bool that has to be passed down through a bunch of layers because that has a habit of turning into three bools and some really convoluted logic.

ironage:
Replaced by https://github.com/realm/realm-core/pull/4306

